PR #35, opened 2016-09-30
Portability

saghul:
Increase the portability by not using some Unix specific utilities and using the libuv provided counterparts.


Globik:
Wow, I think it's great.

ibc:
Works nice, merged. Thanks a lot!

PR #48, opened 2017-01-04
Rtcp: use a static buffer for RTCP serialization

jmillan:


worker/include/RTC/Peer.h
```diff
@@ -67,6 +67,9 @@ namespace RTC
 		RTC::RtpReceiver* GetRtpReceiverFromRequest(Channel::Request* request, uint32_t* rtpReceiverId = nullptr);
 		RTC::RtpSender* GetRtpSenderFromRequest(Channel::Request* request, uint32_t* rtpSenderId = nullptr);
 
+	private:
+		static uint8_t rtcpBuffer[];
+
 	/* Pure virtual methods inherited from RTC::Transport::Listener. */
```
ibc: Please, let's place `static` members before instance members.

worker/include/RTC/RTCP/CompoundPacket.h
```diff
@@ -12,21 +12,17 @@ namespace RTC { namespace RTCP
 	class CompoundPacket
 	{
 	public:
-		~CompoundPacket();
 
-		uint8_t* GetRaw();
```
ibc: I think `GetRaw()` must not be removed. Both `GetRaw()` and `GetLength()` should be the public `CompoundPacket` interface for `Transport` to get the raw content of the compound packet and send it.

worker/include/RTC/RTCP/CompoundPacket.h
```diff
@@ -12,21 +12,17 @@ namespace RTC { namespace RTCP
...
 		size_t GetSize();
```
ibc: Does `GetSize()` return the size in bytes of the raw compound packet? If so, could it be `GetLength()` for consistency with `RtpPacket` and others?

worker/include/RTC/RTCP/CompoundPacket.h
```diff
@@ -12,21 +12,17 @@ namespace RTC { namespace RTCP
...
 		void Dump();
 		void AddSenderReport(SenderReport* report);
 		void AddReceiverReport(ReceiverReport* report);
 		void AddSdesChunk(SdesChunk* chunk);
-		void Serialize();
+		void Serialize(uint8_t* data);
 
 	private:
-		uint8_t* raw = nullptr;
-		size_t size = 0;
```
ibc: Same as above:

* `this->raw` must exist (can be assigned once `Serialize()` is called) so `GetRaw()` returns it.
* Not sure why `size` has been removed. Also, if it represents the size in bytes, could it be renamed to `length`?

worker/include/RTC/RtpReceiver.h
```diff
@@ -56,6 +56,9 @@ namespace RTC
 	private:
 		void FillRtpParameters();
 
+	private:
+		static uint8_t rtcpBuffer[];
```
ibc: Same as above: let's move `static` members declaration to the top of the class, before instance members.

worker/include/RTC/Transport.h
```diff
@@ -50,8 +50,7 @@ namespace RTC
 		void AddRtpReceiver(RTC::RtpReceiver* rtpReceiver);
 		void RemoveRtpReceiver(RTC::RtpReceiver* rtpReceiver);
 		void SendRtpPacket(RTC::RtpPacket* packet);
-		void SendRtcpPacket(RTC::RTCP::Packet* packet);
-		void SendRtcpPacket(RTC::RTCP::CompoundPacket* packet);
+		void SendRtcp(const uint8_t* data, size_t len);
```
ibc: I don't like this signature. `Transport.SendRtpPacket()` receives a `RtpPacket` and uses its `GetRaw()` and `GetLength()`. I would like the same for RTCP `CompoundPacket` and `Packet` (for consistency).

worker/src/RTC/RTCP/CompoundPacket.cpp
```diff
@@ -8,70 +8,69 @@ namespace RTC { namespace RTCP
...
 		{
-			this->senderReportPacket.Serialize(this->raw);
+			this->senderReportPacket.Serialize(data);
 			offset = this->senderReportPacket.GetSize();
 
 			if (this->receiverReportPacket.GetCount())
 			{
 				// Fix header length.
-				Packet::CommonHeader* header = (Packet::CommonHeader*) this->raw;
+				Packet::CommonHeader* header = (Packet::CommonHeader*) data;
```
ibc: Let's please always use `static_cast` when casting pointers to classes.

worker/include/RTC/Transport.h
ibc: I understand that there should be two methods here, one to send RTCP packets and other to send RTCP compound packets. Yes, they can be joined into a single one (by just passing the pointer and the length) but IMHO it does not look nice. I suggest having both `SendRtcpPacket(RTC::RTCP::Packet)` and `SendRtcpCompoundPacket(RTC::RTCP::CompoundPacket)`.

worker/src/RTC/RtpReceiver.cpp
```diff
@@ -302,14 +307,34 @@ namespace RTC
...
+		{
+			// Ensure that the RTCP packet fits into the RTCP buffer.
+			if (packet->GetSize() > MS_RTCP_BUFFER_SIZE)
+			{
+				MS_WARN_TAG(rtcp, "cannot send RTCP packet, size too big (%zu bytes)", packet->GetSize());
+				return;
+			}
+
+			packet->Serialize(RtpReceiver::rtcpBuffer);
+			this->transport->SendRtcp(RtpReceiver::rtcpBuffer, packet->GetSize());
```
ibc: Same as above. The signature should be clear. `SendRtcpPacket(packet)`.

worker/src/RTC/RtpReceiver.cpp
```diff
@@ -302,14 +307,34 @@ namespace RTC
...
```
ibc: Same.

worker/include/RTC/RTCP/CompoundPacket.h
jmillan: In RTCP packet implementation the `raw` attribute holds an internal buffer and `GetRaw` returns a pointer to it. This buffer is allocated when calling `Serialize()` (with no argument).

In `CompoundPacket` there is no reason for having an internal buffer, so `raw` attribute will point to the buffer provided in `Serialize(uint8_t buffer)`, Ok?

worker/include/RTC/RTCP/CompoundPacket.h
jmillan: Ok, I'll make the change for all RTCP then.

ibc: Wait wait, may be we can rename `GetLength` to `GetSize` everywhere.

worker/include/RTC/RTCP/CompoundPacket.h
jmillan: The reason is because a `CompoundPacket` is created to be immediately sent. There is no need to locally store the serialised data since an external buffer is used for that.

worker/include/RTC/Transport.h
jmillan: Yep, I agree.

worker/include/RTC/RTCP/CompoundPacket.h
ibc: It's just that, for example, `SdesItem` has a `GetSize` and `GetLength`, and I think we use both `GetSize` and `GetLength` for the same purpose in some places.

worker/include/RTC/RTCP/CompoundPacket.h
ibc: Sure.

worker/src/RTC/RTCP/CompoundPacket.cpp
jmillan: In this case `Packet::CommonHeader` is a struct.

ibc: Opps, right.

worker/include/RTC/RTCP/CompoundPacket.h
jmillan: Not really.

`GetLength()` indicates a static value in `SdesItem`. It represents the length of the item value, which cannot be changed.

`GetSize()` is dynamic in the sense that a when a RTCP packet is modified, ie: adding an item, removing an item, etc, the size of the element changes.

jmillan: s/size of the element/size of the packet/

jmillan: We can perfectly rename `GetSize()` by `GetLength`

worker/include/RTC/RTCP/CompoundPacket.h
jmillan: yes, sounds good.

worker/include/RTC/RTCP/CompoundPacket.h
ibc: Don't worry, we can do a consistent renaming later. It's just that I don't yet understand the difference between `SdesItem.GetSize` and `SdesItem.GetLength`:

```c++
inline
size_t SdesItem::GetSize()
{
	return 2 + size_t(this->header->length);
}

inline
uint8_t SdesItem::GetLength()
{
	return this->header->length;
}
```

worker/include/RTC/RTCP/Packet.h
```diff
@@ -140,9 +137,9 @@ namespace RTC { namespace RTCP
 	}
 
 	inline
-	uint8_t* Packet::GetRaw()
+	const uint8_t* Packet::GetData()
 	{
-		return this->raw;
+		return (uint8_t*)this->header;
 	}
 }}
```
ibc: The method is `const uint8_t*` but it returns a `uin8_t*), doesn't it cause a warning?

worker/include/RTC/Transport.h
```diff
@@ -51,7 +51,7 @@ namespace RTC
 		void RemoveRtpReceiver(RTC::RtpReceiver* rtpReceiver);
 		void SendRtpPacket(RTC::RtpPacket* packet);
 		void SendRtcpPacket(RTC::RTCP::Packet* packet);
-		void SendRtcpPacket(RTC::RTCP::CompoundPacket* packet);
+		void SendRtcpCompoundPacket(RTC::RTCP::CompoundPacket* packet);
 		RTC::RtpReceiver*  GetRtpReceiver(uint32_t ssrc);
```
ibc: There are two spaces there XD

PR #65, opened 2017-03-13
Add support for sending packets from the Javascript side.

ey6es:
In order to allow replaying packet streams recorded from the 'rtpraw' event, this provides a means to create a Peer and RtpReceiver without setting the transport or capabilities.   The Javascript playback code can then create a Peer and an RtpReceiver, initializing the latter with previously recorded rtpParameters, and call RtpReceiver.receiveRtpPacket to "play" recorded packets.

This is pretty rough, and thus I wouldn't expect you to merge it directly, but I thought you'd be interested.  The background is that we wanted to play with this functionality sooner rather than later for our project, and we weren't sure where it was on your road map.

In testing, I've managed to get it working (that is, playing recorded audio) on Firefox and Chrome, though there may be some timing issues (sometimes the streams don't register--haven't yet delved into WebRTC debugging to figure out why), and there are certainly issues with playing the stream more than once (I am currently generating switching the stream to use random ssrc and msid values to work around this, which seems to work on Chrome but not Firefox).

### DIFF
--- a/lib/Channel.js
+++ b/lib/Channel.js
@@ -180,22 +180,32 @@ class Channel extends EventEmitter
 		}, 250);
 	}
 
-	request(method, internal, data)
+	request(method, internal, data, binary)
 	{
 		let id = utils.randomNumber();
 
 		logger.debug('request() [method:%s, id:%s]', method, id);
 
-		let request = { id, method, internal, data };
+		let request = { id, method, internal, data, binary: binary !== undefined };
 		let ns = netstring.nsWrite(JSON.stringify(request));
 
 		if (Buffer.byteLength(ns) > NS_MAX_SIZE)
 			return Promise.reject(new Error('request too big'));
 
+    let binaryNs;
+    if (binary !== undefined) {
+      binaryNs = netstring.nsWrite(binary);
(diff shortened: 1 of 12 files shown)


notedit:
cool    i need this too.

ibc:
Thanks a lot. This is really interesting and there is lot of nice code in this PR, although I'm not sure whether the implementation is my preferred one. It has some issues due the lack of a real `Transport` for those "peers" just sending RTP packets via JavaScript. This means problems when it comes to generate RTCP Sender Reports and RTCP feedback packets. I think we need a different kind of `Transport` for this purpose, some kind of `JsTransport` (not sure yet).

Let me please come back to this once 1.0.0 is out (which is very close right now).

notedit:
any  plan on this feature?

ibc:
This will be done in [2.0.0](https://github.com/versatica/mediasoup/milestone/2) using a much more flexible approach.

ibc:
We are adding specialized `Producers` and `Consumers` (such as the new `RtpStreamer`) in mediasoup v2. This feature will be added eventually.

PR #71, opened 2017-03-17
MODIFY reduce redundancy accessing nested members

ggarber:
Minor style check, I hope you like it.   I only did it in the class I was looking at but if you like it maybe you can start using that still to simplify the code.

BTW If you are still in time I would recommend using a trailing underscore (f.e. rtp_parameters_) to differentiate instance variables vs local variables (f.e. instead of this->rtpParameters)

### DIFF
--- a/worker/src/RTC/RtpSender.cpp
+++ b/worker/src/RTC/RtpSender.cpp
@@ -196,12 +196,13 @@ namespace RTC
 		// Remove RTP parameters not supported by this Peer.
 
 		// Remove unsupported codecs.
-		for (auto it = this->rtpParameters->codecs.begin(); it != this->rtpParameters->codecs.end();)
+		auto codecs = this->rtpParameters->codecs;
+		for (auto it = codecs.begin(); it != this->rtpParameters->codecs.end();)
 		{
 			auto& codec = *it;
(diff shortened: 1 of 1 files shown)


ibc:
Done, thanks!

> BTW If you are still in time I would recommend using a trailing underscore (f.e. rtp_parameters_) to differentiate instance variables vs local variables (f.e. instead of this->rtpParameters)

That's a hard subject IMHO. Honestly I prefer `this->xxxx` rather than special chars in member names. Said that, why not `_xxxx` rather than `xxxx_`?

However, I'm fully open to discuss about coding styles in *mediasoup* and I'm able to change it. But let's discuss about that after 1.0.0 :)

ggarber:
@ibc That's why I prefer to get an external referee :)    

Probably this is the best referee to make these decision (after 1.0 if you wish :)):
https://google.github.io/styleguide/cppguide.html#Variable_Names

Also because cpplint validates those rules so it allows you to not have to argue about style anymore.   travis will argue with everybody :P

Thx for merging!

ibc:
I checked the Google C++ Guide in the past... and I did not like it much :)
But it may change in the future.

PR #76, opened 2017-03-25
Proper handling of RTP extensions

ibc:
Includes a much better API to get RTP header extension values.


ibc:
Thanks. However I'm fixing a small bug (will commit the fix in master).

jmillan:
Ok

Issue #82, opened 2017-03-29
Installation fails on arch linux


bf:
When installing on my arch linux system with both python2 and python3 installed, I get the following error:

```
bf@xxx> npm i --save mediasoup 

> mediasoup@0.8.0 postinstall /foo/bar/node_modules/mediasoup
> make Release

cd worker &&  ./scripts/configure.py -R mediasoup-worker
  File "./scripts/configure.py", line 48
    print 'Error running GYP'
                            ^
SyntaxError: Missing parentheses in call to 'print'
make: *** [Makefile:19: Release] Error 1
```

It looks like `PYTHON?=$(type -p python2 || echo python)` in `Makefile` returns an empty string. When I do `type -p python2 || echo python` in bash it returns `/usr/bin/python2`. The shell from which I am using npm is fish.

Can anybody help me?

ibc:
Interesting. No idea about `fish`, but I've installed it to check and indeed the problem is there:

```bash
ibc@ibc-macbook ~> echo $(type -p python2 || echo python)
Unsupported use of '||'. In fish, please use 'COMMAND; or COMMAND'.
fish: echo $(type -p python2 || echo python)
                              ^
```

@saghul help please :)

ibc:
To be clear: using `or` instead of `||` would not work in bash/sh, so that's not a solution.

bf:
@ibc how about we add a `#!/bin/bash` on top of the Makefile? Would it then run in bash?

ibc:
No, Makefiles are not scripts but just text files read and interpreted by the `make` command.

ibc:
@bf I don't know how to solve it right now. In the meanwhile, may you try moving to bash or sh before `npm install mediasoup`?

ibc:
Just asked in [StackOverflow](https://stackoverflow.com/questions/43102552/same-fish-and-bash-command-for-detecting-python-executable-name), let's see.

bf:
@ibc you're right. I shouldn't worry about this because the servers aren't running this weird shell :-) They run plain bash. I'll try installing within bash on my dev machine.

Thanks also for your immediate response, much appreciated. Keep up the good work. Will post again to let you know if it worked.

bf:
@ibc Dang. I tried in `bash` and `sh`, both return the same error. I also did `su root` and npm install there, also same error. I tried `sudo -u foo npm i --save ...` but it didn't work because of missing `nvm` initialization.

Not sure what I can do?

[...]

PR #83, opened 2017-03-29
Fix bug where Makefile can't handle empty $PYTHON

bf:
References https://github.com/ibc/mediasoup/issues/82


bf:
It doesnt work like that -.-

bf:
Okay so this issue is too weird for me tonight. I have installed manually on dev machine now. We will see what the server does.

ibc:
Fine, let's move back to #82.

Issue #89, opened 2017-03-31
"invalid TMMB bitrate value"


ibc:
In master branch:

```
RTC::RTCP::FeedbackRtpTmmb::FeedbackRtpTmmbItem() | invalid TMMB bitrate value : 43461 *2^63
```

jmillan:
The value may indeed be invalid :-). I'll check it.

ibc:
NOTE: I also see this:

```
RTC::Peer::onTransportRtcpPacket() | ignoring unsupported Feedback packet 'TMMBN' [sender_ssrc:0, media_ssrc:0]
```

`media_ssrc:0`?

jmillan:
It's Okey, the useful info is in the packet items, which contains the SSRCs they refer to, each. This packet type to be properly handled (forwarded) as commented this morning.

ibc:
opps, ok ok, sorry.

ibc:
I don't think this is ok:

```c++
	template <typename T>
	FeedbackRtpTmmbItem<T>::FeedbackRtpTmmbItem(const uint8_t* data)
	{

		this->ssrc = Utils::Byte::Get4Bytes(data, 0);

		// Read the 4 bytes block.
		uint32_t compact = Utils::Byte::Get4Bytes(data, 4);

		// Read each component.
		uint8_t exponent = compact >> 26;              // 6 bits.
		uint64_t mantissa = (compact >> 9) & 0x1ffff;  // 17 bits.
		this-> overhead = compact & 0x1ff;             // 9 bits.
```

Here we are reading bytes from 5 to 8 as a `uint32_t` number, and then split it into components. Is that right?

ibc:
What mostly worries me (but I'd may be 100% wrong) is this:

```c++
template <typename T>
FeedbackRtpTmmbItem<T>::FeedbackRtpTmmbItem(const uint8_t* data)
{
	this->ssrc = Utils::Byte::Get4Bytes(data, 0);

	// Read the 4 bytes block.
	uint32_t compact = Utils::Byte::Get4Bytes(data, 4);

	// Read each component.
	uint8_t exponent = compact >> 26;              // 6 bits.
	uint64_t mantissa = (compact >> 9) & 0x1ffff;  // 17 bits.
	this-> overhead = compact & 0x1ff;             // 9 bits.

	// Get the bitrate out of exponent and mantissa.
	this->bitrate = (mantissa << exponent);

	if ((this->bitrate >> exponent) != mantissa)
	{
[...]

PR #91, opened 2017-04-01
Forward TMMBR and TMMBN packets

ibc:
Forward TMMBR and TMMBN feedback packets to the corresponding `RtpReceiver`.

### DIFF
--- a/worker/src/RTC/Peer.cpp
+++ b/worker/src/RTC/Peer.cpp
@@ -202,7 +202,7 @@ namespace RTC
 				Json::Value data = this->capabilities.toJson();
 
 				// NOTE: We accept the request *after* calling onPeerCapabilities(). This
-				// guarantees that the Peer will receive a 'newrtpsender' event for all its
+				// guarantees that the Peer will receive a "newrtpsender" event for all its
 				// associated RtpSenders *before* the setCapabilities() Promise resolves.
 				// In other words, at the time setCapabilities() resolves, the Peer already
 				// has set all its current RtpSenders.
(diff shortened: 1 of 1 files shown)


ibc:
Question: should we also check possible TMMB packets into received RTCP compound packets?

ibc:
Also, this depends on #89

ibc:
NOTE: I do know that the PR is wrong, as it must extract all the items from the TMMB packet and get the `RtpReceiver` associated to each one. But, since I've also improved something in the logs in the same `Peer.cpp` file, may we keep the work in this branch?

worker/src/RTC/Peer.cpp
```diff
@@ -906,15 +914,38 @@ namespace RTC
...
+									"no RtpSender found for received NACK Feedback packet [sender_ssrc:%" PRIu32 ", media_ssrc:%" PRIu32 "]",
+									feedback->GetMediaSsrc(), feedback->GetMediaSsrc());
 							}
 
 							break;
 						}
 
 						case RTCP::FeedbackRtp::MessageType::TMMBR:
 						case RTCP::FeedbackRtp::MessageType::TMMBN:
+						{
```
jmillan: TMMBR is sent by a remote `receiver` as a feedback to the `sender` of the indicated SSRC.
TMMBN is sent by the remote `sender` as response to the previous one.

They cannot be threaten the same way.

ibc: Wow, that's also right... disaster... I will just cancel this PR and delete the branch (I will copy the logs improvement into master branch).

jmillan:
> Question: should we also check possible TMMB packets into received RTCP compound packets?

Compound packets are decomposed on reception, just when parsing. After parsing them Mediasoup does not know whether they came in a compound or separately. It just processes the packets separatelly.

ibc:
Yes, right... I always forget that our RTCP design is so cool :)

PR #92, opened 2017-04-03
Constness

jmillan:


jmillan:
Some const ness applied to Listeners.

PR #95, opened 2017-04-03
Add transport.setMaxBitrate() API

ibc:
Generated REMB bitrate for that a Transport won't exceed such a value (if given).


jmillan:
:+1: super

PR #96, opened 2017-04-04
Transport: trigger PLI if uplink bitrate abruptly decreases

ibc:
Also, sorry for some pure cosmetic changes in REMB C++ files...


ibc:
This improves a bit the "frozen video" when the available bandwidth is artificially and abruptly decreased. Still some improvements come to my mind, such as clearing the associated buffer in the `RtpSenders` to avoid sending retransmitted packets once we have requested a PLI to the `RtpReceiver`.

PR #132, opened 2017-09-07
Consumer: generate self RTP sequence numbers and timestamps.

jmillan:


worker/include/RTC/Consumer.hpp
```diff
@@ -83,6 +83,12 @@ namespace RTC
 		// RTP counters.
 		RTC::RtpDataCounter transmittedCounter;
 		RTC::RtpDataCounter retransmittedCounter;
+		// RTP sequence number and timestamp.
+		uint16_t seqNum{ 0 };
+		uint32_t rtpTimestamp{ 0 };
+		uint16_t lastRecvSeqNum{ 0 };
+		uint32_t lastRecvRtpTimestamp{ 0 };
+		bool syncRequired{ true };
```
ibc: Should this be true?

worker/src/RTC/Consumer.cpp
```diff
@@ -267,6 +272,38 @@ namespace RTC
...
+			this->seqNum += 1;
+
+			uint64_t now = DepLibUV::GetTime();
+
+			if (now > this->rtpTimestamp)
+				this->rtpTimestamp = now;
+
+			this->syncRequired = false;
+
+			RequestFullFrame();
```
ibc: Is this call required? Note than, upon resumed, the Consumer already requests a full frame.

worker/src/RTC/Consumer.cpp
```diff
@@ -267,6 +272,38 @@ namespace RTC
...
+
+		// Update packet sequence number.
+		packet->SetSequenceNumber(this->seqNum);
+
+		// Update packet timestamp.
+		packet->SetTimestamp(this->rtpTimestamp);
+
 		// Process the packet.
 		if (!this->rtpStream->ReceivePacket(packet))
 			return;
```
ibc: If, for any reason, `rtpStream->ReceivePacket()` returns false, the seq and timestamp of the packet won't be reverted so other Consumers will fail. seq and timestamp should always be reverted, no matter what happens in each Consumer.

worker/include/RTC/Consumer.hpp
jmillan: We want the very first packet being relayed to have a fresh new sequence number and timestamp, otherwise we cannot calculate it out of a 'non-received' packet.

worker/src/RTC/Consumer.cpp
jmillan: Does it also cover the case where the remote endpoint changes the video device or updates the resolution?

worker/src/RTC/Consumer.cpp
jmillan: True

worker/include/RTC/Consumer.hpp
ibc: right

worker/src/RTC/Consumer.cpp
ibc: If the remote Producer changes the video stream then it will send a full frame by itself.

jmillan: Okey

jmillan: For some reason requesting it from the Producer does not work 100% or times. It does when requesting from Cosumer.

I'll investigate further to know the reason. Meanwhile it's functional.

PR #136, opened 2017-09-16
Fix serialization of RTX decoded packets.

ibc:
Clone decodec RTX packets so they become later serializable by just getting their `GetData()` and `GetSize()`.

NOTE: This fix also required this [change](9f47d41aee1641cf1e7cf1493b95ee9daa8c34d8) in `RtpPacket::Clone()`.

### DIFF
--- a/TODO_V2.md
+++ b/TODO_V2.md
@@ -1,9 +1,5 @@
 # TODO in mediasoup v2 (server-side)
 
-* `RtpPacket::RtxEncode()` should remove the original padding (also the header flag by using `SetPayloadPaddingFlag(false)`), then set the RTX payload, then check whether padding is required and, if so, set the padding flag and bytes.
-  - Similar stuff (inverse) with `RtxDecode()`.
-  - NOTE: Nobody does this. Also, it's dangerous since after `RtxDecode()` it may happen that N bytes of padding are needed, but those exceed the original RTX packet size.
-
 * In `RtpStreamSend::StorePacket()` we should remove old packets (as `NackGenerator` does).
 
 * Remove `producer.on('rtprawpacket')` and, instead, create a special `RtpConsumer` or something like that.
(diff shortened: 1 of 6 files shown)


jmillan:
What do you think about this modification on the RtpPacket::DecodeRtx instead?

It is more in sync with RtpPacket::EncodeRtx implementation and does not require cloning since the decoded RTX packet becames the 'usable' packet for all the consumers, as we were already assuming.

```diff
diff --git a/worker/src/RTC/RtpPacket.cpp b/worker/src/RTC/RtpPacket.cpp
index 13af6c0..6db32e3 100644
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -421,8 +421,10 @@ namespace RTC
 		// Rewrite the sequence number.
 		this->SetSequenceNumber(Utils::Byte::Get2Bytes(this->payload, 0));

-		// Fix the payload and payload length.
-		this->payload += 2;
+		// Shift the payload to its original place.
+		std::memmove(this->payload, this->payload + 2, this->payloadLength - 2);
+
+		// Fix the payload length.
 		this->payloadLength -= 2;
```

ibc:
Yep, it seems much efficient. I will check the payload padding stuff and apply that.

ibc:
PLZ PLZ review

TODO_V2.md
```diff
@@ -1,5 +1,9 @@
 # TODO in mediasoup v2 (server-side)
 
+* `RtpPacket::RtxEncode()` should remove the original padding (also the header flag by using `SetPayloadPaddingFlag(false)`), then set the RTX payload, then check whether padding is required and, if so, set the padding flag and bytes.
```
jmillan: How does a RTX packet receiver suppose to know that padding is required on the original RTP?

ibc:
Done in v2.

PR #137, opened 2017-09-19
Virtual RTC::Transport and RTC::WebRtcTransport descendant

ibc:
In order to provide future "plugins" such as specialized `Producers` or `Consumers` (write to disk, broadcast in plain RTP, etc) we don't need specialized classes of `Producer` or `Consumer`, but specialized `Transport` classes.

The current PR makes `RTC::Transport` a virtual class and provides a `RTC::WebRtcTransport` descendant class so in the future it should be easier to add more in/out stuff.


jmillan:
@ibc, I find it really nice. Please merge it into v2 at your will.

PR #142, opened 2017-09-21
Add RtpPacket::CreateProbationPacket()

ibc:


ibc:
NOTE: If probation packets are inserted within the real video RTP stream, we must check whether it's valid for them to have the `marker` bit set to 0 (or stuff like that...), as we may break things.

ibc:
From RFC 3551:

>    Most of these video encodings also specify that the marker bit of the
   RTP header SHOULD be set to one in the last packet of a video frame
   and otherwise set to zero.  Thus, it is not necessary to wait for a
   following packet with a different timestamp to detect that a new
   frame should be displayed.

So it's ok to send probation packets (with marker bit 0) between video packets sharing `timestamp` (given that the `timestamp` of the probation packets matches the same value).

PR #150, opened 2017-10-09
GetStats implementation

jmillan:


jmillan:
Comments handled and merged.

PR #180, opened 2018-01-03
Added missing include cmath so that std::round can be used

jacobEAdamson:
std::round needs `#include <cmath>` to work in g++ 6.3.0 and clang 9.0.0

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,12 @@
+{
+  "files.associations": {
+    "__config": "cpp",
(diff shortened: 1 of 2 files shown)


ibc:
Thanks. I've commited 66ded49155d18c33def49b2ccc08ac97fadfbdf5 since your PR contains a `.vscode` folder.

ibc:
Available in mediasoup 2.0.7, published right now.
Thanks again.

PR #190, opened 2018-02-07
Fix #188. Increase profiles one by one unless explicitly forced

### LINKED ISSUE #188: Strange behavior in "low or bad" bandwidth links
Hi,

following rules from Jose Luis, i put this issue.

When a client connect to server with an "low or very bad" bandwidth link, the server is always (each 2-3 seconds) sending "consumerEffectiveProfileChanged" notification to cliente, jumping to profile "high", "medium" and "low". When client is using "medium/low" profile, the stream is possible, but when the server jump to "high" profile ... the client freeze the stream. I uploaded a STAT trace, calling consumer.enableStats() from client side.

Begin lines with "cliente.js:88" show the "consumerEffectiveProfileChanged" notification, and "clientRoom.js:159" all the DATA RTP. I hope this can help you to find the source of this behavior.

MacBook Air - MacOS 10.12.6
node v7.6.0
npm mediasoup-2.0.11
mediasoup-client-2.0.8 
Disabled "maxBitrate" from config.js in mediasoup-server
The link used it's a IPSec VPN in a 4G link

Regards,
Heri

[traceSTATs.txt](https://github.com/versatica/mediasoup/files/1688891/traceSTATs.txt)

jmillan:
@ibc,

This was already there and remove by https://github.com/versatica/mediasoup/commit/b291ef4ed7dbda3f824731cdad7746b404f2af38#diff-89773978bd7f2fe02e0919ba2011c27aL1006

It prevents the levels to be upgraded one by one as it should which causes the issue #188

### DIFF
--- a/worker/src/RTC/Consumer.cpp
+++ b/worker/src/RTC/Consumer.cpp
@@ -934,12 +934,15 @@ namespace RTC
 			// Downgrade the target profile.
 			newTargetProfile = (std::prev(it))->first;
 		}
-		// If there is no preferred profile, get the highest one.
+		// If there is no preferred profile, get the next higher one available.
 		else if (GetPreferredProfile() == RTC::RtpEncodingParameters::Profile::DEFAULT)
 		{
-			auto it = this->mapProfileRtpStream.crbegin();
+			auto it = this->mapProfileRtpStream.upper_bound(this->effectiveProfile);
 
-			newTargetProfile = it->first;
+			if (it != this->mapProfileRtpStream.end())
+				newTargetProfile = it->first;
+			else
+				newTargetProfile = this->effectiveProfile;
 		}
 		// Try with the closest profile to the preferred one.
 		else


ibc:
LGTM. Is it tested? If so let's merge it.

jmillan:
> LGTM. Is it tested? If so let's merge it.

Difficult to test from a single device. Limiting resources to one participant affects both in the testing room. We can give it a quick shoot later today.

ibc:
This PR basically changes:

> "If there is no preferred profile, get the highest one"

into this:

> "If there is no preferred profile, get the next higher one available"

In v3 we no longer have profiles (we have spatial layers instead). Anyway, the important thing is that in v3 there is no chance for not having "preferred profile" (or preferred spatial layer). This is, the preferred spatial layer is always set to the maximum value and it cannot be "unset" by the user.

PR #191, opened 2018-02-07
Issue 174

jmillan:


worker/src/RTC/RtpDataCounter.cpp
```diff
@@ -57,7 +62,7 @@ namespace RTC
 		// A whole window size time has elapsed since last entry. Reset the buffer.
 		if (newOldestTime > this->oldestTime + this->windowSize)
 		{
-			this->Reset(newOldestTime);
+			this->Reset(now);
```
ibc: Let's avoid `this->` in methods.

jmillan: Ups.., right

PR #199, opened 2018-04-05
Add localIP option for RtpStreamer and RTP mirroring (fixes #198)

### LINKED ISSUE #198: Add support for `RtpStreamer` and mirroring binding in a different IP
Typical use case:

* Use the public IP of the server for RTP with clients.
* Use the private IP as source IP for forwarding RTP to another server in the LAN via the private interface.

ibc:
Related to #198.

Adds a new `localIP` optional setting for [RtpStreamerOptions](https://mediasoup.org/documentation/mediasoup/api/#RtpStreamer-RtpStreamerOptions) and [MirroringOptions](https://mediasoup.org/documentation/mediasoup/api/#WebRtcTransport-MirroringOptions).

If set, such a IP will be used as local IP rather than the `rtcIPv4` or `rtcIPv6`.

Note that when it's set, a random port will be chosen and this won't be necessarily in the range of `rtcMinPort`-`rtcMaxPort`.

### DIFF
--- a/lib/Room.js
+++ b/lib/Room.js
@@ -299,13 +299,15 @@ class Room extends EnhancedEventEmitter
 
 				return rtpStreamer;
 			})
-			.catch(() =>
+			.catch((error) =>
 			{
 				if (plainRtpTransport)
 					plainRtpTransport.close();
 
 				if (consumer)
 					consumer.close();
+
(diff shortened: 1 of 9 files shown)


jmillan:
LGTM :+1:

PR #203, opened 2018-05-18
getStats: add kind as alias for mediaType

fippo:
see https://github.com/w3c/webrtc-stats/issues/301

:nail_care: 

I did not see any `how to run tests` but it still compiles...


ibc:
> I did not see any how to run tests but it still compiles...

`npm test` runs some stuff.

Merged, thanks!

PR #211, opened 2018-07-19
support multi ips

yangjinecho:


ibc:
Wow! this is an awesome feature and PR!

We'll check it as far as we can. Just a few comments:

* Could you describe here the changes in the JavaScript side? I mean: which new fields should be given to use a custom IPv4 or IPv6?
* Also, I like this multi-IP feature and I think that the existing API should change. Instead of providing a default IPv4 and IPv6 in the Server settings, we could just require passing a `ipv4` and/or `ipv6` in the `createTransport()` JavaScript API.
    * This would require some changes to JavaScript layer.
* What about the `PlainRtpTransport.cpp`? It should also be able to use a custom IP.

Thanks!

yangjinecho:
Will you modify JavaScript  api to adapt  multi-IP feature？

jmillan:
Very nice work @yangjinecho!

I think that `ipv4` and `ipv6` settings can change to accept:

 * `Boolean` (for auto-detect)
 * `String` (single IPv[46] address)
 * `Array<IPv[46] address>`

ibc:
@yangjinecho yes, we can do the JS part (we must think about how to do it). We'll check this PR on next week and update here. Thanks!

ibc:
Hi @yangjinecho, it will take us longer than expected before we can work on this PR. Will notify here when we start working on it.

ibc:
Hi. Thanks for this PR. However we are gonna delay this feature for v3, as described it the v3 roadmap: #227

ibc:
Hi @yangjinecho, just wanted to let you know that mediasoup v3 (work in progress) no longer accepts RTC ipv4/ipv6 within the `Server` settings but, instead, requires passing the desired IP (or IPs) when creating each `Transport`. For that, I've created a `PortManager` class that manages all the IPs and ports:

* https://github.com/versatica/mediasoup/blob/v3/worker/include/RTC/PortManager.hpp
* https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/PortManager.cpp

PR #218, opened 2018-10-18
Fix reading out of buffer when parsing malformed H.264 packets

ggarber:
If naluSize value would be `len - 1` you would end up reading out of buffer because sizeof(naluSize) is 2.

### DIFF
--- a/worker/src/RTC/Codecs/H264.cpp
+++ b/worker/src/RTC/Codecs/H264.cpp
@@ -51,7 +51,7 @@ namespace RTC
 					}
 
 					// Check if there is room for the indicated NAL unit size.
-					if (len <= naluSize)
+					if (len < (naluSize + sizeof(naluSize)))
 						break;
 
 					offset += naluSize + sizeof(naluSize);


worker/src/RTC/Codecs/H264.cpp
```diff
@@ -51,7 +51,7 @@ namespace RTC
 					}
 
 					// Check if there is room for the indicated NAL unit size.
-					if (len <= naluSize)
+					if (len < (naluSize + sizeof(naluSize))
```
jmillan: such check is redundant since it is present in the loop condition.

```while (len > 3) // 3 => sizeof(naluSize) /* 2 */ + NAL unit header size /* 1 */```

Comment (ibc) [reply worker/src/RTC/Codecs/H264.cpp]: Also there is a missing `)`:

```
../src/RTC/Codecs/H264.cpp: In static member function ‘static RTC::Codecs::H264::PayloadDescriptor* RTC::Codecs::H264::Parse(uint8_t*, size_t)’:
../src/RTC/Codecs/H264.cpp:55:7: error: expected ‘)’ before ‘break’
       break;
       ^
../src/RTC/Codecs/H264.cpp:55:12: warning: suggest braces around empty body in an ‘if’ statement [-Wempty-body]
       break;
            ^
make[1]: *** [/home/travis/build/versatica/mediasoup/worker/out/Release/obj.target/mediasoup-worker/src/RTC/Codecs/H264.o] 
```

:)

ggarber: if len is 100 and naluSize is 99 then...   

`len -= naluSize + sizeof(naluSize);`  // 100 - 101

Because len is unsigned its new value will be a very big number bigger than 3 and it will enter in the `while (len > 3)` condition

Sorry for being picky and/or if my comment doesn't make sense.

jmillan: You're right,

> Sorry for being picky and/or if my comment doesn't make sense.

Nothing to sorry about!

ibc: I'm lost. So what? is there something to fix or not? and if so, is this PR valid for that?

NOTE: As said above there is a missing `)`.

jmillan: Yes, the PR is correct.

ibc: It's correct but it does not compile :)

jmillan: yes, you pointed that. Let me fix it

PR #223, opened 2018-11-16
Plain RTP transport

jmillan:


ibc:
LGTM, however travis fails because `gulp lint:worker` (`gulp format` missing):

https://travis-ci.org/versatica/mediasoup/jobs/455954648

jmillan:
true, done

PR #265, opened 2019-01-21
KeyFrameRequestManager

jmillan:


worker/include/RTC/KeyFrameRequestManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+	public:
+		class Listener
+		{
+		public:
+			virtual void OnKeyFrameNeeded(uint32_t ssrc) = 0;
+		};
+
+	public:
+		explicit KeyFrameRequestManager(Listener* listener);
+		~KeyFrameRequestManager() = default;
```
ibc: I suggest removing all those useless destructors that do nothing. I've already removed all non needed virtual methods (child classes do not need to define their methods as virtual at all).

worker/include/RTC/KeyFrameRequestManager.hpp
```diff
@@ -0,0 +1,98 @@
+#ifndef MS_KEY_FRAME_REQUEST_MANAGER_HPP
+#define MS_KEY_FRAME_REQUEST_MANAGER_HPP
+
+#include "handles/Timer.hpp"
+#include <map>
+
+static uint16_t KeyFrameWaitTime{ 2000 };
```
ibc: This is creating a global constant visible to any source file including this header. I think we should define the class constructor in the CPP file and have declare there required constants.

worker/include/RTC/KeyFrameRequestManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+{
+	return this->ssrc;
+}
+
+inline void RTC::PendingKeyFrameInfo::OnTimer(Timer* timer)
+{
+	if (timer == this->timer)
+	{
+		this->listener->OnKeyFrameRequestTimeout(this);
+	}
```
ibc: Unneeded brackets.

worker/include/RTC/KeyFrameRequestManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+		void KeyFrameNeeded(uint32_t ssrc);
+		void KeyFrameReceived(uint32_t ssrc);
+
+		/* Pure virtual methods inherited from PendingKeyFrameInfo::Listener. */
+	public:
+		void OnKeyFrameRequestTimeout(PendingKeyFrameInfo* keyFrameRequestInfo) override;
+
+	private:
+		Listener* listener{ nullptr };
+		std::map<uint32_t, PendingKeyFrameInfo*> mapSsrcPendingKeyFrameInfo;
```
ibc: May be this could be `std::map<uint32_t, PendingKeyFrameInfo>` (no pointer) so we avoid so much allocation/deallocation? Of course it would require in-place constructor when inserting a key/value into the map.

Just wondering. Not important.

worker/include/RTC/KeyFrameRequestManager.hpp
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+/* Inline KeyFrameRequestManager methods */
+
+inline RTC::KeyFrameRequestManager::KeyFrameRequestManager(KeyFrameRequestManager::Listener* listener)
+  : listener(listener)
+{
+}
+
```
ibc: As told above, I suggest moving common methods to the CPP file so we can add `MS_TRACE()` and debugging logs if needed.

worker/src/RTC/KeyFrameRequestManager.cpp
```diff
@@ -0,0 +1,43 @@
+#define MS_CLASS "KeyFrameRequestManager"
+
+#include "RTC/KeyFrameRequestManager.hpp"
+#include "DepLibUV.hpp"
+#include "Logger.hpp"
+
+void RTC::KeyFrameRequestManager::KeyFrameNeeded(uint32_t ssrc)
+{
```
ibc: Missing `MS_TRACE()` (here and in all methods defined in the CPP file).

worker/src/RTC/KeyFrameRequestManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+void RTC::KeyFrameRequestManager::OnKeyFrameRequestTimeout(PendingKeyFrameInfo* pendingKeyFrameInfo)
+{
+	auto it = this->mapSsrcPendingKeyFrameInfo.find(pendingKeyFrameInfo->GetSsrc());
+
+	MS_ASSERT(it != this->mapSsrcPendingKeyFrameInfo.end(), "PendingKeyFrameInfo not present in the map")
+
+	delete pendingKeyFrameInfo;
+
+	this->mapSsrcPendingKeyFrameInfo.erase(it);
```
ibc: Shouldn't here call to `this->listener->OnKeyFrameNeeded(ssrc);`?

worker/test/src/RTC/TestKeyFrameRequestManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+			this->onKeyFrameNeededTimesCalled++;
+		}
+
+	public:
+		size_t onKeyFrameNeededTimesCalled{ 0 };
+	};
+
+	static TestKeyFrameRequestManagerListener listener;
+
+	SECTION("KeyFrame requested once, not received on time")
```
ibc: Can we reword "KeyFrame" to "key frame" everywhere in the text? This is how we name it in other places.

worker/test/src/RTC/TestKeyFrameRequestManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+		KeyFrameRequestManager keyFrameRequestManager(&listener);
+
+		keyFrameRequestManager.KeyFrameNeeded(1111);
+		keyFrameRequestManager.KeyFrameNeeded(1111);
+		keyFrameRequestManager.KeyFrameNeeded(1111);
+		keyFrameRequestManager.KeyFrameNeeded(1111);
+
+		DepLibUV::RunLoop();
+
+		REQUIRE(listener.onKeyFrameNeededTimesCalled == 2);
```
ibc: It would be simpler if `TestKeyFrameRequestManagerListener` had a `Reset()` method that set `onKeyFrameNeededTimesCalled` to 0, so each test does not depend on he previous one (so here we can verify that `onKeyFrameNeededTimesCalled == 1` instead of 2).

This is: calling `listener.Reset()` before each test.

worker/test/src/RTC/TestKeyFrameRequestManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+		KeyFrameRequestManager keyFrameRequestManager(&listener);
+
+		keyFrameRequestManager.KeyFrameNeeded(1111);
+		keyFrameRequestManager.KeyFrameReceived(1111);
+
+		DepLibUV::RunLoop();
+
+		REQUIRE(listener.onKeyFrameNeededTimesCalled == 3);
+	}
+}
```
ibc: Is this test testing the `OnKeyFrameRequestTimeout` case? Since such a callback does not even fire the public `OnKeyFrameNeeded()` I think it's not tested.

jmillan: It does, otherwise the libuv loop would not stop and the test would never finish.

worker/src/RTC/KeyFrameRequestManager.cpp
```diff
@@ -101,6 +101,9 @@ void RTC::KeyFrameRequestManager::OnKeyFrameRequestTimeout(PendingKeyFrameInfo*
 
 	MS_ASSERT(it != this->mapSsrcPendingKeyFrameInfo.end(), "PendingKeyFrameInfo not present in the map")
 
+	// Best effort in case the PLI/FIR was lost
```
ibc: dot dot dot

worker/src/RTC/KeyFrameRequestManager.cpp
```diff
@@ -101,6 +101,9 @@ void RTC::KeyFrameRequestManager::OnKeyFrameRequestTimeout(PendingKeyFrameInfo*
...
+	this->listener->OnKeyFrameNeeded(pendingKeyFrameInfo->GetSsrc());
```
ibc: NOTE: Better read the ssrc before into a variable and use it. And then invoke the listener at the end of the function. Why? Just imagine that the listener (Producer) decides to delete the `KeyFrameManager` within the `OnKeyFrameNeeded` cb. It would call the destructor which would delete all the `pendingKeyFrameInfos` (including this one) and the next line would crash.

It's safe to place all the calls to listener's callback at the end of the methods.

ibc:
The signature of the public listener callback does not match what we usually do. It's: 

```c++
OnKeyFrameNeeded(uint32_t ssrc)
```

It should include `this` as first argument, this is:

```c++
OnKeyFrameNeeded(KeyFrameRequestManager* keyFrameRequestManager, uint32_t ssrc)
```

It also makes the methods more readable once in the Producer.

PR #270, opened 2019-02-09
Merge RtpMonitor into RtpStream

ibc:
Before this PR:

* `RtpStream` has a `RtpMonitor`.
* Such a `RtpMonitor` is provided with the `RtpStream` itself. Not good.
  * So it needs class forward declaration in the hpp.
  * So it needs lot of public methods/getters in `RtpStream` just because the `RtpMonitor` needs to call them.
* `RtpStream` behaves as a wrapper between `Producer`/`Consumer` and `RtpMonitor`:
   * `GetScore()`
   * `OnScoreChange()` event.
* `RtpStream` has a `packetRepaired` numeric number, and `RtpMonitor` has a `repairedPackets` map. In order to fill it, the `RtpStream` must tell the `RtpMonitor` about "repaired packets" even if it's clear that both entities represent the very same.

In this PR:

* `RtpMonitor` no longer exists. It was just a "code split". All what the monitor does is 100% based on the parameters/data of the stream. Exposing it as an external class is just an artifact to split code into two files (plus the required wrappers for everything and public members).
* `RtpStreamRecv` and `RtpStreamSend` call now to `RtpStream::UpdateScore(receiverReport)` and the `UpdateScore()` methods does all score related stuff in a single method (no need to jump).
* Also, remove some useless private methods in `RtpStream` (such as `GetExpectedPackets()`) that were just called once.


ibc:
Sure? any consideration?

jmillan:
I find it OK, really.

PR #303, opened 2019-05-13
RtpProbator: use a different ssrc and sequential seq

ibc:
- Sending a packet with same ssrc and seq does not increase the bwe in  the receiver.
- By using a separate (random) ssrc and sequential seq, the remote reports terribly higher bwe values in REMB fb packets.
- Also make it possible to send up to 4 probation packets per original packet.

This can be tested as follows using the demo app:

* Unset `initialAvailableOutgoingBitrate` in `config.js`.
* Run tab A with VP8 simulcast and just let the spatial layer 0 enabled (simulcast "low" stream).
* In tab 2 run a consumer.
* Initial bwe is 600000 bps (default value) during 6-7 seconds.
* After it, it honors the last REMB sent by the consumer browser, which is ~320000 bps (**NOT GOOD**).

Repeat the test with this PR applied:

* After it, it honors the last REMB sent by the consumer browser, which is ~780000 bps (**YES!**).

### DIFF
--- a/worker/include/RTC/RtpProbator.hpp
+++ b/worker/include/RTC/RtpProbator.hpp
@@ -32,6 +32,8 @@ namespace RTC
 		uint32_t probationTargetBitrate{ 0 };
 		RTC::RtpDataCounter transmissionCounter;
 		RTC::RtpDataCounter probationTransmissionCounter;
+		uint32_t ssrc{ 0 };
+		uint16_t seq{ 0 };
 	};
 } // namespace RTC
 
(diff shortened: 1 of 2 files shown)


jmillan (approved):
Marvelous 👍

jmillan:
> Also make it possible to send up to 4 probation packets per original packet.

Those 4 are:

* The original one sent out of the probator
* The original one sent from the probator
* 2 packets with the probator ssrc and the corresponding sequence number

yes?

If the retransmission of the original one does not count for the REMB calculation, should we maybe just not send it from the probator?

https://github.com/versatica/mediasoup/pull/303/files#diff-6bcf83cdfcdf9c6466e9a169620a665dR126

ibc:
Let's close this and go for #304 instead.

Issue #306, opened 2019-05-20
Improve H264 temporal layer switching


ibc:
Once frame marking is implemented (#298) there are two pending subjects to improve/fix in `H264::PayloadDescriptorHandler::Process()`:

1. We should check incremental "picture id" somehow and let the packet pass if its picture id is older than highest seen. However no idea how to check it in H264 frame-marking.
   * NOTE: Better don't do it. We cannot check `tl0picidx` since many packets with different `tid` would have the same value.
2. Eventually (and this happens in all simulcast streams) there are no H264 packets with the `framingMarking->base` (Base Layer Sync) set, so we cannot wait for it in order to let higher temporal layers pass.
   * We may need [The Layer Refresh Request (LRR) RTCP Feedback](https://tools.ietf.org/html/draft-ietf-avtext-lrr) for it, but libwebrtc does not support it.
   * Issue reported in libwebrtc: https://bugs.chromium.org/p/webrtc/issues/detail?id=10651
   * This issue depends entirely on libwebrtc.

ibc:
Bullet 1 above may has been implemented in #308, but must be reviewed (see the PR details).

ibc:
Let's keep this issue open until libwebrtc people fixes the h264 [issue](https://bugs.chromium.org/p/webrtc/issues/detail?id=10651#c5).

ibc:
Confirmed that this is fixed in Chrome 78.0.3904.97 (OSX), so eventually we should uncomment [this block](https://github.com/versatica/mediasoup/blob/3.3.3/worker/src/RTC/Codecs/H264.cpp#L203). However, we must be sure that it would not affect older libwebrtc based devices.

PR #308, opened 2019-05-24
Improve codecs (related to #306)

ibc:
- VP8: Check that the payload has pictureId and tl0PictureIndex before using them.
- H264: Check tl0picidx is greater than highest seen (related to #306).

However, note that while in `VP8::PayloadDescriptorHandler::Encode()` we check whether `pictureId` is greater than highest seen, we don't have "picture Id" in H264, but just the `tl0picidx`. So I've implemented the same logic but by checking `tl0picidx`. I hope it's also valid.


ibc:
NOTE: It strongly worries me the fact that we have a `currentTemporalLayer` in both `SimulcastConsumer` and `EncodingContext` classes, and they are not in sync but independently managed.

jmillan:
> NOTE: It strongly worries me the fact that we have a currentTemporalLayer in both SimulcastConsumer and EncodingContext classes, and they are not in sync but independently managed.

Agreed, I think we could use the `EncodingContext` to set, get the current temporal layer from `SimulcastConsumer`

ibc:
> However, note that while in VP8::PayloadDescriptorHandler::Encode() we check whether pictureId is greater than highest seen, we don't have "picture Id" in H264, but just the tl0picidx. So I've implemented the same logic but by checking tl0picidx. I hope it's also valid.

And it was not valid so has been reverted. This was letting MOST of the packets pass. This is because many H264 packets with different `tid` have the same `tl0picidx`. So check removed.

PR #309, opened 2019-05-27
Skip worker compilation

RoCkAs:
Some systems do not have gcc >4.8 which is required, e.g. Centos. Thus `postinstall` of mediasoup fails. In order to avoid this, we can skip the worker compilation providing an environmental variable, like the `MEDIASOUP_WORKER_BIN` that is used in `mediasoup/lib/Worker.js`

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -13,18 +13,24 @@ MEDIASOUP_BUILDTYPE ?= Release
 	docker-build docker-run
 
 default:
-	$(PYTHON) ./scripts/configure.py -R mediasoup-worker
-	$(MAKE) BUILDTYPE=$(MEDIASOUP_BUILDTYPE) -C out
+	ifeq ($(MEDIASOUP_WORKER_BIN),)
+		$(PYTHON) ./scripts/configure.py -R mediasoup-worker
+		$(MAKE) BUILDTYPE=$(MEDIASOUP_BUILDTYPE) -C out
+	endif
 
 test:
-	$(PYTHON) ./scripts/configure.py -R mediasoup-worker-test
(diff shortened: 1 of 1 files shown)


ibc:
I really do not understand. You need GCC>4.8 anyway to build `mediasoup-worker`, so what your PR does is basically not compiling `mediasoup-worker` if `MEDIASOUP_WORKER_BIN` env variable is given during `npm install mediasoup`. Which solves nothing because later you must go into mediasoup sources and run `make` manually with GCC>4.8.

So I assume that, somehow, you **can** install GCC>4.8 in CentOS, and once done you can do:

```bash
$ CC=gcc-4.9 CPP=g++-4.9 npm install mediasoup
```

I also do not like the fact that you changed tabs with spaces instead of following the Makefile syntax when `ifeq` is used (such as in the `docker-build` task).

antoniom:
@ibc In our case gcc update cannot be performed because no root privileges are available on the machine. Moreover, the development is done behind a closed (not connected on the internet) network.

Instead we get a pre-compiled worker binary from another machine which has the appropriate gcc version and this is used on our app by setting the `MEDIASOUP_WORKER_BIN` env var.

We had a similar issue with [node-sass](https://github.com/sass/node-sass) and the way they solved it was to supply a `SASS_BINARY_DIR` as described in https://www.npmjs.com/package/node-sass#binary-configuration-parameters

ibc:
Ok. Can you please fix the indentation issue?

RoCkAs:
> Ok. Can you please fix the indentation issue?

Ok, i will fix it now.

ibc:
Merged, thanks.

ibc:
Will be available in mediasoup 3.0.11 (to be released in next days after some other changes are done).

PR #319, opened 2019-07-10
Remove unneeded ntohl in SctpAssociation

ggarber:


ibc:
Not agreed. Such a value is not converted to host order by usrsctp and we want to properly log it.

jmillan:
> Not agreed. Such a value is not converted to host order by usrsctp and we want to properly log it.

Correct, PPID (uint16_t) is not converted to host byte order by usrsctp.

ggarber:
@jmillan @ibc 
Then why are you assigning it to an uint8_t variable?  I'm confused:
`uint8_t ppid      = ntohl(rcv.rcv_ppid);`

Looking at the code it is possible that an older version of usrsctp used uint8 for ppid and latest version uses uint32.

jmillan:
`recv_ppid` is a `uint32_t` https://github.com/sctplab/usrsctp/blob/master/usrsctplib/usrsctp.h#L162

We are casting it to `uint8_t` in the line you point, which is wrong according to the data type, but covers all the [specified values](https://www.iana.org/assignments/sctp-parameters/sctp-parameters.xhtml)

There's typo `uint8_t/uint32_t` [here](https://github.com/versatica/mediasoup/pull/319/files#diff-010f14a142ae5e8db9b04e782881233eL55)

ibc:
In fact PPID is 32 bits long: https://tools.ietf.org/html/rfc4960#section-3.3.1

This must be fixed in many files in mediasoup (including `Router.hpp`, etc). It does not matter that there are IANA specified values. The application should be able to use `ppid = 65535` if desired as the meaning of PPID is up to the application.

jmillan:
> It does not matter that there are IANA specified values.

Of course, as I said it is a typo, or we can call it a bug

ibc:
yep, it's a bug.

jmillan:
fixing

jmillan:
Fixed, thanks for the point @ggarber!

PR #324, opened 2019-07-17
RateCalculator: improve efficiency

ibc:
* Store `this->lastTime` and `this->lastRate` when calling `GetRate(now)` and return `this->lastRate` if `GetRate(now)` is called with `now == this->lastTime` (instead of calculating the rate again).
* Reset both `this->lastTime` and `this->lastRate` to 0 when calling `Reset()` or `Update()` (to allow that immediate new calls to `GetRate()` with same `now` trigger a new rate calculation).
* Avoid unnecessary calls to `GetRate()` in `RemoteBitrateEstimatorAbsSendTime` (even if no longer necessary with the changes above, but looks better).

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "devDependencies": {
     "eslint": "^6.0.1",
-    "eslint-plugin-jest": "^22.9.0",
+    "eslint-plugin-jest": "^22.10.0",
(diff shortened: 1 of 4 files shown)


jmillan (approved):

jmillan:
:+1: LGTM

PR #329, opened 2019-08-29
win support

haiyangwu:
## Installation for Win
ensure your host satisfies the following **requirements**:
* Visual Studio 2015/2017
* Add the path of MSBuild.exe to System Environment Variable (eg. C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin)
* Add GYP_MSVS_VERSION to System Environment Variable (eg. 2017 for visual studio 2017 )

## key changes:
* using gulp shell and MSBuild to build project
* **using two pipes to replace the duplex pipe to avoid deadlock**, the related bug in NodeJS: https://github.com/nodejs/node/issues/29238

## supported scripts:
* lint
* format:worker
* test:node
* coverage:node
* postinstall

## not supported scripts:
*  test:worker (haven't found an alternative tool to replace lcov in windows, but running in visual studio is ok )

## not sure
* changes about channels and pipes


saghul:
>     * **using two pipes to replace the duplex pipe to avoid deadlock**, the related bug in NodeJS: [nodejs/node#29238](https://github.com/nodejs/node/issues/29238)

That is solved by passing the UV_OVERLAPPED_PIPE flag when creating the pipe. It's a noop on Unix. See: https://github.com/libuv/libuv/commit/62a0f763a7d1a5bbaa2670b9d2687ba183d2c528

.eslintrc.js
```diff
@@ -172,6 +172,7 @@ module.exports =
 		'spaced-comment': [ 2, 'always' ],
 		'strict': 2,
 		'valid-typeof': 2,
-		'yoda': 2
+		'yoda': 2,
+		"linebreak-style": 0
```
ibc: Please use single quotes here.

.gitignore
```diff
@@ -22,3 +22,9 @@
 
 # Mac Stuff.
 .DS_Store
+
+# Win Stuff
+configure.pyc
```
ibc: Where is this file generated? If it's generated in root folder (mediasoup folder) please indicate it with `/configure.pyc` instead.

.gitignore
```diff
@@ -22,3 +22,9 @@
...
+
+# VS stuff
+/worker/.vs/
```
ibc: Missing line break here.

gulpfile.js
```diff
@@ -1,5 +1,7 @@
 const gulp = require('gulp');
+const shell = require('gulp-shell');
 const clangFormat = require('gulp-clang-format');
+const os = require('os');
```
ibc: I don't like this approach much. In fact we already removed building stuff from `gulpfile.js` to `package.json` "scripts" time ago and I don't want to depend on `gulp` for the basic building process.

Take into account that this PR also makes "scripts" in `package.json` run `gulp` instead of make, so `gulp` should now be a "dependency" rather than a "devDependency" (otherwise `npm install mediasoup --production` will fail).

Can we figure out a different approach rather than depending on `gulp` for building on Windows? I really want that when executing `npm install mediasoup`, the "postinstall" script calls `make -C worker/` directly (at least in Linux/Mac). Can we do something in the "postinstall" script to make it do something different when in Windows? (without depending on `gulp` I mean).

lib/Channel.js
```diff
@@ -147,8 +148,11 @@ class Channel extends EnhancedEventEmitter
 			}
 		});
 
-		this._socket.on('end', () => this._logger.debug('Channel ended by the worker process'));
-		this._socket.on('error', (error) => this._logger.error('Channel error: %s', String(error)));
+		this._consumerSocket.on('end', () => this._logger.debug('Consumer channel ended by the worker process'));
```
ibc: Let's log "Consumer Channel ..." and "Producer Channel..." instead (here and below).

package.json
```diff
@@ -46,13 +43,18 @@
     "h264-profile-level-id": "^1.0.0",
     "netstring": "^0.3.0",
     "random-number": "^0.0.9",
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
+    "gulp": "^4.0.2",
+    "gulp-clang-format": "^1.0.27",    
+    "gulp-eslint": "^6.0.0",
```
ibc: As far as I see we do not need `gulp-eslint`, `gulp-rename` and `gulp-touch-cmd` at all. They are not used anywhere in this PR.

Also, `gulp-clang-format` must remain in "devDependencies".

Regarding `gulp` in "dependencies" let's discuss the topic about gulpfile as mentioned above.

test/test-Worker.js
```diff
@@ -3,6 +3,7 @@ const { toBeType } = require('jest-tobetype');
 const mediasoup = require('../');
 const { createWorker, observer } = mediasoup;
 const { InvalidStateError } = require('../lib/errors');
+const os = require('os');
```
ibc: Let's move system/core modules to the top, this, to the first line of this file.

test/test-Worker.js
```diff
@@ -194,6 +195,11 @@ test('Worker emits "died" if worker process died unexpectedly', async () =>
 
 test('worker process ignores PIPE, HUP, ALRM, USR1 and USR2 signals', async () =>
 {
+	if (os.platform() === 'win32') 
+	{
+		return;
+	}
```
ibc: Why this?

BTW: Let's not use `{` `}` when the `if` condition takes a single line:

```js
if (os.platform() === 'win32') 
  return;
```

worker/include/Channel/Request.hpp
```diff
@@ -67,7 +67,7 @@ namespace Channel
 		static std::unordered_map<std::string, MethodId> string2MethodId;
 
 	public:
-		Request(Channel::UnixStreamSocket* channel, json& jsonRequest);
+		Request(json& jsonRequest);
```
ibc: The constructor declaration must have `explicit` keyword in front (since it has a single argument).

worker/include/Channel/Request.hpp
```diff
@@ -76,15 +76,18 @@ namespace Channel
...
 		MethodId methodId;
 		json internal;
 		json data;
 		// Others.
 		bool replied{ false };
+
+	public:
+		static void ClassInit(Channel::UnixStreamSocket* channel);
+		// Passed by argument.
+		static Channel::UnixStreamSocket* channel;
```
ibc: Why has this been converted into a `static` member? Also, it's no longer "passed by argument", right?

If so, it must be separated from the static methods and added into a new `public:` group below just for variable static members. However, first I need to understand why it's now a `static` member.

worker/include/Channel/UnixStreamSocket.hpp
```diff
@@ -38,6 +38,19 @@ namespace Channel
 		// Others.
 		size_t msgStart{ 0 }; // Where the latest message starts.
 	};
+
+	class ChannelWrapper
+	{
+	public:
+		explicit ChannelWrapper(int consumerFd, int producerFd)
```
ibc: IMHO this must not be `explicit` (no need for it since it has two arguments).

worker/include/Channel/UnixStreamSocket.hpp
```diff
@@ -38,6 +38,19 @@ namespace Channel
...
+		  : consumerSocket(consumerFd, ::UnixStreamSocket::SocketRole::CONSUMER),
+		    producerSocket(producerFd, ::UnixStreamSocket::SocketRole::PRODUCER)
+		{
+		}
+
```
ibc, line 48: Let's add a `public` line here to separate methods and members.

worker/include/RTC/RTCP/FeedbackPsFir.hpp
```diff
@@ -27,7 +27,12 @@ namespace RTC
 			{
 				uint32_t ssrc;
 				uint8_t sequenceNumber;
+#ifdef _MSC_VER
+				// alignment
+				uint8_t reserved[3];
```
ibc: Why is this required in Windows? (just wondering). And, if so, is alignment just required in this file?

BTW: Please always capitalize comments, add a dot at the end, and put them in the same line if possible:

```c++
uint8_t reserved[3]; // Alignment.
```

worker/include/handles/UnixStreamSocket.hpp
```diff
@@ -15,8 +15,14 @@ class UnixStreamSocket
 		uint8_t store[1];
 	};
 
+	enum SocketRole
+	{
+		CONSUMER,
+		PRODUCER
+	};
```
ibc, line 22: We always use `class enum` when possible. Can it be moved to this?

```c++
enum class Role
{
	PRODUCER = 1,
	CONSUMER
};
```

Note that I've also renamed it from `SocketRole` to just `Role`.  Being a `class enum` you need to later reference it as `UnixStreamSocket::Role`.

worker/include/opt/getopt.h
```diff
@@ -0,0 +1,101 @@
...
+	extern int
+		getopt(int argc,
+			char *const *argv,
+			const char *optstring);
+
+	#ifdef __cplusplus
+}
+#endif 
+
+#endif /* _GETOPT_H */
```
ibc: I'd like that `getopt` is moved to `worker/deps` which is corresponding entry in `worker/scripts/get-dep.sh`.

And then, we should create a `getopt.gyp` and place it in `worker/deps/getopt/`, and have a separate folder `worker/deps/getopt/xxxxxx` with the external getopt GIT repository.

This is easy:

* Take as example the [netstring.gyp](https://github.com/versatica/mediasoup/tree/v3/worker/deps/netstring) we wrote.
* And add [some getopt](https://github.com/kimgr/getopt_port) GIT repository to `get-dep.sh`.

And then, we need to make `mediasoup-worker.gyp` to just use `getopt.gyp` in "dependencies" when in Windows. I hope that's possible with some condition.

worker/mediasoup-worker.gyp
```diff
@@ -271,6 +271,13 @@
         'ldflags': [ '-Wl,--export-dynamic' ]
       }],
 
+      [ 'OS == "win"', {
+        'sources': [
+          'include/opt/getopt.h',
+          'opt/getopt.cpp'
+        ]
+      }],
```
ibc, line 290: As told above, I'd prefer that  `getopt` is under `worker/deps` as many other dependencies.

worker/mediasoup-worker.gyp
```diff
@@ -298,7 +305,13 @@
       [
         # C++ source files.
         'src/main.cpp'
-      ]
+      ],
+      # handle multi files with same name
```
ibc: "Handle multi files with same name."

:)

worker/mediasoup-worker.gyp
```diff
@@ -356,6 +369,12 @@
         'OTHER_LDFLAGS': [
           '--coverage'
         ]
+      },
+      # handle multi files with same name
```
ibc: Same.

worker/mediasoup-worker.gyp
```diff
@@ -419,6 +438,13 @@
       [
         'fuzzer/include'
       ],
+      # handle multi files with same name
```
ibc: Same.

worker/scripts/configure.py
```diff
@@ -85,6 +85,11 @@ def run_gyp(args):
     args.append('-Dclang=%d' % int(is_clang))
     if is_clang == False and major == 4 and minor <= 8:
       raise RuntimeError('gcc <= 4.8 not supported, please upgrade your gcc')
+  else:
+    if 'ninja' not in args:
+      args.extend(['-Goutput_dir=' + output_dir])
+      args.extend(['--generator-output', output_dir])
+
```
ibc: Avoid two empty lines together.

worker/src/Channel/Request.cpp
```diff
@@ -57,9 +57,22 @@ namespace Channel
 	};
 	// clang-format on
 
+	/* Class variables. */
+
+	Channel::UnixStreamSocket* Request::channel{ nullptr };
```
ibc: As told before, I need to understand why we need to make `Request::channel` static for this PR to work.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1335,6 +1335,10 @@ namespace RTC
 		  srtpRemoteMasterKey,
 		  srtpMasterLength,
 		  this->remoteCert);
+
+		delete[] srtpMaterial;
+		delete[] srtpLocalMasterKey;
+		delete[] srtpRemoteMasterKey;
```
ibc: Amazing catch! so we were leaking right???

ibc: Maybe we should use

```c++
uint8_t srtpMaterial[srtpMasterLength * 2];
```

instead of:

```c++
uint8_t* srtpMaterial = new uint8_t[srtpMasterLength * 2];
```

?

worker/src/Settings.cpp
```diff
@@ -10,9 +10,13 @@
 #include <cerrno>
 #include <iterator> // std::ostream_iterator
 #include <sstream>  // std::ostringstream
+#ifdef _MSC_VER
+#include "opt/getopt.h"
```
ibc: Once in `worker/deps/getopt/` it should be `#include <getopt.h>`.

worker/src/Settings.cpp
```diff
@@ -10,9 +10,13 @@
...
+#else
 extern "C" {
 #include <getopt.h>
 }
+#endif // _MSC_VER
```
ibc: So we can remove the if else here, since it would always be `#include <getopt.h>`:

* In Linux/Mac it will take the system header since `getopt` is not included in gyp "dependencies".
* In Windows it will take the `getopt` in deps/ folder since `getopt` is included in gyp "dependencies".

worker/src/Worker.cpp
```diff
@@ -118,10 +119,13 @@ RTC::Router* Worker::GetRouterFromRequest(Channel::Request* request) const
 	return router;
 }
 
-inline void Worker::OnChannelRequest(Channel::UnixStreamSocket* /*channel*/, Channel::Request* request)
+inline void Worker::OnChannelRequest(Channel::UnixStreamSocket* socket, Channel::Request* request)
 {
 	MS_TRACE();
 
+	MS_ASSERT(
+	  socket != &this->channel->producerSocket, "Producer Socket should not used as request channel.");
```
ibc: Remove dot at the end in logs.

worker/src/handles/TcpConnection.cpp
```diff
@@ -332,7 +332,7 @@ void TcpConnection::Write(const uint8_t* data1, size_t len1, const uint8_t* data
 
 void TcpConnection::ErrorReceiving()
 {
-	MS_ERROR();
+	MS_ERROR("");
```
ibc: Please, replace this with `MS_TRACE()`. It was a typo.

worker/test/include/helpers.hpp
```diff
@@ -10,6 +10,11 @@ namespace helpers
 	inline bool readBinaryFile(const char* file, uint8_t* buffer, size_t* len)
 	{
 		std::string filePath = "test/" + std::string(file);
+#ifdef _MSC_VER
+		// visual studio solution and project files generated in out dir. so resources nav to parent dir.
```
ibc: Rewrite to:

> // Visual Studio solution and project files generated in out dir so resources nav to parent dir.

jmillan:
looks great @haiyangwu,

>> * **using two pipes to replace the duplex pipe to avoid deadlock**, the related bug in NodeJS: 
>>[nodejs/node#29238](https://github.com/nodejs/node/issues/29238)

>That is solved by passing the UV_OVERLAPPED_PIPE flag when creating the pipe. It's a noop on >Unix. See: libuv/libuv@62a0f76

Could you please try the UV_OVERLAPPED_PIPE flag and adapt the PR accordingly?

ibc (changes requested):

ibc:
It's not enough. The bug is also in Node usage of libuv.

worker/include/RTC/RTCP/FeedbackPsFir.hpp
```diff
@@ -27,7 +27,12 @@ namespace RTC
 			{
 				uint32_t ssrc;
 				uint8_t sequenceNumber;
+#ifdef _MSC_VER
```
saghul: Is this a compiler specific requirement? If not I'd suggest using `ifdef _WIN32` for Windows code. It's not impossible the compiler changes and grows native supoport for Clang on Windows.

worker/src/RTC/DtlsTransport.cpp
ibc, line 1402: Sorry, I don't understand, it seems that there was no leak at all. It happens that you have replaced:

```c++
uint8_t srtpMaterial[srtpMasterLength * 2];
```

with this:

```c++
uint8_t* srtpMaterial = new uint8_t[srtpMasterLength * 2];
// [...]
delete[] srtpMaterial;
```

Why? is it needed by Visual Studio? Doesn't it allocate arrays of variable length in the stack?

**UPDATE:** Yes, it seems that Visual Studio compiler does not allow variable-length C arrays: https://stackoverflow.com/questions/5246900/enabling-vlas-variable-length-arrays-in-ms-visual-c

ibc (changes requested):

worker/include/opt/getopt.h
ibc: It should be like this in `mediasoup-worker.gyp`, within the `'conditions'` block:

```gyp
# Platform-specifics.

[ 'OS == "win"', {
  'dependencies': [ 'deps/getopt/getopt.gyp:getopt' ]
}],
```

ibc (changes requested):

.eslintrc.js
haiyangwu: yep, already changed.

.gitignore
haiyangwu: generated by python. the path is `/worker/scripts/configure.pyc`. already add explicit path.

.gitignore
haiyangwu: yep, already added.

gulpfile.js
haiyangwu: agree. I'll have a try.

lib/Channel.js
haiyangwu: yep. already changed.

package.json
haiyangwu: unused deps already removed.

test/test-Worker.js
haiyangwu: agree.

test/test-Worker.js
haiyangwu: Windowns has no `SIGPIPE/SIGALRM/SIGUSR1/SIGUSR2` signals,  and haven't found alternatives for `sigfillset` and `sigaction`. in `/worker/src/main.cc#IgnoreSignals` already use `_WIN32` marco to skip these logic code. so  in test, also skip.

worker/include/Channel/Request.hpp
haiyangwu: agree. already add `explicit` keyword.

worker/include/Channel/Request.hpp
haiyangwu: now that we use two separate pipe. After getting request from **Consumer Channel**,  we'll send response ( accept or reject) to JS which should use **Producer Channel**, but **Consumer Channel* has not relation to **Producer Channel**, so use a `static` member store the  **Producer Channel** just like` Channel::Notifier`.

worker/include/Channel/UnixStreamSocket.hpp
haiyangwu: already removed.

worker/include/Channel/UnixStreamSocket.hpp
haiyangwu: yep. already added.

worker/include/RTC/RTCP/FeedbackPsFir.hpp
haiyangwu: visual studio seems do not support alignment like this `uint32_t reserved : 24;`, so just use `uint8_t reserved[3];` here.

worker/include/handles/UnixStreamSocket.hpp
haiyangwu: agree. already changed.

worker/include/opt/getopt.h
haiyangwu: agree. already move `getopt` into deps.

worker/mediasoup-worker.gyp
haiyangwu: already move getopt into deps.

worker/mediasoup-worker.gyp
haiyangwu: already add ` .`. 😄

worker/mediasoup-worker.gyp
haiyangwu: already add ` .`. 😄

worker/mediasoup-worker.gyp
haiyangwu: already add ` .`. 😄

worker/src/RTC/DtlsTransport.cpp
haiyangwu: yep.  just adapte to Visual Studio compiler.

worker/src/Settings.cpp
haiyangwu: already changed.

worker/src/Settings.cpp
haiyangwu: already changed.

worker/src/Worker.cpp
haiyangwu: yep. removed.

worker/include/RTC/RTCP/FeedbackPsFir.hpp
haiyangwu: for better compatibility use `_WIN32`, already renamed.

worker/test/include/helpers.hpp
haiyangwu: ok. already configured.

gulpfile.js
haiyangwu: whether https://www.npmjs.com/package/run-script-os is a better solution ?

ibc: **run-script-os** looks indeed a perfect solution :)

NOTE: It must be added into "dependencies" since it's required for "postinstall".

gulpfile.js
```diff
@@ -1,5 +1,7 @@
...
```
ibc: Let's move `const os` to the top since it's a Node core module.

package.json
```diff
@@ -46,13 +43,14 @@
     "h264-profile-level-id": "^1.0.0",
     "netstring": "^0.3.0",
     "random-number": "^0.0.9",
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
+    "gulp": "^4.0.2",
+    "gulp-clang-format": "^1.0.27",
```
ibc: "gulp-clang-format" must be in "devDependencies" (as it was).

test/test-Worker.js
```diff
@@ -194,6 +195,9 @@ test('Worker emits "died" if worker process died unexpectedly', async () =>
 
 test('worker process ignores PIPE, HUP, ALRM, USR1 and USR2 signals', async () =>
 {
+	if (os.platform() === 'win32')
```
ibc, line 200: Please, add a comment here explaining why such a `return` in case of Windows.

worker/deps/getopt/getopt.gyp
```diff
@@ -0,0 +1,18 @@
...
+        'getopt.h',
+        'getopt.cpp'
+      ],
+      'direct_dependent_settings':
+      {
+        'include_dirs': ['.']
+      },
+    }
+  ]
+}
```
ibc: Add line break at the end of all files :)

worker/include/Channel/Request.hpp
ibc: I understand, however I'd prefer to avoid static members as far as possible. I've seen that you have created a `ChannelWrapper` class which holds both the Producer Channel and Consumer Channel. Perhaps we could manage a single instance of `ChannelWrapper` and pass it to other classes.

worker/src/Channel/Request.cpp
ibc: (let's discuss this topic on the comment above)

ibc (changes requested):

worker/src/handles/TcpConnection.cpp
ibc: Don't forget to replace it, please :)

ibc (changes requested):

worker/deps/getopt/getopt.gyp
ibc: NOTE: You have added all `getopt` dep into `worker/deps/getopt/` folder, however there is no "getopt" entry in `worker/scripts/get-dep.sh`. We prefer to have all deps with their corresponding entry in `get-dep.sh` pointing to a GIT repository and specific branch/version.

Instead of copying `getopt` files manually, is it possible to use some GIT repo instead?

And, if so, since such a repo probably won't have any `getopt.gyp`, that's why I told about creating a main `worker/deps/getopt/` folder as follows:

```
worker/
  deps/
    getopt/
      getopt.gyp  <-- The one you have created (must change paths).
      getopt/     <-- The result of git clone (by scripts/get-dep.sh).
        getopt.c
        getopt.h
```

ibc (changes requested):

ibc:
@haiyangwu, this is a great PR and I like much how you are doing it. There are some minor changes I've requested above. Other than those, I want to take a double look to the Channel, Producer/Consumer Channel and ChannelWrapper stuff (I want to avoid having a static member).

I'll come back to this on next Monday. Thanks a lot for your nice contribution.

worker/deps/getopt/getopt.gyp
haiyangwu: Got. updated.

worker/src/handles/TcpConnection.cpp
haiyangwu: already merged the  'fix typo' commit from current v3 branch.

gulpfile.js
haiyangwu: yep. updated.

test/test-Worker.js
haiyangwu: yep. a comment added already.

package.json
haiyangwu: yep.  "gulp-clang-format" already moved into "devDependencies".

gulpfile.js
haiyangwu: oops, **run-script-os** seems ok, but can't get `process.env.MEDIASOUP_BUILDTYPE ` which used for judging to build a Debug or Release version.  I'll still work on it.

ibc: Oh, may be a bug in run-script-os (it overrides env variables)?

worker/scripts/get-dep.sh
```diff
@@ -197,6 +206,9 @@ case "${DEP}" in
 	fuzzer-corpora)
 		get_fuzzer_corpora
 		;;
+	get-win-opt)
```
ibc: Please rename to just `win-getopt)`.

worker/scripts/get-dep.sh
```diff
@@ -158,6 +158,15 @@ function get_fuzzer_corpora()
...
+	GIT_REPO="git@github.com:alex85k/wingetopt.git"
+	GIT_TAG="master"
+	DEST="deps/getopt/getopt"
+
+	get_dep "${GIT_REPO}" "${GIT_TAG}" "${DEST}"
+}
+
 case "${DEP}" in
 	'-h')
 		echo "Usage:"
```
ibc, line 185: Append `|win-getopt"` to the next echo usage line.

ibc (changes requested):

gulpfile.js
ibc: No. By looking at the [source code](https://github.com/charlesguse/run-script-os/blob/master/index.js) I see no reason for  env variables not to be bypassed. It calls `spawn()` without specifying `env` so, according to the [doc](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options), `process.env` should be used.

ibc: `process.env.MEDIASOUP_BUILDTYPE` may exist or not. If not, "Release" is assumed. Otherwise it may be set to "Debug".

Regarding run-script-os, I've created the following simple use case and it seems that env variables are properly propagated:

```json
{
  "name": "foo",
  "version": "1.0.0",
  "dependencies": {
    "run-script-os": "^1.0.7"
  },
  "scripts": {
  	"foo": "echo \"--- FOO:${FOO}\"",
  	"test": "run-script-os",
  	"test:darwin": "echo \"--- darwin FOO:${FOO}\"",
  	"test:linux": "echo \"--- linux FOO:${FOO}\""
  }
}
```

and it works fine:

```bash
$ FOO=1234 npm test

> kk@1.0.0 test /private/tmp/kk
> run-script-os


> kk@1.0.0 test:darwin /private/tmp/kk
> echo "--- darwin FOO:${FOO}"

--- darwin FOO:1234
```

```bash
$ FOO=1234 npm run foo

> kk@1.0.0 foo /private/tmp/kk
> echo "--- FOO:${FOO}"

--- FOO:1234
```

worker/scripts/get-dep.sh
haiyangwu: yep. updated.

worker/scripts/get-dep.sh
haiyangwu: yep. updated.

gulpfile.js
haiyangwu: already updated. but had to use a fork version, this version solves an [issue](https://github.com/charlesguse/run-script-os/issues/20) which caused error after run 'npm install' in windows platform. have better solution?

package.json
```diff
@@ -28,10 +28,14 @@
     "lint:worker": "gulp lint:worker",
     "format:worker": "gulp format:worker",
     "test": "npm run test:node && npm run test:worker",
-    "test:node": "gulp build && jest",
-    "test:worker": "gulp test:worker",
-    "coverage:node": "gulp build && jest --coverage && open-cli coverage/lcov-report/index.html",
-    "postinstall": "gulp build"
+    "test:node": "npm run postinstall && jest",
```
ibc, line 35: Why this new "npm run postinstall" here? I prefer to leave it as it was, without referencing another "script":

```json
"test:node": "make -C worker && jest",
```

package.json
```diff
@@ -28,10 +28,14 @@
...
+    "test:worker": "run-script-os",
+    "test:worker:win32": "echo test:worker not support yet!!! run test in visual studio!!!",
+    "test:worker:default": "make test -C worker",
+    "coverage:node": "npm run postinstall && jest --coverage && open-cli coverage/lcov-report/index.html",
+    "postinstall": "run-script-os",
+    "postinstall:win32": "python ./worker/scripts/configure.py --format=msvs && IF %MEDIASOUP_BUILDTYPE% == Release ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Release -t:mediasoup-worker ) ELSE ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Debug -t:mediasoup-worker )",
```
ibc: The problem here is that this line is trying to replace the whole `make` task in `worker/Makefile` which does much more that this. For instance, `make` allows using a custom `PYTHON` env variable to choose the preferred python executable. It also allows setting `MEDIASOUP_WORKER_BIN` env variable (for development purposes):

https://github.com/versatica/mediasoup/blob/v3/worker/Makefile#L20

May be we could create a separate shell/SuperShell/whatever script in `worker/` folder just for Windows and call it from `"potinstall:win32"? Something like:

```json
"postinstall:win32": "./worker/win-tasks.sh make"
```

(no idea about the script extension). Such a script (which is just for Windows) should take care of the PYTHON,  MEDIASOUP_BUILDTYPE  and MEDIASOUP_WORKER_BIN environment variables (if set) or use default values. It should also read the single argument it's called with (in this case "make"), do a `switch` internally and run the corresponding code. This is, something like a replacement for Windows of the `worker/Makefile` file.

Does this make sense?

gulpfile.js
ibc: Oh, sad. Ok, honestly I don't want that mediasoup depends on GIT. I've commented in the issue you referenced asking the author for a new version. Not sure if this will happen, so here a different proposal:

What about if we have a `worker/scripts/npm-script-os.js` (a single file hopefully) with the fixed code of that lib (mentioning the author, of course) and use it in "scripts"?:

```json
  "postinstall": "$NODE ./worker/scripts/npm-script-os.js",
  "postinstall:win32": "./worker/win-tasks.sh make",
  "postinstall:default": "make -C worker"
```

NOTE: Such a `./worker/scripts/npm-script-os.js` script should be called with `$NODE` instead of just `node`. As the [doc](https://docs.npmjs.com/cli/run-script) says:

> npm run sets the NODE environment variable to the node executable with which npm is executed.

ibc (changes requested):
Good work, thanks a lot. So there are two remaining issues: the issue in `run-script-os` and the loss of features when calling `npm install` in Windows. I've written proposals for both. Let me know your thoughts.

ibc:
BTW all tests are failing in Travis: https://travis-ci.com/versatica/mediasoup/jobs/229954211

> "spawn /home/travis/build/versatica/mediasoup/worker/out/Release/mediasoup-worker ENOENT"

And also this (confirmed that it fails in OSX too):

```
ibc at ibc-macbook in /tmp/mediasoup-win on v3
$ npm run test

> mediasoup@3.2.5 test /private/tmp/mediasoup-win
> npm run test:node && npm run test:worker

> mediasoup@3.2.5 test:node /private/tmp/mediasoup-win
> npm run postinstall && jest

> mediasoup@3.2.5 postinstall /private/tmp/mediasoup-win
> run-script-os

run-script-os was unable to execute the script 'undefined'
```

This seems to happen because "postinstall:default" script is not called when doing `npm install` into mediasoup folder. May be this is the issue you referenced above?

gulpfile.js
kawatzaki: Hi guys! I was kinda just lurking around a little bit and found you're having issues with run-script-os. I had the same issue, so I made the fork.

There's also an npm package under my profile so you can add the dependency based on that...
https://www.npmjs.com/package/@dr.amaton/run-script-os

Good luck!

ibc: Thanks @kawatzaki, that's great!

ibc: I'm afraid that the same issue happens (or at least one of the issues I told about above). Basically calling `npm run xxxxxx` within a "scripts" item that uses `npm-script-os` does not work (it does nothing):

```json
"scripts": {
    "test": "npm run test:node && npm run test:worker",
    "test:node": "npm run postinstall && jest",
    "test:worker": "run-script-os",
    "test:worker:win32": "echo test:worker not support yet!!! run test in visual studio!!!",
    "test:worker:default": "make test -C worker",
    "postinstall": "run-script-os",
    "postinstall:win32": "python ./worker/scripts/configure.py --format=msvs && IF %MEDIASOUP_BUILDTYPE% == Release ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Release -t:mediasoup-worker ) ELSE ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Debug -t:mediasoup-worker )",
    "postinstall:default": "make -C worker"
},
```

Here, if I call `npm test` or `npm run test` this happens:

```
ibc at ibc-macbook in /tmp/mediasoup-win on v3*
$ npm test

> mediasoup@3.2.5 test /private/tmp/mediasoup-win
> npm run test:node && npm run test:worker

> mediasoup@3.2.5 test:node /private/tmp/mediasoup-win
> npm run postinstall && jest


> mediasoup@3.2.5 postinstall /private/tmp/mediasoup-win
> run-script-os

run-script-os was unable to execute the script 'undefined'
```

As you can see, when it calls `npm run postinstall` it does nothing. It is supposed to run `npm-script-os` so it should run the "postinstall:default" script, but it does not.

ibc: Thanks @kawatzaki.

However I'm afraid that the same issue happens (or at least one of the issues I told about above). Basically calling `npm run xxxxxx` within a "scripts" item that uses `npm-script-os` does not work (it does nothing):

```json
"scripts": {
    "test": "npm run test:node && npm run test:worker",
    "test:node": "npm run postinstall && jest",
    "test:worker": "run-script-os",
    "test:worker:win32": "echo test:worker not support yet!!! run test in visual studio!!!",
    "test:worker:default": "make test -C worker",
    "postinstall": "run-script-os",
    "postinstall:win32": "python ./worker/scripts/configure.py --format=msvs && IF %MEDIASOUP_BUILDTYPE% == Release ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Release -t:mediasoup-worker ) ELSE ( MSBuild ./worker/mediasoup-worker.sln /p:Configuration=Debug -t:mediasoup-worker )",
    "postinstall:default": "make -C worker"
},
```

Here, if I call `npm test` or `npm run test` this happens:

```
ibc at ibc-macbook in /tmp/mediasoup-win on v3*
$ npm test

> mediasoup@3.2.5 test /private/tmp/mediasoup-win
> npm run test:node && npm run test:worker

> mediasoup@3.2.5 test:node /private/tmp/mediasoup-win
> npm run postinstall && jest

> mediasoup@3.2.5 postinstall /private/tmp/mediasoup-win
> run-script-os

run-script-os was unable to execute the script 'undefined'
```

As you can see, when it calls `npm run postinstall` it does nothing. It is supposed to run `npm-script-os` so it should run the "postinstall:default" script, but it does not.

kawatzaki: Oh I see. Ok, let me check on that and I'll see if I can reproduce the issue. I'll add it to my fork repo, be sure to add it to your watch list.

ibc: Thanks!

ibc: How to track your repo? Your NPM package does not specify any GitHub (or similar) repository.

kawatzaki: I'm sorry about that, yeah, here's the repo:

https://github.com/kawatzaki/run-script-os

Anyway, I think the issue is that the 'default' script was not being handled, I fixed it and will be updating the npm package soon enough

kawatzaki: yup, npm package update to version 1.1.2 please let me know if that fixed it up, if it didn't you can make an issue on my repo and we'll continue from there!

package.json
ibc: The same in "coverage:node" script. Let's use "make -C worker && jest ..." instead of "npm run postinstall && jest ...".

ibc: Ah ok, I see. If we use "make -C worker" then it won't work in Windows...

ibc: Ok, let's use "npm run postinstall" then.

gulpfile.js
ibc: Confirmed that it works now, @kawatzaki. Thanks!

kawatzaki: No problemo, glad I could help!

haiyangwu: Confirmed, it works. @kawatzaki 👍

package.json
haiyangwu: do not make much sense for me. already test both python2 and python3, those two work fine. **MEDIASOUP_BUILDTYPE** already checked in `postinstall:win32`. **MEDIASOUP_WORKER_BIN** is checked to skip some tasks if my understanding is right.

haiyangwu: If we want to replace all `make` tasks,  create a `./worker/win-make.js` seems a better solution. Shell script's syntax is complicated for me 😅.

ibc: You are right, if `MEDIASOUP_WORKER_BIN` is set, Makefile tasks are skipped.

Regarding python 2 or 3, yes, it may work fine. The thing is that it can not be chosen by the user. In Linux/OSX the user can set `PYTHON=xxxx` env variable and then run `npm install` or `make` and `xxxx` will be used as python executable. That's not possible in Windows with the above solution.

> If we want to replace all `make` tasks,  create a `./worker/win-make.js` seems a better solution. Shell script's syntax is complicated for me 😅.

Oh, sure. I just don't know how it must be, but of course it would be cool to have this:

```json
"postinstall:win32": "$NODE ./worker/win-tasks.js make"
```

:)

ibc:
BTW travis-ci tasks pass again. This is due the usage of `"@dr.amaton/run-script-os": "^1.1.2"` :)

package.json
haiyangwu: already had [a draft version ](https://github.com/haiyangwu/mediasoup-win/commit/7b2053498b1ea681f188d79613072c844428e51f), and the postinstall script is
````
"postinstall:win32": "\"\"%NODE%\"\" ./worker/win-tasks.js make"
````

but these two tasks `bear` and `tidy` used for what? I can't get the exact point. 😅

Comment (ibc) [reply package.json]: Looks nice! I won't be able to properly review it until next week. For now just a few considerations:

* Forget about `bear`, `tidy` and other Makefile tasks. All them are [documented](https://github.com/versatica/mediasoup/blob/v3/doc/Building.md#makefile) but they must not be added to `win-tasks.js` (they are for development purposes).
* The style of the JS does not conform to mediasoup eslint rules. Try adding it to package.json:
```
"lint:node": "eslint -c .eslintrc.js gulpfile.js lib test worker/win-tasks.js",
```
* Let's not use "use strict". It's already implicit (it should produce an eslint warning in fact).

Comment (haiyangwu) [reply package.json]: @ibc updated.

File: test/test-Worker.js [hunk 37]
```diff
@@ -194,6 +195,11 @@ test('Worker emits "died" if worker process died unexpectedly', async () =>
 
 test('worker process ignores PIPE, HUP, ALRM, USR1 and USR2 signals', async () =>
 {
+	// windowns doesn't have some signals such as SIGPIPE, SIGALRM, SIGUSR1, SIGUSR2
+	// so we just skip this test in windows
+	if (os.platform() === 'win32')
```
ibc, line 200: Typo: "windowns" :)

Please, let start comments always with capital letter ("Windows") and end with dot ("in Windows"). Also use capital letter for words like Windows.

.gitignore
```diff
@@ -22,3 +26,13 @@
 
 # Mac Stuff.
 .DS_Store
+
+# python win generated Stuff.
```
ibc: Write "Python generated stuff in Windows." instead, please.

.gitignore
```diff
@@ -22,3 +26,13 @@
...
+/worker/scripts/configure.pyc
+
+# vistual studio generated Stuff.
```
ibc: Write "Visual Studio" instead, please.

.gitignore
```diff
@@ -22,3 +26,13 @@
...
+/worker/**/Debug/
+/worker/**/Release/
+
+# VS stuff.
```
ibc: Write "Visual Studio" instead, please.

.eslintrc.js
```diff
@@ -172,6 +172,7 @@ module.exports =
 		'spaced-comment': [ 2, 'always' ],
 		'strict': 2,
 		'valid-typeof': 2,
-		'yoda': 2
+		'yoda': 2,
+		'linebreak-style': 0
```
ibc: Please remove this. There is already a `'linebreak-style': [ 2, 'unix' ]` line above that must be respected. All the source code must use Unix style line-breaks. This is: `LF` instead of `CRLF`.

And ensure please `npm run lint` works after removing such a rule.

gulpfile.js
```diff
@@ -36,4 +36,4 @@ gulp.task('format:worker', () =>
 	return gulp.src(src, { base: '.' })
 		.pipe(clangFormat.format('file'))
 		.pipe(gulp.dest('.'));
-});
+});
```
ibc: Missing line jump at the end of the file. In fact, lint task should fail due to missing empty line:

`'no-multiple-empty-lines': [ 1, { max: 1, maxEOF: 0, maxBOF: 0 } ]`

worker/include/Channel/Request.hpp
ibc: Can we work on this?

worker/mediasoup-worker.gyp
```diff
@@ -271,6 +271,23 @@
         'ldflags': [ '-Wl,--export-dynamic' ]
       }],
 
+      [ 'OS == "win"', {
+        'dependencies': [ 'deps/getopt/getopt.gyp:getopt' ],
+        
+        # handle multi files with same name.
```
ibc: "Handle" please (capital initial letter).

worker/src/Worker.cpp
```diff
@@ -118,10 +119,13 @@ RTC::Router* Worker::GetRouterFromRequest(Channel::Request* request) const
 	return router;
 }
 
-inline void Worker::OnChannelRequest(Channel::UnixStreamSocket* /*channel*/, Channel::Request* request)
+inline void Worker::OnChannelRequest(Channel::UnixStreamSocket* socket, Channel::Request* request)
 {
 	MS_TRACE();
 
+	MS_ASSERT(
+	  socket != &this->channel->producerSocket, "Producer Socket should not used as request channel");
```
ibc: Let's write it as `&(this->channel->producerSocket)` for readability, please.

worker/src/main.cpp
```diff
@@ -57,7 +62,7 @@ int main(int argc, char* argv[])
 	}
 
 	// Initialize the Logger.
-	Logger::ClassInit(channel);
+	Logger::ClassInit(&channel->producerSocket);
```
ibc: Also here, use `&(channel->producerSocket)`.

worker/src/main.cpp
```diff
@@ -107,7 +112,8 @@ int main(int argc, char* argv[])
 		Utils::Crypto::ClassInit();
 		RTC::DtlsTransport::ClassInit();
 		RTC::SrtpSession::ClassInit();
-		Channel::Notifier::ClassInit(channel);
+		Channel::Notifier::ClassInit(&channel->producerSocket);
+		Channel::Request::ClassInit(&channel->producerSocket);
```
ibc: The same.

worker/src/Channel/Request.cpp
```diff
@@ -57,9 +57,22 @@ namespace Channel
 	};
 	// clang-format on
 
+	/* Class variables. */
```
ibc: This comment `/* Class variables. */` is duplicated above. Please remove it here.

worker/src/Channel/Request.cpp
```diff
@@ -57,9 +57,22 @@ namespace Channel
...
+
+	Channel::UnixStreamSocket* Request::channel{ nullptr };
+
+	/* Static methods. */
```
ibc: Replace with `/* Class methods */`.

worker/include/Channel/Request.hpp
```diff
@@ -76,15 +76,20 @@ namespace Channel
...
-		Channel::UnixStreamSocket* channel{ nullptr };
 		uint32_t id{ 0u };
 		std::string method;
 		MethodId methodId;
 		json internal;
 		json data;
 		// Others.
 		bool replied{ false };
+
+	public:
```
ibc: Move static methods and members above, below the `UnixStreamSocket` constructor.

worker/src/Channel/Request.cpp
```diff
@@ -112,6 +125,8 @@ namespace Channel
 	{
 		MS_TRACE();
 
+		MS_ASSERT(Request::channel != nullptr, "channel unset");
```
ibc: Let's replace it with `MS_ASSERT(Request::channel, "channel unset");` and let's remove the empty line between both `MS_ASSERT()`s.

worker/src/Channel/Request.cpp
```diff
@@ -121,13 +136,15 @@ namespace Channel
...
-		this->channel->Send(jsonResponse);
+		Request::channel->Send(jsonResponse);
 	}
 
 	void Request::Accept(json& data)
 	{
 		MS_TRACE();
 
+		MS_ASSERT(Request::channel != nullptr, "channel unset");
+
```
ibc: Let's replace it with `MS_ASSERT(Request::channel, "channel unset");` and let's remove the empty line between both `MS_ASSERT()`s.

worker/src/Channel/Request.cpp
```diff
@@ -140,13 +157,15 @@ namespace Channel
...
-		this->channel->Send(jsonResponse);
+		Request::channel->Send(jsonResponse);
 	}
 
 	void Request::Error(const char* reason)
 	{
 		MS_TRACE();
 
+		MS_ASSERT(Request::channel != nullptr, "channel unset");
+
```
ibc: Let's replace it with `MS_ASSERT(Request::channel, "channel unset");` and let's remove the empty line between both `MS_ASSERT()`s.

worker/src/Channel/Request.cpp
```diff
@@ -159,13 +178,15 @@ namespace Channel
...
-		this->channel->Send(jsonResponse);
+		Request::channel->Send(jsonResponse);
 	}
 
 	void Request::TypeError(const char* reason)
 	{
 		MS_TRACE();
 
+		MS_ASSERT(Request::channel != nullptr, "channel unset");
+
```
ibc: Let's replace it with `MS_ASSERT(Request::channel, "channel unset");` and let's remove the empty line between both `MS_ASSERT()`s.

worker/src/Worker.cpp
ibc: And fix the message as follows: ""Producer Socket should not be used for receiving requests"

worker/src/main.cpp
ibc: Also here, use &(channel->xxxxxSocket)`.

worker/src/main.cpp
```diff
@@ -137,6 +143,7 @@ int main(int argc, char* argv[])
 
 void IgnoreSignals()
 {
+#ifndef _WIN32
```
ibc, line 145: I'd prefer to do it as follows:

```c++
void IgnoreSignals()
{
#ifdef _WIN32
  return;
#endif

  // Here the original code with its original indentation.
}
```

worker/win-tasks.js
```diff
@@ -0,0 +1,81 @@
...
+}
+/* eslint-enable no-console */
+
+function execute(command) 
+{
+	const childProcess = exec(command);
+	
+	childProcess.stdout.pipe(process.stdout);
+	childProcess.stderr.pipe(process.stderr);
+}
```
ibc, line 81: Add line jump here. And make sure all new files use LF (`\n`)  as line separator instead of CRLF (`\r\n`).

ibc (changes requested):
I think this is getting cool. Let's handle some changes.

worker/src/main.cpp
haiyangwu: If did like this, it won't be compiled for SIGPIPE, SIGALRM, SIGUSR1 and SIGUSR2 undefined.

worker/include/Channel/Request.hpp
haiyangwu: `UnixStreamSocket` constructor ? I didn't get it.  change like this?
```
	class Request
	{
	public:
		static void ClassInit(Channel::UnixStreamSocket* channel);

	public:
		// Passed by argument.
		static Channel::UnixStreamSocket* channel;
		//....
	}
```

worker/win-tasks.js
haiyangwu: Already add another LF after function scope. now has two LF, is this expected?
```
function execute(command) 
{
	const childProcess = exec(command);
	
	childProcess.stdout.pipe(process.stdout);
	childProcess.stderr.pipe(process.stderr);
}// LF here
// LF here

```

ibc: Umm, I think there should be just one. May be it was already present, if so, sorry :)

ibc: So you can remove it.

worker/src/main.cpp
ibc: you are completely right, my fault.

worker/include/Channel/Request.hpp
haiyangwu: The problem is that `ChannelWrapper` depends on `Producer Channel ` and `Consumer Channel`, `Consumer Channel` create `request` object. so making a single instance of `ChannelWrapper` and passing it `Consumer Channel` seems weird.

If we wan't to avoid static members, there are two approaches I can think of.
* wrap `consumerFd` and `producerFd` into `::UnixStreamSocket`, when invoking `read` or `write` we chose the right one.  I already implemented  this [here](https://github.com/haiyangwu/mediasoup-win/commit/5892905f8f22cf479848f5823aa340c0ab0692ee)
* create two subclass (`::ConsumerSocket` and `::ProducerSocket` ) of `::UnixStreamSocket`. `Channel::UnixStreamSocket` instead  of inheriting `::UnixStreamSocket`, it composite `::ConsumerSocket` and `::ProducerSocket`

IMO, I like the second approach.  I need some suggestion from you. 😘

ibc: I think you were right and we can have a static `channel` member in `Request.hpp` (as we have in `Notifier.hpp`.  Just please ensure both are written in a similar way.

haiyangwu: >* create two subclass (`::ConsumerSocket` and `::ProducerSocket` ) of `::UnixStreamSocket`. `Channel::UnixStreamSocket` instead  of inheriting `::UnixStreamSocket`, it composite `::ConsumerSocket` and `::ProducerSocket`

[Implementation](https://github.com/haiyangwu/mediasoup-win/commit/6fbcc2aff91dc42ad8044935fe308a8d21a3990a). IMO, this seems better.

ibc: I've done some comments in that commit: https://github.com/haiyangwu/mediasoup-win/commit/6fbcc2aff91dc42ad8044935fe308a8d21a3990a

ibc: Specially [this one](https://github.com/haiyangwu/mediasoup-win/commit/6fbcc2aff91dc42ad8044935fe308a8d21a3990a#r35040725).

BTW I like your new approach. I've just "cleaned" it a bit to follow mediasoup guidelines :)

worker/scripts/configure.py
ibc: this :)

haiyangwu: Already removed these two lines for two reasons.
*  To generate Visual Studio solution file (.sln) and projects files (.vcxproj) just under `worker` directory. Then we can run `mediasoup-worker-test.exe` without changing the directory hierarchy of test resource files.
* The parameter `-Goutput_dir=` doesn't work, added settings into ` mediasoup-worker.gyp` 
```
# Output Directory setting for msvc. 
# https://github.com/nodejs/node-gyp/issues/1242#issuecomment-310921441
'msvs_configuration_attributes': {
     'OutputDirectory': '$(SolutionDir)\\out\\$(Configuration)\\'
 }
```

ibc (approved):

ibc:
Ok, so 2000 millions of thanks, @haiyangwu, this is an amazing work and I think we are done. Now the "problem"... :)

We are working on adding sender side Transport-Wide Congestion Control to mediasoup. We are doing it in a private branch for now but will move it to a public "devel" branch here in GitHub next week. The thing is that such a branch includes many new files of libwebrtc (those related to congestion control and so on).

So the problem is that, if we merge this PR right now into "v3" branch:

* For a while (several weeks) we are not gonna have time to maintain it in case it needs some modifications.
* We are spending all our time in the private "devel" branch (to be public next week).
* We'll probably have tons of git conflicts when we later merge "devel" branch into "v3" branch.
* It may happen that the "devel" branch needs additional changes to compile on Windows.

So, what about if next week we make our "devel" branch public and request you to merge this PR into it and, probably, make some additional changes to make it work on Windows?

haiyangwu:
> Ok, so 2000 millions of thanks, @haiyangwu, this is an amazing work and I think we are done. Now the "problem"... :)
> 
> We are working on adding sender side Transport-Wide Congestion Control to mediasoup. We are doing it in a private branch for now but will move it to a public "devel" branch here in GitHub next week. The thing is that such a branch includes many new files of libwebrtc (those related to congestion control and so on).
> 
> So the problem is that, if we merge this PR right now into "v3" branch:
> 
> * For a while (several weeks) we are not gonna have time to maintain it in case it needs some modifications.
> * We are spending all our time in the private "devel" branch (to be public next week).
> * We'll probably have tons of git conflicts when we later merge "devel" branch into "v3" branch.
> * It may happen that the "devel" branch needs additional changes to compile on Windows.
> 
> So, what about if next week we make our "devel" branch public and request you to merge this PR into it and, probably, make some additional changes to make it work on Windows?

I'll handle the conflicts as soon as "devel" branch merged into "v3" branch.  😄

ibc:
Sure, then let's us some weeks before we merge "devel" into "v3".

ibc:
Hi, @haiyangwu. The "devel" branch with work-in-progress transport-cc support is now in GitHub:

https://github.com/versatica/mediasoup/tree/devel

May you please try to merge your branch with Win support into the new "devel" branch? Some notes:

* If you need to modify any file under `worker/deps/libwebrtc`, please let's know first as those are files from Google's libwebrtc.
* I hope you don't need to modify anything in `worker/deps/libwebrtc/deps/abseil-cpp`, but if so, please let's also know. May be modifications into its .gyp` file are required.

haiyangwu:
@ibc Merged and all tests passed.  Just like your comment, require some [modifications](https://github.com/versatica/mediasoup/pull/329/commits/7ac3a9995be7ca3f00070d2b3b7dabdf3385e08e) in `worker/deps/libwebrtc/deps/abseil-cpp`'s gyp file which is also described [abseil-cpp/CMake/README.md](https://github.com/versatica/mediasoup/blob/devel/worker/deps/abseil-cpp/abseil-cpp/CMake/README.md)

haiyangwu:
@ibc 🤔Checks failed.  Is the config wrong? May you please check it.

ibc:
The failing travis task is [this](https://travis-ci.com/versatica/mediasoup/jobs/241780829):

```
../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/base/policy_checks.h:77:2: error: #error "C++ versions less than C++11 are not supported."
 #error "C++ versions less than C++11 are not supported."
  ^
In file included from ../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/strings/string_view.h:51:0,
                 from ../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/strings/match.h:38,
                 from ../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/strings/match.cc:15:
../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/base/internal/throw_delegate.h:39:1: error: expected unqualified-id before ‘[’ token
 [[noreturn]] void ThrowStdLogicError(const std::string& what_arg);
 ^
```

It's like we need to force C++11 in the gyp of abseil-cpp, am I right?

ibc:
NOTE: We did **never** run the "deve" branch in Travis, so...

worker/deps/abseil-cpp/abseil-cpp.gyp
```diff
@@ -39,6 +39,16 @@
...
+        'defines': [
+          'NOMINMAX', # Don't define min and max macros (windows.h)
+          # Don't bloat namespace with incompatible winsock versions.
+          'WIN32_LEAN_AND_MEAN',
+          # Don't warn about usage of insecure C functions.
+          '_CRT_SECURE_NO_WARNINGS',
+          '_SCL_SECURE_NO_WARNINGS',
+          # Introduced in VS 2017 15.8, allow overaligned types in aligned_storage
+          '_ENABLE_EXTENDED_ALIGNED_STORAGE'
+        ]
```
ibc: NOTE: `worker/deps/abseil-cpp` has been removed in "devel" branch. It's now just in `worker/deps/libwebrtc/deps/abseil-cpp`.

ibc:
> It's like we need to force C++11 in the gyp of abseil-cpp, am I right?

I've added it in devel branch: https://github.com/versatica/mediasoup/commit/780d1c18b1194138b2ab4c19f75997ae118d4211

ibc:
BTW, regarding `run-script-os` dep now I'm 100% confused. I understood in the past that we should use [this](https://github.com/kawatzaki/dr.amaton-run-script-os) one, however the same author is also updating [this](kawatzaki/run-script-os) one.

I've asked the author (@kawatzaki) [here](https://github.com/kawatzaki/run-script-os/pull/1#issuecomment-538514411) :)

kawatzaki:
Hi fellas, don't worry about @dr.amaton/run-script-os, we just updated to 1.1.3, which shouldn't break anything at all, I was just clarifying which repo should be updated and which should be left alone....

Keep this project going!

ibc:
Thanks a lot for the clarification, @kawatzaki :)

worker/deps/abseil-cpp/abseil-cpp.gyp
haiyangwu: Ok... New commits in `devel` branch merged.

ibc:
@haiyangwu can you please make this PR against the "devel" branch instead? We don't want to merge "devel" into "v3" yet (it's not ready).

haiyangwu:
> @haiyangwu can you please make this PR against the "devel" branch instead? We don't want to merge "devel" into "v3" yet (it's not ready).

This PR's base branch already switched to "devel" branch.  BTW, doc works about "requirements" as below still need to add, please let me know when the proper time and tell me where it should be put.

> ensure your host satisfies the following **requirements**:
> 
> * Visual Studio 2015/2017
> * Add the path of MSBuild.exe to System Enviroment (eg. C:\Program Files (x86)\MSBuild\14.0\Bin)
> * Add GYP_MSVS_VERSION to System Enviroment (eg. 2017 for visual studio 2017 )

worker/deps/libwebrtc/deps/abseil-cpp/abseil-cpp.gyp
```diff
@@ -49,6 +49,16 @@
...
           '/wd4702'  # unreachable code
         ],
+        'defines': [
+          'NOMINMAX', # Don't define min and max macros (windows.h)
+          # Don't bloat namespace with incompatible winsock versions.
+          'WIN32_LEAN_AND_MEAN',
+          # Don't warn about usage of insecure C functions.
+          '_CRT_SECURE_NO_WARNINGS',
+          '_SCL_SECURE_NO_WARNINGS',
+          # Introduced in VS 2017 15.8, allow overaligned types in aligned_storage
```
ibc: Please dot at the end of the comment :)

worker/include/common.hpp
```diff
@@ -1,14 +1,29 @@
...
+#include <cstddef>    // size_t
+#include <cstdint>    // uint8_t, etc
+#include <functional> // std::function
+#include <memory>     // std::addressof()
+#ifdef _WIN32
+#include <winsock2.h>
+// https://stackoverflow.com/a/27443191/2085408
+#undef max
+#undef min
+// avoid uv/win.h: error C2628 'intptr_t' followed by 'int' is illegal
```
ibc: Dot at the end :)

ibc (changes requested):
Just two minor changes, please (two dots) :)

ibc:
> > @haiyangwu can you please make this PR against the "devel" branch instead? We don't want to merge "devel" into "v3" yet (it's not ready).
> 
> This PR's base branch already switched to "devel" branch.

Great, please just add two missing dots (as requested) and I'll merge it.


> BTW, doc works about "requirements" as below still need to add, please let me know when the proper time and tell me where it should be put.
> 
> > ensure your host satisfies the following **requirements**:
> > 
> > * Visual Studio 2015/2017
> > * Add the path of MSBuild.exe to System Enviroment (eg. C:\Program Files (x86)\MSBuild\14.0\Bin)
> > * Add GYP_MSVS_VERSION to System Enviroment (eg. 2017 for visual studio 2017 )

Don't worry, I'll add those requirements in the mediasoup website :)

ibc (approved):

ibc:
Hi @haiyangwu, merged! so many thanks! This will be eventually merged into v3 once we end dealing with transport-cc. BTW let me two questions:

1) Is this ok and accurate?: https://mediasoup.org/documentation/v3/mediasoup/installation/#windows

2) I've added you to [Acknowledgements](https://mediasoup.org/about/#acknowledgements). Is Haiyang Wu your name? and do you have a preferred personal site?

haiyangwu:
@ibc My pleasure. 

> 1. Is this ok and accurate?: https://mediasoup.org/documentation/v3/mediasoup/installation/#windows

For accurate, please change a little bit.

* Append the path of **MSBuild.exe** folder to **PATH** in Windows  environment variable (e.g. C:\Program Files (x86)\MSBuild\14.0\Bin).
* Create a new Windows environment variable **GYP_MSVS_VERSION** to demostrate Vistual Studio's version (e.g. 2017 for Visual Studio 2017).

> 2. I've added you to [Acknowledgements](https://mediasoup.org/about/#acknowledgements). Is Haiyang Wu your name? and do you have a preferred personal site?

Thanks again and yes my name is Haiyang Wu. Linking to Github is ok.

ibc:
Thanks, done.

ibc:
@haiyangwu may you please check whether the new "scripts" in `package.json` in the [typescript](https://github.com/versatica/mediasoup/tree/typescript) branch run fine in Windows?

ibc:
Also @haiyangwu, there is an important issue due the usage of "@dr.amaton/run-script-os":

https://github.com/kawatzaki/dr.amaton-run-script-os/issues/1

I hope @kawatzaki can take a look to it since this's a show stopper for me right now :)

kawatzaki:
I'll be checking it out, I'm currently at lunch!

But I'll get it fixed today

haiyangwu:
> Also @haiyangwu, there is an important issue due the usage of "@dr.amaton/run-script-os":
> 
> [kawatzaki/dr.amaton-run-script-os#1](https://github.com/kawatzaki/dr.amaton-run-script-os/issues/1)
> 
> I hope @kawatzaki can take a look to it since this's a show stopper for me right now :)

@ibc I saw the issue closed (add extra `--no-arguments`. It works in OSX now? I just want to confirm and I'll check it in Windows.

ibc:
Yes, with `--no-arguments` it works (pushed to "devel" branch).

PR #349, opened 2019-12-04
fix(worker): cpu cores detection using sh and default to 4

smoke:
In case bash is missing from the system the `cpu_cores.sh` fails with an
error and returns empty string resulting in command like
`make -j BUILDTYPE=Release -C out` that uses unlimited jobs and could
lead to signifficant CPU and RAM usage during build

The cpu_cores.sh is now changed to using `sh` instead of `bash`, so it
should be more portable. Additionally if the cpu detection fails the
jobs default to 4, that should be safe enough.


ibc:
Thanks, it makes sense. Released in mediasoup 3.4.1.

PR #371, opened 2020-03-30
C++: Remove inline methods

### LINKED ISSUE #366: C++: Remove inline methods
We use too many `inline` methods, declared and **defined** in the corresponding `.hpp` files.

Using `inline` is not the panacea. In increases compilation time and executable size (the latter is not a big issue for us). But, overall, we have ended with many classes having half of their methods defined in the `.cpp` file and the other half in the `.htpp` file. And we cannot use `MS_XXXX` logging functions in `.hpp` files.

@jmillan what about getting rid of **all** `inline` methods? If so, we may have to create new `.cpp` files (for instance, for all classes in `Utils/` folder).

ibc:
Let's close #366 when merged.

No more `inline` methods anywhere. Some concerns:

- `include/RTC/Codecs/Codecs.hpp` (which defined inline functions within the `Codecs` namespace) has been renamed to `Tools.hpp` and now defines a `Tool` class with static methods.

- I do **not** understand why if I write `MS_TRACE()` into any static method in `include/RTC/Codecs/Tool.hpp` it does not fail to compile. No .hpp file includes `Logger.hpp` so this makes no sense (unless this happens because those are static methods and behave different for whatever reason).


jmillan:
> I do not understand why if I write MS_TRACE() into any static method in include/RTC/Codecs/Tool.hpp it does not fail to compile. No .hpp file includes Logger.hpp so this makes no sense (unless this happens because those are static methods and behave different for whatever reason).

Does it actually print the trace? If so, what value does it show in MS_CLASS?

ibc:
> Does it actually print the trace? If so, what value does it show in MS_CLASS?

There is no `MS_CLASS` defined in `Tool.hpp`. I didn't try to compile everything with trace support to see it.

ibc:


```diff

Clones added
============
- worker/include/RTC/RtxStream.hpp  1
- worker/include/handles/UdpSocket.hpp  1
- worker/include/RTC/RTCP/ReceiverReport.hpp  3
- worker/include/RTC/Codecs/VP8.hpp  1
- worker/include/RTC/RTCP/XR.hpp  2
- worker/include/RTC/RTCP/SenderReport.hpp  2
- worker/include/RTC/Codecs/VP9.hpp  1
- worker/include/RTC/RtpStream.hpp  1
- worker/include/RTC/SimulcastConsumer.hpp  1
- worker/include/RTC/SimpleConsumer.hpp  1
- worker/include/RTC/SvcConsumer.hpp  1
- worker/include/handles/TcpServer.hpp  1
         

Clones removed
==============
+ worker/include/RTC/RtpStreamRecv.hpp  -1
+ worker/include/RTC/RTCP/Sdes.hpp  -1
+ worker/include/RTC/RTCP/FeedbackRtpTransport.hpp  -3
+ worker/include/RTC/RtpStreamSend.hpp  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/versatica/mediasoup/pullRequest?prid=5271650&bid=17266864)

jmillan:
LGTM, also curious about the MS_TRACE in Tools.hpp

jmillan (approved):

PR #377, opened 2020-04-09
Use primitive values instead of boxed values

JamesChenX:


ibc:
I think this is just cosmetic and there is no much difference using both. And I like to be explicit rather than using magic syntax.

Thanks but I don't think this fixes anything and it does not really improve anything.

PR #389, opened 2020-04-26
Remove useless assignment

universeroc:
`dtlsCertificateFile` and `dtlsPrivateKeyFile`  are never to be modified, so make them `const` and also the latest two lines are useless


ibc:
thanks

Issue #395, opened 2020-05-11
Another bug in seqManager and performence increase


penguinol:
Hi, guys!
There seem to be another bug in seqManager.
See such case:
```
	SECTION("sync and drop some input near max-value")
	{
		// clang-format off
		std::vector<TestSeqManagerInput<uint16_t>> inputs =
		{
			{ 65530,  1,  true, false },
			{ 65531,  2, false, false },
			{ 65532,  3, false, false },
			{ 65533,  0, false, true  },
			{ 65534,  0, false, true  },
			{ 65535,  4, false, false },
			{     0,  5, false, false },
			{     1,  6, false, false },
			{     2,  7, false, false },
			{     3,  8, false, false }	
		};
		// clang-format on

		SeqManager<uint16_t> seqManager;
		validate(seqManager, inputs);
	}
```

When we droped some input near the max-value, after input back to 0, the output secquence num will be discoutinued.

I think it's because we simple use `<` to compare the input with the dropped packets, as follow:
in `bool SeqManager<T>::Input(const T input, T& output)`
```
size_t dropped = std::count_if(
			  this->dropped.begin(), this->dropped.end(), [&input](T i) { return i < input; });
```
In this case, in fact, 65535 is smaller than 0 in sequence num, but larger in mathematics.
So we should use `IsSeqLowerThan` instead.

And also, i found that the performence of `count_if` is really low, here is a flame graph when about 20 peers in a conference, cpu is near 100%.
![perf](https://user-images.githubusercontent.com/14129614/81541230-6022c600-93a5-11ea-8320-c87b8d3cf295.png)
`count_if` takes more than 60% cpu time.
I think it's because `count_if` simply travel the map and find out how many items match the condition. When there are a lot of dropped packets, it will cost a lot of time.
Since we should use `IsSeqLowerThan` instead of `<`,and  `IsSeqLowerThan` is the defalut compare function of dropped map, we can just use `this->dropped.lower_bound(input)` to find the iterator, and use `std::distance` to calculate the number of dropped packets which are "smaller" than input.
Futhermore, in most of case, the number of dropped packets which are larger than input is far more less than those are smaller.
So `this->dropped.size() - std::distance(this->dropped.upper_bound(input), this->dropped.end())` might be faster than using ` std::distance(this->dropped.begin(), this->dropped.upper_bound(input))`, because `map::iterator` is not a random access iterator.

ibc:
Amazing catch!

[...]

PR #396, opened 2020-05-11
SeqManager.cpp: fix num discontinued bug and increase performance

penguinol:
#395


jmillan (approved):

ibc (commented):
The test does not compile for me. It looks like you are not synced with `v3` branch:

```
../test/src/RTC/TestSeqManager.cpp:409:15: error: use of class template 'TestSeqManagerInput' requires template
      arguments
                std::vector<TestSeqManagerInput> inputs =
```

ibc:
Please take a look to the current `TestSeqManager.cpp` in mediasoup `v3` branch:

* https://github.com/versatica/mediasoup/blob/v3/worker/test/src/RTC/TestSeqManager.cpp

All `inputs` are created with the proper template syntax:

```c++
std::vector<TestSeqManagerInput<uint16_t>> inputs =
```

However the diff in your PR does not include those templates.

ibc:
Also, I think `SeqMaanger.cpp` should include `#include <iterator>` which exposes `std::distance()`.

ibc:
> Also, I think `SeqMaanger.cpp` should include `#include <iterator>` which exposes `std::distance()`.

Please add this and we are done. And let me know your name/URL or something to add into mediasoup Ackknowledges.

worker/src/RTC/SeqManager.cpp
```diff
@@ -1,6 +1,7 @@
 #define MS_CLASS "RTC::SeqManager"
 // #define MS_LOG_DEV_LEVEL 3
 
+#include <iterator>
```
ibc, line 4: This must be at the end. Otherwise `make lint` will fail. `make format` will fix it for you.

ibc: Will do myself, don't worry.

ibc:
Merged, thanks a lot!

penguinol:
> > Also, I think `SeqMaanger.cpp` should include `#include <iterator>` which exposes `std::distance()`.
> 
> Please add this and we are done. And let me know your name/URL or something to add into mediasoup Ackknowledges.

JunLong Zou
www.lonny.me

ibc:
Thanks, added here: https://mediasoup.org/about/#acknowledgements

PR #398, opened 2020-05-12
improve a little bit performance

penguinol:
We can reuse the result of `set::lower_bound` as `set::find`, so that it will improve a little performace.
In my case, it's about 2.7%.


worker/src/RTC/SeqManager.cpp
```diff
@@ -77,28 +77,30 @@ namespace RTC
 		if (!this->dropped.empty())
 		{
 			// Delete dropped inputs older than input - MaxValue/2.
-			size_t droppedSize = this->dropped.size();
-			auto it            = this->dropped.lower_bound(input - MaxValue / 2);
+			size_t dropped = this->dropped.size();
```
ibc: Can you rename this `size_t dropped` variable to `size_t droppedCount`?

penguinol: yep

jmillan (approved):

jmillan:
I'm OK with this, lets see Travis results.

worker/src/RTC/SeqManager.cpp
```diff
@@ -77,28 +77,30 @@ namespace RTC
 		if (!this->dropped.empty())
 		{
 			// Delete dropped inputs older than input - MaxValue/2.
-			size_t droppedSize = this->dropped.size();
-			auto it            = this->dropped.lower_bound(input - MaxValue / 2);
+			size_t droppedCount = this->dropped.size();
+			auto it        = this->dropped.lower_bound(input - MaxValue / 2);
```
ibc: Please run `make lint` in `worker/` directory,. This will fail. Auto-fix it by running `make format`.

ibc (changes requested):

ibc (approved):

worker/src/RTC/SeqManager.cpp
```diff
@@ -77,28 +77,30 @@ namespace RTC
 		if (!this->dropped.empty())
 		{
 			// Delete dropped inputs older than input - MaxValue/2.
-			size_t droppedSize = this->dropped.size();
-			auto it            = this->dropped.lower_bound(input - MaxValue / 2);
+			size_t droppedCount = this->dropped.size();
+			auto it             = this->dropped.lower_bound(input - MaxValue / 2);
 
 			this->dropped.erase(this->dropped.begin(), it);
```
ibc, line 83: Here it no longer removes entries from `this->dropped`. Is this ok?

penguinol: The code you quoted is the same as before, only rename `droppedSize` to `droppedCount`
We first remove items lower than (input-MaxValue/2) and recalculate this->base. 
Then count the items lower than input and use this->base minus it.

penguinol: This commit works the same as before, only replace`find` & `upper_bound` with `lower_bound`.
Because when input is in dropped set, `lower_bound` will return as 'find', otherwise, it will return as `upper_bound`.

ibc: sorry, my fault.

PR #403, opened 2020-05-14
ConsumerOptions.rtpCapabilities field is not optional

nazar-pc:
It is also not optional according to documentation

### DIFF
--- a/src/Consumer.ts
+++ b/src/Consumer.ts
@@ -18,7 +18,7 @@ export type ConsumerOptions =
 	/**
 	 * RTP capabilities of the consuming endpoint.
 	 */
-	rtpCapabilities?: RtpCapabilities;
+	rtpCapabilities: RtpCapabilities;
 
 	/**
 	 * Whether the Consumer must start in paused mode. Default false.


ibc:
Not accurate. In fact, if you run  `npm run typescript:watch` it will fail:

```
src/Router.ts:743:53 - error TS2345: Argument of type '{ producerId: string; }' is not assignable to parameter of type 'ConsumerOptions'.
  Property 'rtpCapabilities' is missing in type '{ producerId: string; }' but required in type 'ConsumerOptions'.

743  			pipeConsumer = await localPipeTransport.consume({ producerId });
     			                                                ~~~~~~~~~~~~~~

  src/Consumer.ts:21:2
    21  rtpCapabilities: RtpCapabilities;
        ~~~~~~~~~~~~~~~
    'rtpCapabilities' is declared here.

[5:13:05 PM] Found 1 error. Watching for file changes.
```

This is because, in [pipeTransport.consume()](https://github.com/versatica/mediasoup/blob/v3/src/PipeTransport.ts#L256), field `rtpCapabilities` is not needed and not given.

nazar-pc:
Then it is documentation issue?
https://mediasoup.org/documentation/v3/mediasoup/api/#ConsumerOptions

Either way having optional field for WebRTC Transport where it is not optional is confusing.

ibc:
Yep, the API doc should be fixed. I've reported it and will fix eventually:

https://github.com/versatica/mediasoup-website/issues/14

Thanks.

PR #410, opened 2020-05-31
fix(worker build): CLT version detect exception

enimo:
XCode **Command Line Tools** info no longer show up when call `CLTVersion() ->  /usr/sbin/pkgutil` on Catalina (OS X 10.15), which break off the XcodeVersion detection with an exception **"gyp: No Xcode or CLT version detected!"** .

This patch adds a additional method to detect CLT version based on **Software Update** history: `/usr/sbin/softwareupdate` .

PR #413, opened 2020-06-02
fix CLT/xcode version detection exception

enimo:
fix(worker build): CLT version detect exception without XCode install, more details refers to #410


ibc (approved):

ibc:
Thanks.

ibc:
Released in 3.6.4.

ibc:
BTW good work here. Indeed it seems that mediasoup did not even install in Mojave if XCode was not installed: https://mediasoup.discourse.group/t/worker-process-failed-pid-undefined/1169/10

enimo:
Glad to hear this :)

PR #416, opened 2020-06-08
Unspecified address support

nazar-pc:
This fixes following line from documentation:
> Do not use “0.0.0.0” as ip value. It must be a specific bindable IP in the host.

Rationale behind this change is that it is very inconvenient to specify exact IPs is dynamic environments like containers.

Implementation supports both `0.0.0.0` and `::`, in second case both IPv4 and IPv6 addresses are included to mimic [Node.js Net module](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback):
> In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

I have also changed TypeScript library and target, because ES2019 and ES2020 were not correct since Node.js v12 and especially v10 are still supported according to Travis CI and having them enabled results in code that compiles, but [doesn't run](https://travis-ci.com/github/nazar-pc/mediasoup/builds/170218653) on older Node.js versions.

Tests were adjusted accordingly and are all green: https://travis-ci.com/github/nazar-pc/mediasoup/builds/170222119


ibc:


```diff

Clones added
============
- lib/Transport.js  1
- lib/PipeTransport.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/versatica/mediasoup/pullRequest?prid=5657001&bid=18518986)

jmillan:
Looks good!

ibc:
Thanks for this, although I have to look it in more details because there is something that I know that maybe problematic. Let me please a few days (too busy now).

lib/WebRtcTransport.d.ts
```diff
@@ -6,7 +6,7 @@ export declare type WebRtcTransportOptions = {
      * Listening IP address or addresses in order of preference (first one is the
      * preferred one).
      */
-    listenIps: TransportListenIp[] | string[];
+    listenIps: Array<TransportListenIp | string>;
```
ibc, line 9: Why this change? It was on purpose. I really prefer to use `[]` syntax rather than `Array`.

src/WebRtcTransport.ts
```diff
@@ -16,7 +16,7 @@ export type WebRtcTransportOptions =
 	 * Listening IP address or addresses in order of preference (first one is the
 	 * preferred one).
 	 */
-	listenIps: TransportListenIp[] | string[];
+	listenIps: Array<TransportListenIp | string>;
```
ibc, line 19: Why this change? It was on purpose. I really prefer to use `[]` syntax rather than `Array`.

nazar-pc: This is a more correct way of typing such things as it allows you to use array methods on `listenIps` without casting it first with `(listenIps as any[])` and loosing type information in the process.

ibc:
This is problematic for some reasons:

1. It's not a good practice to tell users _"since you don't know (and don't event want to figure out) the bindable private IP of your server, then pass "0.0.0.0" or "::" and mediasoup will bind in all available IPs". This means that, if the server has a complex networking with 4 IPs, for every call to `router.createWebRtcTransport()` mediasoup will bind in all those 4 IPs. I don't find this elegant from the security point of view.

2. `os.networkInterfaces()` provides IPv6 entries that seem to be bindable but, sometimes, they are not. If `router.createWebRtcTransport()` is called with '::' and one the found IPv6 is one of them, the method will throw (it must be able to bind in all given IPs, not just in a one or some of them).

3. In mediasoup v2 we did "local IP auto-discovery" (for IPv4 and IPv6) in C++ `[1]`. Result? it did work most of the times, but not always. And when it failed (due to the networking settings in the host) people reported it as a bug in mediasoup. That's why we force the application developer to provide the exact IP or IPs in which it wishes to bind.

4. The approach in this PR is just for `WebRtcTransport`. So users will eventually ask why something similar is not done for `router.createPlainTransport()` but, instead of returning an array of all available IPs just return one, the "best one", the "one that works"). And here we go to bullet 3 again.

5. The idea behind mediasoup is do less and do it fine. Auto-discovery of "bindable IPs" never fits into that slogan because a library cannot be aware of the networking complexity in every host. Adding sugar APIs (just to avoid that the user codes it into his app) means more bug reports.

6. I should not share this but who cares. Here a 64 lines script to get the proper bindable IP in a host:
  * https://gist.github.com/ibc/b828c55a99b313cd7133c38c30dcd68d

Given all this rationale, my position is that I don't want to include more sugar API or utilities into mediasoup, because I know what they mean at the end.

`[1]`: https://github.com/versatica/mediasoup/blob/v2/worker/src/Settings.cpp#L359

nazar-pc:
1. It is not elegant, but at the same time it is not enforced. If I have a container with just loopback and eth0 is it quite annoying to write bash script around an app just to figure out with what IP container started with this time. Literally all TCP/HTTP servers allow binding to `0.0.0.0` and `::`, so I think it makes sense to do the same here.
2. I think this is acceptable with a note in the docs, certainly better than nothing IMHO.
3. Well, there should be a balance between automatic behavior and being explicit for sure.
4. Again, I think it still makes sense to add note to docs just like not all methods are available for all transports in general.
5. I know mediasoup is not TCP/HTTP server, but it is natural to assume that `0.0.0.0` just works, so much that there is already a note about this in the docs

ibc:
> I know mediasoup is not TCP/HTTP server, but it is natural to assume that 0.0.0.0 just works, so much that there is already a note about this in the docs

This is not entirely true. Servers listening in 0.0.0.0 do not care about the **real** local IP in which they have received a specific TCP connection or UDP packet. However mediasoup needs to care about that because that's the way ICE works.

Said that, I'll be honest and say that I don't want to add this feature (for all the reasons exposed above, which are not new reasons, I insist that mediasoup v2 was already designed that way so we know what happens).

However, an idea coming to my mind would be:

* Allow passing `0.0.0.0` or `::` as bindable IP **just** if `announcedIp` is given (which is the IP that will be present in `transport.iceCandidates` signaled to the peer.
* And then, the C++ transport would **indeed** bind into `0.0.0.0` or `::` directly (instead of passing an array of IPs from JavaScript to C++).
* It may require some additional changes in C++ side but it may work.
* I think this would indeed be a safe, simple, non convoluted and useful feature (because when the server does not have a public IP you need to provide it in `announcedIp` anyway, and that's the only real use case for "bindable IP autodiscovery", which in this case wouldn't be needed because we would really bind in `0.0.0.0` or `::`).

Thoughts?

nazar-pc:
I think that would be useful, but not in my case. In my case only TURN server is exposed publicly and WebSocket endpoint, everything else is in private network, including SFU. So I would still need to write a bash script or add application logic to determine `announcedIp`.

What about having `0.0.0.0` bind to all IPs and select the first one that is not loopback as `announcedIp`? Or the first that is publicly routable and fallback to non-loopback? (this option seems too complex to explain I think).

ibc:
Actually this is funny because, after rechecking the code, mediasoup **already** allows passing `listenIp: '0.0.0.0'` or `listenIp: '::'`.

ibc:
Documented: https://github.com/versatica/mediasoup-website/commit/817cb3673072c1cb14a31fb4ea7a9e37bf48458d

ibc:
And confirmed that it works (as far as you also provide `announcedIp`.

nazar-pc:
Is there any chance to have a reasonable default for `announcedIp` in that case then?

ibc:
> In my case only TURN server is exposed publicly and WebSocket endpoint, everything else is in private network, including SFU. So I would still need to write a bash script or add application logic to determine `announcedIp`.

I don't understand, which IP do you use to communicate your TURN server and mediasoup? Obviously you need to know a proper routable IP (routable between both servers).

> What about having `0.0.0.0` bind to all IPs and select the first one that is not loopback as `announcedIp`? Or the first that is publicly routable and fallback to non-loopback? (this option seems too complex to explain I think).

So that means doing "bindable IP auto-discovery" again, which is what I want to avoid. And here the trick for your scenario:

- You can just create the WebRtcTransport with `ip: "0.0.0.0"` and `announcedIp: "127.0.0.1"`.
- A browser will discard 127.0.0.1 for WebRTC UDP (interestingly it will accept it for WebRTC TCP), but this is not the point, since your browser is supposed to just contact your TURN server, so you may even create the local side transport with `icePolicy: "relay"`.
- The point here is that your TURN server **can** send media to mediasoup using 127.0.0.1.

Do you know what I mean?

ibc:
> Is there any chance to have a reasonable default for `announcedIp` in that case then?

Really, I want to avoid that. Please read my previous long comment, it should work for you perfectly.

nazar-pc:
> Obviously you need to know a proper routable IP (routable between both servers).

Both containers are in shared private network.

> Do you know what I mean?

Yes, but TURN server and SFU are separate containers and do not share `127.0.0.1`.

Well, looks like I'll have to copy contents of this PR into my app.
Do you want me to submit a separate PR with just TypeScript changes?

ibc:
> Well, looks like I'll have to copy contents of this PR into my app.

To clarify, the only thing you miss at this point is the "magic optimistic may-work-or-not bindable IP auto-discovery", right?

> Do you want me to submit a separate PR with just TypeScript changes?

It seems you needed it in your fork due to `array.flatten` lack of support in Node 10, but it's not used in mediasoup v3 branch (Travis CI confirms it), so why is it needed?

PR #422, opened 2020-06-11
TypeScript fixes

nazar-pc:
Minor tweaks from #416 that I believe are worth including, rationale is well described there

### DIFF
--- a/src/WebRtcTransport.ts
+++ b/src/WebRtcTransport.ts
@@ -16,7 +16,7 @@ export type WebRtcTransportOptions =
 	 * Listening IP address or addresses in order of preference (first one is the
 	 * preferred one).
 	 */
-	listenIps: Array<TransportListenIp | string>;
+	listenIps: (TransportListenIp | string)[];
 
 	/**
 	 * Listen in UDP. Default true.
--- a/src/WebRtcTransport.ts
+++ b/src/WebRtcTransport.ts
@@ -16,7 +16,7 @@ export type WebRtcTransportOptions =
 	 * Listening IP address or addresses in order of preference (first one is the
 	 * preferred one).
 	 */
-	listenIps: TransportListenIp[] | string[];
+	listenIps: Array<TransportListenIp | string>;
 
 	/**
 	 * Listen in UDP. Default true.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,8 +2,8 @@
 	"compileOnSave": true,
 	"compilerOptions":
 	{
-		"lib": [ "es2017" ],
-		"target": "es2017",
+		"lib": [ "es2018" ],
+		"target": "es2018",
 		"module": "commonjs",
 		"moduleResolution": "node",
 		"strict": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,8 +2,8 @@
 	"compileOnSave": true,
 	"compilerOptions":
 	{
-		"lib": [ "es2019" ],
-		"target": "es2020",
+		"lib": [ "es2017" ],
+		"target": "es2017",
 		"module": "commonjs",
 		"moduleResolution": "node",
 		"strict": true,
--- a/src/Router.ts
+++ b/src/Router.ts
@@ -359,7 +359,7 @@ export class Router extends EnhancedEventEmitter
 		else if (appData && typeof appData !== 'object')
 			throw new TypeError('if given, appData must be an object');
 
-		listenIps = (listenIps as any[]).map((listenIp: TransportListenIp | string) =>
+		listenIps = listenIps.map((listenIp) =>
 		{
 			if (typeof listenIp === 'string' && listenIp)
 			{


src/WebRtcTransport.ts
```diff
@@ -16,7 +16,7 @@ export type WebRtcTransportOptions =
 	 * Listening IP address or addresses in order of preference (first one is the
 	 * preferred one).
 	 */
-	listenIps: TransportListenIp[] | string[];
+	listenIps: Array<TransportListenIp | string>;
```
ibc: Now I understand this :)
However, can we do this instead?:

```suggestion
	listenIps: (TransportListenIp | string)[];
```

I've tested it and seems to be the very same.

tsconfig.json
```diff
@@ -2,8 +2,8 @@
 	"compileOnSave": true,
 	"compilerOptions":
 	{
-		"lib": [ "es2019" ],
-		"target": "es2020",
+		"lib": [ "es2017" ],
+		"target": "es2017",
```
ibc: I still don't know why this is needed. In your previous PR you changed this due to the usage of `array.flatten()` (not available in Node 10 AFAIR). There is no `array.flatten()` here. Why is this change required? Obviously CI is working and it tests Node 10, 12 and 14:

- https://travis-ci.com/github/versatica/mediasoup
- https://travis-ci.com/github/versatica/mediasoup/jobs/347408876#L1423

src/WebRtcTransport.ts
nazar-pc: You probably can, but that is not idiomatic in TypeScript to do

tsconfig.json
nazar-pc: It doesn't matter whether you are using something that is forbidden or not now, the issue is that you can introduce breaking change at any point and compiler will not stop you (yes, Rust had influence on me in this direction).

You can put `??` operator in the code now and only Node.js 14 will be able to handle it, while compiler will tell nothing. So target should represent actual target you expect code to run on. Same about libraries.

src/WebRtcTransport.ts
ibc: Ok, but I prefer to not introduce exceptions. The whole TS code in mediasoup right now uses `[]` syntax instead of `Array`. We can change the whole thing in the future.

tsconfig.json
ibc: I see, the problem is the `target` and not `lib`. And since we are supposed to support Node 10 we must produce code valid for Node 10, and that means  `target: "es2017"`, right? However:

* Why not `target: "es2018"`? It seems to be fully supported by Node 10:
  - https://stackoverflow.com/questions/51716406/typescript-tsconfig-settings-for-node-js-10

* Why must `lib` also be `es2017` or `es2018`? Why not `es2019` or `es2020` so I can use `??` eventually?

src/WebRtcTransport.ts
nazar-pc: tslint's default codestyle argues that `T[]` is for simple types and `Array<T | R>` or `Array<T<R>>` is for complex cases

tsconfig.json
nazar-pc: I don't mind about `es2018` target, but didn't check if everything is supported in TS. If it is feel free to edit it.
`lib` is not just syntax, but also things like JS standard library. If you specify `es2019` things like that `flatten()`/`flatMap()` will be left as is and will not be polyfilled for you. So it will break in Node.js 10.

nazar-pc: `??` was not polyfilled either if I remember correctly.

src/WebRtcTransport.ts
ibc: `tslint` is deprecated. Using `eslint` with `@typescript-eslint/xxxx` packages does not complain about it (tested right now).

tsconfig.json
ibc: Tested right now with:

```
"lib": [ "es2019" ],
"target": "es2018",
```

TypeScript code:

```ts
let foo: string | undefined;

console.warn('foo.length:', foo?.length ?? 'UNSET');
```

Compiled code:

```js
let foo;
console.warn('foo.length:', (_a = foo === null || foo === void 0 ? void 0 : foo.length) !== null && _a !== void 0 ? _a : 'UNSET');
```

nazar-pc: Yet, it will not work with JS API, so... https://travis-ci.com/github/nazar-pc/mediasoup/jobs/345707263

src/WebRtcTransport.ts
nazar-pc: Well, if you're so opposed to it... `(TransportListenIp | string)[]` looks completely alien to me.

ibc: I'm not against it. I just mean that I don't want to use two separate syntaxes. We can move all `[]` to `Array<>` or keep using `[]`. I just don't want mixed code. Since this is just cosmetic it can be done later, right?

tsconfig.json
ibc: That's unrelated IMHO. There you are using `array.flatten()` which does not exist in Node 10, that's the problem. TypeScript does not even try to polyfill it.

src/WebRtcTransport.ts
nazar-pc: I don't see why `string[]` needs to change to `Array<string>`, it is also too verbose. Anyway, yes, that is just cosmetic, the point is to have an array as outer data structure.

tsconfig.json
nazar-pc: > TypeScript does not even try to polyfill it.

Right, but with `"lib": [ "es2017" ],` it wouldn't even compile, isn't than nice to have?

ibc: But we are not using `array.flatten()` in the code :)

src/WebRtcTransport.ts
ibc: Ok, I promise I'll revisit this when it comes to modernize TS stuff.

tsconfig.json
nazar-pc: That is not the point. The point is that compiler proactively prevents you from even putting things that don't work into the repository.

ibc: So do you mean that if we use `lib: [ "es2018" ]` instead of `es2020`, then the compiler will complain if I try to use `array.flatten()`?

nazar-pc: Exactly

ibc: Good reason. Can we then use `es2018` for lib and target? As I pointed out before, it's supposed to be perfectly supported by Node 10 (no need to use es2017 anywhere).

nazar-pc: I think so

ibc: Cool. Can you update your PR with just that, please?

nazar-pc: Done

src/WebRtcTransport.ts
ibc: so can you please also change this to `listenIps: Array<TransportListenIp | string>;` for now?

ibc: sorry, I mean:

`listenIps: (TransportListenIp | string)[];`

nazar-pc: Done

ibc:
Thanks

ibc:
Interesting: I've run `npm run typescript:build` and the new code under `lib/` code does not even change.

nazar-pc:
Because you were careful and didn't use anything unsupported in Node 10, so makes perfect sense.

ibc:
:)

PR #427, opened 2020-06-24
fix srtp_steam leak when closing producer and consumers

penguinol:
#426

PR #428, opened 2020-06-24
Remove ended SRTP streams from SrtpSession

### LINKED ISSUE #426: srtp streams are not removed after removing consumer
Hi guys, it seems we do not remove strp streams after removing conusmers and producers.
We made a test which let about 10 client enter and leave the meeting repeatedly when the others keep stay in the meeting, after one night, the cpu usage was up to 100%.

By falme graph, i found `strp_get_stream` cost a lot of cpu cycles.
image

```
srtp_stream_ctx_t *srtp_get_stream(srtp_t srtp, uint32_t ssrc)
{
    srtp_stream_ctx_t *stream;

    /* walk down list until ssrc is found */
    stream = srtp->stream_list;
    while (stream != NULL) {
        if (stream->ssrc == ssrc)
            return stream;
        stream = stream->next;
    }

    /* we haven't found our ssrc, so return a null */
    return NULL;
}
```
It's just a loop to find the stream by ssrc from a steam list.

I think it's beacuse when we call `srtp_protect` and `srtp_unprotect` to decrypt or encrypt steam, it will automatically create a srtp_stream, but we do not remove the stream when closing consumers and producers, so the list is becoming larger and large, and finally used up all cpu.

ibc:
Fixes #426. 

This PR overrides #427 by simplifying the code. In summary, we don't want that `Transport` child classes are aware of `Producers` or `Consumers`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 ### 3.6.8 (WIP)
 
+* Fix SRTP leak due to streams not being removed when a `Producer` or `Consumer` is closed.
+  - PR #428 (fixes issues #426). 
+  - Credits to credits to @penguinol for reporting and initial work at PR #427.
 * Update `nlohmann::json` C++ dep to 3.8.0.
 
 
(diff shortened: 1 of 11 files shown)


jmillan (approved):
Perfect :+1:

ibc:
Merging, thanks a lot @penguinol

PR #441, opened 2020-08-03
Don't build with MEDIASOUP_WORKER_BIN set

universeroc:
Don't build worker after npm install on all platforms not only Windows with MEDIASOUP_WORKER_BIN environment variable is set


ibc (approved):
Looks legitimate. Thanks.

PR #445, opened 2020-08-17
Fix get consumer status abort

aggresss:
When large member in a room，may abort when get consumer status


ibc:
I'm afraid mediasoup v2 is no longer maintained so we won't take care of issues or new releases, sorry.

Anyway, if there was an issue, it just be fixed (rather than just avoided). Your PR does not fix it but just avoids it.

Please, use mediasoup v3.

PR #447, opened 2020-08-19
fix issue 408

Innovationzk:
#408 send an empty packet when switching layer.

### DIFF
--- a/worker/include/RTC/SimulcastConsumer.hpp
+++ b/worker/include/RTC/SimulcastConsumer.hpp
@@ -101,6 +101,7 @@ namespace RTC
 		std::vector<RTC::RtpStreamSend*> rtpStreams;
 		std::vector<RTC::RtpStream*> producerRtpStreams; // Indexed by spatial layer.
 		bool syncRequired{ false };
+		bool lastSentRtpPacketHasMarker{ true };
 		RTC::SeqManager<uint16_t> rtpSeqManager;
 		int16_t preferredSpatialLayer{ -1 };
 		int16_t preferredTemporalLayer{ -1 };
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -870,6 +870,44 @@ namespace RTC
 
 		if (isSyncPacket)
 		{
+			if (!lastSentRtpPacketHasMarker && shouldSwitchCurrentSpatialLayer)
+			{
+				RTC::RtpPacket::Header header;
+				header.csrcCount = 0u;
+				header.extension = 0u;
+				header.padding = 0u;
+				header.version = 2u;
+				header.payloadType = packet->GetPayloadType();
+				header.marker = 1u;
+				header.sequenceNumber = seq;
+				header.timestamp = this->rtpStream->GetMaxPacketTs();
+				header.ssrc = packet->GetSsrc();
+
+				RTC::RtpPacket *emptyPacket = RTC::RtpPacket::Parse(reinterpret_cast<uint8_t*>(&header), sizeof(RTC::RtpPacket::Header));
+
+				if (this->rtpStream->ReceivePacket(emptyPacket))
+				{
+					this->listener->OnConsumerSendRtpPacket(this, emptyPacket);
+
+					this->rtpSeqManager.Offset(1);
+					packet->SetSequenceNumber(seq+1);
+				}
+				else
+				{
+					MS_WARN_TAG(
+					rtp,
+					"failed to send empty packet [ssrc:%" PRIu32 ", seq:%" PRIu16 ", ts:%" PRIu32
+					"] from original [ssrc:%" PRIu32 ", seq:%" PRIu16 ", ts:%" PRIu32 "]",
+					emptyPacket->GetSsrc(),
+					emptyPacket->GetSequenceNumber(),
+					emptyPacket->GetTimestamp(),
+					origSsrc,
+					origSeq,
+					origTimestamp);
+				}
+				delete emptyPacket;
+			}
+			
 			MS_DEBUG_TAG(
 			  rtp,
 			  "sending sync packet [ssrc:%" PRIu32 ", seq:%" PRIu16 ", ts:%" PRIu32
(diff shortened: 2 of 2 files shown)


worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -870,6 +870,45 @@ namespace RTC
 
 		if (isSyncPacket)
 		{
+			if (!lastSentRtpPacketHasMarker && shouldSwitchCurrentSpatialLayer)
```
ibc, line 873: We always use `this->` for class instance members:

```suggestion
			if (!this->lastSentRtpPacketHasMarker && shouldSwitchCurrentSpatialLayer)
```

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -870,6 +870,45 @@ namespace RTC
...
+					  "failed to send empty packet [ssrc:%" PRIu32 ", seq:%" PRIu16 ", ts:%" PRIu32
+					  "] from original [ssrc:%" PRIu32 ", seq:%" PRIu16 ", ts:%" PRIu32 "]",
+					  emptyPacket->GetSsrc(),
+					  emptyPacket->GetSequenceNumber(),
+					  emptyPacket->GetTimestamp(),
+					  origSsrc,
+					  origSeq,
+					  origTimestamp);
+				}
+				delete emptyPacket;
```
ibc, line 909:
```suggestion
				}

				delete emptyPacket;
```

ibc (changes requested):
Two cosmetic required changes for now.

ibc:
In theory it looks good (thanks for this) although I'm not sure about the implications of sending an empty packet without any RTP extension (abs-send-time, MID, etc). In theory nothing will happen. However, assuming you are using Chrome (so MID is in use for reception) to receive the simulcast based video stream, can you please dump the RTP packets sent by the Consumer and print here the dump of a normal RTP packet and also the empty packet with marker bit set?

In `Transport.cpp` in `OnConsumerSendRtpPacket()` method, just add this at the beginning:

```c++
packet->Dump();
```

@jmillan thoughs?

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -870,6 +870,45 @@ namespace RTC
...
+				header.ssrc           = packet->GetSsrc();
+
+				RTC::RtpPacket* emptyPacket =
+				  RTC::RtpPacket::Parse(reinterpret_cast<uint8_t*>(&header), sizeof(RTC::RtpPacket::Header));
+
+				if (this->rtpStream->ReceivePacket(emptyPacket))
+				{
+					this->listener->OnConsumerSendRtpPacket(this, emptyPacket);
+
+					this->rtpSeqManager.Offset(1);
```
jmillan, line 893:
```suggestion
					this->listener->OnConsumerSendRtpPacket(this, emptyPacket);
					this->rtpSeqManager.Offset(1);
```

worker/include/RTC/SimulcastConsumer.hpp
```diff
@@ -101,6 +101,7 @@ namespace RTC
 		std::vector<RTC::RtpStreamSend*> rtpStreams;
 		std::vector<RTC::RtpStream*> producerRtpStreams; // Indexed by spatial layer.
 		bool syncRequired{ false };
+		bool lastSentRtpPacketHasMarker{ true };
```
jmillan, line 104: I would say this should be initialized to *false*

ibc: If initialized to `true`, the first RTP packet the receiver endpoint will receive will be an empty packet. IMHO it's better to have it initialized to true so we avoid sending such a useless packet when the consumer sends its first packet.

Said that, I'm not sure if we should also unset this flag in other circumstances (when the consumer is paused/resumed, etc) or not.

penguinol: > If initialized to `true`, the first RTP packet the receiver endpoint will receive will be an empty packet. IMHO it's better to have it initialized to true so we avoid sending such a useless packet when the consumer sends its first packet.
> 
> Said that, I'm not sure if we should also unset this flag in other circumstances (when the consumer is paused/resumed, etc) or not.

You are right, that's why we inited it as true. we will do some test tomorrow see what will happen when pasuing and resuming.

jmillan: :+1:

jmillan:
> In theory it looks good (thanks for this) although I'm not sure about the implications of sending an empty packet without any RTP extension (abs-send-time, MID, etc).

Good point. Since we are only sending this empty RTP packet *after* other RTP packets have already been sent, the RTP de-multiplexing on the receiver side should already be done and sending this packet with the correct SSRC and PT should be enough for the receiver to identify this packet.

ibc:
I think that within `SimulcastConsumer::UserOnPaused()` and `SimulcastConsumer::UserOnTransportDisconnected()` the flag `this->lastSentRtpPacketHasMarker` should be set to `true` since later, when resumed or transport reconnected, it would be an scenario similar to when the Consumer is created so the flag should have the same value.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -870,6 +870,45 @@ namespace RTC
...
+				header.padding        = 0u;
+				header.version        = 2u;
+				header.payloadType    = packet->GetPayloadType();
+				header.marker         = 1u;
+				header.sequenceNumber = seq;
+				header.timestamp      = this->rtpStream->GetMaxPacketTs();
+				header.ssrc           = packet->GetSsrc();
+
+				RTC::RtpPacket* emptyPacket =
+				  RTC::RtpPacket::Parse(reinterpret_cast<uint8_t*>(&header), sizeof(RTC::RtpPacket::Header));
```
ibc, line 887: Note for us: we should provide a proper `RtpPacket` factory function. This code sucks (but this is the only that is possible for now).

To clarify, nothing to be done within this PR. This is a task we should deal with in the future.

penguinol: Sad that the issus can still be reproduced after applying the patch.

We have two chrome browsers of the same version as video source. And another chrome broswer to watch them.
One of the video source had mosics when switching layers, but the other did't.

ibc: And what is the behavior in the same scenario if you don't apply this patch?

PR #450, opened 2020-08-24
SimulcastConsumer: prefer the highest spatial layer initially

jmillan:
If none of the provisional stream and the one being inspected have been active for the minimum time.


ibc (approved):

jmillan:
Let's postpone this a bit.

I want to test it a bit more.

jmillan:
merged

PR #454, opened 2020-08-29
Only print logs when errors occur

gnauhtt:
Before outputting the error log, it was not judged whether there was an error here.


ibc (changes requested):
Thanks, but this PR does not pass the linting tasks.

ibc:
The issue has been fixed in v3 branch with proper code syntax.
Thanks.

PR #457, opened 2020-09-08
Remove non-existing fields from interface

nazar-pc:
These are not present anywhere else and were probably added accidentally.


ibc:
Yep, they are not used. However, eventually we may be ready to detect RTP probation packets sent by endpoints to mediasoup but we don't do it yet. So ok, let's remove. Thanks!

PR #463, opened 2020-09-14
Fix closed consumer leak

dimes:
When a Consumer is instantiated, it adds a listener on to the payload channel. This listener is not removed when the consumer is closed, leading to a leak.


ibc:
Thanks, you found a bug (fixes in v3 branch right now and 3.6.21 released). However the way to fix it is not correct, check  this commit instead: https://github.com/versatica/mediasoup/commit/fbba811c253221bc2fca9d903afc568f52783d50

Thanks again.

dimes:
@ibc thanks for the fast fix and release!

PR #464, opened 2020-09-18
Transport.cpp: provide transport congestion client with RTCP Receiver…

jmillan:
… Reports

### DIFF
--- a/worker/src/RTC/Transport.cpp
+++ b/worker/src/RTC/Transport.cpp
@@ -1875,8 +1875,6 @@ namespace RTC
 			case RTC::RTCP::Type::RR:
 			{
 				auto* rr = static_cast<RTC::RTCP::ReceiverReportPacket*>(packet);
-				// Retrieve the RTT from one of the consumers in order to provide it to tccClient.
-				float rtt = 0;
 
 				for (auto it = rr->Begin(); it != rr->End(); ++it)
 				{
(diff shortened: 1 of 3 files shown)


worker/src/RTC/Transport.cpp
```diff
@@ -1914,9 +1916,13 @@ namespace RTC
 						continue;
 					}
 
+					rtt = consumer->GetRtt();
```
ibc: What about the commented block above? Shouldn't we also pass RR when it belongs to the probator?

jmillan: correct.

jmillan: In fact the API in `RtpTransportControllerSend` expects a list of reports, not just the one of the probator (as suggested in the commented block above) and the libwebrtc code does not distinguish between probator or other streams in order to provide with the RR's. I'll consider the probator consumer `rtt` as well.

jmillan: Indeed there is no consumer for probation.

jmillan: comment block removed.

ibc: There is no consumer for probator, right, but we should be able to read RR from probator and pass it to the bwe, am I wrong?

jmillan: oh, that's right. I've just ignore probator RR...

jmillan: > There is no consumer for probator, right, but we should be able to read RR from probator and pass it to the bwe, am I wrong?

It's done already.

ibc: Where?

jmillan: after the loop: https://github.com/versatica/mediasoup/pull/464/files#diff-d9a2ff41c9bac4d9444a0ea711677398R1906

ibc: If the RR belongs to the probator we do `continue` so `rtt` is 0., It is ok?

jmillan: the whole RTCP packet is provided after iterating.

jmillan: We take the RTT from one of the Consumers, probator does no have any so it's not considered for the RTT.

jmillan: > If the RR belongs to the probator we do continue so rtt is 0., It is ok?

It's a continue, not a break, we keep iterating the Receiver Reports.

PR #468, opened 2020-09-24
libwebrtc: BWE tweaks and debug logs

jmillan:
* Enabled WebRTC-Bwe-AlrLimitedBackoff/Enabled/ experiment.
  * It makes the down peaks smoother.
* Changed the trendline estimator to be more resilent upon sudden RTP delta increases/decreases.


ibc:
Status of this?

jmillan:
In our tests it behaves better within this PR. I would say it's overall not 100% conclusive.

ibc:
Can we proceed with this PR already?

PR #472, opened 2020-10-13
SctpAssociation: increase sctpBufferedAmount before sending any data

jmillan:
Prevents increasing the value 'after' send_cb is called before returning
from usrsctp_sendv.

### DIFF
--- a/worker/src/RTC/SctpAssociation.cpp
+++ b/worker/src/RTC/SctpAssociation.cpp
@@ -425,20 +425,21 @@ namespace RTC
 			}
 		}
 
+		this->sctpBufferedAmount += len;
+
 		int ret = usrsctp_sendv(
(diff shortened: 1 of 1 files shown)


ibc:
Rationale plz?

ibc (approved):

jmillan:
Before `usrsctp_sendv` returns, SCTP data is sent  (the corresponding static function registered for that event is called) and potentially (I can't say at 100%, but better preventing than curing)  `send_cb` could be called too (we are updating the buffered amount upon such last event). If that occurs and we increase the buffered amount as we do now we are simply setting a wrong value to the buffered amount which is the base for the calculations in apps that need this value on their sending algorithms.

ibc:
let's go

PR #474, opened 2020-10-27
Fix DirectTransport to make producer.send() typically work

dabbler0:
Formerly, `producer.send()` on media packets would typically fail on DirectTransports. This is because the `RtpPacket` instance passed from JavaScript would have no extension data (`mid`, `rid`, etc.) and the Transport would consequently fail to identify what producer is supposed to receive it, and give up.

But we know what producer is supposed to receive it; it's the producer that we called `send` on and it's stored in the `notification->internal` field coming from the `PayloadChannel`. This pull request adds an additional method `Transport::ReceiveRtpPacketForProducer`, and changes the `DirectTransport `handler so that this producer is identified at the proper time and the packet can go all the way through. It also adds a test for this functionality.

I would appreciate advice on how to make the test for this more elegant; it currently contains a hardcoded buffer for an RTP packet that was logged from a test file from a different project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "gulp-clang-format": "^1.0.27",
     "jest": "^26.6.0",
     "jest-tobetype": "^1.2.3",
+    "node-rtp": "git+https://github.com/dabbler0/node-rtp.git",
     "open-cli": "^6.0.1",
     "sctp": "^1.0.0",
     "tsc-watch": "^4.2.9",
--- a/test/test-DirectTransport.js
+++ b/test/test-DirectTransport.js
@@ -2,6 +2,30 @@ const { toBeType } = require('jest-tobetype');
 const mediasoup = require('../');
 const { createWorker } = mediasoup;
 
+// "standard" parameters for an audio-only
+// test of producers and consumers
+const mediaCodecs = [
+  {
+    kind: 'audio',
+    mimeType: 'audio/opus',
+    clockRate: 48000,
+    channels: 2
+  },
+  {
+    kind: 'video',
+    mimeType: 'video/VP8',
+    clockRate: 90000,
+    parameters:
+      {
+        'x-google-start-bitrate': 1000
+      }
+  }
+];
+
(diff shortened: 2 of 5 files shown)


ibc:


```diff

Clones added
============
- test/test-DirectTransport.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/versatica/mediasoup/pullRequest?prid=6406283&bid=20764668)

ibc:
> Formerly, `producer.send()` on media packets would typically fail on DirectTransports. This is because the `RtpPacket` instance passed from JavaScript would have no extension data (`mid`, `rid`, etc.) and the Transport would consequently fail to identify what producer is supposed to receive it, and give up.

This is not true. The `RtpPacket` instance must honor the RTP send parameters given to its associated `Producer`, and this means that those `rtpParameters` must have `mid` or `rid` or `ssrc`, and hence the `RtpPacket` as well.

Check the documentation:

- `RtpSendParameters`: https://mediasoup.org/documentation/v3/mediasoup/rtp-parameters-and-capabilities/#RtpParameters
- `transport.produce()`: https://mediasoup.org/documentation/v3/mediasoup/api/#ProducerOptions

dabbler0:
`producer.send()` only accepts a Buffer instance as an argument, and the `RtpPacket` on the C++ side consequently never has any `mid` or `rid` attached to it (as a C++ object) even if it is present in the binary data. The packet in the test here was copied out of a WebRTCTransport that mediasoup also managed and definitely had a `mid` and `rid` when it came in. But the instance of `RtpPacket` that the `DirectTransport` dealt with later on returns nothing when you call `packet->ReadMid()`, even though its payload data is exactly the same.

To be more concrete, the real-world use case that caused it to fail was this:
  1. Create a `WebRTCTransport` and connect to it with a client and produce, as usual and as documented. This creates a producer; we'll call it `producer1`.
  2. Create a `DirectTransport`, and use it to consume `producer1`, creating `consumer1`. Also create `producer2` on the DirectTransport, using identical `rtpParameters` to `producer1`.
  3. Run `consumer1.on('rtp', (packet) => producer2.send(packet));`. This will fail; no consumers of `producer2` will ever receive any packets.

To try to make this into a minimal example, I wrote the test in this PR. The test that I added in this PR does not pass without the changes that I made to `worker/src/RTC/Transport.cpp` and `worker/src/RTC/DirectTransport.cpp`. If the code beforehand was working as intended, then how should that test change so as to be proper usage? Is there example code somewhere of `producer.send(rtpPacket)` working?

ibc:
> * Create a `WebRTCTransport` and connect to it with a client and produce, as usual and as documented. This creates a producer; we'll call it `producer1`.
> * Create a `DirectTransport`, and use it to consume `producer1`, creating `consumer1`. Also create `producer2` on the DirectTransport, using identical `rtpParameters` to `producer1`.

Well, that's the problem. You are assuming that the `rtpParameters` of `producer1` are the same as those in `consumer1`, and that's not true at all. Check this (please, really check it):

* https://mediasoup.org/documentation/v3/mediasoup/rtp-parameters-and-capabilities/#RtpReceiveParameters

The `rtpParameters` of `consumer1` will have a single `encodings` with a random `ssrc`. So create `producer2` with the real RTP parameters, this is, `consumer1.rtpParameters`.

So there is no issue here in mediasoup but a wrong usage in your app. If you have further questions please let's move this discussion to the mediasoup forum.

dabbler0:
Thanks so much!

PR #476, opened 2020-11-05
Pipe consumers everywhere

dabbler0:
Adds a `pipe` boolean option to `Transport.consume()`, which if enabled creates a `PipeConsumer` instead of using the consumer type associated with the producer. This is for use in transports that intend to forward packets through to other producers and therefor need simulcast streams to be kept separate.


ibc:
I don't forget about this, I just need a few spare days.

ibc:
Closing in favour of #494 (which is the similar but with some changes and fixes).

PR #479, opened 2020-11-11
sctp fixes

jmillan:
- DataConsumer: make 'setBufferedAmountLowThreshold' trigger 'bufferedamountlow'
    if the buffered amount data is less or same than the given threshold.

- SctpAssociation: Increase buffered amount data before sending it in
  order to avoid inconsistencies if 'onSendSctpData' fires while 'usrsctp_senv'
  is being executed. Be conservative.

### DIFF
--- a/worker/src/RTC/DataConsumer.cpp
+++ b/worker/src/RTC/DataConsumer.cpp
@@ -164,6 +164,24 @@ namespace RTC
 
 				request->Accept();
 
+				// There is less or same buffered data than the given threshold.
+				// Trigger 'bufferedamountlow' now.
+				if (this->bufferedAmount <= this->bufferedAmountLowThreshold)
+				{
+					// Notify the Node DataConsumer.
+					json data = json::object();
+
+					data["bufferedAmount"] = this->bufferedAmount;
+
+					Channel::Notifier::Emit(this->id, "bufferedamountlow", data);
+				}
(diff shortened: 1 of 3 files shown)


jmillan:
We don't have full control on the real buffered data in usrsctp. With this changes we can subscribe to 'bufferedamountlow' event and set the buffered amount low threshold and we will be sure that:

* the event will fire if there is less or same buffered buffer as indicated.
* otherwise the event will fire as soon as there is less or same buffered buffer as indicated.

worker/src/RTC/DataConsumer.cpp
```diff
@@ -164,6 +164,24 @@ namespace RTC
...
+					data["bufferedAmount"] = this->bufferedAmount;
+
+					Channel::Notifier::Emit(this->id, "bufferedamountlow", data);
+				}
+				// Force the trigger of 'bufferedamountlow' once there is less or same
+				// buffered data than the given threshold.
+				else
+				{
+					this->forceTriggerBufferedAmountLow = true;
+				}
```
ibc, line 183: Why is this required at all? Isn't this the normal & expected behavior? Why do we need a "force" flag for that?

ibc: I understand now.

ibc (approved):

PR #480, opened 2020-11-17
Added possibility to use LD_LIBRARY_PATH

vitali2y:
Suggested quick fix allows to compile _mediasoup_ with non-standard or upgraded libs (e. g. I use it for running _mediasoup_ on _CentOS 7_ , which has deprecated _gcc_ (and, correspondent "_RuntimeError: gcc <= 4.8 not supported, please upgrade your gcc_" error), and, therefore it's required to manually compile and install a new _gcc-10.2.0_). So, preliminary setting _LD_LIBRARY_PATH_ (e. g. `LD_LIBRARY_PATH=/home/vyermolenko/build/gcc-10.2.0/lib64/:$LD_LIBRARY_PATH`) resolves the problem with this fix.

### DIFF
--- a/src/Worker.ts
+++ b/src/Worker.ts
@@ -269,7 +269,8 @@ export class Worker extends EnhancedEventEmitter
 			{
 				env :
 				{
-					MEDIASOUP_VERSION : '__MEDIASOUP_VERSION__'
+					MEDIASOUP_VERSION : '__MEDIASOUP_VERSION__',
+					LD_LIBRARY_PATH: process.env["LD_LIBRARY_PATH"]
 				},
 
 				detached : false,
--- a/src/Worker.ts
+++ b/src/Worker.ts
@@ -270,7 +270,7 @@ export class Worker extends EnhancedEventEmitter
 				env :
 				{
 					MEDIASOUP_VERSION : '__MEDIASOUP_VERSION__',
-					LD_LIBRARY_PATH: process.env["LD_LIBRARY_PATH"]
+					LD_LIBRARY_PATH   : process.env['LD_LIBRARY_PATH']
 				},
 
 				detached : false,


src/Worker.ts
```diff
@@ -269,7 +269,8 @@ export class Worker extends EnhancedEventEmitter
 			{
 				env :
 				{
-					MEDIASOUP_VERSION : '__MEDIASOUP_VERSION__'
+					MEDIASOUP_VERSION : '__MEDIASOUP_VERSION__',
+					LD_LIBRARY_PATH: process.env["LD_LIBRARY_PATH"]
```
ibc: You have not run `npm run lint` task.

ibc:
Is this really needed? Actually you can choose compiler via the `CC` environment variable:

```bash
cd mediasoup/worker
CC=/foo/gcc-xxx make
```

Is this enough or is `LD_LIBRARY_PATH` needed for something else at runtime?

ibc:
Any response to my comment please?

vitali2y:
First of all, pardon for the delay in reply.

> Is this really needed? Actually you can choose compiler via the `CC` environment variable:
> 
> ```shell
> cd mediasoup/worker
> CC=/foo/gcc-xxx make
> ```

I did not know about such `CC=xxx` possibility. But with using this approach I dunno whether it will help to find all compiled libs from `lib64` dir under "fresh" `/home/vyermolenko/build/gcc-10.2.0` compilation in my case.

> Is this enough or is `LD_LIBRARY_PATH` needed for something else at runtime?

I did not need `LD_LIBRARY_PATH` for runtime usage, but adding next two lines before official `npm install mediasoup@3 --save` installation command helped me in installing the `mediasoup` @ `CentOS 7` with "deprecated" `gcc 4.8.5` by default:

```
export PATH=/home/vyermolenko/build/gcc-10.2.0/bin/:$PATH
export LD_LIBRARY_PATH=/home/vyermolenko/build/gcc-10.2.0/lib64/:$LD_LIBRARY_PATH
```

src/Worker.ts
vitali2y: Fixed.

ibc:
Thanks, I'll comment next week.

vitali2y:
> I did not need `LD_LIBRARY_PATH` for runtime usage

Correction: after `mediasoup@3` installation the `node server.js` script cannot be started ("`Error: /lib64/libstdc++.so.6: version 'CXXABI_1.3.9' not found`") without preliminary exporting the `LD_LIBRARY_PATH` variable (as above).

ibc:
Hi, instead of passing individual environment variables to the mediasoup-worker process in runtime, I've made it inherit all envs here:

https://github.com/versatica/mediasoup/commit/52ae40ea92c8c7583a9be83bf974c199b7deb430

Will be available in 3.6.27 to be released in a few minutes.

PR #493, opened 2020-12-24
simplify header extension logic

woshiniuren:
simplify header extension logic


nazar-pc:
It was much more readable before

ibc:
> It was much more readable before

I agree. Closing.

PR #494, opened 2020-12-25
Add pipe boolean option to transport.consume()

ibc:
- So the receiver will get all streams.
- It works for any kind of transport (but PipeTransport which is always like this by design).
- Replaces PR #476.
- **TODO:** Document it.


src/Transport.ts
```diff
@@ -543,26 +544,29 @@ export class Transport extends EnhancedEventEmitter
...
+			}
 		}
 
 		const internal = { ...this._internal, consumerId: uuidv4(), producerId };
 		const reqData =
 		{
 			kind                   : producer.kind,
 			rtpParameters,
-			type                   : producer.type,
+			type                   : !pipe ? producer.type : 'pipe',
```
jmillan: cosmetic, just for readability:

`type: pipe ? pipe : producer.type`

jmillan (approved):
:+1:

src/Transport.ts
ibc: done

nazar-pc:
Does this mean that `PipeTransport` is now similar to `PlainTransport` with all consumers having `pipe: true` or there are some other differences?

src/ortc.ts
```diff
@@ -1043,53 +1044,74 @@ export function getConsumerRtpParameters(
...
 
-	// Set a single encoding for the Consumer.
-	consumerParams.encodings!.push(consumerEncoding);
+			if (rtxSupported)
+				encoding.rtx = { ssrc: baseRtxSsrc + i };
+			else
+				delete encoding.rtx;
 
-	// Copy verbatim.
-	consumerParams.rtcp = consumableParams.rtcp;
```
nazar-pc, line 1092: Was RTCP removed intentionally here?

ibc: It was already copied in the object created above at the beginning of the method.

ibc:
> Does this mean that `PipeTransport` is now similar to `PlainTransport` with all consumers having `pipe: true` or there are some other differences?

Right, differences are:

- plain transport still needs receiver rtpCapabilities in its consume() API.
- plain transport allows TCC and REMB

In v4 we are gonna merge both somehow.

nazar-pc:
I already have single `consume()` implementation in Rust port instead of separate for PipeTransport and with growing number of `if` statements it is harder to manage (updating it to match latest `v3` changes right now).

ibc:
That's why we want to merge both in v4 and simplify things. They have ended being the same thing with artificial constraints in both.

src/ortc.ts
albyho: Got it.

PR #498, opened 2020-12-28
worker: handle SIGILL signal in *nix

ibc:
As discussed in https://mediasoup.discourse.group/t/hardware-exceptions-handling-in-mediasoup-workers/.

**TODO:** Unclear whether `std::_Exit(EXIT_FAILURE)` is needed in the SIGILL signal hander. Probably not.

**TODO:** No idea how to do the Windows stuff since `__except (EXCEPTION_EXECUTE_HANDLER)` does not receive the exact error, so I cannot print it as we currently do.
   **NOTE:** May be done, see PR code.

### DIFF
--- a/worker/src/main.cpp
+++ b/worker/src/main.cpp
@@ -35,6 +35,9 @@ static constexpr int PayloadProducerChannelFd{ 6 };
 void IgnoreSignals();
 void HandleSignals();
 void SignalHandler(int signal, siginfo_t* info, void* ucontext);
+#ifdef _WIN32
+int WinExceptionHandler(unsigned int code, struct _EXCEPTION_POINTERS* ep);
+#endif
 
 int main(int argc, char* argv[])
 {
@@ -122,7 +125,11 @@ int main(int argc, char* argv[])
 	Settings::PrintConfiguration();
 	DepLibUV::PrintVersion();
 
+#ifndef _WIN32
 	try
+#else
+	__try
+#endif
 	{
 		// Initialize static stuff.
 		DepOpenSSL::ClassInit();
@@ -156,12 +163,19 @@ int main(int argc, char* argv[])
 
 		std::_Exit(EXIT_SUCCESS);
 	}
+#ifndef _WIN32
 	catch (const MediaSoupError& error)
 	{
 		MS_ERROR_STD("failure exit: %s", error.what());
 
 		std::_Exit(EXIT_FAILURE);
 	}
+#else
+	__except (WinExceptionHandler(GetExceptionCode(), GetExceptionInformation()))
+	{
+		std::_Exit(EXIT_FAILURE);
+	}
+#endif
 }
 
 void IgnoreSignals()
@@ -263,3 +277,13 @@ void SignalHandler(int signal, siginfo_t* info, void* /*ucontext*/)
 	}
 #endif
 }
+
+#ifdef _WIN32
+int WinExceptionHandler(unsigned int code, struct _EXCEPTION_POINTERS* /*ep*/)
+{
+	MS_ERROR_STD("failure exit [code:%u]", code);
+
+	// Execute exception handler.
+	return EXCEPTION_EXECUTE_HANDLER;
+}
+#endif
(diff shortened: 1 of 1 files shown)


snnz:
> **TODO:** No idea how to do the Windows stuff since `__except (EXCEPTION_EXECUTE_HANDLER)` does not receive the exact error, so I cannot print it as we currently do.

There is an exact error, it is just stored in some structure that is not accessed directly, but through the `GetExceptionCode` function (actually a macro): https://docs.microsoft.com/en-us/windows/win32/debug/getexceptioncode
It covers quite a lot of possible errors. And then there is GetExceptionInformation function that provides more information about an exception, rather technical.

ibc:
Can you provide example code of using those macros to print error name/message/info as we do now on the catch()?

ibc:
I don't understand anything. I see things like this and I don't know how to deal with them:

```c++
} __except (ExceptionFilter (GetExceptionInformation (), GetExceptionCode ())) {
```

NOTE: I don't have Windows.

ibc:
Ok, this may be done. Can you review @snnz?

BTW I'd also need help with this:

> **TODO:** Unclear whether `std::_Exit(EXIT_FAILURE)` is needed in the *unix SIGILL signal hander. Probably not.

See [here](https://github.com/versatica/mediasoup/pull/498/files#diff-f15b6b2f6a023cdae80e1bd4b4f994deaa515a18c700cc250e1baa7d4a633c6fR271).

snnz:
SEH (structural exception handling) is not a complete replacement for the regular C++ exception handling, so in Windows, it is better to leave try/catch blocks as they are, including the one in main(), but move all code that is currently in main into a separate function, and envelope its call in __try/__except that will catch exceptions unhandled by C++ RTL (separate functions for try/catch and __try/__except are necessary to avoid using expensive /EHa option).

Here is the sketch:

```
#ifndef _WIN32
  int main()
#else
  int win32_main()
#endif
  {
     try
     {
        ...
     }
     catch
     {
        ...
     }
  }

#ifdef _WIN32
  int main()
  {
    ___try
    {
       return win32_main();
    }
    ___except (EXCEPTION_EXECUTE_HANDLER)
    {
       // Report exception
    }
  }
#endif
```

Getting intelligible messages for the codes returned by GetExceptionCode() is rather tricky, because normal formatting of these errors is broken (believe it or not, they haven't fixed it for the past 25+ years!). If something more than just a bare code is necessary, then here is the way to do this:

```
#include <windows.h>
#include <tchar.h>

DWORD FormatNtStatus(ULONG nsCode, LPTSTR* ppszMessage)
{
	auto hNtDll = LoadLibrary(_T("NTDLL.DLL"));
	if (!hNtDll) return 0;
	typedef ULONG(*RtlNtStatusToDosErrorProc)(ULONG Status);
	auto RtlNtStatusToDosError = (RtlNtStatusToDosErrorProc)GetProcAddress(hNtDll, "RtlNtStatusToDosError");
	auto n = FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_FROM_HMODULE, hNtDll, RtlNtStatusToDosError(nsCode), MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPTSTR)ppszMessage, 0, NULL);
	FreeLibrary(hNtDll);
	return n;
}
```

And then, for example:

```
int main()
{
	__try
	{
		int *p = nullptr;
		return *p;
	}
	__except (EXCEPTION_EXECUTE_HANDLER)
	{
		LPTSTR pszMessage;
		if (FormatNtStatus(GetExceptionCode(), &pszMessage))
		{
			_fputts(pszMessage, stderr);
			LocalFree(pszMessage);
		}
		else _ftprintf(stderr, _T("Exception: 0x%08lX\n"), GetExceptionCode());
	}

}
```

will write something like "Invalid access to memory location".

snnz:
> **TODO:** Unclear whether `std::_Exit(EXIT_FAILURE)` is needed in the SIGILL signal hander. Probably not.

I am not that familiar with POSIX signal handling, but don't standards (POSIX and C) say that if handler function returns normally when handling SIGFPE, SIGILL or SIGSEGV, the behavior is undefined?

ibc:
Closing due to inactivity.

PR #500, opened 2021-01-06
Fix "SimulcastConsumer cannot switch layers if initial tsReferenceSpa…

### LINKED ISSUE #492: SimulcastConsumer cannot switch layers if initial tsReferenceSpatialLayer disappears
## Bug Report

SimulcastConsumer cannot switch layers if initial tsReferenceSpatialLayer disappears

### Your environment

- Operating system: Linux (multiple)
- Node version:  N/A
- npm version: N/A
- gcc/clang version: N/A
- mediasoup version: 3.6.24, 3.6.29 (Probably since 3.6.17, but only tested those)
- mediasoup-client version: 3.6.11

### Issue description

The problem happens if the higher spatial layer is never resumed. For example:

* Start with 3 spatial layers
* SimulcastConsumer will select layer 2 for tsReferenceSpatialLayer and currentSpatialLayer. (I believe this is only since the change in 3.6.17 (SimulcastConsumer: Prefer the highest spatial layer initially (PR #450).)
* Client sets max spatial layer to 0 before any Sender Reports are sent. (And leaves it at this forever).
* SimulcastConsumer will eventually notice layer 2 as inactive and set currentSpatialLayer to -1, but the logic in CanSwitchToSpatialLayer will not allow any other layer to be selected since GetProducerTsReferenceRtpStream()->GetSenderReportNtpMs() will remain false.

LewisWolfgang:
…tialLayer disappears" (#492)

* Set tsReferenceSpatialLayer later, only after receiving an RTCP Sender Report.
* Track and carry forward offset from tsReferenceSpatialLayer.

### DIFF
--- a/worker/include/RTC/SimulcastConsumer.hpp
+++ b/worker/include/RTC/SimulcastConsumer.hpp
@@ -111,7 +111,8 @@ namespace RTC
 		int16_t currentSpatialLayer{ -1 };
 		int16_t tsReferenceSpatialLayer{ -1 }; // Used for RTP TS sync.
 		std::unique_ptr<RTC::Codecs::EncodingContext> encodingContext;
-		uint32_t tsOffset{ 0u }; // RTP Timestamp offset.
+		uint32_t tsOffset{ 0u };          // RTP Timestamp offset.
+		uint32_t tsReferenceOffset{ 0u }; // RTP Timestamp offset from the tsReferenceSpatialLayer.
 		bool keyFrameForTsOffsetRequested{ false };
 		uint64_t lastBweDowngradeAtMs{ 0u }; // Last time we moved to lower spatial layer due to BWE.
 	};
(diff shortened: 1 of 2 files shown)


LewisWolfgang:
I had an oversight in my first commit: we need to only set tsReferenceSpatialLayer when currentSpatialLayer and targetSpatialLayer are the same, in order to always maintain consistency with the conditions defined in CanSwitchToSpatialLayer(). 
We are continuing to test on our side, but I hope there are no remaining issues with this fix.

ibc:
Actually we had something a bit different in mind. Unfortunately we are not gonna be able to review this PR no our planned approach for a while. Will notify here when possible.

ibc:
Sorry for the delay. Finally the issue this PR addressed was also solved in a merged PR https://github.com/versatica/mediasoup/pull/1098

PR #504, opened 2021-01-21
Fix lint

universeroc:
Use eslint-disable-line to bypass eslint for special case

I check the commit history find the changes is made in 2020, but no one found it breaks down the eslint.

I suggest to make a pre-commit hook to run lint at least.

### DIFF
--- a/src/Consumer.ts
+++ b/src/Consumer.ts
@@ -224,7 +224,7 @@ export class Consumer extends EnhancedEventEmitter
 			appData,
 			paused,
 			producerPaused,
-			score = { score: 10, producerScore: 10, producerScores: [] },
+			score = { score: 10, producerScore: 10, producerScores: [] }, // eslint-disable-line
 			preferredLayers
 		}:
 		{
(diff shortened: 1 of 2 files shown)


ibc:
Thanks, but the problem is in `@typescript-eslint/no-use-before-define` and `no-undef` ESLint rules, not in mediasoup. This did not happen before. On it, but we don't want to ignore tons of lines just to make `eslint` or `@typescript-eslint` buggy versions happy.

ibc:
Here is the problem BTW:

- https://github.com/typescript-eslint/typescript-eslint/issues/2941
- https://github.com/typescript-eslint/typescript-eslint/issues/2942

universeroc:
> Thanks, but the problem is in `@typescript-eslint/no-use-before-define` and `no-undef` ESLint rules, not in mediasoup. This did not happen before. On it, but we don't want to ignore tons of lines just to make `eslint` or `@typescript-eslint` buddy versions happy.

Yes, agree with your reply, and current settings not work when running lint, so what's your advice to pass the linters?

Maybe downgrade the eslint version firstly to go through or wait for its patch, we dependent the linter to make sure the coding style.

ibc:
Here: https://github.com/versatica/mediasoup/issues/507

universeroc:
> Here: #507

OK

Issue #397, opened 2020-05-11
Move to GitHub Actions?


ibc:
Just wondering if there is any benefit in moving from Travis-CI to GitHub Actions. It seems that GitHub Actions makes it easier to test in Mac and Windows. Example:

* https://github.com/eslint/eslint/blob/master/.github/workflows/ci.yml

jmillan:
I was thinking on the very same.

ibc:
Thanks to @nazar-pc for PR #521 that fixes this issue.

PR #518, opened 2021-02-27
Rust port of mediasoup!

nazar-pc:
What this massive PR includes:
* Complete Rust port of the TypeScript mediasoup library
* Port of the relevant tests from TypeScript
* Some Rust-specific tests
* Built-in documentation (as generally expected in Rust ecosystem)
* Echo example app (server in Rust + client in TypeScript)
* CI config that checks formatting with rustfmt, lints code with clippy and runs tests

There are minor cleanups in worker and existing tests, but nothing significant (let me know if that is an issue).
They are primarily related to stricter typing generally used in Rust.

Since there is #397 open and adding CI for Rust into existing setup would be challenging, I decided to use GitHub actions instead, hopefully that is OK.

Code style used is default for Rust ecosystem with no tweaks. For TypeScript code used in echo example (the only one so far) I used already established code style from this repository to make it consistent.

I acknowledge how big this is and would be happy to address any feedback there might be. Please start with documentation on docs.rs, it explains some high-level information about this port.

References:
* discussion on forum: https://mediasoup.discourse.group/t/mediasoup-library-in-rust/1803
* releases on crates.io: https://crates.io/crates/mediasoup
* documentation on docs.rs: https://docs.rs/mediasoup

As discussed with Iñaki, there will be some complementary activities related to this around website documentation and such.


.gitignore
```diff
@@ -37,3 +37,10 @@
 
 # Vistual Studio Code stuff.
 /.vscode/
+
+# JetBrains IDE stuff
```
ibc:
```suggestion
# JetBrains IDE stuff.
```

test/test-Consumer.js
```diff
@@ -681,10 +681,7 @@ test('consumer.dump() succeeds', async () =>
 				{
 					ssrc : videoConsumer.rtpParameters.encodings[0].rtx.ssrc
 				},
-				scalabilityMode : 'S4T1',
-				spatialLayers   : 4,
-				temporalLayers  : 1,
-				ksvc            : false
```
ibc, line 687: Why are these fields removed? Actually they represent real fields that must be set.

test/test-WebRtcTransport.js
```diff
@@ -96,15 +96,12 @@ test('router.createWebRtcTransport() succeeds', async () =>
...
-			OS                 : 2048,
-			MIS                : 2048,
-			maxMessageSize     : 1000000,
-			isDataChannel      : true,
-			sctpBufferedAmount : 0,
-			sendBufferSize     : 262144
+			port           : 5000,
+			OS             : 2048,
+			MIS            : 2048,
+			maxMessageSize : 1000000
```
ibc, line 104: Same here, why so many fields removed?

ibc (commented):
Amazing. I must do a proper review and it will take some days before I finish. For now just a few comments:

- I'd like to move current Travis-CI to GitHub CI entirely and not just for Rust.
- I don't like the idea of including frontend examples in `mediasoup/rust/` folder.
- I don't understand why there is `rust/.gitignore` instead of just adding entries to main `.gitignore`.
- Why is mediasoup rust version 0.5.0?

worker/src/RTC/SctpAssociation.cpp
```diff
@@ -356,15 +356,6 @@ namespace RTC
 
 		// Add maxMessageSize.
 		jsonObject["maxMessageSize"] = this->maxSctpMessageSize;
-
```
jmillan: Why this removals?

nazar-pc:
> I'd like to move current Travis-CI to GitHub CI entirely and not just for Rust.

Files for Rust and Node.js will most likely be different, so one doesn't depend on the other. I will try to find time for Node.js workflow in a separate PR. Here I just didn't want to submit it without CI in place.

> I don't like the idea of including frontend examples in `mediasoup/rust/` folder.
> I don't understand why there is `rust/.gitignore` instead of just adding entries to main `.gitignore`.

Moved `.gitignore` entries to the root in last commit

nazar-pc:
> I'd like to move current Travis-CI to GitHub CI entirely and not just for Rust.

Files for Rust and Node.js will most likely be different, so one doesn't depend on the other. I will try to find time for Node.js workflow in a separate PR. Here I just didn't want to submit Rust version without CI in place from the beginning.

> I don't like the idea of including frontend examples in `mediasoup/rust/` folder.

There is a convension in Rust (supported by Cargo) to have `examples` folder, so you can do `cargo run --example echo`.
So it felt natural to have frontend directory nearby.

There is no identical backend example in TypeScript so it seems fine to me to have it there.

I have no stong feelings about it if you want me to move it somewhere.
I can even move Rust examples and corresponding frontend to the root or maybe even to separate repo if you want (but that would be harder to maintain, cargo workspaces are great and everything is compiled together in CI).

> I don't understand why there is `rust/.gitignore` instead of just adding entries to main `.gitignore`.

Moved `.gitignore` entries to the root in last commit

> Why is mediasoup rust version 0.5.0?

This is less similar to TypeScript library and more similar to mediasoup-client case.
Rust version will have different things that will be considered a breaking change (for instance #503 broke dump in Rust because they are heavily typed there) and I don't think versioning will line-up perfectly.
And there will probably be more things that will break until it reaches the level of API stability of mediasoup itself.

I am open to suggestions how to make it clearer which version of the worker/mediasoup Rust version corresponds to (I'm also exploring how to turn worker into Rust crate to it can be installed as a dependency automatically).

Updated ticket with similarities/differences with version info.

test/test-Consumer.js
nazar-pc: The reason is that these are inconsistent with defined types.
https://mediasoup.org/documentation/v3/mediasoup/rtp-parameters-and-capabilities/#RtpParameters doesn't have `spatialLayers`, `temporalLayers` and `ksvc` in it and thus identical test in Rust wasn't possible to write because it just wouldn't compile.

Since this is a dump and the only place where I found these properties are used were tests, I decided to remove them from tests and from corresponding place in the worker itself. At the end of the day `scalabilityMode` already encapsulates all three and you can parse it into those fields if really needed.

test/test-WebRtcTransport.js
nazar-pc: This property represents https://mediasoup.org/documentation/v3/mediasoup/sctp-parameters/#SctpParameters struct, which doesn't have fields `isDataChannel`, `sctpBufferedAmount` and `sendBufferSize`. Identical Rust version of the test again didn't compile due to type mismatch.

These were again only used in tests and never described in public API, so I decided to clean that up and remove both from tests and worker itself for consistency.

If those properties are important to check, they need to be extracted from dump or elsewhere, not placed into `SctpParameters` where they don't belong.

worker/src/RTC/SctpAssociation.cpp
nazar-pc: Answered in reply to corresponding test change above.

test/test-Consumer.js
ibc: It's not a dump, these fields are part of "our" `RtpEncodingParameters` type. We may have to properly document them indeed, but they are needed. In fact we use them in production.

BTW `scalabilityMode` is a real W3 field defined in an additional spec: https://www.w3.org/TR/webrtc-svc/

ibc:
> > I'd like to move current Travis-CI to GitHub CI entirely and not just for Rust.
> 
> Files for Rust and Node.js will most likely be different, so one doesn't depend on the other. I will try to find time for Node.js workflow in a separate PR. Here I just didn't want to submit Rust version without CI in place from the beginning.

Thanks. In fact we are having problems with Travis-CI due to its new limits.
 
> > I don't like the idea of including frontend examples in `mediasoup/rust/` folder.
> 
> There is a convension in Rust (supported by Cargo) to have `examples` folder, so you can do `cargo run --example echo`.
> So it felt natural to have frontend directory nearby.

Perfect then :)

> > Why is mediasoup rust version 0.5.0?
> 
> This is less similar to TypeScript library and more similar to mediasoup-client case.
> Rust version will have different things that will be considered a breaking change (for instance #503 broke dump in Rust because they are heavily typed there) and I don't think versioning will line-up perfectly.
> And there will probably be more things that will break until it reaches the level of API stability of mediasoup itself.

Ok, let's do this way for now and let's see in the future.

NOTE: I still have to review more things and won't be able to finish this week. Will do next one, promise.

test/test-Consumer.js
nazar-pc: Hm... this particular `data` is a dump as you can see from the call above and `data.rtpParameters` indeed comes from `RtpEncodingParameters::FillJson` in the worker, which is an implementation for dumping said object.

Neither documentation nor TypeScript code has `spatialLayers`, `temporalLayers` or `ksvc` here.

I did NOT remove `scalabilityMode` (which I think is the source of confusion here). It is red on this side of the diff, but it is also green on the other side. Is that your concern here?

nazar-pc:
Tests are still a bit flaky at times, but I'll probably be ironing them out in next version instead of introducing more changes to this PR.

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
```diff
@@ -151,9 +151,6 @@ namespace RTC
 		if (!this->scalabilityMode.empty())
 		{
 			jsonObject["scalabilityMode"] = this->scalabilityMode;
-			jsonObject["spatialLayers"]   = this->spatialLayers;
-			jsonObject["temporalLayers"]  = this->temporalLayers;
-			jsonObject["ksvc"]            = this->ksvc;
```
ibc: Can we please keep those? We don't need to remove them from `dump()` method whose output is not even documented (private API just for debugging). We really need them for debugging.

worker/src/RTC/SctpAssociation.cpp
```diff
@@ -356,15 +356,6 @@ namespace RTC
...
-		// Add sendBufferSize.
-		jsonObject["sendBufferSize"] = this->sctpSendBufferSize;
-
-		// Add sctpBufferedAmountLowThreshold.
-		jsonObject["sctpBufferedAmount"] = this->sctpBufferedAmount;
-
-		// Add isDataChannel.
-		jsonObject["isDataChannel"] = this->isDataChannel;
```
ibc: Same. Can we please keep those? We don't need to remove them from `dump()` method whose output is not even documented (private API just for debugging). We really need them for debugging.

ibc (changes requested):
Please address just these two comments and let's merge.

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
ibc: I'm gonna do it

worker/src/RTC/SctpAssociation.cpp
ibc: I'm gonna do it

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
ibc: done here: https://github.com/versatica/mediasoup/pull/518/commits/b2047f73512197086ec1129098c4be8e6fdf8e2b

worker/src/RTC/SctpAssociation.cpp
ibc: done here: https://github.com/versatica/mediasoup/pull/518/commits/b2047f73512197086ec1129098c4be8e6fdf8e2b

ibc (approved):

ibc:
waiting for tests to pass in CI

ibc (approved):

PR #520, opened 2021-03-03
Prefer python2, but fallback to python3 if not found

nazar-pc:
This doesn't break anything that works already, but allows installation on Ubuntu 20.10 (maybe older too) without having to specify `PYTHON=python3` every time. I had to do this way too often already :slightly_smiling_face:

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -4,7 +4,7 @@
 
 # Best effort to prefer Python 2 executable since there are yet pending issues
 # with gyp and Python3.
-PYTHON ?= $(shell command -v python2 2> /dev/null || echo python)
+PYTHON ?= $(shell command -v python2 2> /dev/null || command -v python 2> /dev/null || echo python3)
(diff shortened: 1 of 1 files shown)


ibc (approved):
nice

nagkumar:
What is the exact use of using the latest python? Does performance issues with python3 impact mediasoup post-install step.
Is python required even at runtime or only at build time, can we take node image without build-essential and python at run and run mediasoup nodejs application?

nazar-pc:
Python is only needed during installation to build the worker

nagkumar:
Thank You Nazar. This info helps a lot to cut our docker image size.

nagkumar:
Also, can you point me to any info, on why python is used by media soup, I would expect it to be written in C or C+ for any complex functionality as it is related to Video and Audio streams. any insights on the worker its role could help.

nazar-pc:
Look at the source code then, it is all there. This PR is not a discussion forum.

Issue #359, opened 2020-02-13
Drop gyp, move to CMake


ibc:
Rationale:

* Nobody is maintaining `gyp`.
* `libuv` is dropping `gyp`.
* We (mediasoup authors) have to provide and maintain our own `xxx.gyp` for `libsrtp`, `usrsctp` and more deps we use, while those libraries do support `CMake` out of the box.
* There is `gyp` issues in Windows that nobody fixes (even if reported in `gyp` project) and affect mediasoup (see #345 and #347).

So let's move to `CMake`:

* Should choose a `CMake` version available in most common platforms (Debian, Ubuntu, OSX, etc). 
* Avoid the very latest and coolest `CMake` version since some of its features are not implemented by most common versions such as 3.7.

Ping @saghul because I can.

saghul:
Current worker deps:

|Name|CMake support upstream?|
|--|--|
|catch| ✅ |
|getopt| ✅ |
|json| ✅ |
|lcov| 🚫 |
|libsrtp| ✅ |
|libuv| ✅ |
|libwebrtc| 🚫  |
|netstring| 🚫  |
|openssl| 🚫 |
|usrsctp| ✅ |

Out of these, I'd say OpenSSL is the trickiest one. An option would be to move to BoringSSL, which does support CMake out of the box. The rest are pretty small and writing CMakefiles for them should be trivial.

ibc:
Wow, thanks! BTW:

* `lcov` is not really used in mediasoup (I mean, it's not built).
* Regarding `libwebrtc` we only use a subset of files so it shouldn't be hard to make a `CMakelists.txt` for it (wait, it depends on `abseil-cpp` lib but it does provide `CMakelists.txt`).
* `netstring` is two files code :)
* `openssl`... wow, didn't know that... Moving to BoringSSL is an option, of course, but somehow I don't want to rely on so many Google-managed projects... sad.

saghul:
>     * `openssl`... wow, didn't know that... Moving to BoringSSL is an option, of course, but somehow I don't want to rely on so many Google-managed projects... sad.

It already has the necessary APIs for QUIC unlike OpenSSL (still in PR) so maybe it's a sign from, the future! 😎

barry-ran:
[OpenSSL-CMake](https://github.com/pol51/OpenSSL-CMake)

ibc:
> [OpenSSL-CMake](https://github.com/pol51/OpenSSL-CMake)

It's a 4 years old project.

[...]

PR #522, opened 2021-03-06
Remove unnecessary files from gyp configs to fix linking issues

nazar-pc:
I was doing some hacking and found that linker is really unhappy about those files. Removing them doesn't change anything negatively, tests still pass, so I think this is a correct change.

The error looked like this:
```
  = note: /usr/bin/ld: /web/github/mediasoup/worker/out/Release/libuv.a(sysinfo-loadavg.o): in function `uv_loadavg':
          /web/github/mediasoup/worker/out/../deps/libuv/libuv/src/unix/sysinfo-loadavg.c:28: multiple definition of `uv_loadavg'; /web/github/mediasoup/worker/out/Release/libuv.a(linux-core.o):/web/github/mediasoup/worker/out/../deps/libuv/libuv/src/unix/linux-core.c:1132: first defined here
          /usr/bin/ld: /web/github/mediasoup/worker/out/Release/libabseil.a(print_hash_of.o): in function `main':
          /web/github/mediasoup/worker/out/../deps/libwebrtc/deps/abseil-cpp/abseil-cpp/absl/hash/internal/print_hash_of.cc:20: multiple definition of `main'; /web/github/mediasoup/target/debug/deps/mediasoup_sys-ad2c49156d6c0bb6.37125jqqwh13ju51.rcgu.o:37125jqqwh13ju51:(.text.main+0x0): first defined here
          collect2: error: ld returned 1 exit status
```

### DIFF
--- a/worker/deps/libuv/uv.gyp
+++ b/worker/deps/libuv/uv.gyp
@@ -245,7 +245,6 @@
             'libuv/src/unix/procfs-exepath.c',
             'libuv/src/unix/random-getrandom.c',
             'libuv/src/unix/random-sysctl-linux.c',
-            'libuv/src/unix/sysinfo-loadavg.c',
           ],
           'link_settings': {
             'libraries': [ '-ldl', '-lrt' ],
@@ -262,7 +261,6 @@
             'libuv/src/unix/procfs-exepath.c',
             'libuv/src/unix/random-getrandom.c',
             'libuv/src/unix/random-sysctl-linux.c',
-            'libuv/src/unix/sysinfo-loadavg.c',
           ],
           'link_settings': {
             'libraries': [ '-ldl' ],
--- a/worker/deps/libwebrtc/deps/abseil-cpp/abseil-cpp.gyp
+++ b/worker/deps/libwebrtc/deps/abseil-cpp/abseil-cpp.gyp
@@ -153,7 +153,6 @@
         'abseil-cpp/absl/synchronization/mutex.h',
         'abseil-cpp/absl/hash/internal/hash.cc',
         'abseil-cpp/absl/hash/internal/spy_hash_state.h',
-        'abseil-cpp/absl/hash/internal/print_hash_of.cc',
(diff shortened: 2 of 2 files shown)


jmillan:
> I was doing some hacking and found that linker is really unhappy about those files.

Can you please give some detail here?

nazar-pc:
> Can you please give some detail here?

Yes, those are the same I mentioned on the forum thread.

In order to make mediasoup-worker automatically installable as a regular dependency in Rust port, I'm working on turning it into a library (it will be built and linked upon installation automatically, like it does when installing NPM package right now, but for Rust and as library, not executable).

For that I started with creating `lib.cpp`, which is basically `main.cpp` with `main()` function renamed to `run()` and some minor tweaks. I then build that into static library and link into Rust library that calls `run()`, here is simple wrapper I have right now: https://github.com/nazar-pc/mediasoup/blob/02529bc44bf1c7ea90b910beac8d0b9c5a0f3979/worker/src/lib.rs

The issue happens when I try to link those static C/C++ libraries into Rust library. Apparently `uv_loadavg` is defined in both `linux-core.c` and `sysinfo-loadavg.c`, so only the first one needs to be included on Linux/Android.

Similarly `main()` function in `print_hash_of.cc` is not necessary too I think.

I don't know about linking mechanics enough to explain why it doesn't error when building worker regularly, but I worker still runs just fine with those changes, so I believe they are safe and correct to do.

ibc:
I'm fine with these changes. Note that `uv.gyp` is written by us (libuv removed GYP support time ago) and we have to switch to CMake anyway.

@jmillan I'll merge, ok?

jmillan:
> @jmillan I'll merge, ok?

:+1:

nazar-pc:
CMake would be great to have! But I do not volunteer for it :wink:

ibc:
> CMake would be great to have! But I do not volunteer for it 😉

There is ongoing work (unknown current status): https://github.com/versatica/mediasoup/issues/359

PR #526, opened 2021-03-10
Added missing delete of the callbacks functions

vpalmisano:
The delete operator is missing in the `OnUvSend` method, leading to a possible memory leak.

### DIFF
--- a/worker/src/handles/UdpSocket.cpp
+++ b/worker/src/handles/UdpSocket.cpp
@@ -287,7 +287,11 @@ inline void UdpSocket::OnUvSend(int status, UdpSocket::onSendCallback* cb)
 	if (status == 0)
 	{
 		if (cb)
+		{
 			(*cb)(true);
+
+			delete cb;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


jmillan:
PR looks good to me. Although I don't think we are leaking a pointer for each RTP packet we are sending.., that should be 100% noticeable.

ibc:
If this was true (we are leaking in **every sent UDP packet**) we would have noticed it, it's just impossible. Also, same would apply to `TcpConnection::OnUvWrite()` method.

**However** it makes sense. We always do `delete cb;` everywhere (in all `XxxxTransport.cpp`).

vpalmisano:
@jmillan I agree, but I see that the memory allocation of the worker process seems increasing of a small amount over time, I'm trying to exclude all the potential memory leaks.

ibc:
@jmillan @vpalmisano, we may NOT have noticed this because such a leak would just happen when a UDP packet cannot be directly sent (due to the kernel UDP buffer being full) which does not happen often.

ibc:
I'm extending code in this PR and will commit soon and announce here.

ibc:
Ok, I've done this: https://github.com/versatica/mediasoup/pull/527. 

Thanks a lot @vpalmisano, amazing catch. I cannot believe that we didn't see it before.

jmillan:
Thanks @vpalmisano!

PR #533, opened 2021-03-13
Mediasoup worker as a library: libmediasoup-worker

nazar-pc:
#### Motivation:
There is an ergonomic issue with #518 that requires user to build worker separately and then specifying a path to its executable during instantiation. Ideally worker would also be a library, which would allow it to be packaged into Rust crate and seamlessly installed as any other dependency.

I worked on achieving this in my fork and succeeded in doing so, this PR is a distilled worker-specific set of changes without anything Rust-specific in order to separate review process for each. I also think this could be useful for those using worker in other languages.

#### This PR contains a few things:
* adds `libmediasoup-worker` static library build, which is the almost identical to `mediasoup-worker`, but presents itself as a library instead of an executable
* regular executable `mediasoup-worker` now shares code with `libmediasoup-worker`
* `mediasoup-worker`'s internals are now aware of multiple threads, namely:
  * truly global resources are guarded by mutexes
  * data specific to worker instance are made thread-local
  * above allows to have multiple concurrent workers running each in their own thread withing parent application
* `worker.close` message added such that it is possible to stop worker that is running in the thread and can't be controlled with signals
* with a thin wrapper above library can be turned into Rust crate (build script as of right now: https://github.com/nazar-pc/mediasoup/blob/a67a5ad535315979e273644a9f215ac24c79e451/worker/build.rs)

API of the introduced `run_worker()` is intentionally very close to `int main(argc, argv)` of the executable itself, more convenient API can be created later if ever needed. PR wasn't squashed to simplify review process.

P.S. This needs careful review and probably some more extensive testing, but it seems to work from my testing so far, passes all TypeScript, C++ and Rust tests in my fork. The primary concern is around data channels as usrsctp internals are really hard to understand and are not documented.

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -16,7 +16,7 @@ DOCKER ?= docker
 
 .PHONY:	\
 	default clean clean-all xcode lint format test bear tidy \
-	fuzzer fuzzer-run-all docker-build docker-run
+	fuzzer fuzzer-run-all docker-build docker-run libmediasoup-worker
 
 default:
 ifeq ($(MEDIASOUP_WORKER_BIN),)
@@ -101,3 +101,7 @@ docker-run:
 		--cap-add SYS_PTRACE \
 		-v $(shell pwd)/../:/mediasoup \
 		mediasoup/docker:latest
+
+libmediasoup-worker:
+	$(PYTHON) ./scripts/configure.py --no-duplicate-basename-check -R libmediasoup-worker
+	$(MAKE) -j$(CORES) BUILDTYPE=$(MEDIASOUP_BUILDTYPE) -C out
--- a/worker/include/RTC/DtlsTransport.hpp
+++ b/worker/include/RTC/DtlsTransport.hpp
@@ -137,15 +137,15 @@ namespace RTC
 		static void GenerateFingerprints();
 
 	private:
-		static X509* certificate;
-		static EVP_PKEY* privateKey;
-		static SSL_CTX* sslCtx;
-		static uint8_t sslReadBuffer[];
-		static std::map<std::string, Role> string2Role;
-		static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
-		static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
-		static std::vector<Fingerprint> localFingerprints;
-		static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
+		thread_local static X509* certificate;
+		thread_local static EVP_PKEY* privateKey;
+		thread_local static SSL_CTX* sslCtx;
+		thread_local static uint8_t sslReadBuffer[];
+		thread_local static std::map<std::string, Role> string2Role;
+		thread_local static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
+		thread_local static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
+		thread_local static std::vector<Fingerprint> localFingerprints;
+		thread_local static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
 
 	public:
 		explicit DtlsTransport(Listener* listener);
--- a/worker/include/RTC/DtlsTransport.hpp
+++ b/worker/include/RTC/DtlsTransport.hpp
@@ -145,7 +145,7 @@ namespace RTC
 		static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
 		static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
 		thread_local static std::vector<Fingerprint> localFingerprints;
-		thread_local static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
+		static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
 
 	public:
 		explicit DtlsTransport(Listener* listener);
--- a/worker/include/RTC/DtlsTransport.hpp
+++ b/worker/include/RTC/DtlsTransport.hpp
@@ -141,9 +141,9 @@ namespace RTC
 		thread_local static EVP_PKEY* privateKey;
 		thread_local static SSL_CTX* sslCtx;
 		thread_local static uint8_t sslReadBuffer[];
-		thread_local static std::map<std::string, Role> string2Role;
-		thread_local static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
-		thread_local static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
+		static std::map<std::string, Role> string2Role;
+		static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
+		static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
 		thread_local static std::vector<Fingerprint> localFingerprints;
 		thread_local static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
 
--- a/worker/src/Channel/UnixStreamSocket.cpp
+++ b/worker/src/Channel/UnixStreamSocket.cpp
@@ -87,7 +87,7 @@ namespace Channel
 
 		if (nsPayloadLen == 0)
 		{
-			nsNumLen       = 1;
+			nsNumLen             = 1;
 			this->WriteBuffer[0] = '0';
 			this->WriteBuffer[1] = ':';
 			this->WriteBuffer[2] = ',';
--- a/worker/src/Channel/UnixStreamSocket.cpp
+++ b/worker/src/Channel/UnixStreamSocket.cpp
@@ -19,18 +19,21 @@ namespace Channel
 	// netstring length for a 4194304 bytes payload.
 	static constexpr size_t NsMessageMaxLen{ 4194313 };
 	static constexpr size_t NsPayloadMaxLen{ 4194304 };
-	static uint8_t WriteBuffer[NsMessageMaxLen];
 
 	/* Instance methods. */
 	UnixStreamSocket::UnixStreamSocket(int consumerFd, int producerFd)
 	  : consumerSocket(consumerFd, NsMessageMaxLen, this), producerSocket(producerFd, NsMessageMaxLen)
 	{
 		MS_TRACE_STD();
+
+		this->WriteBuffer = (uint8_t*)std::malloc(NsMessageMaxLen);
 	}
 
 	UnixStreamSocket::~UnixStreamSocket()
 	{
 		MS_TRACE_STD();
+
+		std::free(this->WriteBuffer);
 	}
 
 	void UnixStreamSocket::SetListener(Listener* listener)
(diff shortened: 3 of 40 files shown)


worker/src/handles/TcpConnection.cpp
```diff
@@ -108,7 +108,7 @@ void TcpConnection::Close()
 		err       = uv_shutdown(
       req, reinterpret_cast<uv_stream_t*>(this->uvHandle), static_cast<uv_shutdown_cb>(onShutdown));
 
-		if (err != 0)
+		if (err != 0 && err != UV_ENOTCONN)
```
ibc: Why this?

nazar-pc: Sorry, I should have probably skipped this here.

Basically I couldn't figure out why it ends up with this error instead of shutting down properly in Rust version no matter the order things are happening in. It never sets `isClosedByPeer` to `true` since `Close()` is called directly. But I though that if on shutdown connection is already dropped, this is not something that is worth killing the process.

It wasn't an issue when worker was a process, but it is when running in a thread since it takes the whole app with it.

I might be understanding it wrong, please correct me if that is the case.

worker/include/RTC/DtlsTransport.hpp
```diff
@@ -137,15 +137,15 @@ namespace RTC
...
-		static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
-		static std::vector<Fingerprint> localFingerprints;
-		static std::vector<SrtpCryptoSuiteMapEntry> srtpCryptoSuites;
+		thread_local static X509* certificate;
+		thread_local static EVP_PKEY* privateKey;
+		thread_local static SSL_CTX* sslCtx;
+		thread_local static uint8_t sslReadBuffer[];
+		thread_local static std::map<std::string, Role> string2Role;
+		thread_local static std::map<std::string, FingerprintAlgorithm> string2FingerprintAlgorithm;
+		thread_local static std::map<FingerprintAlgorithm, std::string> fingerprintAlgorithm2String;
```
jmillan: The above 3 maps don't need to be thread local.

worker/src/Channel/UnixStreamSocket.cpp
```diff
@@ -19,18 +19,21 @@ namespace Channel
 	// netstring length for a 4194304 bytes payload.
 	static constexpr size_t NsMessageMaxLen{ 4194313 };
 	static constexpr size_t NsPayloadMaxLen{ 4194304 };
-	static uint8_t WriteBuffer[NsMessageMaxLen];
```
ibc, line 22: Instead of removing this, why not making it thread static?

If not possible (why?) then let's use modern C++ cast instead of `(uint8_t*)std::malloc(NsMessageMaxLen);`. Let's use `<static_cast>` or whatever (I don't know right now). Legacy C style cast produces errors when running `make tidy` in `worker/` folder.

worker/src/DepLibSRTP.cpp
```diff
@@ -45,17 +49,31 @@ void DepLibSRTP::ClassInit()
...
 
-	srtp_err_status_t err = srtp_init();
+			srtp_err_status_t err = srtp_init();
 
-	if (DepLibSRTP::IsError(err))
-		MS_THROW_ERROR("srtp_init() failed: %s", DepLibSRTP::GetErrorString(err));
+			if (DepLibSRTP::IsError(err))
+				MS_THROW_ERROR("srtp_init() failed: %s", DepLibSRTP::GetErrorString(err));
+		}
+		++globalInstances;
```
ibc: Please blank line after code clocks everywhere:

```suggestion
		}

		++globalInstances;
```

Also, is the purpose of those `globalSyncMutex` just protect the logging based on `globalInstances`?

worker/src/DepUsrSCTP.cpp
```diff
@@ -62,14 +65,21 @@ void DepUsrSCTP::ClassInit()
...
-	// Disable explicit congestion notifications (ecn).
-	usrsctp_sysctl_set_sctp_ecn_enable(0);
+			// Disable explicit congestion notifications (ecn).
+			usrsctp_sysctl_set_sctp_ecn_enable(0);
 
 #ifdef SCTP_DEBUG
-	usrsctp_sysctl_set_sctp_debug_on(SCTP_DEBUG_ALL);
+			usrsctp_sysctl_set_sctp_debug_on(SCTP_DEBUG_ALL);
 #endif
+		}
```
ibc, line 81: same

worker/src/DepLibSRTP.cpp
```diff
@@ -45,17 +49,31 @@ void DepLibSRTP::ClassInit()
...
 
 void DepLibSRTP::ClassDestroy()
 {
 	MS_TRACE();
 
-	srtp_shutdown();
+	{
+		std::lock_guard<std::mutex> lock(globalSyncMutex);
+		--globalInstances;
+		if (globalInstances == 0)
```
ibc: same, blank line before block codes, please

worker/src/DepUsrSCTP.cpp
```diff
@@ -62,14 +65,21 @@ void DepUsrSCTP::ClassInit()
 
 	MS_DEBUG_TAG(info, "usrsctp");
 
-	usrsctp_init_nothreads(0, onSendSctpData, sctpDebug);
+	{
+		std::lock_guard<std::mutex> lock(globalSyncMutex);
+		if (globalInstances == 0)
```
ibc: same

worker/src/DepUsrSCTP.cpp
```diff
@@ -78,7 +88,19 @@ void DepUsrSCTP::ClassDestroy()
 {
 	MS_TRACE();
 
-	usrsctp_finish();
+	{
+		std::lock_guard<std::mutex> lock(globalSyncMutex);
+		--globalInstances;
+		if (globalInstances == 0)
```
ibc: same

worker/src/Settings.cpp
```diff
@@ -62,6 +67,10 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 	/* Parse command line options. */
 
+	// getopt_long_only() is not thread-safe
+	std::lock_guard<std::mutex> lock(globalSyncMutex);
+
+	optind = 1; // Set explicitly, otherwise subsequent runs will fail
```
ibc: Dot at the end of comments, please. Also, why this? Can you explain please?

```suggestion
	optind = 1; // Set explicitly, otherwise subsequent runs will fail.
```

worker/src/handles/TcpConnection.cpp
ibc: Ok, we just call `MS_ABORT` when something should NOT happen at all (or we think that it should never happen). Isn't this true? is it legitimate for this to happen? when?

ibc: May be `uv_shutdown()` can exit with `UV_ENOTCONN` if the process is closed before it completes?

worker/Makefile
```diff
@@ -101,3 +101,7 @@ docker-run:
 		--cap-add SYS_PTRACE \
 		-v $(shell pwd)/../:/mediasoup \
 		mediasoup/docker:latest
+
+libmediasoup-worker:
```
ibc, line 105: Do we have to call `make libmediasoup-worker` manually? or how is it called? If it makes sense to call it manually, can you document it in `doc/Building.md`?

worker/include/RTC/DtlsTransport.hpp
nazar-pc: Agreed, reverted

worker/src/Channel/UnixStreamSocket.cpp
nazar-pc: First of all: global state is bad, less global state - better.
But more significant reason is that thead_local storage is limited and the same for all threads of the eventual app if I understand it correctly. So by storing there something that could just as well be stored in object's property, we're wasting precious resource.
Sure, it is possible to store it on the heap, but what is the benefit there?

worker/Makefile
nazar-pc: Yes, defaults didn't change, needs to be called manually. Added to the doc.

worker/src/DepLibSRTP.cpp
nazar-pc: That and `srtp_init()` call, since it only needs to be called once globally.

Linter didn't complain about that :shrug:
Fixed :slightly_smiling_face:

worker/src/Settings.cpp
nazar-pc: I was very surprised that `getopt_long_only()` uses those global variables.
`optind` is initialized as `1` by default and is incremented as `getopt_long_only()` iterates through `argv`. Since now we are expecting this function to be called multiple times we need to reset `optind` back to its default value before use or it doesn't work properly.

I really truly do not understand why this trivial function needs global state, but hey, it does.

worker/src/handles/TcpConnection.cpp
nazar-pc: I will need to dig into `uv_shutdown()` sources before I can answer, docs didn't have explanation as to why this may happen.

nazar-pc: So the code looks like this:
```c
int uv_shutdown(uv_shutdown_t* req, uv_stream_t* stream, uv_shutdown_cb cb) {
  assert(stream->type == UV_TCP ||
         stream->type == UV_TTY ||
         stream->type == UV_NAMED_PIPE);

  if (!(stream->flags & UV_HANDLE_WRITABLE) ||
      stream->flags & UV_HANDLE_SHUT ||
      stream->flags & UV_HANDLE_SHUTTING ||
      uv__is_closing(stream)) {
    return UV_ENOTCONN;
  }
```

And in my case it is `!(stream->flags & UV_HANDLE_WRITABLE)`. Why it is not writable I'm really not sure as of right now.

nazar-pc: Hm... maybe it should have `this->role == UnixStreamSocket::Role::PRODUCER` in the condition there?

worker/Makefile
ibc: But somewhere else? Otherwise "npm install mediasoup" won't build it. See npm-scripts.js please.

nazar-pc: You wouldn't use it with NPM.

The primary use case I see right now it to call it from build script in Rust crate, see: https://github.com/nazar-pc/mediasoup/blob/a67a5ad535315979e273644a9f215ac24c79e451/worker/build.rs#L5-L15

worker/src/handles/TcpConnection.cpp
nazar-pc: Yep, for some reason it tries to call that on reading end of the pipe, which according to the `uv_shutdown()` should fail with `UV_ENOTCONN`. How does it not reach that place with Node.js library I don't know.

Applied what I believe is the correct patch in the latest commit.

worker/src/Channel/UnixStreamSocket.cpp
```diff
@@ -19,18 +19,21 @@ namespace Channel
...
 
 	/* Instance methods. */
 	UnixStreamSocket::UnixStreamSocket(int consumerFd, int producerFd)
 	  : consumerSocket(consumerFd, NsMessageMaxLen, this), producerSocket(producerFd, NsMessageMaxLen)
 	{
 		MS_TRACE_STD();
+
+		this->WriteBuffer = (uint8_t*)std::malloc(NsMessageMaxLen);
```
ibc:
```suggestion
		this->WriteBuffer = static_cast<uint8_t*>(std::malloc(NsMessageMaxLen));
```

jmillan (approved):

ibc:
I'm running `make tidy` to see if this PR introduces something wrong according to our `clang-tidy` rules.

ibc:
But I can because I have to upgrade to Catalina to run `bear`...

ibc:
Will merge this tomorrow morning (here) plus the Rust related one.

lugy-bupt:
Hi ibc, jmillan & nazar-pc,

Thank for your good project and good contribution. I am interested about mediasoup and happy to see that it works as a library with other application. Can you kind to give a simple example about the usage of the library in future? I think it is good for the mediasoup newbies just like me.

nazar-pc:
@lugy-bupt just check lib.hpp in this PR for the single exported public function.
But you wouldn't really use it like this in your own project unless it is also a mediasoup library that wraps the worker.
Use TypeScript or Rust library if you need to add SFU functionality to your app instead.

lugy-bupt:
> @lugy-bupt just check lib.hpp in this PR for the single exported public function.
> But you wouldn't really use it like this in your own project unless it is also a mediasoup library that wraps the worker.
> Use TypeScript or Rust library if you need to add SFU functionality to your app instead.

@nazar-pc  Thank you for the good answer. And just confirm another thing. There is the UDS socket in the mediasoup-library still. Does it means the *caller* program should access the mediasoup-worker library throught UDS still? And how about that the library is called through API? Maybe the mutex or other synchronization ways are needed when calling these APIs

nazar-pc:
> should access throught UDS still?

Yes, nothing changes there.

Since you ask this question libmediasoup-worker is 99% not what you need, this is not the public API of the mediasoup you're looking for.

lugy-bupt:
@nazar-pc You are right, the libmediasoup-worker is not the one I hope so. Anyway, thank you for the quick clarification.

ibc (approved):

PR #534, opened 2021-03-15
Consumer: make IsActive() return true (even if Producer's score is 0) when DTX is enabled

### LINKED ISSUE #532: Issue with key frame requests on producers with little to no activity (hence score 0)
### Issue description
Producer for `<canvas>` capture at 1fps framerate (where most of the time nothing happens anyway) can get score of 0.

This in turn causes https://github.com/versatica/mediasoup/blob/7a60eb815de7925e60683afff7047bb8eaf9d7f0/worker/include/RTC/SimpleConsumer.hpp#L31 to return `false` from `IsActive()`, making it impossible for consumers to request key frames when they need them (like when consumer is resumed or even explicit request).

I worked around this by ignoring `IsActive()` in strategic places as I don't know whether removing score check there will have significant negative effects:
https://github.com/nazar-pc/mediasoup/commit/8e22c7a049954e05565db63fce22966900ad0173

I think this is a bug and needs to be dealt with somehow.

The bug is racy and I was only able to reproduce it once out of ~100 attempts on my machine, but happens every second attempt on the machine of my colleague.

ibc:
- Should fix #532

Now, `Consumer::IsActive()` returns `true` in that case, which makes `RequestKeyFrame()` work. Although this requires a full review since we check `IsActive()` for many other things.

### DIFF
--- a/worker/include/RTC/RtpStream.hpp
+++ b/worker/include/RTC/RtpStream.hpp
@@ -103,6 +103,10 @@ namespace RTC
 		{
 			return this->params.spatialLayers;
 		}
+		bool HasDtx() const
+		{
+			return this->params.useDtx;
+		}
 		uint8_t GetTemporalLayers() const
 		{
 			return this->params.temporalLayers;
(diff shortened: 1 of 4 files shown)


nazar-pc (approved):
Makes sense, thanks!

jmillan (approved):
LGTM

PR #538, opened 2021-03-31
Rate calculator optimization

vpalmisano:
In the current implementation of `RateCalculator` class, the `RemoveOldData` has an high CPU usage because it should iterate over an array where each item index represents a time moment with a resolution of 1ms. As a consequence, in most cases we have a sparse array. This PR changes the calculator algorithm, assigning each sample to an array position in a contiguous way, avoiding large iterations when removing the expired elements. The unit test has been modified accordingly.

Test is needed!
- [ ] no downsides 
- [ ] no bad corner cases
- [ ] no bugs

Callgrind outputs:
- v3:
image

- with optimization:
image

### DIFF
--- a/worker/include/RTC/RateCalculator.hpp
+++ b/worker/include/RTC/RateCalculator.hpp
@@ -43,13 +43,13 @@ namespace RTC
 		void Reset(uint64_t nowMs)
 		{
 			this->buffer.reset(new BufferItem[this->windowItems]);
-			this->latestTime  = 0u;
-			this->latestIndex = -1;
-			this->oldestTime  = 0u;
-			this->oldestIndex = -1;
-			this->totalCount  = 0u;
-			this->lastRate    = 0u;
-			this->lastTime    = 0u;
+			this->newestTime      = 0u;
+			this->newestTimeIndex = -1;
+			this->oldestTimeIndex = 0u;
+			this->oldestIndex     = -1;
+			this->totalCount      = 0u;
+			this->lastRate        = 0u;
+			this->lastTime        = 0u;
 		}
 
 	private:
@@ -69,11 +69,11 @@ namespace RTC
 		// Buffer to keep data.
 		std::unique_ptr<BufferItem[]> buffer;
 		// Time (in milliseconds) for last item in the time window.
-		uint64_t latestTime{ 0u };
+		uint64_t newestTime{ 0u };
 		// Index for the last item in the time window.
-		int32_t latestIndex{ -1 };
+		int32_t newestTimeIndex{ -1 };
 		// Time (in milliseconds) for oldest item in the time window.
-		uint64_t oldestTime{ 0u };
+		uint64_t oldestTimeIndex{ 0u };
 		// Index for the oldest item in the time window.
 		int32_t oldestIndex{ -1 };
 		// Total count in the time window.
--- a/worker/include/RTC/RateCalculator.hpp
+++ b/worker/include/RTC/RateCalculator.hpp
@@ -17,7 +17,10 @@ namespace RTC
 		static constexpr uint16_t DefaultWindowItems{ 1000u };
 
 	public:
-		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale, uint16_t windowItems = DefaultWindowItems)
+		RateCalculator(
+		  size_t windowSize    = DefaultWindowSize,
+		  float scale          = DefaultBpsScale,
+		  uint16_t windowItems = DefaultWindowItems)
 		  : windowSize(windowSize), scale(scale), windowItems(windowItems)
 		{
 			Reset();
--- a/worker/include/RTC/RateCalculator.hpp
+++ b/worker/include/RTC/RateCalculator.hpp
@@ -14,10 +14,11 @@ namespace RTC
 	public:
 		static constexpr size_t DefaultWindowSize{ 1000u };
 		static constexpr float DefaultBpsScale{ 8000.0f };
+		static constexpr uint16_t DefaultWindowItems{ 1000u };
 
 	public:
-		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale)
-		  : windowSize(windowSize), scale(scale)
+		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale, uint16_t windowItems = DefaultWindowItems)
+		  : windowSize(windowSize), scale(scale), windowItems(windowItems)
 		{
 			Reset();
 		}
@@ -38,9 +39,11 @@ namespace RTC
 		void RemoveOldData(uint64_t nowMs);
 		void Reset(uint64_t nowMs)
 		{
-			this->buffer.reset(new BufferItem[this->windowSize]);
-			this->oldestTime  = nowMs - this->windowSize;
-			this->oldestIndex = 0u;
+			this->buffer.reset(new BufferItem[this->windowItems]);
+			this->latestTime  = 0u;
+			this->latestIndex = -1;
+			this->oldestTime  = 0u;
+			this->oldestIndex = -1;
 			this->totalCount  = 0u;
 			this->lastRate    = 0u;
 			this->lastTime    = 0u;
@@ -50,6 +53,7 @@ namespace RTC
 		struct BufferItem
 		{
 			size_t count{ 0u };
+			uint64_t time{ 0u };
 		};
 
 	private:
@@ -57,12 +61,18 @@ namespace RTC
 		size_t windowSize{ DefaultWindowSize };
 		// Scale in which the rate is represented.
 		float scale{ DefaultBpsScale };
+		// Window Size (number of items).
+		uint16_t windowItems{ DefaultWindowItems };
 		// Buffer to keep data.
 		std::unique_ptr<BufferItem[]> buffer;
+		// Time (in milliseconds) for last item in the time window.
+		uint64_t latestTime{ 0u };
+		// Index for the last item in the time window.
+		int32_t latestIndex{ -1 };
 		// Time (in milliseconds) for oldest item in the time window.
 		uint64_t oldestTime{ 0u };
 		// Index for the oldest item in the time window.
-		uint32_t oldestIndex{ 0u };
+		int32_t oldestIndex{ -1 };
 		// Total count in the time window.
 		size_t totalCount{ 0u };
 		// Total bytes transmitted.
--- a/worker/src/RTC/RateCalculator.cpp
+++ b/worker/src/RTC/RateCalculator.cpp
@@ -41,7 +41,7 @@ namespace RTC
 				this->oldestIndex = 0;
 		}
 
-		// Update the latest item.
+		// Update the newest item.
 		BufferItem& item = buffer[this->newestTimeIndex];
 		item.count += size;
 		item.time        = nowMs;
--- a/worker/src/RTC/RateCalculator.cpp
+++ b/worker/src/RTC/RateCalculator.cpp
@@ -83,7 +83,7 @@ namespace RTC
 
 		uint64_t newOldestTime = nowMs - this->windowSize;
 
-		// Oldest item already removed
+		// Oldest item already removed.
 		if (newOldestTime <= this->oldestTime)
 			return;
 
--- a/worker/src/RTC/RateCalculator.cpp
+++ b/worker/src/RTC/RateCalculator.cpp
@@ -20,11 +20,11 @@ namespace RTC
 
 		RemoveOldData(nowMs);
 
-		// Increase the index
+		// Increase the index.
 		if (++this->latestIndex >= this->windowItems)
 			this->latestIndex = 0;
 
-		// Latest index overlaps with the oldest one, remove it
+		// Latest index overlaps with the oldest one, remove it.
 		if (this->latestIndex == this->oldestIndex && this->oldestIndex != -1)
 		{
 			BufferItem& oldestItem = buffer[this->oldestIndex];
@@ -41,7 +41,7 @@ namespace RTC
 		item.time        = nowMs;
 		this->latestTime = nowMs;
 
-		// Set the oldest item index and time, if not set
+		// Set the oldest item index and time, if not set.
 		if (this->oldestIndex < 0)
 		{
 			this->oldestIndex = this->latestIndex;
--- a/worker/src/RTC/RateCalculator.cpp
+++ b/worker/src/RTC/RateCalculator.cpp
@@ -35,7 +35,7 @@ namespace RTC
 				this->oldestIndex = 0;
 		}
 
-		// Update the latest item
+		// Update the latest item.
 		BufferItem& item = buffer[this->latestIndex];
 		item.count += size;
 		item.time        = nowMs;
--- a/worker/src/RTC/RateCalculator.cpp
+++ b/worker/src/RTC/RateCalculator.cpp
@@ -30,7 +30,7 @@ namespace RTC
 			BufferItem& oldestItem = buffer[this->oldestIndex];
 			this->totalCount -= oldestItem.count;
 			oldestItem.count = 0u;
-			oldestItem.time = 0u;
+			oldestItem.time  = 0u;
 			if (++this->oldestIndex >= this->windowItems)
 				this->oldestIndex = 0;
 		}
@@ -38,14 +38,14 @@ namespace RTC
 		// Update the latest item
 		BufferItem& item = buffer[this->latestIndex];
 		item.count += size;
-		item.time = nowMs;
+		item.time        = nowMs;
 		this->latestTime = nowMs;
 
 		// Set the oldest item index and time, if not set
 		if (this->oldestIndex < 0)
 		{
 			this->oldestIndex = this->latestIndex;
-			this->oldestTime = nowMs;
+			this->oldestTime  = nowMs;
 		}
 
 		this->totalCount += size;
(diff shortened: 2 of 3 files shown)


worker/include/RTC/RateCalculator.hpp
```diff
@@ -14,10 +14,11 @@ namespace RTC
 	public:
 		static constexpr size_t DefaultWindowSize{ 1000u };
 		static constexpr float DefaultBpsScale{ 8000.0f };
+		static constexpr uint16_t DefaultWindowItems{ 1000u };
 
 	public:
-		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale)
-		  : windowSize(windowSize), scale(scale)
+		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale, uint16_t windowItems = DefaultWindowItems)
```
ibc: Please run `npm run lint:worker` (it will fail) and fix with `npm run format:worker`.

worker/src/RTC/RateCalculator.cpp
```diff
@@ -20,14 +20,34 @@ namespace RTC
...
-		// Set data in the index before the oldest index.
-		uint32_t offset = this->windowSize - 1;
-		uint32_t index  = this->oldestIndex + offset;
+		// Increase the index
+		if (++this->latestIndex >= this->windowItems)
+			this->latestIndex = 0;
 
-		if (index >= this->windowSize)
-			index -= this->windowSize;
+		// Latest index overlaps with the oldest one, remove it
```
ibc:
```suggestion
		// Latest index overlaps with the oldest one, remove it.
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -20,14 +20,34 @@ namespace RTC
...
+		{
+			BufferItem& oldestItem = buffer[this->oldestIndex];
+			this->totalCount -= oldestItem.count;
+			oldestItem.count = 0u;
+			oldestItem.time = 0u;
+			if (++this->oldestIndex >= this->windowItems)
+				this->oldestIndex = 0;
+		}
+
+		// Update the latest item
```
ibc:
```suggestion
		// Update the latest item.
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -20,14 +20,34 @@ namespace RTC
...
+		BufferItem& item = buffer[this->latestIndex];
+		item.count += size;
+		item.time = nowMs;
+		this->latestTime = nowMs;
+
+		// Set the oldest item index and time, if not set
```
ibc:
```suggestion
		// Set the oldest item index and time, if not set.
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -57,25 +77,18 @@ namespace RTC
...
-
-		// Should never happen.
-		if (newOldestTime < this->oldestTime)
-		{
-			MS_ERROR(
-			  "current time [%" PRIu64 "] is older than a previous [%" PRIu64 "]",
-			  newOldestTime,
-			  this->oldestTime);
-
+		// No item set
```
ibc, line 86:
```suggestion
		// No item set.
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -57,25 +77,18 @@ namespace RTC
...
+		if (this->latestIndex < 0 || this->oldestIndex < 0)
 			return;
-		}
 
-		// We are in the same time unit (ms) as the last entry.
-		if (newOldestTime == this->oldestTime)
+		uint64_t newOldestTime = nowMs - this->windowSize;
+
+		// Oldest item already removed
```
ibc:
```suggestion
		// Oldest item already removed.
```

worker/include/RTC/RateCalculator.hpp
vpalmisano: Done

penguinol:
I've also tried to optimize rate calcautor by using deque，see  https://github.com/penguinol/mediasoup/commit/009f78b6d2d4ca68f568bd6638cf2a30754e769e, but the performence improvement is not as good as i excepted.
The performence under -O3 is quite different from which under -O0.
In our test, the original rate calcautor cost about 5% cpu time under -O3.

vpalmisano:
@penguinol right, I've run the same comparison with -O3:
- v3: 1.45%
- with my optimization: 0.67%

Anyway, `RateCalculator::Update()` is the mediasoup function with the greatest self CPU usage:
v3:
image

with my optimization:
image

ibc:
Thanks for this. How to test this? and how to test that there are no downsides? are unit tests enough?

@jmillan can you please take a look to this?

vpalmisano:
I have used this env variable to trace the process:
```MEDIASOUP_USE_VALGRIND=true MEDIASOUP_VALGRIND_OPTIONS='--tool=callgrind --dump-instr=yes --simulate-cache=yes'```

Looking at the source code, the `RateCalculator` class is used only to estimate the `recvTransmission` and `sendTransmission` to be reported into the transport statistics. If this is true, I see no potential critical issues.

**Comment by ibc** [comment]
Ok, let's merge next week (bank holidays here).

**Comment by jmillan** [comment]
We are currently using `oldestTime` and `latestTime` names. Same for indexes. This IMHO makes difficult readability.

I propose using: `newestXXX` & `oldestXXX` or `firstXXX` & `lastXXX` instead.

Personally I prefer `newestXXX`& `oldestXXX`. Being related to time, it makes more sense to me. What do you think @ibc, @vpalmisano?

File: worker/include/RTC/RateCalculator.hpp [hunk 7]
```diff
@@ -14,10 +14,14 @@ namespace RTC
...
 		static constexpr float DefaultBpsScale{ 8000.0f };
+		static constexpr uint16_t DefaultWindowItems{ 1000u };
 
 	public:
-		RateCalculator(size_t windowSize = DefaultWindowSize, float scale = DefaultBpsScale)
-		  : windowSize(windowSize), scale(scale)
+		RateCalculator(
+		  size_t windowSize    = DefaultWindowSize,
+		  float scale          = DefaultBpsScale,
+		  uint16_t windowItems = DefaultWindowItems)
```
jmillan, line 23: Is there any reason/case for `windowItems` value to be different than `windowSize` value?

I see no setter for it, and it takes the same value as `windowSize` in constructor, by default.

ibc:
> Personally I prefer `newestXXX`& `oldestXXX`. Being related to time, it makes more sense to me.

Agreed

worker/include/RTC/RateCalculator.hpp
ibc: I wonder the very same.

vpalmisano: The `windowItems` value depends on both the rate evaluation window period (`windowSize`) and the events frequency. 
For instance, if the measured event frequency is 1 every 2ms, you can set `windowSize=1000` and `windowItems=500`.

vpalmisano:
> We are currently using `oldestTime` and `latestTime` names. Same for indexes. This IMHO makes difficult readability.
> 
> I propose using: `newestXXX` & `oldestXXX` or `firstXXX` & `lastXXX` instead.
> 
> Personally I prefer `newestXXX`& `oldestXXX`. Being related to time, it makes more sense to me. What do you think @ibc, @vpalmisano?

Or maybe we can use `oldestTime`, `latestTime` and `oldestTimeIndex`, `latestTimeIndex`.

vpalmisano:
For the same reason it will be better to rename `windowSize` -> `windowDuration`.

jmillan:
> Or maybe we can use oldestTime, latestTime and oldestTimeIndex, latestTimeIndex.

The thing is that `oldest` and `latest` are not antonyms. `oldest` and `newest` are, `first` and `last` are antonyms too. Being antonyms makes it easier to understand and relate each other.

jmillan:
Wrapping up with your proposal @vpalmisano:

`oldestTime`, `newestTime` and `oldestTimeIndex`, `newestTimeIndex` would make it.

vpalmisano:
> Wrapping up with your proposal @vpalmisano:
> 
> `oldestTime`, `newestTime` and `oldestTimeIndex`, `newestTimeIndex` would make it.

Done with the latest commit.

worker/include/RTC/RateCalculator.hpp
jmillan: `windowSize` is only used to get the `scale` factor and to retrieve the new `oldestTimeIndex` whereas `windowItems` is the range (from  0 to `windowItems`) where the items are really stored in the `buffer`.

Having a `windowSize` bigger than the `windowItems` does not provide accurate results I would say. Ie:

````
windowSize = 1000
|-----------------------------------------------------|

windowItems = 500
|------------------------|

````

We are calculating the rate for a time window of 1 second, and only considering the data of the last 1/2 second, which is not accurate.

Due to that rationale I think `windowItems` should be removed. Any comment on this @vpalmisano?

Comment (vpalmisano) [reply worker/include/RTC/RateCalculator.hpp:23]: The idea is using different variables to represent the maximum window size expressed in seconds and in number of items, without mixing the two things.
The `windowSize` is expressed in seconds, while the `windowItems` is the maximum number of items that can be stored at the same time in the time window, so in your example you must take into account also the frequency of the data sample.
If you receive 1 sample per millisecond, you will fill the array with 500 items considering only 1/2 second of data. 
Instead, if the data sample frequency is 1 every 2 milliseconds, a 500 items array will correspond to 1 second of data.
I can perform same tests evaluating if the window is filled or not when changing the `windowItems` value.

Comment (vpalmisano) [reply worker/include/RTC/RateCalculator.hpp:23]: I've added a warning to detect when the calculation buffer is filled.
With the default `windowSize = 2500` used in `RtpDataCounter` with `windowItems >= 700` I get no warnings. I think that `windowItems = 1000` is a safe value that we can keep.

**Comment by ibc** [comment]
Are we ready to merge this? or is there something missing?

**Comment by vpalmisano** [comment]
I've replaced some variable names and added a warning log in the case the calculation buffer is full. I think there are no other changes to be done. @jmillan ?

File: worker/src/RTC/RateCalculator.cpp [hunk 8]
```diff
@@ -12,22 +12,48 @@ namespace RTC
...
-		// Set data in the index before the oldest index.
-		uint32_t offset = this->windowSize - 1;
-		uint32_t index  = this->oldestIndex + offset;
+		// Increase the index.
+		if (++this->newestTimeIndex >= this->windowItems)
+			this->newestTimeIndex = 0;
 
-		if (index >= this->windowSize)
-			index -= this->windowSize;
+		// Latest index overlaps with the oldest one, remove it.
```
jmillan:
```suggestion
		// Newest index overlaps with the oldest one, remove it.
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -12,22 +12,48 @@ namespace RTC
...
+
+			BufferItem& oldestItem = buffer[this->oldestIndex];
+			this->totalCount -= oldestItem.count;
+			oldestItem.count = 0u;
+			oldestItem.time  = 0u;
+			if (++this->oldestIndex >= this->windowItems)
+				this->oldestIndex = 0;
+		}
+
+		// Update the latest item.
```
jmillan:
```suggestion
		// Update the newest item.
```

worker/include/RTC/RateCalculator.hpp
jmillan: Thanks @vpalmisano,

I see. Usually we don't control the frequency at which we update the rate calculator, that's why there was 1 item representing each millisecond.

jmillan (approved):

jmillan:
LGMT, thanks @vpalmisano!

worker/include/RTC/RateCalculator.hpp
vpalmisano: Yes, using the log warning we will able to see if the buffer size is underestimated.

ibc:
Merging!

ibc (approved):

ibc:
Just a thing:

```
  CXX(target) /Users/ibc/src/v3-mediasoup/worker/out/Release/obj.target/mediasoup-worker-test/src/RTC/RtpObserver.o
../src/RTC/RateCalculator.cpp:33:6: warning: format specifies type 'unsigned long long' but the
      argument has type 'size_t' (aka 'unsigned long') [-Wformat]
                          this->windowSize,
                          ^~~~~~~~~~~~~~~~~
../include/Logger.hpp:435:22: note: expanded from macro 'MS_WARN_TAG'
        #define MS_WARN_TAG MS_WARN_TAG_STD
                            ^
../include/Logger.hpp:227:90: note: expanded from macro 'MS_WARN_TAG_STD'
  ..._MS_LOG_STR_DESC desc _MS_LOG_SEPARATOR_CHAR_STD, _MS_LOG_ARG, ##__VA_ARGS__); \
                      ~~~~                                            ^~~~~~~~~~~
```

ibc:
fixing in master

worker/include/RTC/RateCalculator.hpp
notorca: In the real world test I'm getting tons of logs saying,
 `mediasoup:WARN:Channel [pid:92750] RTC::RateCalculator::Update() | calculation buffer full, windowSize:2500 ms windowItems:1000 +0ms`

ibc: I assume this is not good news. Is there something than can be fixed/improved here?

notorca: I suppose the old approach with fixed side buckets is much reliable. It can be tuned a bit to help compiler to vectorise calculation sum of removed buckets and memset. 

There is a simple math here:
2 Mbit/s video stream gives on average 200 packets per sec, when keyframe is generated, it can be up to 300. 5 participants room can generate 1200 packets per sec for the rate calculator if transport-cc is enabled. That can already overfill the calculation buffer, especially for window size 2.5 sec. And that is us just for 5 participants.

The old fixed size bucket implementation is resistant to this, if window size of 2.5 sec is need, it can be achieved by using 2.5 ms buckets without increasing the buffer size.

vpalmisano: This means that with more than 10 participants and/or more than 2mbps also with the old approach we were underestimating the calculated rates. We need to find a different algorithm that allows to increase the number of packets per second without increasing the bucket size.

vpalmisano: On the meantime we can set the bucket size to 2500, to obtain at least the same performance of the old approach.

notorca: Old algorithm has no issues with multiple packets with the same ms timestamp, those packets would go to the same bucket, and bucket would contain sum of all those packets sizes. Buckets are per timeslot, not per packet.

New algorithm with bucket size 2500 would have worse performance, then the old one, it tracks timestamp together with packet's sizes and less SIMD friendly.

vpalmisano: We can modify the new algorithm to sum more packets with timestamps that not differ from a minimum delta to the same array position, keeping the array size to a small value.

vpalmisano: We have to decide the calculation sliding window granularity. E.g. with a 2500ms window we can use a 250 items array, obtaining a granularity of 10ms. I will push a commit very soon.

notorca: Same can be applied to the old algorithm. Bucket duration can be any integer. For example 2.5 second window can be handled by 500 buckets with 5ms duration or with 250 buckets with 10ms duration.

ibc: Thanks for the info, guys. So AFAIU the problem has always been there (even before this PR): if bitrate is very high it may happen that there is no buffer enough to store N seconds of data, is that right? And the only "problem" after this PR is that the size of such a buffer is now smaller so making it bigger would resolve the problem. Did I understand it properly?

vpalmisano: I've added some modifications here (https://github.com/versatica/mediasoup/compare/v3...vpalmisano:optimize-rate-calculator) to sum more packets with timestamps that not differ from a minimum delta to the same array position.
I'm testing it with a large (>100) number of consumers, I will post the results very soon.

jmillan: > We have to decide the calculation sliding window granularity. E.g. with a 2500ms window we can use a 250 items array, obtaining a granularity of 10ms. I will push a commit very soon.

Right, based on the window size (ms) and number of items in the window we must compute how much time each item represents. :+1:

jmillan: > So AFAIU the problem has always been there (even before this PR):

No, we didn't have that problem before because each item represented a millisecond and each item would hold the sum of the sizes represented in that millisecond.

vpalmisano: @jmillan right, the problem was not there before this PR. The drawback of the old approach is that it is using a number of items == window size in ms and in some cases this size could be equal to 2500 or 6000 packets.

Can we re-open this PR or should I open a new one with the latest fixes?

ibc: PRs cannot be reopened AFAIK so create a new one, please.

MrSurana:
Too many buffer full warnings after this commit.

```
2021-04-29T09:58:02.592Z mediasoup:WARN:Channel [pid:29676] RTC::RateCalculator::Update() | calculation buffer full, windowSize:2500 ms windowItems:1000
```

ibc:
> Too many buffer full warnings after this commit.
> 
> ```
> 2021-04-29T09:58:02.592Z mediasoup:WARN:Channel [pid:29676] RTC::RateCalculator::Update() | calculation buffer full, windowSize:2500 ms windowItems:1000
> ```

https://github.com/versatica/mediasoup/pull/547

worker/include/RTC/RateCalculator.hpp
vpalmisano: @notorca thanks for reporting the bug and comments.
Fixed now in: https://github.com/versatica/mediasoup/pull/547

notorca: Nice, I will test it soon.

notorca: @vpalmisano Works good for me.

PR #540, opened 2021-03-31
Rust version with thread-based worker

nazar-pc:
This PR builds on top of capabilities introduced in #533.

Main changes here:
* workers are now started in threads instead of processes, so the logic and API around that had to be changed
* minor tweaks to worker's makefile to clean some more files and support custom output directory that facilitates mediasoup-sys crate creation
* mediasoup-sys crate itself, which bundles C++  code, has build instructions for it and exposes libmediasoup-worker using FFI to regular mediasoup crate as plain standard dependency
* disable Rust Nightly in CI as it may introduce new clippy lints and such at any time and would cause annoying false-positives more often
* improved code quality with additional lints that enforce more of best practices
* rust-specific changelog with important code changes

Everything above greatly improves usability of mediasoup crate, there is no need to have NPM installed anymore or build worker manually.

The bulk of the diff is moving tests around since I could no longer send signal to worker to trigger all kinds of edge cases, instead there is a private API now used for testing and not available from integrations tests, so those test cases had to be moved into internal unit tests.


ibc (approved):
Nice. Merging. Do we need a new mediasoup release? or just a new `rust-X.X.X` (which value) git tag? or both?

nazar-pc:
Git tag `rust-0.7.1` would be nice to have, thanks!

ibc:
Tag pushed

PR #541, opened 2021-04-02
Fix miscalculation when increasing layer

penguinol:
We should subtract the whole bitrate of provisional layer when increasing to a higher spatial layer.


ibc:
Please run `npm run lint` (there will be errors) and then `npm run worker:format` to fix them.

Also, could you please detail the rationale of this change? I fail to see it.

penguinol:
When we increse to a higher spatial layer, we need to know how mush additional bitrate is needed.
But GetLayerBitrate only return the bitrate of one temporal layer.
For example, we have a S2T2 stream.
When we are sending S1T2, means we are both sending packets of S1T1 and S1T2
Then we tried to increase to S2T1, so the addtional bitrate = bitrate of S2T1 - (bitrate of S1T1 + bitrate of S1T2)
And GetBirate(0,1) = GetLayerBitrate(0, 0) + GetLayerBitrate(0, 1), so here should be GetBitrate

ibc:
You are 100% right. Thanks. Merging.

ibc (approved):
releasing a new version in a few minutes.

jmillan (commented):
:+1:

PR #546, opened 2021-04-18
feat: gyp move to camke

barry-ran:


nazar-pc:
Looks like some of the dependencies were updated here. I think it would be much easier to review if those updates were submitted separately so that it is possible to focus on just CMake support instead.
Also it would be nice to have comments in English and support for libmediasoup-worker build that was recently introduced to the codebase.

jmillan:
Absolutely,

Please @barry-ran, let's have a PR just with CMake related changes. We can always have another one upgrading the deps if needed.

barry-ran:
I split commit, you can cherry-pick what you need

ibc:
> I split commit, you can cherry-pick what you need

I appreciate your effort, but taking just some commits of a PR is not the way to go. I have many questions about changes in this PR but I'm not sure if you are wishing to accept change requests or not. Are you?

Going to the questions: 

- There are many changes in this PR into C/C++ dependencies. The idea is that we use `worker/scripts/get-dep.sh xxx` to update a specific dep. However if manual changes have been done on current deps, that's not possible. Could you please explain which changes you did into deps and why?

- Regarding OpenSSL, does this PR include an official version of OpenSSL? Or is it a forked version to work with CMake?

barry-ran:
> > I split commit, you can cherry-pick what you need
> 
> I appreciate your effort, but taking just some commits of a PR is not the way to go. I have many questions about changes in this PR but I'm not sure if you are wishing to accept change requests or not. Are you?
> 
> Going to the questions:
> 
> * There are many changes in this PR into C/C++ dependencies. The idea is that we use `worker/scripts/get-dep.sh xxx` to update a specific dep. However if manual changes have been done on current deps, that's not possible. Could you please explain which changes you did into deps and why?
> * Regarding OpenSSL, does this PR include an official version of OpenSSL? Or is it a forked version to work with CMake?

- libsrtp use dep/openssl not system openssl
- it is a forked version to work with CMake

yes, taking just some commits of a PR is not the way to go, But I don’t have more time to do this, you can just use it as a reference, sorry

ibc:
Thanks for the clarifications.

PR #547, opened 2021-04-27
Optimize rate calculator (v2)

vpalmisano:
With this PR we are fixing issues generated by the new RateCalculator implementation:
- it uses a window granularity `windowSize / windowItems` to sum up multiple samples into the newest item;
- in the default configuration we are using a total items count of `100`, this means that with a window size of `1000ms` each item will sum up the samples arriving in a `10ms`  time window;
- avoids to fill the calculation buffer;
- fixed oldestTime/Index naming


worker/src/RTC/RateCalculator.cpp
```diff
@@ -12,46 +12,57 @@ namespace RTC
...
-			oldestItem.time  = 0u;
-			if (++this->oldestIndex >= this->windowItems)
-				this->oldestIndex = 0;
+			this->newestTimeIndex++;
+			this->newestTime = nowMs;
+			if (this->newestTimeIndex >= this->windowItems)
+				this->newestTimeIndex = 0;
+
+			// Newest index overlaps with the oldest one, remove it.
+			if (this->newestTimeIndex == this->oldestTimeIndex && this->oldestTimeIndex != -1)
```
vpalmisano: This condition should never happen with the new implementation.

worker/test/src/RTC/TestRateCalculator.cpp
```diff
@@ -61,7 +61,7 @@ SCENARIO("Bitrate calculator", "[rtp][bitrate]")
 
 	SECTION("slide")
 	{
-		RateCalculator rate;
+		RateCalculator rate(1000, 8000, 1000);
```
vpalmisano, line 64: With the default window items value (`100`) we have this:
```
		RateCalculator rate(1000, 8000, 100);

		std::vector<data> input =
		{
			{ 0,    5, 40 },
			{ 999,  2, 56 },
			{ 1001, 1, 24 }, // merged inside 999
			{ 1001, 1, 32 }, // merged inside 999
			{ 2000, 1, 24 } // will erase the item with timestamp=999, removing also the next two samples. The end estimation will include only the last sample 
		};
```

worker/src/RTC/RateCalculator.cpp
```diff
@@ -12,46 +12,57 @@ namespace RTC
...
 
-		// Increase the index.
-		if (++this->newestTimeIndex >= this->windowItems)
-			this->newestTimeIndex = 0;
-
-		// Newest index overlaps with the oldest one, remove it.
-		if (this->newestTimeIndex == this->oldestIndex && this->oldestIndex != -1)
+		// If the elapsed time from the last used item is greater than the
+		// window granularity, increase the index.
+		if (this->newestTimeIndex < 0 || nowMs - this->newestTime >= windowSize / windowItems)
```
penguinol: maybe we can save the result of windowSize/windowItem to avoid dividing it every time

worker/include/RTC/RateCalculator.hpp
```diff
@@ -46,7 +46,7 @@ namespace RTC
 			this->newestTime      = 0u;
 			this->newestTimeIndex = -1;
 			this->oldestTimeIndex = 0u;
-			this->oldestIndex     = -1;
+			this->oldestTimeIndex = -1;
```
jmillan: This and previous line set the same member.

jmillan:
I'll properly review this tomorrow.

worker/include/RTC/RateCalculator.hpp
vpalmisano: Fixed.

worker/src/RTC/RateCalculator.cpp
jmillan: Yes, it indicates the milliseconds represented in each item (ie: `this->itemSize = this->windowSize/this->windowItems`). It should be a class member IMHO.

vpalmisano: Added with the last commits.

worker/src/RTC/RateCalculator.cpp
```diff
@@ -12,46 +12,57 @@ namespace RTC
...
 		RemoveOldData(nowMs);
 
-		// Increase the index.
-		if (++this->newestTimeIndex >= this->windowItems)
-			this->newestTimeIndex = 0;
-
-		// Newest index overlaps with the oldest one, remove it.
-		if (this->newestTimeIndex == this->oldestIndex && this->oldestIndex != -1)
+		// If the elapsed time from the last used item is greater than the
+		// window granularity, increase the index.
```
jmillan: The condition below is not accurate.

Each item in the buffer must contain the data for a certain period of time. Ie:

Scenario where each item holds 10ms of time.

`windowSize = 1000`
`windowItems = 100`
`oldestTime = 2000`

|Index|0    |1       |2        |3          | n|        
|------|----|-----|------|-------|--|
| Time Range|2000-2009|2010-2019|2020-2029|2030-2039|.. |

So, instead of having `newestTime`, we should have the start time for the newest item, and check whether the incoming data should be added to the same item or not.

vpalmisano: It is already there, the newestTime update is performed only when a new item is added: https://github.com/vpalmisano/mediasoup/blob/optimize-rate-calculator/worker/src/RTC/RateCalculator.cpp#L28

jmillan: Right.

I'm having a hard time reading the PR due to the member naming. I propose some naming enhancement:

newestTimeIndex -> newestItemIndex (Buffer index holding the newest item)
oldestTimeIndex -> oldestItemIndex (Buffer index holding the oldest item)
newestTime -> newestItemStartTime (Start time of the newest item)
oldestTime -> oldestItemStartTime (Start time of the oldest item)
windowSize -> windowSizeMs (Window size in milliseconds)
itemSize -> itemSizeMs (Item size in milliseconds)

It would certainly help with the code readability.

What do you think?

vpalmisano: I agree, much better. Done.

worker/src/RTC/RateCalculator.cpp
```diff
@@ -12,46 +12,57 @@ namespace RTC
...
```
jmillan:
```suggestion
		// If the elapsed time from the newest item start time is greater than the
		// item size (in milliseconds), increase the item index.
```

jmillan:
@vpalmisano,

The PR looks good to me. The new variable names makes it much more readable, thanks!

Just one thing. Please, instead of modifying the existing tests, keep them, and create new ones with the new `windowSizeMs` and `windowItems` values.

vpalmisano:
> Just one thing. Please, instead of modifying the existing tests, keep them, and create new ones with the new `windowSizeMs` and `windowItems` values.

The only modification I've done is here (https://github.com/vpalmisano/mediasoup/blob/optimize-rate-calculator/worker/test/src/RTC/TestRateCalculator.cpp#L73) and it is required to make the test pass as explained in the comment.

jmillan:
What I mean is, keep those two tests with the original construction values `(1000, 8000, 1000)` and a create two new ones with the new values.

The old tests should keep passing providing the correct values in construction.

Screenshot 2021-04-29 at 09 33 30

vpalmisano:
> What I mean is, keep those two tests with the original construction values `(1000, 8000, 1000)` and a create two new ones with the new values.
> 
> The old tests should keep passing providing the correct values in construction.

Yes, the "slide" test will pass using `(1000, 8000, 1000)`, I've added a new one with `(1000, 8000, 100)` values.

The "wrap" test is useful only if new samples are allocated to different items, for this reason I've modified the samples timestamps.

jmillan (approved):

ibc (approved):

jmillan:
@vpalmisano,

This warning: https://github.com/vpalmisano/mediasoup/blob/ff423f531b9a526ef45f71e72c7255d21aa402af/worker/src/RTC/RateCalculator.cpp#L35

Shouldn't be so. It's a legit situation. Any reason why you set a warning there?

vpalmisano:
> @vpalmisano,
> 
> This warning: https://github.com/vpalmisano/mediasoup/blob/ff423f531b9a526ef45f71e72c7255d21aa402af/worker/src/RTC/RateCalculator.cpp#L35
> 
> Shouldn't be so. It's a legit situation. Any reason why you set a warning there?

The if condition should never happen with the latest modification. 
I left the warning in the case we have missed something.

PR #556, opened 2021-05-14
fix rr report

penguinol:
Previous code is as follow:
```
uint32_t newLostInterval     = (worstRemoteFractionLost * expectedInterval) >> 8;
uint32_t newReceivedInterval = expectedInterval - newLostInterval;
this->reportedPacketLost += (receivedInterval - newReceivedInterval);
```
When consumer franction lost >> producer fraction lost, (receivedInterval - newReceivedInterval) is ok.
But when consumer franction lost is only a litte bit higher than producer fraction lost,  (receivedInterval - newReceivedInterval) will be very small even if both consumer franction lost and producer fraction are very large.


ibc:
@jmillan can you please take a look to this? My eyes are not good at this.

jmillan:
I need to look at this with caution too :-)

jmillan (approved):
It's totally correct.

Thanks!

ibc:
Nice, let's merge.

ibc:
Thanks @penguinol!

PR #557, opened 2021-05-14
Avoid duplicated basenames so that libmediasoup-worker is compilable on macOS

nazar-pc:
Without these changes libmediasoup-worker compilation on macOS (and I think Windows too) fails with this:
```
  static library libmediasoup-worker has several files with the same basename:
    TcpConnection: src/handles/TcpConnection.cpp src/RTC/TcpConnection.cpp
    UnixStreamSocket: src/handles/UnixStreamSocket.cpp src/Channel/UnixStreamSocket.cpp src/PayloadChannel/UnixStreamSocket.cpp
    Request: src/Channel/Request.cpp src/PayloadChannel/Request.cpp
    Notifier: src/Channel/Notifier.cpp src/PayloadChannel/Notifier.cpp
    UdpSocket: src/handles/UdpSocket.cpp src/RTC/UdpSocket.cpp
    TcpServer: src/handles/TcpServer.cpp src/RTC/TcpServer.cpp
  libtool on OS X will generate warnings for them.
  Error running GYP
```

This was previously suppressed with `--no-duplicate-basename-check` to get Linux build working, which is no longer needed.

UPD: This is one of a few PRs that will together bring full macOS support for Rust version.


ibc:
I don't understand. I compile mediasoup-worker in macOS and Linux let's say... 30 times a week, and this has never been an issue.

nazar-pc:
It is only about libmediasoup-worker. I do not fully understand why, but building static C++ library isn't happening unless all file names are distinct.

ibc:
Please provide exacts steps to reproduce and OS/gcc/clang version.

nazar-pc:
Just run this:
```
cd worker
make libmediasoup-worker
```
You'll get the error right away and I don't think it depends on OS/compiler, but I'm running macOS 11.3.1 with Clang 12.0.5 as it comes with OS I believe.

ibc:
Sorry for the delay. I strongly **hate** these changes, but I don't have any good argument against them.

Just a question: can we please do a bit more research before merging this? I'd really like to know **why** we cannot have duplicate file names when building a static library in macOS.

nazar-pc:
From the search I did it looks like everyone deals with the same thing by renaming files.

Apparently libtool that Apple has is not the same as GNU libtool.

Those warnings shouldn't really be errors, but GYP doesn't allow static library compilation even when the error is suppressed with CLI switch, so I'm not sure what else to do here. I wish this wasn't necessary.

Similar issue will likely happen on Windows, see https://github.com/grpc/grpc/issues/7814 for instance and PR there that fixes it.

ibc:
I see. This is weird... Thanks for your effort.

PR #558, opened 2021-05-15
Update modules structure and other minor improvements for Rust version

nazar-pc:
Changes:
* Refactored modules structure: it still looks the same to developers and docs, but doesn't confuse IDE as much so autocomplete works well
* `NonClosingProducer` renamed to `PipeProducer` with better documentation about why it exists (old name is still available)
* Minor fix in libmediasoup-worker that avoids printing irrelevant messages

I will update changelog and bump versions in upcoming PRs once macOS support is merged upstream (hopefully very soon).


jmillan (commented):
I don't know Rust myself, but the changes seem to enhance readability and inline documentation.

Considering tests pass I'm OK with the PR.

nazar-pc:
Pushed an update to fix some of the lints.

Do you plan to update libwebrtc to fix `numeric_limits` issues in CI with Ubuntu 18.04/20.04 and Clang (if it does fix it)?
Looks like an import is missing of something like that.

jmillan:
> Do you plan to update libwebrtc to fix numeric_limits issues in CI with Ubuntu 18.04/20.04 and Clang (if it does fix it)?

The problem is within the abseil dependency used by libwebrtc. I Don't think we want to clone it just to avoid this issue, since also there are being some efforts to remove libwebrtc completely from mediasoup.

ibc (approved):
I don't see any bad words in the changes so I assume it's ok. Merging.

PR #562, opened 2021-05-19
Thread and memory safety fixes

nazar-pc:
First commit adds some more `thread_local` and other minor tweaks to make sure things don't go wrong when multiple workers are running in different threads concurrently, I believe all cases should now be covered.

Next 2 commits fix some interesting use-after-free memory issues that were reported by `valgrind`. They weren't causing crashes under Linux, at least most of the time, but were crashing on macOS most of the time. Now things should work well, but please look carefully because I don't know what I'm doing when it comes to C++ :sweat_smile: 

This is the last prerequisite to introduction of macOS support in Rust version, which will be a separate PR.

PR #566, opened 2021-05-28
Stop failing MacOSX builds.

mstyura:
It seems like the original reason for not build on MacOS X is resolved (something with `libc++.a` according to comment).
With these changes I could build media soup on my mac.


nazar-pc:
It may compile, but not work like that. The work on macOS support is already done and is pending review.
#562 fixes memory safety issues without which macOS version will not work anyway and there will be one more PR that introduces full support for macOS.

Panic is there for a reason and related to mediasoup-sys crate and you did nothing to address it.

If you need macOS support today use this branch until next release of Rust version on crates.io: https://github.com/nazar-pc/mediasoup/tree/rust-macos-support

PR #567, opened 2021-05-28
Rust support on macOS

nazar-pc:
This implements macOS support in Rust crates and adds macOS to CI for Rust.

What it boils down to is that on macOS there is no static version of C++ standard library available out of the box, so build.rs will download LLVM 12.0.0 tarball and will extract `libc++.a` + `libc++abi.a` and will link those into final mediasoup-sys library alongside with libmediasoup-worker itself.

### DIFF
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -7,10 +7,17 @@ fn main() {
         return;
     }
 
+    let current_dir = std::env::current_dir()
+        .unwrap()
+        .into_os_string()
+        .into_string()
+        .unwrap();
+    let out_dir = env::var("OUT_DIR").unwrap();
+
     // Add C++ std lib
     #[cfg(target_os = "linux")]
     {
-        let path = Command::new(env::var("c++").unwrap_or_else(|_| "c++".to_string()))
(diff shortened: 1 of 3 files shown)


worker/build.rs
```diff
@@ -7,10 +7,17 @@ fn main() {
...
+        .into_os_string()
+        .into_string()
+        .unwrap();
+    let out_dir = env::var("OUT_DIR").unwrap();
+
     // Add C++ std lib
     #[cfg(target_os = "linux")]
     {
-        let path = Command::new(env::var("c++").unwrap_or_else(|_| "c++".to_string()))
+        let path = Command::new(env::var("CXX").unwrap_or_else(|_| "c++".to_string()))
```
ibc, line 20: 👍

ibc (approved):

ibc:
ready to merge? and if so, anything else missing? can we release and announce "support for mediasoup Rust in macOS"?

nazar-pc:
Yes, this is ready and let's get #568 in so I can publish a proper release.

ibc:
Released in mediasoup 3.7.7

PR #571, opened 2021-06-01
Allow to reconctruct IDs from UUIDs.

mstyura:
It is possible to convert various `MediaSoup` IDs to `uuid::UUID` via `from/into`, but I can't find a way to reconstruct `ID` from `uuid::UUID`. This PR should resolve this problem (if it was not intentional restriction).

### DIFF
--- a/rust/src/macros.rs
+++ b/rust/src/macros.rs
@@ -32,6 +32,12 @@ macro_rules! uuid_based_wrapper_type {
             }
         }
 
+        impl From<uuid::Uuid> for $struct_name {
+            fn from(id: uuid::Uuid) -> Self {
+                $struct_name(id)
+            }
+        }
+
         impl $struct_name {
             pub(super) fn new() -> Self {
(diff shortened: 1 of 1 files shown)


ibc:
Please explain why is this needed (given that it's just for Rudy).

ibc:
@nazar-pc can you take a look please?

nazar-pc:
This is intentional API design, you shouldn't need this.
You get `ProducerId` and such produced by mediasoup or get it from API (where Serde helps you to deserialize Uuid).

You shouldn't be able to convert `ConsumerId` into `Uuid` and then into `ProducerId` because it makes no sense, those are intentionally new types to improve misuse resistance. In other words, `ProducerId` is always `Uuid` inside, but not every `Uuid` is a valid `ProducerId`.

Can you describe the use case where you legitimately need this first?

mstyura:
@nazar-pc thank you very much for explanation. Closing then.

PR #572, opened 2021-06-01
Remove clang tools binaries from regular installation

jmillan:
Move them to worker/scripts

Bonus point: Remove Gulp dependency too from regular installation.


.gitignore
```diff
@@ -29,6 +29,8 @@
 
 # Python generated stuff in Windows.
 /worker/scripts/configure.pyc
+# worker scripts node-modules.
+/worker/scripts/node_modules
```
ibc:
```suggestion
/worker/scripts/node_modules/
```

npm-scripts.js
```diff
@@ -72,21 +72,21 @@ switch (task)
 	case 'lint:node':
 	{
 		execute('cross-env MEDIASOUP_NODE_LANGUAGE=typescript eslint -c .eslintrc.js --max-warnings 0 --ext=ts src/');
-		execute('cross-env MEDIASOUP_NODE_LANGUAGE=javascript eslint -c .eslintrc.js --max-warnings 0 --ext=js --ignore-pattern \'!.eslintrc.js\' .eslintrc.js gulpfile.js npm-scripts.js test/');
+		execute('cross-env MEDIASOUP_NODE_LANGUAGE=javascript eslint -c .eslintrc.js --max-warnings 0 --ext=js --ignore-pattern \'!.eslintrc.js\' .eslintrc.js npm-scripts.js test/');
```
ibc: Why is gulpfile.js removed from lint?

worker/Makefile
```diff
@@ -77,8 +77,8 @@ bear:
 tidy:
 	$(SED) "s|PATH|$(PWD)|g" compile_commands_template.json > compile_commands.json
 	$(PYTHON) ./scripts/clang-tidy.py \
-		-clang-tidy-binary=../node_modules/.bin/clang-tidy \
-		-clang-apply-replacements-binary=../node_modules/.bin/clang-apply-replacements \
+		-clang-tidy-binary=./scripts/node_modules/.bin/clang-tidy \
+		-clang-apply-replacements-binary=./scripts/node_modules/.bin/clang-apply-replacements \
```
ibc, line 81: Are those still work if I do `cd worker/ && make tidy`?

ibc:
CI failing

ibc:
So, in order for this to work locally, we must enter `worker/` and run `npm install` before running usual npm/make commands?

worker/Makefile
jmillan: yes, it does.

jmillan:
> So, in order for this to work locally, we must enter worker/ and run npm install before running usual npm/make commands?

Correct. Not enter `worker` but `worker/scripts`

jmillan:
> CI failing

On it

jmillan:
ooh, my...

npm-scripts.js
jmillan: It was previously *excluded*. Now it's moved to `worker/scripts`

.github/workflows/mediasoup-worker.yaml
```diff
@@ -69,8 +69,8 @@ jobs:
 
       - run: npm install
       - run: |
-          pushd worker/scripts
+          cd worker/scripts
           npm install
-          popd
+          cd ../..
```
nazar-pc: I think it runs every command under the same directory regardless of `cd` in previous command...

jmillan: oh, makes sense. Thanks

jmillan: So probably this is executing `npm install` in the root directory again instead of in `worker/scripts`, right?

jmillan: Fixed, let's see

npm-scripts.js
```diff
@@ -163,6 +161,13 @@ switch (task)
 		break;
 	}
 
+	case 'install-clang-tools':
```
ibc, line 164: Why is this needed at all? Nobody calls `node npm-scripts.js install-clang-tools`.

ibc: ok, yes, package.json does. What I mean is: does it make sense to call something from.... forget it, it's ok.

ibc (approved):

PR #575, opened 2021-06-01
Fix memory leak caused by incomplete libuv closing procedure

nazar-pc:
Turns out `uv_loop_close()` can return `UV_EBUSY`, which means there are still handles that are not closed. Correct behavior is to wait for all handles to stop or else memory will be leaked.

### DIFF
--- a/worker/src/DepLibUV.cpp
+++ b/worker/src/DepLibUV.cpp
@@ -9,20 +9,6 @@
 
 thread_local uv_loop_t* DepLibUV::loop{ nullptr };
 
-void on_uv_close(uv_handle_t* handle)
-{
-	if (handle != NULL)
-	{
-		delete handle;
-	}
-}
-
-void on_uv_walk(uv_handle_t* handle, void* arg)
-{
-	if (!uv_is_closing(handle))
-		uv_close(handle, on_uv_close);
-}
-
 /* Static methods. */
 
 void DepLibUV::ClassInit()
@@ -45,17 +31,12 @@ void DepLibUV::ClassDestroy()
 	{
 		int err;
 
-		uv_stop(DepLibUV::loop);
-		uv_walk(DepLibUV::loop, on_uv_walk, NULL);
-
 		while (true)
 		{
 			err = uv_loop_close(DepLibUV::loop);
 
 			if (err != UV_EBUSY)
-			{
 				break;
-			}
 
 			uv_run(DepLibUV::loop, UV_RUN_NOWAIT);
 		}
--- a/worker/src/DepLibUV.cpp
+++ b/worker/src/DepLibUV.cpp
@@ -9,6 +9,20 @@
 
 thread_local uv_loop_t* DepLibUV::loop{ nullptr };
 
+void on_uv_close(uv_handle_t* handle)
+{
+	if (handle != NULL)
+	{
+		delete handle;
+	}
+}
+
+void on_uv_walk(uv_handle_t* handle, void* arg)
+{
+	if (!uv_is_closing(handle))
+		uv_close(handle, on_uv_close);
+}
+
 /* Static methods. */
 
 void DepLibUV::ClassInit()
@@ -30,7 +44,26 @@ void DepLibUV::ClassDestroy()
 	// This should never happen.
 	if (DepLibUV::loop != nullptr)
 	{
-		uv_loop_close(DepLibUV::loop);
+		int err;
+
+		uv_stop(DepLibUV::loop);
+		uv_walk(DepLibUV::loop, on_uv_walk, NULL);
+
+		while (true)
+		{
+			err = uv_loop_close(DepLibUV::loop);
+
+			if (err != UV_EBUSY)
+			{
+				break;
+			}
+
+			uv_run(DepLibUV::loop, UV_RUN_NOWAIT);
+		}
+
+		if (err != 0)
+			MS_ABORT("failed to close libuv loop: %s", uv_err_name(err));
+
 		delete DepLibUV::loop;
 	}
 }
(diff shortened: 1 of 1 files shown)


ibc:
How did you get into this? As the comment says:

"This should never happen"

nazar-pc:
> This should never happen

This is about `DepLibUV::loop != nullptr` check if I understand correctly.

> How did you get into this?

I ran one of Rust's integration tests under valgrind and was analyzing the output for potential memory leaks. Turned out there was one :slightly_smiling_face: 
Obviously doesn't affect TypeScript as process stops shortly after.

nazar-pc:
Hm... now I get the question, interesting...

nazar-pc:
I think `DepLibUV::loop` should always be non-null, otherwise how would anything work at all?

nazar-pc:
Found it, that comment remained after refactoring, see 768b411e0c697b112f0ccc75cdb279586ee80a1d:
```cpp
	// This should never happen.
	if (DepLibUV::loop == nullptr)
		MS_ABORT("DepLibUV::loop was not allocated");
```

Removed outdated comment.

ibc:
It doesn't make sense. We do not force the loop to close. We just close all UV handles so main.cpp continues and ends calling to Worker::Destroy() which should just free the loop. There shouldn't be any running handle at that time when Destroy() is called. And if there was any, we wouldn't be there yet.

ibc:
Ok ok, maybe there are no running UV handles but they have not yet had their onClose callbacks called? What does uv_walk() do?

ibc:
No, it doesn't make sense. Those UV onClose callbacks are called by the loop, of course.

nazar-pc:
OK, thanks for hints, simplified and it still achieves the same result

ibc:
Can you please check the return value of uv_run() in RunLoop() method? It should be 0 meaning that no active handles left, and it must be 0 because we never call uv_stop() to force the uv_run() to return.

nazar-pc:
It returns `0` :shrug:

ibc:
There is a handle API function in libuv to make uv_run() terminate even if such a handle is active, however we don't use it and if so I assume uv_run() should return != 0 but I'm not sure.

ibc:
> ```c++
> // This should never happen.
> ```

Sorry, this comment was right. Take a look to `lib.cpp`. It may happen that `DepLibUV::ClassInit()` is **NOT** called (if something else throws first). So please do not remove it.

ibc:
> There is a handle API function in libuv to make uv_run() terminate even if such a handle is active, however we don't use it and if so I assume uv_run() should return != 0 but I'm not sure.

`uv_unref()`, but we do not use it.

nazar-pc:
> Sorry, this comment was right.

`DepLibUV::ClassInit()` is the very first call ever, nothing can physically throw before it

ibc:
> > Sorry, this comment was right.
> 
> `DepLibUV::ClassInit()` is the very first call ever, nothing can physically throw before it

True.

BTW, a uv handle that is NOT active (example: a non running uv_timer) also makes uv_run() exit with 0 but the handle is still there and must be closed. So your PR is identifying some bug in mediasoup.

Pretty sure we CLOSE all timers but we do NOT close the SignalHandler in Worker.cpp.

I'm gonna test uv_walk and print handle->type.

ibc:
I've pushed a little change to v3 to assert that uv_run() returns 0

worker/src/DepLibUV.cpp
```diff
@@ -30,7 +44,26 @@ void DepLibUV::ClassDestroy()
 	// This should never happen.
 	if (DepLibUV::loop != nullptr)
 	{
-		uv_loop_close(DepLibUV::loop);
```
ibc, line 33: This must be called before `delete DepLibUV::loop;`

ibc: ok ok, ignore please

ibc:
> > Sorry, this comment was right.
> 
> `DepLibUV::ClassInit()` is the very first call ever, nothing can physically throw before it

Mmmm, no:

```c++
	try
	{
		// Initialize static stuff.
		DepOpenSSL::ClassInit();
		DepLibSRTP::ClassInit();
		DepUsrSCTP::ClassInit();
		DepLibWebRTC::ClassInit();
		Utils::Crypto::ClassInit();
		RTC::DtlsTransport::ClassInit();
		RTC::SrtpSession::ClassInit();
		Channel::ChannelNotifier::ClassInit(channel);
		PayloadChannel::PayloadChannelNotifier::ClassInit(payloadChannel);

#ifdef MS_EXECUTABLE
		{
			// Ignore some signals.
			IgnoreSignals();
		}
#endif

		// Run the Worker.
		Worker worker(channel, payloadChannel);

		// Free static stuff.
		DepLibSRTP::ClassDestroy();
		Utils::Crypto::ClassDestroy();
		DepLibWebRTC::ClassDestroy();
		RTC::DtlsTransport::ClassDestroy();
		DepUsrSCTP::ClassDestroy();
		DepLibUV::ClassDestroy();
```

for example, `DepUsrSCTP::ClassInit();` can throw.

nazar-pc:
But `DepLibUV::ClassInit()` that creates `DepLibUV::loop` happens before that

ibc:
ok, I have some news:



> But `DepLibUV::ClassInit()` that creates `DepLibUV::loop` happens before that

you are super right. Let me remove the comment in v3 branch.

BTW: can we please move to https://github.com/versatica/mediasoup/pull/576? I've done some "progress" in there.

ibc:
Let's close this since PR https://github.com/versatica/mediasoup/pull/576/ fixes the real design issue.

PR #579, opened 2021-06-02
Skip mode modules in mediasoup-sys publishing

nazar-pc:
Sorry, another one due to recent changes

PR #581, opened 2021-06-05
Fix memory leaks on error exit, minor tweaks

nazar-pc:
The primary goal here is to fix memory leaks that happen when library user provides incorrect worker settings. Since that is essentially a user input, it needs to be handled gracefully.

Previously worker would just exit without de-allocating channels and event loop, which was a problem in Rust case since it leaked memory forever.

Valgrind summary before after running all integration tests:
```
==652620== LEAK SUMMARY:
==652620==    definitely lost: 2,472 bytes in 87 blocks
==652620==    indirectly lost: 17,168 bytes in 271 blocks
==652620==      possibly lost: 25,167,822 bytes in 23 blocks
==652620==    still reachable: 33,705,987 bytes in 106 blocks
==652620==         suppressed: 0 bytes in 0 blocks
```

And after:
```
==688087== LEAK SUMMARY:
==688087==    definitely lost: 2,088 bytes in 87 blocks
==688087==    indirectly lost: 12,528 bytes in 261 blocks
==688087==      possibly lost: 456 bytes in 5 blocks
==688087==    still reachable: 142,774 bytes in 31 blocks
==688087==         suppressed: 0 bytes in 0 blocks
```

There are some one-time globally initialized stuff on Rust side in dependencies, so I think that non-zero result is expected.

Then I checked if SCTP-related cleanup was problematic and turns out it no longer is, I guess it was accidentally caused by other (now resolved) multi-threaded memory issues.

And lastly CI was failing because Ubuntu 20.04 on GitHub Actions got Clang update and cached object files were slightly incompatible, so I just removed them from cache. Errored build for reference: https://github.com/nazar-pc/mediasoup/runs/2754266004

### DIFF
--- a/worker/src/DepLibUV.cpp
+++ b/worker/src/DepLibUV.cpp
@@ -20,8 +20,8 @@ inline static void onWalk(uv_handle_t* handle, void* arg)
 {
 	// Must use MS_ERROR_STD since at this point the Channel is already closed.
 	MS_ERROR_STD(
-	  "alive UV handle found (this shouldn't happen) [type:%d, active:%d, closing:%d, has_ref:%d]",
-	  handle->type,
+	  "alive UV handle found (this shouldn't happen) [type:%s, active:%d, closing:%d, has_ref:%d]",
+	  uv_handle_type_name(handle->type),
 	  uv_is_active(handle),
 	  uv_is_closing(handle),
 	  uv_has_ref(handle));
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -32,8 +32,6 @@ jobs:
         with:
           path: |
             ~/.npm
-            worker/deps
-            worker/out
           key: ${{ matrix.os }}-node-${{ hashFiles('**/package.json') }}
           restore-keys: |
             ${{ matrix.os }}-node-
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -61,8 +61,6 @@ jobs:
         with:
           path: |
             ~/.npm
-            worker/deps
-            worker/out
           key: ${{ matrix.build.os }}-node-${{matrix.build.cc}}-${{ hashFiles('**/package.json') }}
           restore-keys: |
             ${{ matrix.build.os }}-node-${{matrix.build.cc}}-
--- a/worker/src/DepUsrSCTP.cpp
+++ b/worker/src/DepUsrSCTP.cpp
@@ -96,12 +96,10 @@ void DepUsrSCTP::ClassDestroy()
 		{
 			usrsctp_finish();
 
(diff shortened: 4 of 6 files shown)


ibc:
> Previously worker would just exit without de-allocating channels and event loop

So I think this is a regression due to previous changes, right? In `ChannelSocket.cpp`:

```c++
	ChannelSocket::~ChannelSocket()
	{
		MS_TRACE_STD();

		std::free(this->writeBuffer);
	}

	void ChannelSocket::Close()
	{
		MS_TRACE_STD();

		this->consumerSocket.Close();
		this->producerSocket.Close();
	}
```

If we don't call `Close()` (but just the destructor) then it leaks. That's why I didn't like much these related changes but now I see the problem I couldn't see then. Am I right?

So, shouldn't we use in the destructor of `ChannelSocket` and `PayloadChannelSocket` the same patter we use in many other classes such as `::UnixStreamSocket`? This is:

```c++
UnixStreamSocket::~UnixStreamSocket()
{
	MS_TRACE_STD();

	delete[] this->buffer;

	if (!this->closed)
		Close();
}

void UnixStreamSocket::Close()
{
	MS_TRACE_STD();

	if (this->closed)
		return;

        // do more things here
```

?

worker/src/DepLibUV.cpp
```diff
@@ -20,8 +20,8 @@ inline static void onWalk(uv_handle_t* handle, void* arg)
 {
 	// Must use MS_ERROR_STD since at this point the Channel is already closed.
 	MS_ERROR_STD(
-	  "alive UV handle found (this shouldn't happen) [type:%d, active:%d, closing:%d, has_ref:%d]",
-	  handle->type,
+	  "alive UV handle found (this shouldn't happen) [type:%s, active:%d, closing:%d, has_ref:%d]",
+	  uv_handle_type_name(handle->type),
```
ibc, line 24: :+1:

ibc (approved):
I love the changes.

ibc:
> So, shouldn't we use in the destructor of `ChannelSocket` and `PayloadChannelSocket` the same patter we use in many other classes such as `::UnixStreamSocket`?

I'm gonna do this.

ibc:
Done: https://github.com/versatica/mediasoup/commit/85be87acd53613815b202d5bc5625c7767ef22de

ibc:
> So I think this is a regression due to previous changes, right?

No, it was not a regression, it was always that way :)

ibc:
Let me know if something is missing, otherwise I'll do a 3.7.12 release.

nazar-pc:
Looks good to me, thanks!
Please also increment patch version of `mediasoup-sys` FFI crate (`worker/Cargo.toml`) and #579 would be nice to get in so I can publish a new release to crates too.

ibc:
mediasoup-sys version set to 0.1.4 and mediasoup 3.7.12 released.

jmillan (commented):
:+1:

PR #585, opened 2021-06-11
Fix build on FreeBSD

smortex:
FreeBSD make and GNU make share a common base but provide extensions
not supported by the other.  This leads to build failures and can be
fixed by using the right version of make.  On FreeBSD, GNU make is
installed as `gmake` so detect and use this.  While here, also check
the environment for allowing overriding the binary to make in case
other platform can benefit from it (e.g. OpenBSD, NetBSD and probably
others).

While here, also catch the right number of CPU for parallel build.

### DIFF
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -4,6 +4,7 @@ const fs = require('fs');
 const { execSync } = require('child_process');
 const { version } = require('./package.json');
 
+const isFreeBSD = os.platform() === 'freebsd';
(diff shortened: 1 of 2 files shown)


jmillan (approved):

ibc (approved):

ibc:
Thanks, will merge later today.

ibc:
Released in 3.7.13.

PR #587, opened 2021-06-14
Improvements to Rust version and various general cleanups

nazar-pc:
This resolves all but last TODO in Rust code (which is Windows support).

Changes that are not specific to Rust:
* `kind` on `RtpHeaderExtension` is no longer optional (it was only in the type system, if someone tried to not specify it, it wouldn't work anyway)
* `fecMechanisms` was removed from `RtpCapabilities` as it wasn't used anywhere
* `producerId` field in RTP observer messages for worker moved from `internal` into `data` field for consistency, this was one exception which annoyed me

I have also updated changelog files.


jmillan (approved):
💯

src/RtpParameters.ts
```diff
@@ -104,7 +99,7 @@ export type RtpHeaderExtension =
 	 * Media kind. If empty string, it's valid for all kinds.
```
ibc, line 104: Description is wrong :)

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
```
ibc, line 3: Empty line after title, please.

rust/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+### 0.8.0
```
ibc, line 3: Same.

ibc (changes requested):
LGTM just cosmetic changes requested. Also note that changes in the website docs are required. I can do that but it needs to be done before a new release since here we are doing "breaking" changes (even if no one will probably notice).

worker/src/RTC/Router.cpp
```diff
@@ -494,14 +494,14 @@ namespace RTC
 		return rtpObserver;
 	}
 
-	RTC::Producer* Router::GetProducerFromInternal(json& internal) const
+	RTC::Producer* Router::GetProducerFromData(json& internal) const
 	{
 		MS_TRACE();
 
 		auto jsonProducerIdIt = internal.find("producerId");
```
ibc: This is wrong, right? I mean, it should be `data` (argument name).

ibc: so also in the hpp file

ibc (changes requested):

worker/src/RTC/Router.cpp
nazar-pc: You are right, sorry

ibc (approved):

ibc:
Waiting for CI before merging. I'm also changing it in mediasoup-client (not sure yet if needed).

nazar-pc:
Not required, but would be nice to align mediasoup-client similarly to this, yes.

ibc:
Released in 3.7.15.

nazar-pc:
Please create Rust tag for 0.8.0 too

ibc:
Done

PR #589, opened 2021-06-14
Update CI setup with Node.js 16 and remove EOL'd versions

nazar-pc:
Changes:
* add Node.js 16 (will become LTS)
* remove Node.js 10 and 15 (they are EOL, so no point to test any longer)
* remove Ubuntu 16.04 (again, it is EOL unless you pay Canonical a lot of money, so no point to test there either)


ibc (approved):
Thanks

PR #591, opened 2021-06-21
Rust 0.8.1

nazar-pc:
Fix for clippy lint, some convenience methods on enums so that there is less need to use match statements externally and version bump for publishing.


ibc (approved):

ibc:
git tag `rust-0.8.1` pushed

PR #592, opened 2021-06-23
Improve typing of PipeTransport.consume() and ConsumerOptions

nazar-pc:
Documentation already said that `ConsumerOptions.rtpCapabilities` is required, but because of `PipeTransport.consume()` it wasn't in code. Now that is fixed with dedicated `PipeConsumerOptions`.


jmillan (approved):

ibc (approved):

ibc:
Thanks. Released in 3.7.17.

PR #594, opened 2021-07-05
add cmake support

johzzy:
base git@github.com:barry-ran/mediasoup.git


ibc:
Which is the motivation of this PR? Is it just a cosmetic change?

johzzy:
> Which is the motivation of this PR? Is it just a cosmetic change?

for step-by-step operation, but now it is all submitted.

nazar-pc:
I see OpenSSL was updated, if that is updated it should be done in a separate PR to make this one much smaller.

johzzy:
> Which is the motivation of this PR? Is it just a cosmetic change?



> I see OpenSSL was updated, if that is updated it should be done in a separate PR to make this one much smaller.

splitted.

nazar-pc:
> splitted.

This PR still contains changes related to congestion control and over 200k LoC of changes, which as you can imaging is not reviewable as is.

It would be nice to have just OpenSSL upgrade separately in its own PR and CMake changes separately, so that OpenSSL upgrade can be verified separately and it is clear what changes are needed for CMake itself, which I would imagine isn't quite as big as 200k LoC.

PR #595, opened 2021-07-05
Optimize TransportCongestionControl handling process

johzzy:


worker/include/RTC/TransportCongestionControlServer.hpp
```diff
@@ -22,65 +21,53 @@ namespace RTC
...
+        return 0u;
+      }
+      virtual void IncomingPacket(uint64_t nowMs, const RTC::RtpPacket* packet) = 0;
+      virtual void TransportConnected() {}
+      virtual void TransportDisconnected() {}
+
+      void SetMaxIncomingBitrate(uint32_t bitrate);
+      void MaySendLimitationRembFeedback();
+
+      static std::unique_ptr<TransportCongestionControlServer::Strategy> Create(
```
nazar-pc, line 36: I suspect this needs to be `thread_local`

nazar-pc:
Can you write down a few paragraphs about what is done, why and how this optimization positively affects something?

johzzy:
> Can you write down a few paragraphs about what is done, why and how this optimization positively affects something?

This PR splits TWCC and REMB so that they are more independent, also make the execution process more concise and efficient.

Such as, `TransportCongestionControlStrategyREMB` is a`webrtc::RemoteBitrateEstimator::Listener`, And `TransportCongestionControlStrategyTWCC` is a` Timer::Listener`

ibc:
BTW we are writing our own BWE so will remove libwebrtc dep.

PR #598, opened 2021-07-07
Upgrade and optimize openssl dependency

vpalmisano:
This PR includes the following changes:
- openssl upgrade to version 1.1.1k; the full deps/openssl folder has beed taken from https://github.com/nodejs/node/tree/v16.4.2/deps/openssl
- enable the compilation of assembly extensions for openssl (https://mediasoup.discourse.group/t/why-compile-openssl-with-no-asm-by-default/1022/4)
- optimizes the build (-O3) and disables the debug flag (-g)

Some test results with openssl 1.1.1k:
1. default build options (with no openssl assembly support, compiled with debug symbols)
01_default-build

2. no openssl assembly support, no debug symbols and optimized with -O3
02_default-build-optimized_-O3

3. with openssl assembly support, no debug symbols and optimized with -O3
03_custom-build-optimized_-O3


nazar-pc:
`-O3` seems to provide a big win (I doubt debug symbols have much impact), but wondering how safe is it to use for OpenSSL.

vpalmisano:
> `-O3` seems to provide a big win (I doubt debug symbols have much impact)

debug symbols don't have impact on CPU usage, but on executable size:
- with: 92M
- without: 7,5MB

> but wondering how safe is it to use for OpenSSL.

Why? Nodejs is using -O3: https://github.com/nodejs/node/blob/master/common.gypi#L180

ibc:
Other than the question by @nazar-pc, I cannot say "no" to this 😅 so thanks a lot. Let's us take a look tomorrow morning and merge.

nazar-pc:
> Why? Nodejs is using -O3: https://github.com/nodejs/node/blob/master/common.gypi#L180

That was just a question, I know Linux kernel maintainers still refuse to use `-O3` due to bugs in some compilers/versions.

Looks great otherwise!

vpalmisano:
Here are the test results using -O2, no much difference:
03_custom-build-optimized_-O2

vpalmisano:
I've tested with -O2 instead of -O3 with these results:

03_custom-build-optimized_-O2

Performance is ~5% worst compared to -O3.
In general the difference in not too big, but I need to collect more data to have a more precise comparison.

jmillan:
Just wondering if removing the the newer version of MSVS is correct here: worker/deps/gyp/pylib/gyp/MSVSVersion.py

Other than that, looks great @vpalmisano

jmillan (approved):

vpalmisano:
> Just wondering if removing the the newer version of MSVS is correct here: worker/deps/gyp/pylib/gyp/MSVSVersion.py

I've updated the gyp package from the origin (https://chromium.googlesource.com/external/gyp.git/+/703706c4995b4c9ed0d5f319d684ba52647ab255%5E%21/#F0), trying to fix the error in lines `'gas_version and v(gas_version) >= v("2.26") or '` (with no success). I can revert the change anyway.

ibc:
@vpalmisano we want to migrate to CMake anyway.

ibc:
mediasoup 3.7.18 released! Thanks!

worker/deps/openssl/openssl.gyp
```diff
@@ -21,12 +21,12 @@
         }, 'target_arch=="arm64" and OS=="win"', {
           # VC-WIN64-ARM inherits from VC-noCE-common that has no asms.
           'includes': ['./openssl_no_asm.gypi'],
-        }, 'gas_version and v(gas_version) >= v("2.26") or '
-           'nasm_version and v(nasm_version) >= v("2.11.8")', {
-           # Require AVX512IFMA supported. See
-           # https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_ia32cap.html
-           # Currently crypto/poly1305/asm/poly1305-x86_64.pl requires AVX512IFMA.
-          'includes': ['./openssl_asm.gypi'],
+        #}, 'gas_version and v(gas_version) >= v("2.26") or '
```
nazar-pc, line 24: Why was this commented-out? It should (theoretically) use even faster ASM if I understand it correctly.

vpalmisano: Sorry for delay, I've seen now your comment. 
The block is commented because the `v()` macro is not available in node-gyp, and we have to assume that gas/nasm installed versions satisfy the requirement.

PR #601, opened 2021-07-08
Cleanup for non-existing header files and potentially uninitialized local variable

nazar-pc:
Those 2 header files do not actually exist (found while playing with GYP's CMake generator :wink: ).

Local variable is, obviously, defined for all practical purposes, but compiler complained so decided to make it happy, I hope that's fine.

### DIFF
--- a/worker/mediasoup-worker.gyp
+++ b/worker/mediasoup-worker.gyp
@@ -438,8 +438,8 @@
         'fuzzer/src/RTC/RTCP/FuzzerXr.cpp',
         # C++ include files.
         'fuzzer/include/FuzzerUtils.hpp',
-        'fuzzer/include/RTC/FuzzerStunMessage.hpp',
         'fuzzer/include/RTC/FuzzerRtpPacket.hpp',
+        'fuzzer/include/RTC/FuzzerStunPacket.hpp',
         'fuzzer/include/RTC/FuzzerTrendCalculator.hpp',
         'fuzzer/include/RTC/RTCP/FuzzerBye.hpp',
         'fuzzer/include/RTC/RTCP/FuzzerFeedbackPs.hpp',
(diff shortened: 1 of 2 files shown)


worker/mediasoup-worker.gyp
```diff
@@ -438,7 +438,6 @@
         'fuzzer/src/RTC/RTCP/FuzzerXr.cpp',
         # C++ include files.
         'fuzzer/include/FuzzerUtils.hpp',
-        'fuzzer/include/RTC/FuzzerStunMessage.hpp',
```
ibc, line 441: Good catch, but this should be `FuzzerStunPacket.hpp` which does exist.

worker/mediasoup-worker.gyp
```diff
@@ -461,7 +460,6 @@
         'fuzzer/include/RTC/RTCP/FuzzerFeedbackRtpTransport.hpp',
         'fuzzer/include/RTC/RTCP/FuzzerPacket.hpp',
         'fuzzer/include/RTC/RTCP/FuzzerReceiverReport.hpp',
-        'fuzzer/include/RTC/RTCP/FuzzerSdesReport.hpp',
```
ibc, line 464: And this should be `FuzzerSdes.hpp`.

ibc (changes requested):
:)

worker/src/RTC/Transport.cpp
```diff
@@ -2371,7 +2371,7 @@ namespace RTC
 				// provide it with more available bitrate to choose its preferred layers.
 				for (uint8_t i{ 1u }; i <= priority; ++i)
 				{
-					uint32_t usedBitrate;
+					uint32_t usedBitrate = 0u;
```
ibc: We use this everywhere AFAIR:

```suggestion
					uint32_t usedBitrate{ 0u };
```

ibc:
Thanks :) waiting for CI

ibc (approved):

PR #602, opened 2021-07-09
add cmake support

johzzy:
base git@github.com:barry-ran/mediasoup.git
ref https://github.com/versatica/mediasoup/pull/594


ibc:
Thanks for this.

There is however something that worries me here. It looks that this PR requires changes in most of the C/C++ dependencies that we use in `mediasoup-worker/deps/` so, how are we supposed to update them in the future?

nazar-pc:
This is a good start, but at least following improvements are needed IMO:
* Comments in English
* CI that is similar to current setup we have instead of something completely different
* Test, fuzzer, libmediasoup-worker targets are needed or else we can't run CI and it isn't generally a usable replacement t GYP
* instructions with straightforward way to upgrade dependencies (not sure if it is possible in all cases, but ideally they should be downloaded and usable unmodified during installation instead of being bundled all in the repo)
* libuv and potentially other libraries should be linked in statically instead of dynamically

james133:
support vc2019,xcode,clion 
https://github.com/yanhua133/mediasoup-sfu-cpp

jmillan:
@james133, what exactly are you sharing?

PR #605, opened 2021-07-11
C++ cleanup

nazar-pc:
This just fixes a bunch of compiler warnings (not all of them, but still)

### DIFF
--- a/worker/deps/libwebrtc/libwebrtc/mediasoup_helpers.h
+++ b/worker/deps/libwebrtc/libwebrtc/mediasoup_helpers.h
@@ -27,20 +27,20 @@ namespace mediasoup_helpers
 			}
 
 			return receivedPackets;
-		};
+		}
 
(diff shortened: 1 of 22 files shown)


jmillan (approved):

jmillan:
Which compiler is giving such warnings @nazar-pc, gcc?

ibc (approved):
thanks

ibc:
> Which compiler is giving such warnings @nazar-pc, gcc?

Good question, because it's long time since I see any related warning using gcc and clang. Merging BTW.

nazar-pc:
It is GCC. I'm experimenting with Meson build system, non-zero log level produces A LOT of warnings, this PR fixes a good bunch of those, but many still remain.

nazar-pc:
These are what Meson enables with warnings level 1: `-Wall -Winvalid-pch -Wnon-virtual-dtor`

PR #606, opened 2021-07-11
RFC: Meson build system

nazar-pc:
### This PR is a request for comments for Meson build system.

I spent a lot of time trying to add CMake and it was so painful that I decided to look at Meson that many projects are moving towards and found it surprisingly expressive and easy to use. This is basically as an alternative solution for #359.

This PR implements Meson support for building `mediasoup-worker`, `libmediasoup-worker` and tests.

If this makes sense to everyone I can things currently missing, switch from GYP and remove things which are GYP-specific. The only concern there is fuzzing-related code that I have never used myself, but it should be doable too.

### Important notes:
* OpenSSL is no longer built from source, it is expected to be installed on the system (it would be possible to add OpenSSL Meson build, but it would take significant amount of time that I don't have at the moment, see https://github.com/mesonbuild/wrapdb/issues/109)
* following dependencies no longer need to be bundled and can be installed on the fly if not already present on the system:
  * abseil-cpp
  * libsrtp2
  * libuv
  * usrsctp
  * catch2
* It is possible to easily make build that relies on system shared libraries instead of building them from scratch or building them statically (if needed), in fact system-wide static OpenSSL library is already used since no version that can be built from source is provided

### How to try it out:
Assuming you have `meson` and `ninja` installed:
```bash
cd worker
# Prepare `build` directory where builds will happen
meson setup build
# Build mediasoup-worker
meson compile -C build mediasoup-worker
# Build static C++ library
meson compile -C build libmediasoup-worker
# Build and run worker tests
meson test -C build
```

### DIFF
--- a/worker/meson.build
+++ b/worker/meson.build
@@ -1,0 +1,285 @@
+project(
+  'mediasoup-worker',
+  ['c', 'cpp'],
+  default_options : [
+    'warning_level=1',
+    'cpp_std=c++11',
+    'default_library=static',
+  ],
+  meson_version: '>= 0.58',
+)
+
+cpp_args = [
+  host_machine.endian() == 'little' ? '-DMS_LITTLE_ENDIAN' : '-DMS_BIG_ENDIAN',
+  '-DABSL_ALLOCATOR_NOTHROW=1',
+]
+
+if host_machine.system() == 'windows'
+  cpp_args += [
+    '-DNOMINMAX', # Don't define min and max macros (windows.h)
+    # Don't bloat namespace with incompatible winsock versions.
+    '-DWIN32_LEAN_AND_MEAN',
+    # Don't warn about usage of insecure C functions.
+    '-D_CRT_SECURE_NO_WARNINGS',
+    '-D_SCL_SECURE_NO_WARNINGS',
+    # Introduced in VS 2017 15.8, allow overaligned types in aligned_storage.
+    '-D_ENABLE_EXTENDED_ALIGNED_STORAGE'
+  ]
+endif
+
+common_sources = [
+  'src/lib.cpp',
+  'src/DepLibSRTP.cpp',
+  'src/DepLibUV.cpp',
+  'src/DepLibWebRTC.cpp',
+  'src/DepOpenSSL.cpp',
+  'src/DepUsrSCTP.cpp',
+  'src/Logger.cpp',
+  'src/MediaSoupErrors.cpp',
+  'src/Settings.cpp',
+  'src/Worker.cpp',
+  'src/Utils/Crypto.cpp',
+  'src/Utils/File.cpp',
+  'src/Utils/IP.cpp',
+  'src/Utils/String.cpp',
+  'src/handles/SignalsHandler.cpp',
+  'src/handles/TcpConnectionHandler.cpp',
+  'src/handles/TcpServerHandler.cpp',
+  'src/handles/Timer.cpp',
+  'src/handles/UdpSocketHandler.cpp',
+  'src/handles/UnixStreamSocket.cpp',
+  'src/Channel/ChannelNotifier.cpp',
+  'src/Channel/ChannelRequest.cpp',
+  'src/Channel/ChannelSocket.cpp',
+  'src/PayloadChannel/Notification.cpp',
+  'src/PayloadChannel/PayloadChannelNotifier.cpp',
+  'src/PayloadChannel/PayloadChannelRequest.cpp',
+  'src/PayloadChannel/PayloadChannelSocket.cpp',
+  'src/RTC/AudioLevelObserver.cpp',
+  'src/RTC/Consumer.cpp',
+  'src/RTC/DataConsumer.cpp',
+  'src/RTC/DataProducer.cpp',
+  'src/RTC/DirectTransport.cpp',
+  'src/RTC/DtlsTransport.cpp',
+  'src/RTC/IceCandidate.cpp',
+  'src/RTC/IceServer.cpp',
+  'src/RTC/KeyFrameRequestManager.cpp',
+  'src/RTC/NackGenerator.cpp',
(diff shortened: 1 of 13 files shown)


worker/meson.build
```diff
@@ -0,0 +1,227 @@
...
+  'openssl',
+  static: get_option('default_library') == 'static',
+  version: '>= 1.1.1'
+)
+
+if cpp.has_header('nlohmann/json.hpp')
+  nlohmann_json_dep = declare_dependency()
+else
+  nlohmann_json_proj = subproject('nlohmann_json')
+  nlohmann_json_dep = nlohmann_json_proj.get_variable('nlohmann_json_dep')
```
eli-schwartz: I'd just like to note, you do not need this...

Simply using `dependency('nlohmann_json')` will first check for nlohmann_json on the system using pkg-config (installed in upstream nlohmann_json's meson.build), then check for it using cmake (installed in upstream nlohmann_json's CMakeLists.txt), then finally check for it via the wrap file which specifies:

```
[provide]
nlohmann_json = nlohmann_json_dep
```

nazar-pc: I wasn't sure how it is called in an OS, this is the first time I use Meson, thanks for suggestion!

worker/meson.build
```diff
@@ -0,0 +1,227 @@
...
+  'src/RTC/RTCP/FeedbackPsRemb.cpp',
+  'src/RTC/RTCP/XR.cpp',
+  'src/RTC/RTCP/XrDelaySinceLastRr.cpp',
+  'src/RTC/RTCP/XrReceiverReferenceTime.cpp',
+]
+
+cpp = meson.get_compiler('cpp')
+
+openssl_dep = dependency(
+  'openssl',
```
eli-schwartz, line 148: Meson has an experimental module for building wraps using configure/make style build systems: https://mesonbuild.com/External-Project-module.html

It will probably work with openssl, assuming you build statically (which third-party vendored dependencies should do anyway regardless of build system. :D)

nazar-pc: Hm, I might try that. I experimented with https://github.com/viaduck/openssl-cmake with CMake integration, but it didn't build anything despite not throwing errors.

On Linux it might be fine to use configure/make, but at the same time it is kind of a pain to setup a lot of components (like Perl for OpenSSL) on Windows. Having just Meson+Ninja+C compiler is much nicer.

I might end up writing meson.build for OpenSSL, I'll see what maintainers think about Meson in general first.

eli-schwartz: Right, it's experimental after all. :D (And a teeny bit of a hack.)

ibc:
I will start mini holidays and won't be able to properly review this, but there is something that worries me too much:

> OpenSSL is no longer built from source

One of the goals of mediasoup is the ability to do `npm install mediasoup` (or `cargo add mediasoup`) and that's all. Others require having specific lib versions installed globally or somewhere.

When it comes to OpenSSL this is even more worried. We have merged recently a PR that updates OpenSSL and optimizes it at maximum. Telling the user that he must install OpenSSL manually and "optimize" it is a no-go for me.

Anyway, I'll take a proper look next week.

nazar-pc:
If OpenSSL is the only concern then I'll try to take care of https://github.com/mesonbuild/wrapdb/issues/109 and have it built from source, it is just not built YET.

This is MVP to show what Meson is and how it comes together, more work is needed before it is ready for merge, I just want to make sure it has a chance of being accepted before investing more time into it (already spent 3 evenings).

ibc:
> This is MVP to show what Meson is and how it comes together, more work is needed before it is ready for merge, I just want to make sure it has a chance of being accepted before investing more time into it (already spent 3 evenings).

Sure, makes sense. @jmillan may you take a look to this these days, please?

jmillan:
Apart from the OpenSSL concern and the missing efforts for the fuzzer it seems that Meson will relief us from the limitations we are finding with gyp. 

This seems a better alternative to the current build system so my opinion is positive on this regard.

Regarding to the installation process we are adding `meson` to initial requirements, which basically consists on running `pip install meson` (we already require Python).  Right?

I'm not saying we should stay away from trying to use CMake, but this is at least a better alternative than gyp for the time being.

nazar-pc:
OpenSSL is something I'm working on already.

I think we'll only add PIP to requirements (likely already installed, but not necessarily as it is a separate package on Ubuntu for instance) and then we can use PIP to install both Meson and Ninja if needed (Meson doesn't generate Makefiles).

Once OpenSSL is done, I don't see any obstacles with adding fuzzer to the mix.

nazar-pc:
Thanks for the feedback!

Closing for now, will re-open once I have it more or less full-featured.

I have already prepared OpenSSL wrap for Meson (which was the major concern), currently waiting for upstream to decide how exactly it is going to be integrated due to large number of generated files (similarly to how it is done in Node.js repo): https://github.com/mesonbuild/wrapdb/pull/113

nazar-pc:
Current progress:
* :heavy_check_mark: GYP is fully replaced with Meson in my branch
* :heavy_check_mark: CI runs for existing targets (except tests fail on one as in #618)
* :warning: Added Windows CI targets
  * :warning: GCC and Clang compile everything, but fail to run tests, probably because of #618, but not sure
  * :hourglass: MSVC almost compiles, waiting for https://github.com/mesonbuild/wrapdb/pull/116 to land
* :heavy_check_mark: All dependencies, including OpenSSL are build from sources
  * :hourglass: OpenSSL and netstring-c have pending Meson wrap PRs (currently they are bundled)
  * :hourglass: Didn't bother with getopts yet, will probably do too
* :heavy_check_mark: All code dependencies except getopts are installed from subprojects (this means downloaded during compilation instead of bundled in the repo)
* :warning: Requirements updates:
  * :warning: Python 3 with PIP on all platforms
  * :warning: `make` is required on Windows (it now works almost identically to Linux/macOS)

PR #607, opened 2021-07-12
Enhance debugging on channel request/notification error

jmillan:
Make the error show the corresponding event method or notification event.


jmillan:
Updating tests...

ibc (commented):
Super tested? what is `error.buffer`?

jmillan:
> Super tested?
yes

> what is error.buffer?

It's the string that was provided to the current `std::exception`. 

I can change it by `error.what()` which will return the same value.

ibc (approved):

PR #608, opened 2021-07-12
Move header includes

nazar-pc:
This contains 2 commits, review them separately, commit messages should make sense.

There is no logic change, just moving things around so that paths are aligned with what they look in Meson and globally installed libraries, so it will make #606 a lot smaller and easier to review.

Even ignoring Meson, these changes to not hurt anyone and do not impact anything, so I hope it is okay.


ibc (approved):
Nice!

jmillan (commented):
:+1:

PR #609, opened 2021-07-13
Add a delay before sending nack

### LINKED ISSUE #511: Check libwebrtc nack_module2.cc (add delay before sending NACK)
## Feature Request
When we were testing mediasoup with network bonding(round-robin), we found that the cpu usage is about 10% higher than normal.

We cap the packet, and found that there are a few disordered rtp packet.

image

Once mediasoup receiving a disordered packet, it will send a nack immediately.

But the arrival time of these packets are very close (<1ms).

So maybe we should add a delay before sending nack.

In libwebrtc there is also a param `WebRTC-SendNackDelayMs` to set the delay time of sending nack.

https://webrtc.googlesource.com/src/+/refs/heads/master/modules/video_coding/nack_module2.cc#303

penguinol:
fix #511


jmillan:
@penguinol we would benefit of some tests (in TestNackGenerator.cpp file) for this PR.

nazar-pc:
And from corresponding Rust updates too :wink:

penguinol:
> @penguinol we would benefit of some tests (in TestNackGenerator.cpp file) for this PR.

Yep, i need to think about how to test it, it's time related.

ibc:
Hi, we don't forget. Just holidays here and need to check this carefully.

nazar-pc:
We discussed this. It makes sense, but we'll likely have the delay hardcoded in the implementation for simplicity.

ibc:
Coming back to this PR, we don't want it to be configurable. It must be hardcoded behavior if we know that it's better.

penguinol:
According to libwebrtc, nack delay should between 0 ~ 20ms, and default value is 10ms
https://webrtc.googlesource.com/src/+/refs/heads/main/modules/video_coding/nack_requester.cc#35
```
int64_t GetSendNackDelay(const FieldTrialsView& field_trials) {
  int64_t delay_ms = strtol(
      field_trials.Lookup("WebRTC-SendNackDelayMs").c_str(), nullptr, 10);
  if (delay_ms > 0 && delay_ms <= 20) {
    RTC_LOG(LS_INFO) << "SendNackDelay is set to " << delay_ms;
    return delay_ms;
  }
  return kDefaultSendNackDelayMs;
}
```

ibc:
We will review this next week.

penguinol:
Well, it seems the test cases should also be motified. Because we will not send nack immediately now.

ibc:
Sorry for delay, handling this PR this week.

ibc:
Some tests are failing: https://github.com/versatica/mediasoup/runs/6342420672?check_suite_focus=true

penguinol:
> Some tests are failing: https://github.com/versatica/mediasoup/runs/6342420672?check_suite_focus=true

Right, i've noticed this. It's because in this pr the nack packets will be sent after at least 10ms. So these test failures mean it works.
We may add some sleep in test cases to fix it.
I will handle this later, a little busy these days.

worker/include/RTC/NackGenerator.hpp
```diff
@@ -30,14 +30,16 @@ namespace RTC
...
+			explicit NackInfo(uint16_t seq, uint16_t sendAtSeq, uint64_t createAtMs)
+			  : seq(seq), sendAtSeq(sendAtSeq), createAtMs(createAtMs)
 			{
 			}
 
 			uint16_t seq{ 0u };
 			uint16_t sendAtSeq{ 0u };
 			uint64_t sentAtMs{ 0u };
 			uint8_t retries{ 0u };
+			uint64_t createAtMs{ 0u };
```
ibc: Can we rename to `createdAtMs`?

ibc:
We need to resolve the tests by waiting some ms between them.

penguinol:
> We need to resolve the tests by waiting some ms between them.

ibc:
@penguinol I'm, working on this in a separate PR with some additions: https://github.com/versatica/mediasoup/pull/827

ibc:
Closing since https://github.com/versatica/mediasoup/pull/827 is merged. Thanks a lot @penguinol!

PR #617, opened 2021-07-22
New mediasoup-sys release

nazar-pc:
Python 3.9 after OS upgrade was very persistent in creation of `*.pyc` files, had to disable it with environment variable, otherwise cargo complains that source directory was modified when crate is built.


ibc (approved):

ibc:
Waiting for CI

PR #626, opened 2021-07-27
fix timestamp when simulcast consumer switching layers

### LINKED ISSUE #408: There are mosaics on video when switching layers
Hi, guys

I found that there are mosaics on video when switching layers, as follow:

I captured packet. It seems because we did not finish sending the last lower layer frame packets when the higher layer i-frame came.
consumer packets:

producer packets:

In consumer packets, there is serial sequence number and the timestamp of two frame is the same.
This might cause the mosics.

### LINKED ISSUE #615: Timestamp is not increased correctly when switching layers.
Look at picture below:
image
The packets on the left is received by mediasoup, and on the right is sent by mediasoup.
The video source is a 2 spatial-layers simulcast stream.
```
seq = 13294, time = 2538709271 -> seq = 30856, time = 2538709271  offset 0
seq = 13295, time = 2538709271 -> seq = 30857, time = 2538709271  offset 0
seq = 13296, time = 2538709271 -> seq = 30858, time = 2538709271  offset 0
switch layer
seq = 26483, time = 2538709271 -> seq = 30859, time = 2538712241 offset 33 * 90
seq = 26484, time = 2538709271 -> seq = 30859, time = 2538712241 offset 33 * 90
```
The timestamp of the two stream is the same, so we set the offset to 33ms when switching layer.
According to https://github.com/versatica/mediasoup/blob/237f7b2ac6aaf68868df413df523e5848a8a2a4e/worker/src/RTC/SimulcastConsumer.cpp#L792-L798

And then we switch layer again
image
```
seq = 26660, time = 2539111121 -> seq = 31036, time = 2539114091 offset 33 * 90
seq = 26661, time = 2539111121 -> seq = 31037, time = 2539114091 offset 33 * 90
seq = 13429, time = 2539111121 -> not sps/pps, drop
switch layer
seq = 13430, time = 2539114001 -> seq = 31038, time = 2539114091 offset 1 * 90
seq = 13431, time = 2539114001 -> seq = 31039, time = 2539114091 offset 1 * 90
```
So we can see the timestamp of two stream is different, but is changed to the same value.
Because of  2539114001 is less than 2539114091, we go into this if case
https://github.com/versatica/mediasoup/blob/d63758499e16e45894fc41bb765145aacae7bb04/worker/src/RTC/SimulcastConsumer.cpp#L747-L750
And `offset  = 2539114091 - 2539114001 = 90` according to 
https://github.com/versatica/mediasoup/blob/d63758499e16e45894fc41bb765145aacae7bb04/worker/src/RTC/SimulcastConsumer.cpp#L758-L759

It lead to decode error on mac os, and is may be one of the causes of #408

penguinol:
Fix #615 

Previous code only add another `33ms` when `current Timestamp = MaxPacketTs`.
And when `current Timestamp < MaxPacketTs`, it add a `tsExtraOffset` and make  `current Timestamp + tsExtraOffset = MaxPacketTs`.

We should also add a `33ms` delay when `current Timestamp < MaxPacketTs`.


worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -753,10 +753,14 @@ namespace RTC
 				// Max delay in ms we allow for the stream when switching.
 				// https://en.wikipedia.org/wiki/Audio-to-video_synchronization#Recommendations
 				static const uint32_t MaxExtraOffsetMs{ 75u };
+				
+				// It's common that, when switching spatial layer, the resulting TS for the
+				// outgoing packet matches the highest seen in the previous stream. Fix it.
```
jmillan:
```suggestion
				// outgoing packet matches the highest seen in the previous stream.
				// Apply an expected offset for a new frame in a 30fps stream.
```

jmillan (approved):
:+1:

ibc:
Thanks for this. So, does this PR also fix #408?

ibc:
Please run `make format` in worker folder.

penguinol:
> Thanks for this. So, does this PR also fix #408?

This can reduce the probability of #408.
But the problem can't be completely solved with this fix.
Skip on rtp secquence also helps.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -754,9 +754,13 @@ namespace RTC
 				// https://en.wikipedia.org/wiki/Audio-to-video_synchronization#Recommendations
 				static const uint32_t MaxExtraOffsetMs{ 75u };
 
+				// outgoing packet matches the highest seen in the previous stream.
```
jmillan:
```suggestion
				// Outgoing packet matches the highest timestamp seen in the previous stream.
```

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -754,9 +754,13 @@ namespace RTC
...
+				// Apply an expected offset for a new frame in a 30fps stream.
+				static const uint8_t MsOffset{ 33u }; // (1 / 30 * 1000).
+
 				int64_t maxTsExtraOffset = MaxExtraOffsetMs * this->rtpStream->GetClockRate() / 1000;
-				uint32_t tsExtraOffset =
-				  this->rtpStream->GetMaxPacketTs() - packet->GetTimestamp() + tsOffset;
+				uint32_t tsExtraOffset   = this->rtpStream->GetMaxPacketTs() - packet->GetTimestamp() +
+				                         tsOffset + MsOffset * this->rtpStream->GetClockRate() / 1000;
 
 				// NOTE: Don't ask for a key frame if already done.
```
ibc, line 761: Shouldn't we move this condition above so we don't do unused computation?

Also, "make format" required in worker/ folder to fix lint issues.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -813,7 +808,12 @@ namespace RTC
 			this->tsOffset = tsOffset;
 
 			// Sync our RTP stream's sequence number.
-			this->rtpSeqManager.Sync(packet->GetSequenceNumber() - 1);
+			// If previous frame has not been sent completely when we switch layer,
+			// We can tell libwebrtc that previous frame is uncompletely by skipping one RTP sequence num.
+			// 'packet->GetSequenceNumber() -2' may increase SeqManager::base and increase the output
+			// seqence num. 
+			// https://github.com/versatica/mediasoup/issues/408
+			this->rtpSeqManager.Sync(packet->GetSequenceNumber() - (this->lastSentPacketHasMarker ? 1 : 2));
```
ibc, line 817: So this is gonna produce a NACK right?

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -891,6 +891,8 @@ namespace RTC
 		// Process the packet.
 		if (this->rtpStream->ReceivePacket(packet))
 		{
+			this->lastSentPacketHasMarker = packet->HasMarker();
```
ibc: I think this is not ok since RTP packets here could have arrived disordered to mediasoup. We should just set this flag (to true or false) if this is the highest seq number we have sent to the endpoint.

jmillan: correct

worker/src/RTC/SimulcastConsumer.cpp
jmillan: can you confirm it @penguinol?

That is the price to pay for this approach.

worker/src/RTC/SimulcastConsumer.cpp
penguinol: You are right, i'll fix it

worker/src/RTC/SimulcastConsumer.cpp
penguinol: Yes, but it's better than a broken video

worker/src/RTC/SimulcastConsumer.cpp
penguinol: Sorry, which condition do u mean?

ibc: Forget it, bad review from mobile.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -891,6 +892,9 @@ namespace RTC
 		// Process the packet.
 		if (this->rtpStream->ReceivePacket(packet))
 		{
+			if (this->rtpSeqManager.GetMaxOutput() == packet->GetSequenceNumber())
```
ibc, line 895: @jmillan this is exactly as it must be, right? Just to double confirm. I think so.

jmillan: Right :+1:

ibc (approved):

ibc:
Thanks a lot. Will merge and release later.

ibc:
Let's merge this

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -813,7 +809,12 @@ namespace RTC
 			this->tsOffset = tsOffset;
 
 			// Sync our RTP stream's sequence number.
-			this->rtpSeqManager.Sync(packet->GetSequenceNumber() - 1);
+			// If previous frame has not been sent completely when we switch layer,
+			// We can tell libwebrtc that previous frame is uncompletely by skipping one RTP sequence num.
+			// 'packet->GetSequenceNumber() -2' may increase SeqManager::base and increase the output
+			// seqence num.
+			// https://github.com/versatica/mediasoup/issues/408
```
```suggestion
			// If previous frame has not been sent completely when we switch layer, we can tell
			// libwebrtc that previous frame is incomplete by skipping one RTP sequence number.
			// 'packet->GetSequenceNumber() -2' may increase SeqManager::base and increase the
			// output sequence number.
			// https://github.com/versatica/mediasoup/issues/408
```

PR #629, opened 2021-07-29
New Rust release 0.8.3

nazar-pc:
The big Rust-specific change is prelude module addition, which as can be seen from `echo` example includes most things you'd need when working with mediasoup.

There are also other minor fixes and improvements + we need to release Dominant Speaker Event.


jmillan (approved):
💯

PR #632, opened 2021-08-04
Do not railse consumer score event when consumer is paused.

mstyura:
Is it possible to consider changing behavior of consumers not to fire score event, when consumer is paused? Currently I see score updated event when create new consumer in paused state.


ibc (commented):
Nope. Consumer score also contains score about the corresponding Producer. If you want to ignore those score events while the consumer is paused, just check consumer.paused flag and ignore the score event if true.

PR #633, opened 2021-08-06
Touch up some grammar in README.md

FreedomBen:


ibc:
Thanks, but we have the same text in other places (website, Rust docs) and I prefer to not change them just in one place. Will fix those typos in the future when we rewrite those docs.

PR #634, opened 2021-08-12
Erase the dominantId after removal from the map.

SteveMcFarlin:
This fixes a bug where by ActiveSpeakerObserver::CalculateActiveSpeaker()
may cause a crash due to a nullptr. While there is a check after getting
the dominantSpeaker from the map we have notice it will pass the check,
and then crash in EvalActivityScores.


ibc:
Thanks, released in 3.8.2

PR #642, opened 2021-08-19
Emit the dominant speaker event at every interval.

SteveMcFarlin:
The dominant speaker event will now be emitted on each 'interval'. The reasoning for doing this is if you have a producer that is already the dominant speaker and a consumer joins, that consumer will not know who in the "room" is the dominant speaker. This solves the issue for client side applications without having to resort to storing the current dominant speaker in the server application code.

### DIFF
--- a/src/Router.ts
+++ b/src/Router.ts
@@ -930,7 +930,7 @@ export class Router extends EnhancedEventEmitter
 	 */
 	async createActiveSpeakerObserver(
 		{
-			interval = 300,
+			interval = 1000,
 			appData = {}
 		}: ActiveSpeakerObserverOptions = {}
 	): Promise<ActiveSpeakerObserver>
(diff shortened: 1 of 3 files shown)


nazar-pc:
Hm, to me it seems like you'll still have to store to to avoid bombarding clients with unnecessary updates every 300ms or whatever interval happens to be.

~3 messages per second per client is a lot, I think current behavior is much more reasonable default.

SteveMcFarlin:
I thought about adding a flag to indicate emitting the event every timer cycle, but decided keeping the interface free from state would be a better choice for now. The AudioLevelObserver emits far more information every 1sec by default. What I can do is change the default to something more appropriate like 1 sec. I think that is reasonable.

ibc:
Wouldn't be better to just have a getter of current speaker in the ActiveSpeakerObserver class? This way the app can get it when needed (i.e. when a client joins the conf).

SteveMcFarlin:
I think I was a little too hasty in opening this PR, and will close it for now. It is easy enough to just cache this, and send the data after a consumer connects. I'll think about adding a getter. Thanks for the review.

PR #643, opened 2021-08-20
New Rust videoroom example, echo example comments

nazar-pc:
Since Rust is a relatively new language for mediasoup I decided to expand on examples provided with mediasoup a bit.

First of all I have added a bunch of comments to existing `echo` example explaining what it does, why and hopefully making things easier to understand, both on frontend and backend.

Then I have added `videoroom` example that allows multiple participants to join the same room and see/hear each other. It is quite basic with under 1k lines of backend code and under 500 lines of frontend code. Along with a bunch of comments I hope it is helpful for educational purposes and should be a useful starting point for those who want to start hacking on mediasoup.

### DIFF
--- a/rust/examples-frontend/echo/src/index.ts
+++ b/rust/examples-frontend/echo/src/index.ts
@@ -114,15 +114,19 @@ async function init()
 			switch (decodedMessage.action)
 			{
 				case 'Init': {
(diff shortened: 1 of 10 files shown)


ibc (approved):
Amazing

PR #648, opened 2021-08-30
Add packetLoss stats to transport

ggarber:
One common requirement is to monitor the quality of the user's connections to notify the user and the rest of participants when their connections are not good enough so that they can understand any audio/video issue they can experience during conversations.

This is typically based on packetLoss and/or roundTripTime.

With existing APIs those stats are available per consumer and producer so the application could request all/some of them and aggregate the stats but that can be more efficient and simpler if some of those stats are exposed at transport level.

This PR adds packetLoss as a stat at transport level.  The application can retrieve this metric every X seconds and notify the user when its uplink or downlink is bad and also notify rest of participants about it.

The implementation is based on transportCc feedback messages (comparing expected and received values and averaging last 24 of those measurements) so these new stats are only available when transportCC is enabled.

### DIFF
--- a/worker/src/RTC/Transport.cpp
+++ b/worker/src/RTC/Transport.cpp
@@ -552,6 +552,14 @@ namespace RTC
 		// Add maxIncomingBitrate.
 		if (this->maxIncomingBitrate != 0u)
 			jsonObject["maxIncomingBitrate"] = this->maxIncomingBitrate;
+
+		// Add packetLossReceived.
+		if (this->tccServer)
+			jsonObject["rtpPacketLossReceived"] = this->tccServer->GetPacketLoss();
+
+		// Add packetLossSent.
+		if (this->tccClient)
+			jsonObject["rtpPacketLossSent"] = this->tccClient->GetPacketLoss();
 	}
 
 	void Transport::HandleRequest(Channel::ChannelRequest* request)
--- a/worker/src/RTC/TransportCongestionControlClient.cpp
+++ b/worker/src/RTC/TransportCongestionControlClient.cpp
@@ -164,7 +164,7 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		// Update packet loss history
+		// Update packet loss history.
 		size_t expected_packets = feedback->GetPacketStatusCount();
 		size_t lost_packets     = 0;
 		for (const auto& result : feedback->GetPacketResults())
--- a/worker/src/RTC/TransportCongestionControlClient.cpp
+++ b/worker/src/RTC/TransportCongestionControlClient.cpp
@@ -1,5 +1,5 @@
 #define MS_CLASS "RTC::TransportCongestionControlClient"
-// #define MS_LOG_DEV_LEVEL 3
+#define MS_LOG_DEV_LEVEL 3
 
 #include "RTC/TransportCongestionControlClient.hpp"
 #include "DepLibUV.hpp"
@@ -16,6 +16,7 @@ namespace RTC
 	static constexpr float MaxBitrateIncrementFactor{ 1.35f };
 	static constexpr float MaxPaddingBitrateFactor{ 0.85f };
 	static constexpr uint64_t AvailableBitrateEventInterval{ 2000u }; // In ms.
+	static constexpr size_t PacketLossHistogramLength{ 24 };
 
 	/* Instance methods. */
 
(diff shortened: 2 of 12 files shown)


worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -1,5 +1,5 @@
 #define MS_CLASS "RTC::TransportCongestionControlServer"
-// #define MS_LOG_DEV_LEVEL 3
+#define MS_LOG_DEV_LEVEL 3
```
ibc: Upps, this cannot be enabled for releases

worker/src/RTC/TransportCongestionControlClient.cpp
```diff
@@ -163,9 +164,57 @@ namespace RTC
...
+		double totalPacketLoss{ 0 };
+
+		for (auto packetLossEntry : this->packetLossHistory)
+		{
+			weight++;
+			samples += weight;
+			totalPacketLoss += weight * packetLossEntry;
+		}
+
+		[[maybe_unused]] double previousPacketLoss = this->packetLoss;
```
nazar-pc: Mediasoup uses C++11 standard, but this is a C++17 feature, I don't think we can use it yet

worker/src/RTC/TransportCongestionControlServer.cpp
nazar-pc: I don't think this change is supposed to be committed

worker/src/RTC/TransportCongestionControlClient.cpp
```diff
@@ -163,9 +164,57 @@ namespace RTC
...
+		this->UpdatePacketLoss(static_cast<double>(lost_packets) / expected_packets);
+
 		this->rtpTransportControllerSend->OnTransportFeedback(*feedback);
 	}
 
+	void TransportCongestionControlClient::UpdatePacketLoss(double packetLoss)
+	{
+		// Add the score into the histogram.
+		if (this->packetLossHistory.size() == PacketLossHistogramLength)
+			this->packetLossHistory.erase(this->packetLossHistory.begin());
```
nazar-pc: I feel like this will start moving things around on every call once the history grows to `packetLossHistory` size, maybe `std::deque` is a better data structure here?

nazar-pc (commented):
`npm run format:worker` can be used to apply correct formatting to worker code

worker/src/RTC/TransportCongestionControlClient.cpp
jmillan: or `std::queue`

worker/src/RTC/TransportCongestionControlClient.cpp
jmillan: `previousPacketLoss` is not used at all, right?

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -280,6 +299,33 @@ namespace RTC
...
+		double totalPacketLoss{ 0 };
+
+		for (auto packetLossEntry : this->packetLossHistory)
+		{
+			weight++;
+			samples += weight;
+			totalPacketLoss += weight * packetLossEntry;
+		}
+
+		[[maybe_unused]] double previousPacketLoss = this->packetLoss;
```
jmillan: same, `previousPacketLoss` is not used.

worker/src/RTC/TransportCongestionControlClient.cpp
ggarber: It was used for a log that I ended up removing, I will remove the calculation. thx

worker/src/RTC/TransportCongestionControlClient.cpp
ggarber: I used the same approach used for the scores history in RtpStream.hpp for consistency.
```std::vector<uint8_t> scores;```

I will change it in this PR to use `std::queue`

File: worker/src/RTC/Transport.cpp [hunk 5]
```diff
@@ -552,6 +552,14 @@ namespace RTC
...
 		if (this->maxIncomingBitrate != 0u)
 			jsonObject["maxIncomingBitrate"] = this->maxIncomingBitrate;
+
+		// Add packetLossReceived.
+		if (this->tccServer)
+			jsonObject["rtpPacketLossReceived"] = this->tccServer->GetPacketLoss();
+
+		// Add packetLossSent.
+		if (this->tccClient)
+			jsonObject["rtpPacketLossSent"] = this->tccClient->GetPacketLoss();
```
nazar-pc, line 562: These properties need to be added to Rust structs in `rust/src/router/*_transport.rs` files in `*TransportStat` data structures similarly to others that are already there.

nazar-pc (approved):
Looks good

jmillan (approved):

ibc (commented):
doing cosmetic changes

worker/src/RTC/TransportCongestionControlClient.cpp
```diff
@@ -163,9 +164,56 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		// Update packet loss history
```
ibc:
```suggestion
		// Update packet loss history.
```

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -223,6 +231,17 @@ namespace RTC
 		this->listener->OnTransportCongestionControlServerSendRtcpPacket(
 		  this, this->transportCcFeedbackPacket.get());
 
+		// Update packet loss history
```
ibc:
```suggestion
		// Update packet loss history.
```

ibc (approved):

ibc:
This is merged, however one request more, please. Could you add an example of this new stat in mediasoup-website project in https://mediasoup.org/documentation/v3/mediasoup/rtc-statistics/#WebRtcTransport-Statistics?

PR #651, opened 2021-09-05
Fix FailedToParse errors during get_stats() for Consumer and Producer in rust API

Hartigan:
Update ```ProducerStat``` and ```ConsumerStat``` according to ```FillJsonStats``` methods in mediasoup C++ library.

1. https://github.com/versatica/mediasoup/blob/v3/worker/include/RTC/RtpStream.hpp#L202  - rtt is ```float```
2. https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/RtpStreamRecv.cpp#L230  - bitrateByLayer is optinal field

### DIFF
--- a/rust/src/router/consumer.rs
+++ b/rust/src/router/consumer.rs
@@ -239,7 +239,7 @@ pub struct ConsumerStat {
     pub packet_count: usize,
     pub byte_count: usize,
     pub bitrate: u32,
-    pub round_trip_time: Option<f64>,
+    pub round_trip_time: Option<f32>,
 }
 
 /// RTC statistics of the consumer, may or may not include producer statistics.
--- a/rust/src/router/consumer.rs
+++ b/rust/src/router/consumer.rs
@@ -239,7 +239,7 @@ pub struct ConsumerStat {
     pub packet_count: usize,
     pub byte_count: usize,
     pub bitrate: u32,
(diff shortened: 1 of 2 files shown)


rust/src/router/consumer.rs
```diff
@@ -239,7 +239,7 @@ pub struct ConsumerStat {
     pub packet_count: usize,
     pub byte_count: usize,
     pub bitrate: u32,
-    pub round_trip_time: Option<u32>,
+    pub round_trip_time: Option<f64>,
```
nazar-pc: This should be `f32` here and below, it is just `float` on C++ side

nazar-pc (changes requested):
Nice catch!

nazar-pc (approved):
Nice, thank you!

rust/src/router/consumer.rs
Hartigan: Changed ```f32```. Thanks

Hartigan:
@nazar-pc  Can you please upload a new version with fixes to crates.io?

nazar-pc:
Will do probably tomorrow, you can use `git = "https://github.com/versatica/mediasoup.git"` in the meantime

nazar-pc:
Rust 0.8.5 is out

Hartigan:
Thanks

PR #654, opened 2021-09-09
Fix build error of Rust version on M1 mac

satoren:
Require Command line tools for Xcode

### DIFF
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -54,56 +54,19 @@ fn main() {
     }
     #[cfg(target_os = "macos")]
     {
-        let clang_llvm_version = "clang+llvm-12.0.0-x86_64-apple-darwin";
-        let status = Command::new("curl")
+        let path = Command::new("xcrun")
             .args(&[
-                "-L",
-                "-s",
-                "-O",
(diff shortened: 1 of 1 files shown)


nazar-pc:
Hm, the idea was to have a static build of the library, which was the reason to download LLVM and download libc++ from it.

I'm not sure it makes much difference though as realistically macOS will only be used for testing/development, not for production. And having 2 branches, one for x86-64 and another for AArch64 probably isn't worth it :thinking:

nazar-pc:
There is warning in macOS build:
> warning: cargo:rustc-cdylib-link-arg was specified in the build script of mediasoup-sys v0.2.5 (/Users/runner/work/mediasoup/mediasoup/worker), but that package does not contain a cdylib target


I guess that argument is not needed?

satoren:
Removed it.

satoren:
> I'm not sure it makes much difference though as realistically macOS will only be used for testing/development, not for production.

I think so too. And I can not found `clang+llvm-12.0.0-aarch64-apple-darwin` in https://github.com/llvm/llvm-project/releases

nazar-pc (approved):
Makes sense to me, thank you!

PR #655, opened 2021-09-10
Fixes for active speaker observer

ggarber:
Changes:
- Fix the case of audio not received every 20 ms (f.e. if producer is paused, dtx is used or packetization time is more than 20ms).   In that case we need to update multiple entries in the audio levels buffer.   For example if we receive one packet every 100ms we update 5 slots.    Otherwise the algorithm is extremely low on detecting changes in those cases.
- The algorithm was developed by jitsi to use volume and not audio levels (volume = 127 - level).   That's fixed in this PR and I think the results should be better now.
- Optimize storage of audio levels to use a circular buffer and not have to do a memmove for every audio packet received.
- All the buffers used (immediates, mediums, longs, scores, levels....) were not initialized to 0s and we had two steps (initialization + resize) instead of doing the resizing during the initialization.

### DIFF
--- a/worker/src/RTC/ActiveSpeakerObserver.cpp
+++ b/worker/src/RTC/ActiveSpeakerObserver.cpp
@@ -426,7 +426,7 @@ namespace RTC
 				b = level;
 			}
 
-			// The algorithm expect to have an update every 20 milliseconds.  If the producer is paused,
+			// The algorithm expect to have an update every 20 milliseconds. If the producer is paused,
 			// using a different packetization time or using DTX we need to update more than one sample
 			// when receiving an audio packet.
 			uint32_t intervalsUpdated =
@@ -436,6 +436,7 @@ namespace RTC
 				this->levels[this->nextLevelIndex] = b;
 				this->nextLevelIndex               = (this->nextLevelIndex + 1) % LevelsBuffLen;
 			}
+
 			UpdateMinLevel(b);
 		}
 	}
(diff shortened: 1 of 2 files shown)


worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -434,19 +426,25 @@ namespace RTC
 				b = level;
 			}
 
-			std::copy(this->levels.begin(), this->levels.end() - 1, this->levels.begin() + 1);
-
-			this->levels[0] = b;
+			// The algorithm expect to have an update every 20 milliseconds.  If the producer is paused,
```
ibc:
```suggestion
			// The algorithm expect to have an update every 20 milliseconds. If the producer is paused,
```

worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -434,19 +426,25 @@ namespace RTC
...
+			// using a different packetization time or using DTX we need to update more than one sample
+			// when receiving an audio packet.
+			uint32_t intervalsUpdated =
+			  std::min(std::max(static_cast<uint32_t>(elapsed / 20), 1U), LevelsBuffLen);
+			for (uint32_t i = 0; i < intervalsUpdated; i++)
+			{
+				this->levels[this->nextLevelIndex] = b;
+				this->nextLevelIndex               = (this->nextLevelIndex + 1) % LevelsBuffLen;
+			}
```
ibc, line 438:
```suggestion
			}

```

worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -458,7 +456,13 @@ namespace RTC
 
 		for (uint32_t i = 0; i < ImmediateBuffLen; ++i)
 		{
-			uint8_t level = this->levels[i];
+			// this->levels is a circular buffer where new samples are written in the
+			// next vector index.   this->immediates is a buffer where the most recent
```
ibc:
```suggestion
			// next vector index. this->immediates is a buffer where the most recent
```

ibc (changes requested):
Cosmetic changes.

ibc (approved):
LGTM. Will release version later today.

ibc:
Released in 3.8.4, THANKS

PR #665, opened 2021-10-06
Limit max temporal layers to preferred temporal layers

ggarber:
Right now there is no way to limit the maximum number of temporal layers forwarded for a simulcast/svc consumer.

I was assuming `consumer.setPreferredLayers({spatialLayers: 2, temporalLayers: 0})` was limiting the spatial layers to [0,1] and the temporal layers to [0] but actually the implementation is allowing all the combinations up to the second spatial layer and the first temporal layer (0:0, 0:1, 0:2, 1:0, 1:1, 1:2, 2:0).

If the feature of limiting temporal layers is useful (it is useful for our use case at least) there are two options:
1/ We change the meaning of preferred temporal layers so that in the example below mediasoup never forwards the second and third temporal layers.
2/ We add a different API for that (maybe called setMaxLayers).

This PR is trying option 1/ but the authors of mediasoup should decide on which option to pursue or if this is a feature that is not interesting enough.

I'm almost sure this PR doesn't contain all the needed changes but it fixes the issues I had and can be good to help with the discussion/decission.

### DIFF
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -487,6 +487,16 @@ namespace RTC
 					continue;
 				}
 
+				if (
+					this->preferredTemporalLayer != -1 &&
+					temporalLayer > this->preferredTemporalLayer
+				)
+				{
+					requiredBitrate = 1u; // Don't set 0 since it would be ignored.
+					// don't use any bandwidth for this one as it will be disabled later on
+					goto done;
+				}
+
 				requiredBitrate = producerRtpStream->GetLayerBitrate(nowMs, 0, temporalLayer);
 
 				// This is simulcast so we must substract the bitrate of the current temporal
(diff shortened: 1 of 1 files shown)


ibc:
I've talked to @jmillan about this. Indeed we agree the behavior should be the once in bullet 1 above, so in summary:

1. Changes to `consumer.setPreferredLayers({ spatialLayer: N, temporalLayer: M })` method:
   - Right now, `spatialLayer` is mandatory while `temporalLayer` is not. Both should be optional.
2. If `temporalLayer` is given, the `SimulcastConsumer` and `SvcConsumer` should never select a temporal layer higher than that, no matter in which spatial layer we are.
3. I'm pretty sure this change requires much more modifications in `SimulcastConsumer` (plus in `SvcConsumer` and `Consumer` TypeScript and C++ API related to `setPreferredLayer()` method). There are plenty of methods and code in `XxxxConsumer.cpp` that checks preferred layers and assumes it can jump from 2:0 to 1:2 in case preferred temporal layer is 0.

@ggarber do you agree then?

ibc:
Plus changes in Rust and [website](https://mediasoup.org/documentation/v3/mediasoup/api/#ConsumerLayers) would be needed :)

piranna:
Why this got closed? Has it been fixed elsewhere?

ibc:
No idea why this PR was closed and I forgot completely about it, but here we are: https://github.com/versatica/mediasoup/pull/1892

PR #666, opened 2021-10-08
Rust updates and optimizations around communication performance

nazar-pc:
Changelog update has all major changes mentioned.

The big topic here is transition from file descriptors usage to native FFI function calls between C++ and Rust for lower latency, lower memory usage and better cross-platform support (Windows support is coming!).

The way it is implemented is the following:
* each channel construction has 2 pairs of optional parameters:
  * `channelReadFn` and `channelReadCtx` represent a function call (this is C ABI, so we can't send closures, we need static function and its context)
    * it is called by C++ when worker wants to read a message
    * with every call C++ sends over `uv_async_t` handle, so Rust (or any other language, really) can ping worker's `libuv` to try to read next message
      * because of this C++ does one initial call proactively to hand over this handle to the other side
  * `ChannelWriteCtx` and `ChannelWriteFn` also represent a similar function
    * it is called by C++ when worker wants to write a message
    * because Rust side here is thread-safe, this is immediately delivered to Rust and does all of the business logic before returning, so no need for additional notifications
* the implementation allows C++ worker to share memory with Rust using raw pointer without even copying it
* Rust though has to allocate for outgoing messages because it doesn't know when worker will be ready to read

I do not like how many arguments there is in worker instantiation now, but at the same time I don't see a huge potential for improvements either as we are limited by shared C ABI here. Open to suggestions.


worker/src/Channel/ChannelSocket.cpp
```diff
@@ -11,18 +12,67 @@
 namespace Channel
 {
 	/* Static. */
+	inline static void onAsync(uv_handle_t* handle)
+	{
+		while (static_cast<ChannelSocket*>(handle->data)->CallbackRead())
+		{
+			// Read while there are new messages
```
ibc: Dot at the end of comments 😀

worker/src/PayloadChannel/PayloadChannelSocket.cpp
```diff
@@ -12,18 +13,67 @@
 namespace PayloadChannel
 {
 	/* Static. */
+	inline static void onAsync(uv_handle_t* handle)
+	{
+		while (static_cast<PayloadChannelSocket*>(handle->data)->CallbackRead())
+		{
+			// Read while there are new messages
```
ibc: Same

worker/src/Channel/ChannelSocket.cpp
```diff
@@ -44,8 +97,19 @@ namespace Channel
...
+		if (this->uvReadHandle)
+		{
+			uv_close(reinterpret_cast<uv_handle_t*>(this->uvReadHandle), static_cast<uv_close_cb>(onClose));
+		}
+
+		if (this->consumerSocket)
+		{
+			this->consumerSocket->Close();
+		}
+		if (this->producerSocket)
```
ibc, line 110: Line break please

ibc (commented):
This is awesome. Definitely I need some time before I can properly review this (Jose is gonna be out for a while for very good reasons). So just a question/concern for now:

Here comm between Rust and worker is sync meaning that if Rust layer (i.e. the Rust app) blocks or makes the CPU be super busy, that would also block and affect the worker loop. Hopefully I'm wrong given that uv_async is used in worker.

nazar-pc:
> Here comm between Rust and worker is sync meaning that if Rust layer (i.e. the Rust app) blocks or makes the CPU be super busy, that would also block and affect the worker loop. Hopefully I'm wrong given that uv_async is used in worker.

You are correct here. When worker sends a response or notification into Rust this is definitely something everyone should keep in mind, it will steal some time from worker, but presumably very little. On the flip side the whole things is significantly more efficient, so developer now can decide whether to do memory copy and send it to another thread or process in place.

Communication from Rust to worker is async, so there is no such problem there.

src/Channel.ts
```diff
@@ -342,7 +342,7 @@ export class Channel extends EnhancedEventEmitter
 			// the response, destroying the ordered delivery. So we must wait a bit
 			// here.
 			// See https://github.com/versatica/mediasoup/issues/510
-			setImmediate(() => this.emit(msg.targetId, msg.event, msg.data));
+			setImmediate(() => this.emit(String(msg.targetId), msg.event, msg.data));
```
ibc: Why this? it should be a string already. And BTW I'm afraid there are gonna be conflicts here now since these files are now in `node/src/` instead.

nazar-pc: The reason is that **technically** `targetId` can be both a string (UUID) and a number (process ID), while function signature is:
```typescript
NodeJS.EventEmitter.emit(event: string | symbol, ...args: any[]): boolean
```

So using `String()` is semantically correct to use to make sure we always end up with a string there. I know it worked before, but it felt wrong.

> there are gonna be conflicts here

It is fine, I will resolve conflicts after your changes are merged.

ibc: My changes are already merged :)

nazar-pc: #673 still isn't :thinking:

ibc: My fault, I was pretty sure I clicked Merge button but I didn't... done now

nazar-pc:
Merge conflicts resolved

jmillan (approved):
👏 Great

nazar-pc:
Merged latest updates from `v3` here.
@ibc do you have any thoughts on this yet?

ibc:
I'll review tomorrow

worker/include/Channel/ChannelNotifier.hpp
```diff
@@ -12,6 +12,7 @@ namespace Channel
 	{
 	public:
 		static void ClassInit(Channel::ChannelSocket* channel);
+		static void Emit(uint64_t targetId, const char* event);
```
ibc, line 15: In which case do we emit with a number instead of a string?

worker/include/common.hpp
```diff
@@ -23,4 +23,43 @@ typedef SSIZE_T ssize_t;
 #include <sys/socket.h> // struct sockaddr, struct sockaddr_storage, AF_INET, AF_INET6
 #endif
 
+using ChannelReadCtx    = void*;
+using ChannelReadFreeFn = void (*)(uint8_t*, uint32_t, size_t);
+// Returns `ChannelReadFree` on successful read that must be used to free `message`
```
ibc: Please dot after all comments here.

ibc (changes requested):

worker/src/Channel/ChannelSocket.cpp
```diff
@@ -11,18 +12,67 @@
 namespace Channel
 {
 	/* Static. */
```
ibc:
```suggestion
	/* Static methods for UV callbacks. */
```

worker/src/Worker.cpp
```diff
@@ -37,7 +37,7 @@ Worker::Worker(::Channel::ChannelSocket* channel, PayloadChannel::PayloadChannel
 	DepUsrSCTP::CreateChecker();
 
 	// Tell the Node process that we are running.
-	Channel::ChannelNotifier::Emit(std::to_string(Logger::pid), "running");
+	Channel::ChannelNotifier::Emit(Logger::pid, "running");
```
ibc, line 40: I see here.

ibc (commented):
Just a cosmetic change. Code looks great.

ibc (approved):
PR 666 approved (once CI doesn't complain)

ibc:
@nazar-pc feel free to merge :)

PR #667, opened 2021-10-09
Remove request timeouts from TypeScript

nazar-pc:
As we discussed, it is not super useful since requests are not cancelled and will be processed by worker anyway


ibc (approved):
Thanks. I'll merge once CI passes (or feel free to do it if you see it passing before me)

jmillan:
:+1:

PR #672, opened 2021-10-10
Remove dependencies that are no longer necessary and scripts for getting them

nazar-pc:
The diff explains well why I didn't want to combine this with other changes


ibc:
I cannot even open diff in mobile so cannot approve. Please merge (BTW ensure you removed gyp also from scripts/get-deps.sh which probably no longer exists anyway...)

ibc:
I see that  `worker/scripts/get-dep.sh` still exists in v3, shouldn't it be removed?

nazar-pc:
It is cleaned up `get-dep.sh` in this PR. It still has a few useful things and thus isn't removed:
* lcov
* clang-fuzzer
* fuzzer-corpora

ibc:
Merge merge 😀

PR #671, opened 2021-10-10
Move TS and JS to node/ folder and and use ES private fields

ibc:
- Add `node` folder and move TypeScript and compiled JavaScript code to it.
- Also move `tests` folder, `.eslintrc.js` and `tsconfig.json`.
- Some minor cosmetic changes.
- Add Nazar into `contributors` field in `package.json`.
- Node: use ES6 private fields.
- Require Node >= 12.

PR #673, opened 2021-10-10
Add node folder and use js private fields

ibc:
Closing https://github.com/versatica/mediasoup/pull/671 because there were conflicts that shouldn't be there...

- Add `node` folder and move TypeScript and compiled JavaScript code to it.
- Also move `tests` folder, `.eslintrc.js` and `tsconfig.json`.
- Some minor cosmetic changes.
- Add Nazar into `contributors` field in `package.json`.
- Node: use ES6 private fields.
- Require Node >= 12.


nazar-pc (approved):
Looks good to me

ibc:
I've pulled v3 in local and still get this error...

`../../meson.build:189:0: ERROR: Subproject exists but has no meson.build file``

:( 

do I miss something?

ibc:
I just have these deps in `worker/` (as expected):

```
ibc at ibc-macbook-pro in ~/src/v3-mediasoup/worker on v3
$ ll deps/
total 0
drwxr-xr-x   4 ibc  staff   128B Sep 27  2019 clang-fuzzer/
drwxr-xr-x  13 ibc  staff   416B Oct 10  2019 lcov/
drwxr-xr-x   8 ibc  staff   256B Oct 10 20:41 libwebrtc/
drwxr-xr-x   6 ibc  staff   192B Feb  6  2020 webrtc-fuzzer-corpora/
```

nazar-pc:
You really shouldn't, try removing `worker/out`, maybe it got inconsistent after last run. Or go to `worker` and run `make clean-all`

ibc:
on it

ibc:
yeah, it was that

j1elo:
May I ask here, what is the reasoning behind making all private fields *hard*-private instead of the Typescript style *soft*-private?

For most use cases (i.e. code correctness), compile-time checks for privacy were usually enough, while the new JS style runtime privacy is considered as beneficial for dealing with "bad actors" that might want to modify code on the fly (not a risk factor for server Node.js apps using mediasoup, afaik)

Also, JS hard privacy incurs in some runtime performance penalties (as now every access needs to be checked by the js engine).

ibc:
Because we don't want to receive complains later since people were, indeed, building mediasoup apps relying on private and undocumented properties, which proves that using TypeScript private fields was not enough to prevent users from accessing those fields 😀

j1elo:
Hahah well I get it, if you had a high amount of requests due to improper accesses to private fields... but at that point I guess it's the devs themselves that are deciding to get into brown waters. By that reasoning we should drop altogether every use of "private" in TS?

Again thanks for the insights

ibc:
> By that reasoning we should drop altogether every use of "private" in TS?

Yes. Problem is that JS doesn't implement "protected" fields so we cannot use `#xxx` JS syntax for some fields that need to be read in child classes. Other than those, everything that can use it already uses it.

piranna:
> Problem is that JS doesn't implement "protected" fields so we cannot use `#xxx` JS syntax for some fields that need to be read in child classes.

It's possible to use the convention of `_*` prefixed ones... I used also to defined them as non-enumerable non-writable properties for better protection. Other option is to make them public, the same as for private collections, things like `router._transports` showed to be useful to me, so a public `router.transports` could make sense to be exposed.

ibc:
Rationale about exposing things not really required here:

https://github.com/versatica/mediasoup/issues/700

ibc:
> Other option is to make them public, the same as for private collections, things like `router._transports` showed to be useful to me, so a public `router.transports` could make sense to be exposed.

Rationale against exposing those things was given here yesterday: https://github.com/versatica/mediasoup/issues/700

piranna:
> > Other option is to make them public, the same as for private collections, things like `router._transports` showed to be useful to me, so a public `router.transports` could make sense to be exposed.
> 
> Rationale against exposing those things was given here yesterday: #700

Yes, I didn't read it before writting the comment, thanks.

Maybe would make sense to write that down at an architecture or project design document that can be used as reference? That way people would not be asking once and again for the same things and have a global view of what aproach to take to the API.

ibc:
I don't want to write opinionated documentation about how users should use mediasoup based on its exposed API. In general I don't want (or I don't have time) to write high level documentation. There are many mediasoup usage examples out there and users are free to contribute with posts in the forum, blogs or whatever.

Really, I don't want more work in my back.

piranna:
> Really, I don't want more work in my back.

I'm with you on this :-/

PR #675, opened 2021-10-11
Memory optimizations

nazar-pc:
This heavy PR introduces a bunch of optimizations as mentioned in https://mediasoup.discourse.group/t/need-help-testing-significant-memory-optimizations/3302

It would be the best to review this commit by commit, otherwise all of the changes combined will be hard to follow.

### LT;DR:

All of the changes together should result in ~6x memory usage reduction and ~3x reduction in calls to system allocator in a room with 5 participants each of which is sending 720p@30 video to each other, savings will grow with number of consumers.

Briefly about changes:
* many data structures from `std` replaced with more efficient ones from `abseil-cpp` that we already depend on
* some data structures are replaced with alternative implementation (like one-byte header extensions that use static array with 14 entries instead of `std::map` resulting in huge savings on system allocator calls)
* some data structures that call system allocator frequently are now reusing pre-allocated memory from a pool
* RTP video packets are no longer cloned for every consumer and are stored for shorter period of time (for RTX purposes) derived from RTT time (following similar logic in Janus)

I'm not great at C++, so I'm sure APIs can be made better, but that is above my head at the moment.

Also this needs really careful review and preferably some extra testing even though I have tested it extensively in local environment.


worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -10,34 +10,150 @@ namespace RTC
...
 
+	thread_local static Utils::ObjectPool<RtpStreamSend::StorageItem> StorageItemPool;
+
 	// 17: 16 bit mask + the initial sequence number.
 	static constexpr size_t MaxRequestedPackets{ 17 };
 	thread_local static std::vector<RTC::RtpStreamSend::StorageItem*> RetransmissionContainer(
 	  MaxRequestedPackets + 1);
+	static constexpr uint32_t MinRetransmissionDelay{ 200 };
 	// Don't retransmit packets older than this (ms).
 	static constexpr uint32_t MaxRetransmissionDelay{ 2000 };
```
nazar-pc, line 19: Janus uses 1000ms max and http://www.rtcbits.com/2017/03/retransmissions-in-webrtc.html says that libwebrtc also stores just 1000ms of packets in outgoing buffer (even though I wasn't able to find corresponding place in libwebrtc codebase, possibly due to not being familiar with it, like at all).

ibc:
Please patience 😅
I'll review this week. And tonight I'll publish a new release with Meson

worker/src/RTC/Router.cpp
```diff
@@ -695,15 +695,23 @@ namespace RTC
...
 
-		for (auto* consumer : consumers)
+		if (!consumers.empty())
 		{
-			// Update MID RTP extension value.
-			const auto& mid = consumer->GetRtpParameters().mid;
+			// Cloned ref-counted packet that consumers will all store for as long as needed
+			// while also avoiding multiple allocations unless absolutely necessary.
+			// Clone only happens if needed though.
+			RtpPacket* clonedPacket{ nullptr };
```
notorca: It is not clear when this packet is removed. `clonedPacket` is not used this scope at all, why is it declared here?

worker/include/RTC/RtpPacket.hpp
```diff
@@ -143,9 +143,16 @@ namespace RTC
 		  uint8_t payloadPadding,
 		  size_t size);
 
-	public:
+		// Use `RtpPacket::DecRefCount()` instead
 		~RtpPacket();
 
+	public:
```
notorca, line 154: Seems shared_ptr with custom allocator for the pool can be a good alternative to to the manual refcounting here.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -9,6 +9,10 @@
 
 namespace RTC
 {
+	thread_local static Utils::ObjectPool<RtpPacket> RtpPacketPool;
+	// Memory to hold the cloned packet (with extra space for RTX encoding).
+	thread_local static Utils::ObjectPool<RtpPacket::RtpPacketBuffer> RtpPacketBufferPool;
+
 	/* Class methods. */
 
 	RtpPacket* RtpPacket::Parse(const uint8_t* data, size_t len)
```
notorca: `RtpPacket` object lifetime is no longer than `*data` lifetime, as soon as RtpPacket refers to the data memory. So it  is dangerous to make the `RtpPacket` ref-countable, because it's lifetime can be extended over `*data`

notorca: Another idea for memory allocation optimization:
as soon as RtpPacke is limited by the ifletime of `*data` it can be just a stack object passed by reference, in this case the lifetime is clear and no memory pools are needed.

worker/include/RTC/RtpPacket.hpp
nazar-pc: It might be. I looked into it, but I don't think my C++ knowledge is sufficient for a better design. I have spent quite a bit of time trying to use allocator API, but I wasn't able to get generic object pool implementation to compile and compiler wasn't as helpful as rustc usually is, so I gave up on that. Certainly an area for future ergonomic improvements.

worker/src/RTC/Router.cpp
nazar-pc: This is just a pointer, whoever creates it is responsible for re-allocating. Cloning happens towards the end of `RtpStreamSend::StorePacket` and de-allocation happens in the same file in `resetStorageItem()` function.

It is declared here such that we don't clone packets that don't need to be cloned (audio packets for instance) and when it is clones once, the rest of consumers will just increase reference count and will not have to clone it until and if they need to send RTX.

worker/src/RTC/RtpPacket.cpp
nazar-pc: > So it is dangerous to make the RtpPacket ref-countable, because it's lifetime can be extended over *data

You are right, I could check and refuse to increase ref count if the packet doesn't own its memory, this is probably a good idea actually. But as far as I'm concerned the whole C++ thing is dangerous (coming from Rust background where this can be captured capture on type system level).

> Another idea for memory allocation optimization

This will require more tricky internals of the RtpPacket, we also have packets that are and need to be heap-allocated. So it is not like we can get rid of memory pool in a generic case.

worker/src/RTC/Router.cpp
notorca: In this case it should be a boolean flag `needClone` or something like that. Usualy if function accepts pointer to a pointer that means it create something and returns the created instance in this parameter. And in this case the caller is responsible for deallocation.

nazar-pc: It is not just a flag, it holds cloned packet such that other consumer can increase ref counter and store it too. This is how significant memory saving is achieved.

worker/src/RTC/RtpPacket.cpp
notorca: C++ type system could also catch such cases, if you help it a bit with proper primitives. But manual refcounting looks a bit odd here, for example `RTC::Transport::ReceiveRtpPacket` gets packet and decreases refcount on it, but it is completely unclear from the method declaration

nazar-pc: Added assertion such that ref count is only increased for packets that own their buffer and also refactored `RTC::Transport::ReceiveRtpPacket` case that, I agree, wasn't good.

Thanks for review BTW!

notorca: So form the design perspective I would recommend to make RtpPacket a stack object and Clone method returning a shared_ptr< RtpPacket >. It can be done with a custom allocator but I really doubt it would give any noticeable performance improvements. Creation and deletion of same size objects is a trivial task for the system allocator.

nazar-pc: We are stretching the limits of the single thread performance here and we definitely know that we will need the same exact piece of memory like next millisecond. And note that in case of Rust app worker is a part of the larger program and that might mix things up in system allocator.

I didn't find massive improvements from this particular change, but 3x reduction in calls to system allocator must be beneficial, I just wish there was a way for me to write a nicer API for that. Like smart pointer that returns memory to the pool instead of system allocator. And I'm fairly certain it is possible to do so.

notorca: Keep in mind drawbacks of pool allocation:

- Memory usage of the thread would be always equal to the maximum memory usage ever happened. Can be bad in case of multithread approach to the sfu workers.
- This also breaks tools like valgrind. 

Consider finer granularity of pools, for example one pool per receiver.

nazar-pc: A pool is in thread-local memory, IIRC it worked fine with valgrind unless I'm missing something.

There is actually efficiency from sharing pool across multiple consumers/producers as degradation in one temporarily will not result in the persistent memory usage increase, variability of different objects will ammortize each other to a degree.

Memory savings with these changes more than compensate the pooling overhead in pathological cases, while I definitely agree that it would be even less if we didn't have pool at all and just re-allocate memory all the time. So this PR is net positive, even though some potential memory usage is still traded for a bit of performance.

notorca: Pool object is thread-local, true, but inside it stores a pointers to the regular heap memory. And errors like use-after-free can't be captured by valgrind, because this heap memory is not freed from valgrind's point of view (it is returned to the pool).

nazar-pc: That is a good point, yes, that is one of the drawbacks of this approach indeed and will be tricky to detect.

worker/include/RTC/RateCalculator.hpp
```diff
@@ -38,7 +37,8 @@ namespace RTC
 		void RemoveOldData(uint64_t nowMs);
 		void Reset()
 		{
-			this->buffer.reset(new BufferItem[this->windowItems]);
+			std::memset(
```
jmillan: Obvious, but, can we add a comment here please?

```c++
// Reset the Buffer.
```

worker/include/Utils.hpp
```diff
@@ -349,6 +349,46 @@ namespace Utils
...
+		}
+
+		// Return allocated memory into internal pool for future use, make sure to run destructor before
+		// returning memory, ObjectPool will only de-allocate memory on exit.
+		void Return(T* ptr)
+		{
+			if (ptr)
+				this->pool.push_back(ptr);
+		}
+		// TODO
```
jmillan: forgotten comment?

worker/src/RTC/RTCP/CompoundPacket.cpp
```diff
@@ -8,8 +8,27 @@ namespace RTC
...
 	namespace RTCP
 	{
+		thread_local static Utils::ObjectPool<CompoundPacket> CompoundPacketPool;
+
 		/* Instance methods. */
 
+		CompoundPacket* CompoundPacket::Create()
+		{
+			auto* packet = CompoundPacketPool.Allocate();
+			new (packet) CompoundPacket();
```
jmillan: I didn't know this syntax existed.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -10,34 +10,150 @@ namespace RTC
...
+		return this->buffer.at(idx);
+	}
+
+	bool RtpStreamSend::StorageItemBuffer::Insert(uint16_t seq, StorageItem* storageItem)
+	{
+		if (this->buffer.empty())
+		{
+			this->startSeq = seq;
+			this->buffer.push_back(storageItem);
+			return true;
```
jmillan, line 75: New line before return please.

jmillan (approved):

worker/include/Utils.hpp
```diff
@@ -349,6 +349,46 @@ namespace Utils
...
+			}
+		}
+
+		// Get pointer to allocated memory. This can be newly allocated memory or re-use of previously
+		// returned object. Object is not initialized and shouldn't be considered to be in a valid state.
+		T* Allocate()
+		{
+			if (this->pool.empty())
+			{
+				return static_cast<T*>(std::malloc(sizeof(T)));
```
ggarber, line 398: This won't call the constructor of the class if there is any, right?

worker/src/RTC/Transport.cpp
```diff
@@ -2285,53 +2306,56 @@ namespace RTC
...
 			auto* consumer = kv.second;
 
-			for (auto* rtpStream : consumer->GetRtpStreams())
+			for (const auto& rtpStream : consumer->GetRtpStreams())
 			{
-				// Reset the Compound packet.
-				packet.reset(new RTC::RTCP::CompoundPacket());
-
-				consumer->GetRtcp(packet.get(), rtpStream, nowMs);
+				RTC::RTCP::CompoundPacket::ReturnIntoPool(packet);
```
ggarber: Having to manually return packets to the pool looks error prone.   Some smart pointer alternative would be ideal.

Also this method is assuming that the packet passed to the method ReceiveRtcpPacket was allocated in a pool.  I think this class shouldn't need to know that if possible.

worker/include/RTC/RateCalculator.hpp
```diff
@@ -65,7 +65,7 @@ namespace RTC
 		// Item Size (in milliseconds), calculated as: windowSizeMs / windowItems.
 		size_t itemSizeMs{ 0u };
 		// Buffer to keep data.
-		std::unique_ptr<BufferItem[]> buffer;
+		std::vector<BufferItem> buffer;
```
ggarber: looks like an std::array could be even better

ggarber (commented):
Wow, amazing work.  The results look great!

If possible I would give it some more thinking (now or later) to the refcounting/pool implementation.  It looks like too manual and error prone.  Using an object pool with smart pointer and custom deleters could be cleaner.

Thank you!

(As a humble suggestion in the future in my opinion this could be split in different PRs (the replacement of absl structures, the object pool / counting and the changes in the nack buffers) and could make the commit history more informative and PRs easier to review.)

worker/include/RTC/RateCalculator.hpp
satoren: Can the size be determined at compile time?

worker/src/RTC/Transport.cpp
satoren: Like this into ObjectPool (Each smart pointer has costs this pointer). 
```c++
		static void Destruct(T* ptr)
		{
			if (ptr)
			{
				ptr->~T();
				Return(ptr);
			}
		}

		template<class... Args>
		std::shared_ptr<T> AllocateShared(Args&&... args)
		{
			auto deleter = [this](T* ptr) { Destruct(ptr); };
			auto memory  = Allocate();
			return std::shared_ptr<T>(new (memory) T(std::forward<Args>(args)...), deleter);
		}

		template<class... Args>
		std::unique_ptr<T> AllocateUnique(Args&&... args)
		{
			auto deleter = [this](T* ptr) { Destruct(ptr); };
			auto memory  = Allocate();
			return std::unique_ptr<T, decltype(&deleter)>(
			  new (memory) T(std::forward<Args>(args)...), deleter);
		}
```
Or create deleter onto classes has thread_local without this pointer.

worker/include/RTC/RateCalculator.hpp
jmillan: Never mind, we are already in the Reset method.

worker/include/RTC/RateCalculator.hpp
jmillan: Seems like we could benefit from an std::array here, and no, we cannot determine it's size at compile time since it can be given on construction.

worker/include/Utils.hpp
nazar-pc: It just allocates required memory, it is the responsibility of the caller to call constructors/destructors when using this object pool. I tried to make it such that it can be both generic and construct the object, kind of like std smart pointers do, but it turned out to be beyond my C++ level unfortunately.

worker/src/RTC/Transport.cpp
nazar-pc: It is error prone, yes. I make destructor for `CompoundPacket` and other things private to mitigate it somewhat. If someone can submit a PR against my branch with some smart pointer abstraction that does the same thing, but in a nicer way - I'm happy to review it.

nazar-pc (commented):
Merged a heap of updates from `v3` branch into this one

worker/src/RTC/Transport.cpp
satoren: I made it. https://github.com/nazar-pc/mediasoup/pull/4

worker/include/RTC/RateCalculator.hpp
Hartigan: @jmillan i tried looking for usages in the mediasoup codebase and found that the default parameter ```100u``` is being used for buffer size(except tests).

jmillan (approved):
Nice!

nazar-pc:
Fixed memory leak accidentally introduced in https://github.com/nazar-pc/mediasoup/pull/5 and merged latest v3 back into this branch

ibc:
Definitely will review this great work this week.

ibc:
We should merge this. Is everyone happy with this PR?

nazar-pc:
I'm still trying to figure out with Paulo Lanzarin an interesting memory leak that only happens for worker that lives for a very long time (days), please wait until we figure it out.

jmillan:
> I'm still trying to figure out with Paulo Lanzarin an interesting memory leak that only happens for worker that lives for a very long time (days), please wait until we figure it out.

Commenting this here because it may not be solely related to this PR, we are seeing some potential memory leak in v3 to.

Need to investigate further, but this are the valgrind logs.

[valgrind.txt](https://github.com/versatica/mediasoup/files/8043035/valgrind.txt)

nazar-pc:
> Commenting this here because it may not be solely related to this PR, we are seeing some potential memory leak in v3 to.

I don't think those are related as comparison was done against v3 branch (release), but still potentially possible.

Hartigan:
@nazar-pc 
I start mediasoup with ```valgrind --tool=massif <executable>``` and found leak of rtp packets.
[Massif data](https://gist.github.com/Hartigan/b77568249e1f5b0f95c0e5db7df5bc97)

You can visualize it with ```ms_print <file>``` or ```massif-visualizer <file>```

nazar-pc:
Thanks @Hartigan, I'll take a look closer to the end of this week

nazar-pc:
@Hartigan is there a way for me to reproduce it myself?

Hartigan:
@nazar-pc It's custom service written on rust. I created 4 echo participants((1 router + 1 webrtc transport + 2 producer + 2 consumer) * 4) and collect valgrind dump.

nazar-pc:
@Hartigan are those routers in the same worker? Are those running over the network or locally, do they do anything special, or you just start them and wait for memory usage to grow?

Hartigan:
@nazar-pc I run locally in docker. Number of router no matter. I knew that there was a leak and started several routers to speed up the leak.

nazar-pc:
How long does it take for things to leak? I was running for tenth of minutes and had constant memory usage. Also which browsers did you use? As soon as I can reproduce it, fix should be fairly quick.

Hartigan:
1. Chromium Version 98.0.4758.80 (Official Build) Arch Linux (64-bit)
2. Single transport on send and receive in echo room
3. We don't use mediasoup's client libs.
4. Media: vp8 with simulcast(3 layers) + opus
5. I ran the test overnight => estimated time < 8h.

Today I gonna try to reproduce on that sample: https://github.com/versatica/mediasoup/blob/v3/rust/examples/echo.rs

jmillan:
Probably related, even the same, I've found a leak in `RtpStreamSend` retransmission buffer. The problem is within `storage` and `buffer`. I'll be checking it tomorrow hopefully. 

Update: the leak is in v3.

Hartigan:
@jmillan yes, it's same problem

nazar-pc:
Retransmission buffer logic is rewritten in this PR, but I'm still curious what it is :slightly_smiling_face:

ibc:
> Retransmission buffer logic is rewritten in this PR, but I'm still curious what it is :slightly_smiling_face: 

Problem is that another leak is also detected in this PR, right? So we don't know yet if it's the same even if the logic changes.

Hartigan:
@nazar-pc I prepared draft version of ```StorageItemBuffer``` based on ```std::array``` with reusable buckets. Size of bucket based on sqrt-decomposition practice. My estimation: 256 packets ~ 1 sec 720p vp8 video.
https://gist.github.com/Hartigan/c29f20ab261fb7983103fa9d28dfafd6

nazar-pc:
@Hartigan why though? This PR already optimizes it and doesn't require fixed buffers at all (and saves a lot of RAM because of that).

Hartigan:
@nazar-pc 
1. std::deque implementation based on fixed-size buffers allocations :D
2. I wanna provide solution with reusable(pooled) buffers and simple navigation(without ```this->startSeq``` calculations)
3. In my case ```Insert```, ```Remove```, ```Get```  has guaranteed ```O(1)``` complexity. In current implementation ```Insert``` and ```Remove``` do many ```pop_back/pop_front``` and ```push_back/push_front``` staff which can cause allocations of fixed-size buffers in ```std::deque```.

jmillan:
Hihi,

Crossposting here just in case we are dealing with the same base issue. The leak I found is quite crazy: https://github.com/versatica/mediasoup/issues/769#issuecomment-1041474158

Hartigan:
@nazar-pc I create pull request https://github.com/nazar-pc/mediasoup/pull/6 with changes for retransmission buffer.
I collected new [massif dump](https://gist.github.com/Hartigan/020a8029e7aac9e4ae9f1c9f250ba208): 4 echo rooms with 720p VP8 simulcast + opus and 8 hours
Leak of cloned RTP packets gone and memory usage <8mb (total usage of application with mediasoup library).

nazar-pc:
`std::deque` was used specifically to avoid reallocations in most cases. push/pop then is basically numeric increment/decrement unless I don't understand it correctly. Deque will grow as needed, but only if needed, saving memory. Of course allocating fixed buffers will save allocations, but it needs to be measured if it actually makes a difference.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -10,34 +10,146 @@ namespace RTC
...
+		storageItem->clonedPacket.Reset();
+		storageItem->originalPacket.Reset();
 		storageItem->resentAtMs = 0;
 		storageItem->sentTimes  = 0;
 		storageItem->rtxEncoded = false;
 	}
 
+	RtpStreamSend::StorageItem* RtpStreamSend::StorageItemBuffer::Get(uint16_t seq)
+	{
+		auto idx{ static_cast<uint16_t>((seq - this->startSeq) % MaxSeq) };
```
Hartigan: @nazar-pc  May be it's should be ```auto idx{ static_cast<uint16_t>((seq - this->startSeq) % SeqRange) };```  Can you check this?

jmillan:
@nazar-pc,

About the `ObjectPool`, we are using it mainly as `thread static` which means that its lifetime is the same as the thread's lifetime. `ObjectPool` will deallocate memory only on destruction so a pool will have, for the whole lifetime of the thread, as much allocated memory as it was required in the moment of maximum demand.

This is part of the pattern, I just bring this up because honestly I have just realised.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -280,118 +402,103 @@ namespace RTC
...
+						// This is the storage item we have just inserted, no need to go further.
+						if (!checkedStorageItem->originalPacket)
+							break;
+
+						uint32_t checkedPacketTs{ checkedStorageItem->originalPacket->GetTimestamp() };
+						uint32_t diffMs{ (packetTs - checkedPacketTs) * 1000 / clockRate };
+
+						// Cleanup is finished if we found an item with recent enough packet, but also account
+						// for out-of-order packets.
+						if (diffMs < this->retransmissionBufferSize || packetTs < checkedPacketTs)
```
Hartigan: @nazar-pc wrong timestamp comparsion. It's should be ```RTC::SeqManager<uint32_t>::IsSeqHigherThan(checkedPacketTs, packetTs)```

nazar-pc:
@jmillan yes, those pools are for the whole worker, making it more efficient in terms of memory usage comparing to having them per-consumer for instance. I think it wouldn't be a bad idea to have a call and for me to explain what kind of changes I did and why so we are all on the same page since this is a really big PR.

Hartigan:
@nazar-pc  @jmillan  I think I figured out problems:
1. https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L459
In long test cause wrong comparsion of timestamps => buffer not cleaned when its needed.
2. https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L41
https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L59
https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L70
https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L74
https://github.com/nazar-pc/mediasoup/blob/memory-optimizations/worker/src/RTC/RtpStreamSend.cpp#L103

```MaxSeq``` used, but should be ```SeqRange```. That problem broke index in long tests
To reproduce problems need run long tests when ```sequence``` and ```timestamp``` pass over upper limit.

**Comment by jmillan** [comment]
> @nazar-pc @jmillan I think I figured out problems:

Makes sense @Hartigan 👍 👏

**Comment by nazar-pc** [comment]
Thanks for the hints, @Hartigan, there was incorrect handling for timestamps when they wrap around. Latest commit should resolve this. I was able to reproduce the issue with videoroom example app and after fixes memory doesn't seem to leak anymore as far as I can tell.

As a sidenote I was also surprised that timestamps on consumers start with 1 instead of random number, makes testing slower that would have been otherwise.

cc @prlanzarin

**Comment by nazar-pc** [comment]
Hm, eventually it started using 100% CPU of one core, though memory usage was only a little bit higher :thinking:

**Comment by ibc** [comment]
> Hm, eventually it started using 100% CPU of one core, though memory usage was only a little bit higher :thinking: 

That usually means an infinite loop.

**Comment by nazar-pc** [comment]
It is, the question is: where? :slightly_smiling_face: 

I'm gonna figure it out next.

**Comment by Hartigan** [comment]
@nazar-pc how long you waiting for that result?

**Comment by nazar-pc** [comment]
About ~3 hours, but I was AFK when it got stuck

**Comment by Hartigan** [comment]
@nazar-pc I think we can create simple solution based on ```MaxRetransmissionDelay```. We can create fixed buffer ```std::array<StorageItem*, 8192>```. 8192 packets it's enough for ```2000ms``` delay. This will allow us to translate ```seq``` to ```idx``` by ```seq % 8192```. It's fast, simple and without reallocations. Price: ```64kbytes``` per ```RtpStreamSend```.

File: worker/src/RTC/RtpStreamSend.cpp [hunk 14]
```diff
@@ -38,7 +38,7 @@ namespace RTC
 
 	RtpStreamSend::StorageItem* RtpStreamSend::StorageItemBuffer::Get(uint16_t seq)
 	{
-		auto idx{ static_cast<uint16_t>((seq - this->startSeq) % MaxSeq) };
+		uint16_t idx{ static_cast<uint16_t>(seq - this->startSeq) };
```
jmillan: why remove auto? using `static_cast` makes it already explicit.

worker/src/RTC/PipeTransport.cpp
```diff
@@ -593,7 +594,7 @@ namespace RTC
 
 		if (HasSrtp() && !this->srtpRecvSession->DecryptSrtp(const_cast<uint8_t*>(data), &intLen))
 		{
-			RTC::RtpPacket* packet = RTC::RtpPacket::Parse(data, static_cast<size_t>(intLen));
+			auto packet = RTC::RtpPacket::Parse(data, static_cast<size_t>(intLen));
```
jmillan, line 597: Let's please use `auto*` for pointer for readability and consistency.

worker/include/RTC/RtpPacket.hpp
```diff
@@ -131,8 +131,89 @@ namespace RTC
 			);
 			// clang-format on
 		}
+		class SharedPtr
```
jmillan: Why is there a custom shared pointer implementation instead of using `std::shared_pointer`?

nazar-pc: The standard one will call `delete` and de-allocate memory. This implementation will call destructor only and return memory to the pool to be consumed by the next packet. All packets are of the same size, so there is no need to re-allocate them all the time.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -280,118 +402,111 @@ namespace RTC
...
+					if (checkedStorageItem)
+					{
+						// This is the storage item we have just inserted, no need to go further.
+						if (!checkedStorageItem->originalPacket)
+							break;
 
-		if (this->storage.empty())
-			return;
+						uint32_t checkedPacketTs{ checkedStorageItem->originalPacket->GetTimestamp() };
+						uint32_t diffMs{ packetTs - checkedPacketTs };
```
jmillan:
```suggestion
						uint32_t diffTs{ packetTs - checkedPacketTs };
```

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -280,118 +402,111 @@ namespace RTC
...
+			storageItem->ssrc           = (*clonedPacket)->GetSsrc();
+			storageItem->sequenceNumber = (*clonedPacket)->GetSequenceNumber();
+		}
+		else
+		{
+			*clonedPacket               = packet->Clone();
+			storageItem->originalPacket = *clonedPacket;
+			storageItem->ssrc           = (*clonedPacket)->GetSsrc();
+			storageItem->sequenceNumber = (*clonedPacket)->GetSequenceNumber();
+		}
```
jmillan:
```suggestion
		if (!*clonedPacket)
		{
		        *clonedPacket  = packet->Clone();
		}
		
		storageItem->originalPacket = *clonedPacket;
		storageItem->ssrc           = (*clonedPacket)->GetSsrc();
		storageItem->sequenceNumber = (*clonedPacket)->GetSequenceNumber();
```

worker/src/RTC/PipeTransport.cpp
mstyura: From the point of view of casual reader both auto/auto* is not really clear here. What we have - function parse which returns something: guesses taking into account class name includes RTPPacket, RTPPacket*, RTPPacket::SharedPtr or RTPPacket::SharedPtr* or std::unique/shared_prr. So, maybe it would be better in cases where it's not 100% obvious what type is  to write explicit type?

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -10,34 +10,146 @@ namespace RTC
...
 	static void resetStorageItem(RTC::RtpStreamSend::StorageItem* storageItem)
 	{
 		MS_TRACE();
 
 		MS_ASSERT(storageItem, "storageItem cannot be nullptr");
 
-		delete storageItem->packet;
-
-		storageItem->packet     = nullptr;
+		storageItem->clonedPacket.Reset();
```
jmillan: I know this has not been changed in this PR, but this method `resetStorageItem` feels like it should be a member method `StorageItem::Reset()` rather than a static method.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -435,15 +550,24 @@ namespace RTC
...
 				uint32_t diffMs;
 
-				// Calculate the elapsed time between the max timestampt seen and the
-				// requested packet's timestampt (in ms).
+				// Calculate the elapsed time between the max timestamp seen and the
+				// requested packet's timestamp (in ms).
 				if (storageItem)
 				{
-					packet = storageItem->packet;
+					packet = storageItem->originalPacket->Clone();
```
jmillan, line 578: If `storageItem` has a `clonnedPacket` we should not create it again but reuse it, since we are in the scenario where the packet was already resent and we did it (clone the packet) before.

worker/include/RTC/RtpPacket.hpp
Hartigan: @nazar-pc you can pass custom deleter into shared_ptr constructor

Hartigan: `
std::shared_ptr<int> pIntPtr(new int(10), 
    [](int *pi) { delete pi; })
    `

worker/include/RTC/RtpStreamSend.hpp
```diff
@@ -20,10 +20,14 @@ namespace RTC
...
+			RTC::RtpPacket::SharedPtr originalPacket{ nullptr };
+			// Correct SSRC since original packet will have original ssrc.
+			uint32_t ssrc{ 0 };
+			// Correct sequence number since original packet will have original sequence number.
+			uint16_t sequenceNumber{ 0 };
 			// Cloned packet.
-			RTC::RtpPacket* packet{ nullptr };
-			// Memory to hold the cloned packet (with extra space for RTX encoding).
-			uint8_t store[RTC::MtuSize + 100];
+			RTC::RtpPacket::SharedPtr clonedPacket{ nullptr };
```
jmillan, line 37: This represents the RtpPacket that is re-sent upon receiving the corresponding NACK.

**Each** RtpStreamSend re-sending this packet creates its own clone of the original RTP packet, modifies the SSRC and Cseq, or creates the corresponding RTX packet if needed.

Instead of doing that yet more memory would be saved if the RTX packet of the original was shared among all the `RtpStreamSend`  instances as it is done with `originalPacket`. Then, before re-sending the packet in `ReceiveNack`, the `originalPacket`s or `clonnedPacket`s (depending on whether RTX is being used) SSRC and Cseq needs to be rewriteen to fit the one of this RtpStreamSend, and later restored.

Following the above rationale `clonedPacket` should be renamed to `rtxPacket`. 
Once at this point, `originalPacket` (which is also a cloned packet, BTW) should be IMO better named to just `packet`.

worker/src/RTC/RtpStreamSend.cpp
jmillan: Anyway, my previous comment [here](https://github.com/versatica/mediasoup/pull/675/files#r810548053) obsoletes this comment.

worker/include/RTC/RtpPacket.hpp
jmillan: Correct,

In `RtpPacket::Clone()`:
```c++
     auto* packet = RtpPacketPool.Allocate();
     new (packet) RtpPacket(header, headerExtension, payload, payloadLength, payloadPadding, len);

     std::shared_ptr<RtpPacket> shared(packet, /*Deleter*/ [](RtpPacket* packet) {
         // Call destructor manually since memory was pre-allocated upfront.
	 packet->~RtpPacket();
	 // Return packet into object pool for future reuse of memory allocation.
	 RtpPacketPool.Return(packet);
     });

     return shared;
```

nazar-pc: I was honestly not aware that existed in C++, but makes sense to me, will apply in upcoming changes

worker/include/RTC/RtpStreamSend.hpp
nazar-pc: Good catch, I'll probably either add a todo or implement this change later :+1:

worker/src/RTC/PipeTransport.cpp
nazar-pc: I tried to use `auto*` where it is a pointer, here it is a different type now, `RTC::RtpPacket::SharedPtr`, not a pointer, hence just `auto`.

worker/src/RTC/RtpStreamSend.cpp
nazar-pc: Indeed, added to my list of TODOs

worker/src/RTC/RtpStreamSend.cpp
nazar-pc: No good reason honestly, it was needed for another revision that wasn't commited, will revert back.

worker/include/RTC/RtpPacket.hpp
jmillan: @nazar-pc, I will create a PR for this one.

nazar-pc: There are several places where pools are used, I think we need to apply this patter in all cases.

nazar-pc:
Man, I love C++. This just loops forever:
```cpp
uint16_t MaxSeq = std::numeric_limits<uint16_t>::max();
for (uint16_t i{ 0 }; i <= MaxSeq; ++i) {
    // Do something
}
```

I mean, thinking some more about it it does make sense, but come on! This should be a compiler error.

Now I know where it loops forever, next step is to understand why does it even go as far as to reach the upper bound.

ibc:
> Now I know where it loops forever

You can do

do
{
   Xxxxx
}
while (i < MAX)

So it will exit the loop once i==MAX is processed.

Hartigan:
@nazar-pc Can you describe how that buffer should works?

nazar-pc:
@Hartigan we could allocate space for 65536 pointers and store each `StorageItem` according to its sequence number, this is what old implementation does.

Alternatively we can take into consideration that we'll not refer to packets that are too far in the past (upper bound is set by `MaxRetransmissionDelay` in ms, but with this PR it is a maximum and much lower in practice under good networking condition) and not store items that we don't need.

It keeps a `std::deque` that is just sufficient to store just the required amount of packets depending on current networking conditions. Once packets age out of the current dynamic retransmission delay they are removed and previously allocated space in `std::deque` can be used for other packets (I was expecting `std::deque` to be a contiguous piece of memory, thus avoiding re-allocations in most cases, will replace with a contiguous data structure later because right now this is not actually the case). Packets can arrive out of order, so the range kept allocated with `std::deque` is sufficient to store packets between min and max relevant `seq` even if some in between are missing. Since memory is supposed to be sequential, contiguous and without reallocation this flat data structure should be the most efficient way to organize the buffer.

Hartigan:
@nazar-pc I asked about implementation. For example:

1. When ```this->startSeq``` should be updated;
2. How paddings  in ```Insert``` and ```Remove```  works;

nazar-pc:
1. `std::deque` contains just the window of actually useful packets out of possible 65536. If the oldest packet potentially useful for RTX has seq 1000, then `this->startSeq == 1000`. If at the same time the newest useful packet is 1009, then `this->buffer.size() == 10`. If after packet with seq 65535 we insert packet with seq 0, start will not change yet, the math is done such that it should roll over the edge.
It is updated every time the oldest packet is removed (in `Remove`) or when we insert out of order packet before the oldest packet stored so far (in `Insert`).
2. As mentioned before, packets can arrive out of order, so `Insert` will reserve slots for corresponding `seq` even if they end up storing `nullptr`, so when we have seq 10 and 12 stored, buffer will have size of 3 with the middle slot holding `nullptr`. In `Remove` we both remove the slot we want and any `nullptr` slots around it, so removing either 10 or 12 will shink the buffer to size 1, removing pre-allocated space for potential packet with seq 11.

Hartigan:
@nazar-pc looks like ```this->bufferStartSeq``` has same meaning with ```this->startSeq```, but ```this->bufferStartSeq``` updates not like ```this>startSeq```. It's confusing. Why not use ```this->startSeq``` everywhere?

nazar-pc:
With latest changes I haven't seen an issue (either memory leak or infinite loop) for over 24 hours cumulatively across 2 runs, but I'd like to see more testing before we ship this. Still running a videoroom with 6 participants though in case I catch something eventually.

Several TODOs I have collected can be implemented in later PRs independently for further performance improvements and even more memory usage savings.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -689,12 +713,22 @@ namespace RTC
...
 		// Create the new RtpPacket instance and return it.
-		auto* packet = new RtpPacket(
+		auto* packet = RtpPacketPool.Allocate();
+		new (packet) RtpPacket(
 		  newHeader, newHeaderExtension, newPayload, this->payloadLength, this->payloadPadding, this->size);
 
+		SharedPtr shared(
+		  packet,
+		  /*Deleter*/
+		  [](RtpPacket* packet)
```
mstyura: @nazar-pc could it be possibly to put this lambda as deleter struct defined once?
```
struct RtpPacketPoolDeleter
{
    void operator()(RtpPacket* packet) const { 
        // Call destructor manually since memory was pre-allocated upfront.
        packet->~RtpPacket();
        // Return packet into object pool for future reuse of memory allocation.
        RtpPacketPool.Return(packet);
    }`
};
```
And then use as `std::shared_ptr<RtpPacket> packet(new RtpPacket(), RtpPacketPoolDeleter());`

nazar-pc: I think so, but I'll leave that for the future. There are still other places that could use shared_ptr with custom deleter that are not refactored yet. Feel free to create a PR if you feel like it.

worker/include/RTC/RtpPacket.hpp
```diff
@@ -24,6 +22,9 @@ namespace RTC
 	class RtpPacket
 	{
 	public:
+		using RtpPacketBuffer = std::array<uint8_t, MtuSize + 100>;
+		using SharedPtr       = std::shared_ptr<RtpPacket>;
```
satoren, line 26: `std::shared_ptr` does an extra memory allocation for the reference counter area. Have you considered this?
I implemented something like shared_ptr myself because I seemed to care about the one time extra global allocation cost as well.

http://www.cplusplus.com/reference/memory/make_shared/
http://www.cplusplus.com/reference/memory/allocate_shared/

jmillan: We are not allocating memory for the shared pointer itself, just for what it contains, so we are safe.

nazar-pc: Reference count variable still needs to be allocated somewhere (older version kept it in the original data structure since code is single-threaded and packet is already heap-allocated), so I think @satoren is right, but since it happens once for all shared instances it is probably worth the tiny performance cost to save on maintenance later.

jmillan:
I'm currently debugging a leak where more than 6seconds of video is being stored in memory.

Hartigan:
@jmillan  are you have massif dump or smth else?

jmillan:
> @jmillan are you have massif dump or smth else?

Nop, It's a problem where we are not deleting old packets in certain circumstances. I'm creating the corresponding test for this.

worker/include/RTC/RtpPacket.hpp
Hartigan: I created [pull request](https://github.com/nazar-pc/mediasoup/pull/8) which removes allocation for ```std::shared_ptr```

jmillan:
I'm still investigating this issue.

STR:
* Add an assert at the end of `RtpStreamSend::StorageItemBuffer::Insert` in order to verify that `this->buffer.size() < 500`  (500 packets is around 5 seconds of video), we should remove RTP older that 2 seconds at the moment.
* join a meeting from mediasoup-demo (video simulcast) (4 tabs in the same meeting)
* wait a bit, reload the tabs from time to time until it crashes.

Printing the content of `this->buffer` shows that may entries with the corresponding data.

jmillan:
Upon asserting the buffer size is < 500, I'm printing the content of each item in the buffer (some extra info added for debugging purposes):

```c++
for (const auto item : this->buffer) {
	if (item == nullptr) {
		MS_ERROR_STD("nullptr item");
		continue;
	}

         MS_ERROR_STD("ssrc: %" PRIu32 ", seq: %" PRIu16 ", timestamp: %" PRIu32 ", originalPacket: %p, originalPacketState: %s",
	    item->ssrc,
	    item->sequenceNumber,
	    item->timestamp,
	    item->originalPacket.get(),
	    itemm->originalPacketState.c_str()
);
```

The packet firing the assertion has an SSRC that differs from all others in the buffer. But it is that last the SSRC of the `RtpStreamSend` in question. I'm asserting that `RtpStreamSend::ReceivePacket` never receives a packet with a SSRC different than the one  in params, and such assertion has never fired.

**EDIT:** Removed the logs previously provided since they were not correct.

Hartigan:
@jmillan @nazar-pc I created new [massif dump](https://gist.github.com/Hartigan/9c0e4837982faeb0041c225be061685e). I think it's same packet leak

Hartigan:
@jmillan your debug-string contains ```originalPacket``` of last ```storageItem```. Also last ```storageItem``` is empty and created by ```new (storageItem) StorageItem{}```.

jmillan:
> @jmillan your debug-string contains originalPacket of last storageItem. Also last storageItem is empty and created by new (storageItem) StorageItem{}.

Yes, it was en error.

Hartigan:
@nazar-pc @jmillan  I think I found out problem. Problem linked with simulcast. For example:
1.  We have 3 rtp streams in simulcast
2. When simulcast layer changed we change stream and new ```seq``` can be ```older``` then current in buffer but with ```newer``` timestamp. Therefore we should track oldest item in buffer by ```timestamp```.

jmillan:
> @nazar-pc @jmillan I think I found out problem. Problem linked with simulcast. For example:

Sounds reasonable. I'm checking that.

Hartigan:
Therefore, deque can allocate many empty slots for paddings during insert

jmillan:
I can confirm that in my test cases all cseq and timestamps are incremental (some timestamps are the same within few packets, of course). I'm recording them as the come to `RtpStreamSend::ReceivePacket`

jmillan:
> I can confirm that in my test cases all cseq and timestamps are incremental (some timestamps are the same within few packets, of course). I'm recording them as the come to RtpStreamSend::ReceivePacket

Even though once putting that incoming data in a unit tests, the issue does not happen.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -10,34 +10,161 @@ namespace RTC
...
+		}
+		else
+		{
+			// Packets can arrive out of order, add blank slots.
+			for (uint16_t i{ 1 }; i < addToFront; ++i)
+				this->buffer.push_front(nullptr);
+
+			this->buffer.push_front(storageItem);
+			this->startSeq = seq;
+		}
```
jmillan: This is definitely not correct. We can't associate the `startSeq` with the new packet being inserted just without invalidating the rest.

```suggestion

```

jmillan: Let's just `push_back()` always.

jmillan: https://github.com/nazar-pc/mediasoup/pull/9

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -280,118 +417,93 @@ namespace RTC
...
+			clonedPacket.swap(clone);
 		}
 
-		// Reset buffer.
-		this->bufferStartIdx = 0;
-		this->bufferSize     = 0;
+		// Store original packet and some extra info into the retrieved storage item.
+		storageItem->originalPacket = clonedPacket;
+		storageItem->ssrc           = clonedPacket->GetSsrc();
+		storageItem->sequenceNumber = clonedPacket->GetSequenceNumber();
```
jmillan: This is **the** or at least **a** culprit. We are setting here the SSRC and Sequence Number of a different `RtpStreamSend`.., the one that cloned the packet. We need to set the item ssrc to `packet->GetSsrc()`, same with sequence number.

I'll add the changes along with some other sanitizations in a PR.

cc: @nazar-pc @Hartigan

Hartigan: @jmillan ```timestamp``` is needed too?

jmillan: yes

nazar-pc: Timestamp wasn't updated before and doesn't need to be updated now, it is just copied from source packets and there is no problem in that.

SSRC and sequence number are also fine, they are changed to correct values before packet gets here in corresponding consumers, for instance: https://github.com/versatica/mediasoup/blob/af3b681f1355835931c9d4a3d960cbe0a42e739f/worker/src/RTC/PipeConsumer.cpp#L238-L240

So I don't believe there is any error here, cloned packet is updated before it is processed, since it can be later modified again we store SSRC and sequence number in storage item (they weren't stored before because each storage item owned exclusive copy of the packet).

prlanzarin:
> Thanks for the hints, @Hartigan, there was incorrect handling for timestamps when they wrap around. Latest commit should resolve this. I was able to reproduce the issue with videoroom example app and after fixes memory doesn't seem to leak anymore as far as I can tell.
> 
> As a sidenote I was also surprised that timestamps on consumers start with 1 instead of random number, makes testing slower that would have been otherwise.
> 
> cc @prlanzarin

@nazar-pc 

Sorry for the late reply.

I'll re-run the load tests (same as the ones I mentioned in the discourse thread) against the tip of memory-optimizations branch and compare it against the latest v3 commit.
I'll also put memory-optimizations running again in a test server with real users compared against v3.9.8. I can get the results to both published either here or in the discourse thread (I'm thinking the latter).

worker/src/RTC/RtpStreamSend.cpp
jmillan: The problem here was that `storageItem->ssrc` was being set with the `clonedPacket's` one instead of with the original packet.

That is handled in the sanitisation branch.

worker/include/RTC/RtpStreamSend.hpp
jmillan: Resolving the comment. This will be done later.

jmillan:
@nazar-pc ,

Could you please split this into two PRs:

* One modifying `RtpStreamSend`, which modifies the `std::map` holding previous 65565 packets for the circular buffer implemented with `std::vector`.
* One adding the `ObjectPool` in different places.

Both are totally independent implementations that deserve their own PR each. We will be able to give them a last review one by one.

I planned to do this last split myself but I'm not finding the time.

Thanks 🙏

nazar-pc:
Hm... I'm overloaded myself at the moment and I see there are some minor merge conflicts already. It might take some time to extract those changes from monolithic PR, I'll let you know if I get to it sooner than you.

ibc:
@jmillan @nazar-pc does it make sense to have this PR open? AFAIU this effort was splitted into two tasks, and the only remaining one is the "ObjectAllocator" one which has its own ongoing PR already.

nazar-pc:
Agree, closing

jmillan:
:+1:

PR #682, opened 2021-10-16
[fix] Emit `close` event in `Worker` objects after `died` event

piranna:
All tests passing, and documentation updated.


ibc:
Please call this.close() instead of just emitting observer "close". Otherwise I'm not sure existing Routers are closed in JS side. I mean, when worker dies, do its routers emit "workerclose"?

piranna:
> I mean, when worker dies, do its routers emit "workerclose"?

Ah, didn't understand until I got to this part. Yeah, that should be emitted too. What's better, call `this.close()`, or have a private method that does the emits, and it's called from `close()` and when died?

ibc:
this.close() should be fine. It closes all resources and also kills the worker (which is already closed but anyway).

piranna:
> this.close() should be fine. It closes all resources and also kills the worker (which is already closed but anyway).

Done. I've seen that `this.close()` was already being called, and that it was emitting the `close` event... just only, it was at begin of the functions instead of after emitting the `died` event, that was the reason why I didn't see it, and also it was emitting the `died` event after the `close` one. I've moved the `this.close()` calls to the end of the functions so now there's no need to explicitly emit `close` event, and also now all events are emitted in the correct order: Worker `died` -> Routers `workerclosed` -> Worker `close`. Tests are fully passing.

ibc:
I think I don't like this change. Right now, in this PR, if you listen for worker "died" event and check worker.closed flag within the listener, it will be still false instead of true (this is because this.close() is internally called **after** worker emits "died".

Previous behavior is better and I think it's perfectly fine to close() internally before emitting "died". Let's not change this, please.

piranna:
My reasoning to have `close` after `died` is to have the "reason" event before the actual close, like in other objects: `workerclose`, `transportclose`, `producerclose`... all are happening before `close`, so make sense to me that `died` is emitted before `close` too.

I think checking `worker.closed` flag inside the `died` event handler is a bit of nosense, because if the Worker is died, obviously is closed, but we can set the `closed` flag and do all the clean-up before emitting the `died`, `workerclose` and `close` event, so the Worker state is uniform when executing all of them. It would be the same as before, just only moving the `close` event to be emitted after `died`. How do you see it?

ibc:
We must be consistent. People may call some external function to manage workers when one of them dies, and such a function may check closed flag in all them, so it must be true already.

I'll do an alternative PR and announce it here to discuss.

piranna:
Make sense to me, let's do that :-)

ibc:
Done: https://github.com/versatica/mediasoup/pull/684

PR #687, opened 2021-10-21
Fix multiopus frontend example compilation, update dependencies

nazar-pc:
There was a strange update in types that somehow was causing compilation errors in multiopus like this:
```
./src/index.ts 157:34-50
[tsl] ERROR in /web/github/mediasoup/rust/examples-frontend/multiopus/src/index.ts(157,35)
      TS2554: Expected 4 arguments, but got 2.
```

Not sure what caused legacy `setLocalDescription` function signature to be used, but we don't really need much types there anyway.

### DIFF
--- a/rust/examples-frontend/multiopus/src/index.ts
+++ b/rust/examples-frontend/multiopus/src/index.ts
@@ -142,7 +142,7 @@ async function init()
 					//  this case properly
 					{
 						// @ts-ignore
-						const pc = consumerTransport._handler._pc as RTCPeerConnection;
+						const pc = consumerTransport._handler._pc;
 						const setLocalDescription = pc.setLocalDescription;
 
 						pc.setLocalDescription = (answer: {type: 'answer'; sdp: string}) =>
--- a/rust/examples-frontend/echo/package.json
+++ b/rust/examples-frontend/echo/package.json
@@ -9,13 +9,13 @@
   },
   "author": "Nazar Mokrynskyi <nazar@mokrynskyi.com>",
   "dependencies": {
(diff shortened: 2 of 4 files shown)


kioviensis (approved):

rust/examples-frontend/multiopus/src/index.ts
```diff
@@ -142,7 +142,7 @@ async function init()
 					//  this case properly
 					{
 						// @ts-ignore
-						const pc = consumerTransport._handler._pc as RTCPeerConnection;
+						const pc = consumerTransport._handler._pc;
```
ibc, line 145: Probably some TS version issue.

ibc (approved):

ibc:
Can you paste the offending setLocalDescription code?

rust/examples-frontend/multiopus/src/index.ts
nazar-pc: Yeah, but for some reason `lib.dom.d.ts` in `node_modules` contains correct types. I have no idea what went wrong, but something broke with some updates in some of dependencies :man_shrugging: 

Here is the snippet:
```typescript
setLocalDescription(description: RTCSessionDescriptionInit): Promise<void>;
```

ibc: Fair. Let's ignore dumb TS things 😅

fippo: maybe https://github.com/microsoft/TypeScript/issues/46379 even though so far that only seems to have broken some getUserMedia constraints

PR #688, opened 2021-10-22
Transport: Hide debug message for RTX RTCP-RR packets

j1elo:
Everything was already in place, only a final check was needed to avoid printing a debug message when RTCP-RR packets are received from the RTX stream, which are not used for anything in mediasoup but should not cause a constant stream of debug lines either.

Related: https://mediasoup.discourse.group/t/no-consumer-found-for-received-receiver-report/3467

### DIFF
--- a/worker/src/RTC/Transport.cpp
+++ b/worker/src/RTC/Transport.cpp
@@ -1931,6 +1931,12 @@ namespace RTC
 							continue;
 						}
 
+						// Special case for (unused) RTCP-RR from the RTX stream.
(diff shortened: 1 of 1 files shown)


j1elo:
Would be cool if this could be backported to mediasoup 3.8 too... 3.9 is too recent and we're still having some work to do with our Docker images to make it work with the new build system

ibc (approved):

ibc:
We cannot backport to 3.8.X, we don't maintain other 3.X.X versions, that would be too much work. I suggest not enabling DEBUG level in workers in production.

PR #691, opened 2021-10-24
Make eslint work universally both for JS and TS code

nazar-pc:
I got tired with IDE not working with eslint config that requires environment variables, so I made it "just work" universally without any.

### DIFF
--- a/node/.eslintrc.js
+++ b/node/.eslintrc.js
@@ -175,71 +175,59 @@ const eslintConfig =
 		'strict'          : 2,
 		'valid-typeof'    : 2,
 		'yoda'            : 2
-	}
+	},
+	overrides : []
 };
 
-switch (process.env.MEDIASOUP_NODE_LANGUAGE)
-{
-	case 'typescript':
-	{
-		eslintConfig.parser = '@typescript-eslint/parser';
-		eslintConfig.plugins =
-		[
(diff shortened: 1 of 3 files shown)


node/.eslintrc.js
```diff
@@ -175,71 +175,59 @@ const eslintConfig =
...
+});
 
-	default:
-	{
-		throw new TypeError('wrong/missing MEDIASOUP_NODE_LANGUAGE env');
-	}
-}
+eslintConfig.overrides.push({
+	files : [ '*.js' ],
+	env   : {
```
ibc: The initial bracket should be in next line according to our ESlint rules so something is not working here if this "error" is not detected and CI passes, do I miss something?

nazar-pc: I do not think it was enforced by eslint before, you might want to add/change rule for that

ibc: Just please edit any .ts file and add a { in the same line as the = and you'll see 😉

nazar-pc: I tried, it didn't complain either (tried before and after this PR).

ibc: Perhaps the rule just applies to TS code

ibc (approved):
We should move Node tests to TS so this PR goes in the good direction. Thanks.

PR #692, opened 2021-10-24
Improve Windows support with some fixes

nazar-pc:
This are library-side fixes that are required (I made sure to run it locally and confirmed that it works on Windows 10 with Visual Studio 2019 Community Edition).


ibc (approved):
Lot of line removals, approved 😊

PR #695, opened 2021-10-27
Avoid build commands when MEDIASOUP_WORKER_BIN is set

### LINKED ISSUE #694: MEDIASOUP_WORKER_BIN still requires Make and Python in v3.9
### Your environment

- Operating system: Debian 10 Buster (Docker)
- Node version: 16
- npm version: 8
- gcc/clang version: N/A
- mediasoup version: 3.9.0

### Issue description

Commit https://github.com/versatica/mediasoup/commit/b447f7bc568168cdc4a96ceee7d607be624daf3f#diff-0e4b0cf2dfea5eb3fd7fcefa330bbcbe8ec35f53426f00d14f34a94281ed7a2aL123-R135 broke the ability of using `MEDIASOUP_WORKER_BIN` to preclude having to install build tools.

**Before the change**, setting `MEDIASOUP_WORKER_BIN=/path/to/mediasoup-worker` would result in the ability to simply run `npm install` in a clean minimal Debian system (tested with Docker, but probably a "real" setup would be similar), where none of the tools used by the build system are installed. This made sense because the usage of `MEDIASOUP_WORKER_BIN` meant that no build-system related operations needed to be done, only exclusively NPM related actions (install JS dependencies).

**After the change**, it is impossible to run `MEDIASOUP_WORKER_BIN=/path/to/mediasoup-worker npm install` without first having to install *Make* or *Python* (including pip!), which won't then really be used, given that the mediasoup binary is already provided somewhere else.

**Proposed solution**: Evaluate the possibility to restore all the `'postinstall'` actions to be inside an `if (!process.env.MEDIASOUP_WORKER_BIN)`, as it was before the mentioned change.

j1elo:
Small fix that prevents needing to even install build tools when `MEDIASOUP_WORKER_BIN` env var is set.

Fixes #694

### DIFF
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -59,10 +59,7 @@ switch (task)
 
 	case 'worker:build':
 	{
-		if (!process.env.MEDIASOUP_WORKER_BIN)
-		{
-			execute(`${MAKE} -C worker`);
-		}
+		execute(`${MAKE} -C worker`);
 
 		break;
 	}
@@ -121,9 +118,12 @@ switch (task)
 
 	case 'postinstall':
 	{
-		execute('node npm-scripts.js worker:build');
-		execute(`${MAKE} clean-pip -C worker`);
-		execute(`${MAKE} clean-subprojects -C worker`);
+		if (!process.env.MEDIASOUP_WORKER_BIN)
+		{
+			execute('node npm-scripts.js worker:build');
+			execute(`${MAKE} clean-pip -C worker`);
+			execute(`${MAKE} clean-subprojects -C worker`);
+		}
 
 		break;
 	}


ibc (approved):
Looks good to me. Tested?

j1elo:
Tests were done manually for this feature. Here are some commands that show the complete test:

First, a *Stage 1* (Docker image: node:16 which is Debian with Node and extra tools preinstalled) is prepared with all build deps and used to build the `mediasoup-worker` file, then that's used on a minimal system in *Stage 2* (node:16-slim, which comes with minimal requirements just to run Node stuff):

```sh
docker run -ti --rm --name stage1 \
    --entrypoint /bin/bash \
    -v /path/to/mediasoup:/host/mediasoup \
    node:16

apt-get update && apt-get install --no-install-recommends --yes \
    build-essential \
    python3 python3-pip

su node

cd /host/mediasoup/
npm install
```

Now, a clean *Stage 2* is used to test installing with and without this PR's change.

**Before the change**

```sh
docker run -ti --rm --name stage2 \
    --entrypoint /bin/bash \
    -v /path/to/mediasoup:/host/mediasoup \
    node:16-slim

export NODE_ENV="production"
export MEDIASOUP_WORKER_BIN="/host/mediasoup/worker/out/Release/mediasoup-worker"

su node

cd /host/mediasoup/
rm -rf node_modules/
npm install
```

This renders the following error:

> /bin/sh: 1: make: not found
> npm ERR! code 1
> npm ERR! path /host/mediasoup
> npm ERR! command failed
> npm ERR! command sh -c node npm-scripts.js postinstall

And installing Make would then complain about missing Python:

```sh
exit # Exit `su node`

apt-get update && apt-get install --no-install-recommends --yes \
    make

su node

cd /host/mediasoup/
rm -rf node_modules/
npm install
```

> /bin/sh: 2: python: not found
> make: python: Command not found
> make: *** [Makefile:56: meson-ninja] Error 127
> make: Leaving directory '/host/mediasoup/worker'
> npm ERR! code 1
> npm ERR! path /host/mediasoup
> npm ERR! command failed
> npm ERR! command sh -c node npm-scripts.js postinstall

**After the change**

Stage 2 runs without errors:

```sh
docker run -ti --rm --name stage2 \
    --entrypoint /bin/bash \
    -v /path/to/mediasoup:/host/mediasoup \
    node:16-slim

export NODE_ENV="production"
export MEDIASOUP_WORKER_BIN="/host/mediasoup/worker/out/Release/mediasoup-worker"

su node

cd /host/mediasoup/
rm -rf node_modules/
npm install
# OK
```

Is this enough? I thought how to easily make an automated test from this but it's a bit contrived thing to make because it requires modifying what is installed or not in the system (hence why I used Docker for clean baseline on each stage)

ibc:
Honestly I don't know if that's enough, probably it's but it's too Docker dependent.

Something easier:

- No Docker at all.
- Build mediasoup normally.
- Then move the binary somewhere else to /tmp.
- Set the env variable before running "npm run postinstall", it should NOT compile anything nor require python/pip/meson.
- Set the env again and run node interactive terminal, require("mediasoup") and create a Worker. It should work normally.

nazar-pc (approved):
This is correct, thanks!

When I was refactoring it I moved check into `worker:build`, but then when I added cleanup into `postinstall` it became incorrect.

j1elo:
> Honestly I don't know if that's enough, probably it's but it's too Docker dependent.
> 
> Something easier:
> 
>     * No Docker at all.
>     * Build mediasoup normally.
>     * Then move the binary somewhere else to /tmp.
>     * Set the env variable before running "npm run postinstall", it should NOT compile anything nor require python/pip/meson.
>     * Set the env again and run node interactive terminal, require("mediasoup") and create a Worker. It should work normally.

All done and yes, without the change `npm run postinstall` does run some make and pip install commands; after the change, it does nothing.

Note that my test was actually better by using Docker, and it's not really dependent on Docker itself at all, it's just the tool that I happen to use to ensure a clean (and repeatable) system, and the errors are more obvious because there is nothing installed. In your step 4,

> * Set the env variable before running "npm run postinstall", it should NOT compile anything nor require python/pip/meson.

running without Docker does actually work in my system, because I have lots of customization, tools and packages installed, included Meson. So for a true, isolated test, I would need to install a clean Debian system or maybe make a virtual machine... but at that point, a clean Docker container is already equivalent to those (and faster)

ibc:
Thanks. Merging.

PR #699, opened 2021-10-29
Emit Worker `died` event before closing Routers

### LINKED ISSUE #698: Emit Worker `died` event before Routers `workerclosed` ones
At https://github.com/versatica/mediasoup/blob/053c2e08c6b7cb1880ce390ad58308d9fdcbf6dc/node/src/Worker.ts#L597-L607 the Worker `died` event is emitted before Worker `close` event, but it's emitted after the Routers `close` events. Being "died" the event reason that makes the Worker to close, it should be emitted before the `close` event of the Routers, so they can also get the "why" before doing the closing, allowing to process it before the close, lefting that one just only for clean-up tasks. This is similar to what happens with Producer `close` event, that's the one dispatching Consumer `producerclose` and `close` events:

https://github.com/versatica/mediasoup/blob/053c2e08c6b7cb1880ce390ad58308d9fdcbf6dc/node/src/Consumer.ts#L585-L603

Fix is just a matter of move `this.safeEmit('died', error);` line before the loop, nothing else. This way flow would be the same as Producer close: Worker `died` -> Routers `close` -> Worker definitely `close`.

As a side note, I don't know if also it would make sense to have a Router `workerdied` event the same that it exists a `transportclosed` one, but that's not needed for this issue, and if so it can be discussed in another new one.

piranna:
This fixes #698. It's done quickly on the web interface, but can improve it on my editor, and also add support for `workerdied` event if you like it.

### DIFF
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -594,14 +594,14 @@ export class Worker extends EnhancedEventEmitter
 		// Close the PayloadChannel instance.
 		this.#payloadChannel.close();
 
+		this.safeEmit('died', error);
+
 		// Close every Router.
 		for (const router of this.#routers)
 		{
 			router.workerClosed();
 		}
 		this.#routers.clear();
-
-		this.safeEmit('died', error);
 
 		// Emit observer event.
 		this.#observer.safeEmit('close');


node/src/Worker.ts
```diff
@@ -594,15 +594,15 @@ export class Worker extends EnhancedEventEmitter
 		// Close the PayloadChannel instance.
 		this.#payloadChannel.close();
 
+		this.safeEmit('died', error);
```
ibc, line 597: Same concern as told before. This change means that, within the worker "died" event, router.closed is false. And that's wrong and inconsistent state.

piranna: Ok, let me take a look and improve the PR. Should `this.#routers.clear();` be executed before, so state is the same for all emitted events (`died`, Router's `close` and `close`)?

ibc: Calling routers.clear() is not gonna change anything. The thing here is that this is good as it is. Yes, in router "workerclose" event we don't know why the worker was closed (it died or what) but that's another story that could be addressed in another way. Events order is correct as it is.

We may add a worker.died getter that is set in its private died() method so the app can check worker.died within the "workerclose" event emitted by the router.

piranna: > We may add a worker.died getter that is set in its private died() method so the app can check worker.died within the "workerclose" event emitted by the router.

I still think that `died` event should be emitted before than Router "workerclose" event, but yes, a `died` getter would allow to check if the Worker closed because it died, so it's ok for me.

piranna: > This change means that, within the worker "died" event, router.closed is false. And that's wrong and inconsistent state.

I think I understood it know, are you talking about the Router could be "open" in a "closed" Worker, isn't it? Yes, I also think this doesn't make sense, closing should be done on inversed order of creation/open, like a Matrioska, totally agree on that...

"Maybe" it could make sense since what we are emitting is `died` event, the "reason" one, not Worker `close` event, that's being emitted after Router ones and that's correct as is it now, but `died` could be still being emitted before them without any inconsistent state, except that there could be unclosed Routers in a died (but also unclosed) Worker... That's similar the same as having a Worker with a `died` getter and read them from the Router `close` events, just only here emitting a `died` event instead, isn't it?

In any case, both solutions, the `died` event or the `died` getter, both of them before Routers `close`, could do the job for me, just only I think in this case the `died` event is a better API.

ibc:
Addressed in 16683acbb18f194a54a3656c11d758a2a5948ef5 via a new `worker.died` getter that it's set to true **before** routers are closed.

PR #703, opened 2021-10-31
Fix drop ordering

nazar-pc:
This was a tricky racy error that resulted in this among other crashes:
```
(ABORT) DepUsrSCTP::DeregisterSctpAssociation() | failed assertion `found > 0': SctpAssociation not found
```

This primarily happened with very fast tests, sometimes under special circumstances.

The root issue is that process was exiting without waiting for worker to finish properly, thus global C++ destructors were racing with worker shutdown, resulting in weird things that do not make sense like above or sometimes in various interesting segfaults.

With these changes when the very last thing holding `WorkerManager` on Rust side is dropped (worker manager is the first thing user of the library needs to create), it will synchronously wait for all workers to finish whatever they are doing before moving any further, making sure everything is de-initialized on C++ side gracefully.

### DIFF
--- a/rust/src/router.rs
+++ b/rust/src/router.rs
@@ -398,14 +398,11 @@ impl Inner {
                 let request = RouterCloseRequest {
                     internal: RouterInternal { router_id: self.id },
                 };
-                let worker = self.worker.clone();
                 self.executor
                     .spawn(async move {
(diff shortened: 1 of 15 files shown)


ibc (approved):
Code looks good

PR #709, opened 2021-11-07
Use more than one core for compilation on Windows

nazar-pc:
MinGW is what we expect people to use right now on Windows, so this should do the trick


ibc (approved):

ibc:
How can clang linter fail here? 😐

nazar-pc:
Hm... looks like clang-format was updated in the package we use

nazar-pc:
Reinstalled locally and applied new rules. Please take a look, maybe you'll want to tweak formatter settings now.

ibc:
It's ok, thanks

PR #713, opened 2021-11-21
worker: backport m1 compilation to v2

Vardiak:
This is a backport of #548 to v2, which we still use in production. Compilation from docker in arm64 works fine but not natively (M1 Pro).


Vardiak:
@ibc Hi, can you checkout this PR ? Thanks!

nazar-pc:
v2 is no longer supported, I would recommend you to use own fork if you need it

PR #714, opened 2021-11-22
Support VP9 profile-id 0 and 2

j1elo:
Different values of VP9 `profile-id` must be provided by endpoints with different Payload Types. Seems that this parameter would work like `packetization-mode` does in H.264, and should be declared multiple times inside mediasoup caps arrays.

Browsers such as Chrome and Safari already generate SDP Offers where VP9 is offered twice, with lines such as:

```
m=video 9 UDP/TLS/RTP/SAVPF 98 100
a=rtpmap:98 VP9/90000
a=fmtp:98 profile-id=0
a=rtpmap:100 VP9/90000
a=fmtp:100 profile-id=2
```

However, `ortc.getExtendedRtpCapabilities()` in mediasoup-client does codec matching [^1] based on the single VP9 entry that exists in the Router [^2], which doesn't contain `packetization-mode`, and thus is assumed to always be `0` [^3]. What this means is that `packetization-mode=2` is always dropped as a valid codec alternative. The Extended capabilities array ends up with a single VP9 entry, with `packetization-mode=0`.

[^1]: https://github.com/versatica/mediasoup-client/blob/3.6.45/src/ortc.ts#L573
[^2]: https://github.com/versatica/mediasoup/blob/3.9.2/node/src/supportedRtpCapabilities.ts#L213-L225
[^3]: https://github.com/versatica/mediasoup-client/blob/3.6.45/src/ortc.ts#L1123-L1124

Firefox doesn't include `profile-id` in its Offers as of writing this PR; however, RFC says that lack of it must be assumed equivalent to `profile-id=0` anyway.

So if it makes sense, I'd like to propose this addition to mediasoup.

Relevant doc:
* RTP Payload Format for VP9, 6.1. SDP Parameters
  https://datatracker.ietf.org/doc/html/draft-ietf-payload-vp9-16#section-6.1

### DIFF
--- a/node/src/supportedRtpCapabilities.ts
+++ b/node/src/supportedRtpCapabilities.ts
@@ -211,9 +211,30 @@ const supportedRtpCapabilities: RtpCapabilities =
 			]
 		},
 		{
-			kind         : 'video',
-			mimeType     : 'video/VP9',
-			clockRate    : 90000,
+			kind       : 'video',
+			mimeType   : 'video/VP9',
+			clockRate  : 90000,
+			parameters :
+			{
+				'profile-id' : 0
+			},
+			rtcpFeedback :
+			[
+				{ type: 'nack' },
+				{ type: 'nack', parameter: 'pli' },
+				{ type: 'ccm', parameter: 'fir' },
+				{ type: 'goog-remb' },
+				{ type: 'transport-cc' }
+			]
+		},
+		{
+			kind       : 'video',
+			mimeType   : 'video/VP9',
+			clockRate  : 90000,
+			parameters :
+			{
+				'profile-id' : 2
+			},
 			rtcpFeedback :
 			[
 				{ type: 'nack' },


nazar-pc (commented):
This shouldn't be required. H264 also has `profile-id`, yet as you can see `supportedRtpCapabilities` doesn't contain every single variant of it. `profile-id` must be specified in router options and then sent by the browser or other client. You don't need to modify mediasoup source code for that.

j1elo:
So then I understood it wrong. Reading the VP9 payload RFC draft I got the impression that this param makes it mandatory to define the codec several times for different PTs, like `packetization-mode` seems to do for H.264.

While portions of the default `supportedRtpCapabilities` array from mediasoup (or equivalently, the `Router.rtpCapabilities` property) can be copied as-is to the application for use in the Router (and I guess that's how people have been filling their RtpCaps arrays for their mediasoup apps), what you say means the same cannot be done for VP9, as applications must know to bring its own `profile-id` not included in mediasoup default values. Which is OK but seems that providing mediasoup Router with an _actually correct_ RtpCaps becomes a bit more and more tricky every day :-S

j1elo:
(a while later)

I just checked and H.264 has `profile-level-id` (https://datatracker.ietf.org/doc/html/rfc6184#section-8.2.2), which doesn't necessarily mandate a different Payload Type for changed values, because this is a compound value with 3 fields, and (seems that only for unicast) the `level` field is actually changeable by an answerer (so 2 different `profile-level-id` values may indeed be negotiated under the same PT). That's why mediasoup uses an additional library to check if profile-level-ids are actually compatible, before deeming two of them as different with a simple `==` operation.

I guess my confusion then comes from this: wouldn't what you say for VP9 be parallel to doing the same with H.264's `packetization-mode`? I mean, having a single definition in mediasoup's `supportedRtpCapabilities`, and asking app writers to set their Router caps with the correct value for `packetization-mode` (i.e. why include H.264 twice in the default array, if apps are expected to put the correct particular parameters anyway)

ibc:
There is neither issue in h264 (the additional library used by mediasoup comes with lot of tests and it's just C++ code of libwebrtc porter to JS) nor in vp9 (the rule is simple: profile-id must match, and it's 0 if not set in RtpParameters or RtpCapabilities). This is super tested.

ibc:
The user app decides which configuration or configurations of VP9 it enabled in the Router. You can add all them to support all them.

j1elo:
I agree if that's the final position as devs of mediasoup. But then my question is not really answered: why the same doesn't apply to H.264?

"*The user app decides which configuration or configurations of H264 it enabled in the Router. You can add all them to support all them.*"

As it is right now, mediasoup makes an explicit decision and already comes with _both_ `packetization-mode=0` and `packetization-mode=1`. Shouldn't these be omitted, and left as an exercise for the application to set them properly?

EDIT - wording and format

ibc:
Do you mean that supportedCapabilities.ts include 2 h264 codec configurations each with a different packetization-mode? If so, indeed there should just be one without such a param.

j1elo:
Yes, I mean that the supportedCapabilities.ts includes things like:

* 3 `audio/multiopus` codecs, with parameters to support 4, 6, and 8 channels.
* 2 `audio/ISAC`, to support rates of `32000` and `16000`.
* 4 `audio/SILK`, again to support a variety of clock rates.
* 3 `audio/CN`, idem.
* 4 `audio/telephone-event`, idem.
* 2 `video/H264`, for configurations with `packetization-mode` 0 and 1.
* 2 `video/H265`, again for `packetization-mode` 0 and 1.

This list is complete, I took the time to include all such repeated codec configurations.

Given that, I assumed the intention is indeed to include there all of the mandatory parameters, and that probably the omission of `profile-id` for VP9 must have been just a mistake.

However if finding about all these repeated codec configs will open a can of worms because they should in fact not be there... then I'd suggest if you remove them, to write a special doc section about them, because after all this issue comes from the fact that all these are very technical and low level details that users won't know how to set... I myself just learned about VP9's `profile-id` after careful inspection of Chrome's SDP, otherwise I would have never guessed it just from the sample values that mediasoup provides.

ibc:
Reopening but considering it an issue instead.

ibc:
`channels` and `clockRates` are mandatory params (even if they have default values in some cases). Those are included in `supportedRtpCapabilities.ts` because those are the only values allowed for those codecs. We cannot do the same for `profile-level-id` in H264 because there are tons of them (plus combinations with `packetization-mode`).

I think we should just remove `packetization-mode` and have a single H264 and H265 codec.  That's all. Would you mind doing that in this PR or another?

nazar-pc:
Hm, I think those are important because they specify technically supported options. From that perspective having `packetization-mode` is helpful because it will make mediasoup reject `packetization-mode: 99` for instance. With that in mind it might make sense to do the same for VP9 if we know the list of profiles is limited and short.

j1elo:
Oh I think both perspectives make sense... although maybe this rises the question about if having duplicate entries in the array for every valid value is the best way to validate the field (vs. a more specific-purpose function that "knows" which values are correct and compares them with the ones provided by the user).

Maybe this should be added now for VP9, but with the pending task of reviewing all this and moving every non-essential parameters to some validation module in the future, along with removing as much duplication as possible from `supportedRtpCapabilities.ts`. 

Re: 
> With that in mind it might make sense to do the same for VP9 if we know the list of profiles is limited and short.

_Today_, Chrome and Safari do support the so-called VP9 profile-id *0* (8-bit colorspace) and *2* (10/12-bit colorspace). In any case, the full list of possible values is not much longer (it goes from 0 to 3):
https://datatracker.ietf.org/doc/html/draft-ietf-payload-vp9-16#section-6

>     +=========+===========+=================+==========================+
>     | Profile | Bit Depth | SRGB Colorspace |    Chroma Subsampling    |
>     +=========+===========+=================+==========================+
>     |    0    |     8     |        No       |        YUV 4:2:0         |
>     +---------+-----------+-----------------+--------------------------+
>     |    1    |     8     |       Yes       | YUV 4:2:2,4:4:0 or 4:4:4 |
>     +---------+-----------+-----------------+--------------------------+
>     |    2    |  10 or 12 |        No       |        YUV 4:2:0         |
>     +---------+-----------+-----------------+--------------------------+
>     |    3    |  10 or 12 |       Yes       | YUV 4:2:2,4:4:0 or 4:4:4 |
>     +---------+-----------+-----------------+--------------------------+

ibc:
Agreed. Instead of listing 2 H264 codecs we could just validate such a param. Same for profile-id in VP9. There is a validateCodec() function in ortc.ts that should do this. Thoughts?

nazar-pc:
I liked how explicit it was, but in case of H264 not really feasible to include all options of profile-id, so yeah, makes sense to me validate it separately and have one entry.

ibc:
Handling this in PR https://github.com/versatica/mediasoup/pull/718

PR #718, opened 2021-11-25
Single H264/H264 codec configuration in supportedRtpCapabilities

ibc:
- Based on discussion in #714.
- We don't need to have 2 H264 and 2 H265 codec configurations in `supportedRtpCapabilities` each with a different `packetization-mode` param.
- `ortc.ts`: Just check `packetization-mode` param when matching codecs in strict mode.

@nazar-pc may you check the Rust changes please?


nazar-pc (approved):
Looks good to me

PR #719, opened 2021-11-26
Windows support in Rust library

nazar-pc:
After Meson and removing file descriptor usage in Rust, Windows support is a fairly simple task.

2 notes:
* ~some of recent Rust changes cause tests to hang occasionally, I will fix those separately, this is in case they happen here~
* hopefully these Windows tests are more reliable than those with Node and we'll not have to disable them in CI, but who knows...

~After debugging tests hangs~ I will create a PR that prepares new Rust release later.


nazar-pc:
Pushed fix for test hang into this branch too

ibc (approved):

PR #722, opened 2021-11-30
fix: Workaround for tsc error

satoren:
fix error on CI https://github.com/versatica/mediasoup/runs/4362731978?check_suite_focus=true
Jest depends to @types/jsdom since[v27.4.0](https://github.com/facebook/jest/pull/11999). I think the problem in jest not considered the environment without the DOM.

Another option: use older jest.
```
"jest": "^26.6.2",
```

### DIFF
--- a/node/tsconfig.json
+++ b/node/tsconfig.json
@@ -2,7 +2,7 @@
 	"compileOnSave": true,
 	"compilerOptions":
 	{
-		"lib": [ "es2021" ],
+		"lib": [ "es2021", "DOM" ],
 		"target": "esnext",
 		"module": "commonjs",
 		"moduleResolution": "node",


nazar-pc (approved):
Thank you!

ibc:
Uppps I don't like this approach. Adding dom library to tsc just due to a test dependency is not nice.

I'll look for other options. @jmillan any thoughts?

nazar-pc:
I guess the better alternative might be to define all those types in the library itself. We use them as interfaces only, so should be fine. Actually that will go nicely with https://mediasoup.discourse.group/t/simplifying-apis-of-mediasoup-based-apps-in-typescript-with-zod/2996

ibc:
It's hard to tell Jest where dom TS definition is, I'm not sure it's possible. Anyway, defining the whole dom lib seems super crazy IMHO.

jmillan:
> I'll look for other options. @jmillan any thoughts?

I've been checking this PR and honestly I can't come with a better alternative than the one applied at the moment.

satoren:
https://github.com/facebook/jest/issues/12098

jmillan:
> facebook/jest#12098

Yes, I'm aware of it.

PR #723, opened 2021-12-02
Revert "fix: Workaround for tsc error"

satoren:
Because it was fixed in upstream https://github.com/facebook/jest/releases/tag/v27.4.3

### DIFF
--- a/node/tsconfig.json
+++ b/node/tsconfig.json
@@ -2,7 +2,7 @@
 	"compileOnSave": true,
 	"compilerOptions":
 	{
-		"lib": [ "es2021", "DOM" ],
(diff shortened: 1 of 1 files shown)


nazar-pc (approved):
Nice

ibc (commented):
I don't see jest dependency being updated in package.json...

nazar-pc:
> I don't see jest dependency being updated in package.json...

But we don't have lock file either (or else CI wouldn't break in the first place)

ibc:
Right but I always try to keep versions in package.json up to date 😀

Don't worry, I'll update it eventually.

Issue #701, opened 2021-10-30
Replace Makefile with Python script


nazar-pc:
I see people struggle on Windows with following installation instructions, it would be possible to simplify things for everyone if instead of using GNU Make and its Makefile (and have to deal with other Make implementations on WIndows and various other platforms) we used a Python script, in that case things would "just work" as soon as people have MSVS and Python installed, no extra steps would be required.

Pros:
* no need to install MinGW/MSYS and edit `Path` for them on Windows
* no need to manually configure MSVS environment before doing `npm i` on Windows
* no dependency and confusion with Make on any platform (GMake, NMake, etc.)
* more flexibility is what we can write there with sane syntax

Cons:
* Script will definitely be much longer than current Makefile as Makefile is specifically designed to execute a series of commands

This primarily affects Windows, but might potentially benefit other platforms too.

ibc:
`Makefile` has a dependency mechanism by which a task depends on others, but those are just executed if needed. For example:

```make
libmediasoup-worker: setup flatc
```

"setup" and "flatc" dependency tasks will just run if they didn't run before, and somehow `make` knows that. 

The question is: how to do that with a Python or whatever script?

nazar-pc:
We'd have to track the state somehow, not very difficult if you ask me

ibc:
I'm pretty sure it's difficult. How would you check if some .cpp file has changed (or a new one has been added) and if so compile it and relink etc?

nazar-pc:
That is an interesting feature of Make, but unnecessary for our use case

ibc:
How is that unnecessary for our use case? That use case is 58% of my whole life.

ibc:
Some stuff I've found, but doing a proper review of them is hard:

- https://github.com/casey/just
- https://www.pyinvoke.org
- https://pydoit.org

nazar-pc:
I mean we don't use it to compile things. Compilation is taken care of by Meson, we just use it to run other apps.

ibc:
> I mean we don't use it to compile things. Compilation is taken care of by Meson, we just use it to run other apps.

Yes, but if we blindly invoke the Meson command that compiles the worker immediately after having called it, it will recompile everything.

Ah no, it won't.

ibc:
[...]

PR #725, opened 2021-12-03
native windows build fixes

kcking:
Hi there! I am using the mediasoup crate on native windows and ran into some issues with the Makefile / build.rs script.

I made the following changes

1. use windows `rmdir` instead of `rm`
2. use backslashes in file paths since `rmdir` can't handle forward slashes
3. move comments to start-of-line so they are not interpreted as part of shell command
4. remove clean-all step from build.rs (note, `MEDIASOUP_OUT_DIR` was not being set in this `Command`, so it _appears_ to be a no-op currently)

Let me know if these are changes you would consider merging upstream. For now they can just live on my personal fork :)

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -16,8 +16,6 @@ MEDIASOUP_BUILDTYPE ?= Release
 GULP = ./scripts/node_modules/.bin/gulp
 LCOV = ./deps/lcov/bin/lcov
 DOCKER ?= docker
-PIP_DIR = $(MEDIASOUP_OUT_DIR)/pip
-MESON ?= $(PIP_DIR)/bin/meson
 MESON_ARGS ?= ""
 
 # Disable `*.pyc` files creation.
@@ -25,11 +23,15 @@ export PYTHONDONTWRITEBYTECODE = 1
 # Instruct `meson` where to look for ninja binary.
 ifeq ($(OS),Windows_NT)
 # Windows is, of course, special.
-	export NINJA = $(PIP_DIR)/bin/ninja.exe
+	PIP_DIR = $(MEDIASOUP_OUT_DIR)\pip
+	export NINJA = $(PIP_DIR)\bin\ninja.exe
+	MESON ?= $(PIP_DIR)\bin\meson
 	RRM = rmdir /q /s
 else
+	PIP_DIR = $(MEDIASOUP_OUT_DIR)/pip
 	export NINJA = $(PIP_DIR)/bin/ninja
 	RRM = rm -rf
+	MESON ?= $(PIP_DIR)/bin/meson
 # Workaround for NixOS and Guix that don't work with pre-built binaries, see:
 # https://github.com/NixOS/nixpkgs/issues/142383.
 	PIP_BUILD_BINARIES = $(shell [ -f /etc/NIXOS -o -d /etc/guix ] && echo "--no-binary :all:")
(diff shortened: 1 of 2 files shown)


ibc:
I just wonder if current build system is designed to support native Windows, and also wonder whether these changes here may break other working platforms, but I have no idea about Windows. @nazar-pc thoughts?

nazar-pc:
Wasn't designed (especially since you'd need MinGW for Makefile anyway), but I'll check these changes in more detail later.
If it doesn't break things, I have nothing against supporting more configurations of course.

kcking:
Thanks for the quick reply!

>  you'd need MinGW for Makefile anyway

Just to be clear I'm actually using `scoop`-installed make with these changes, so no need for mingw as far as I can tell.

```
# cmd.exe
C:\Users\kevin\...>where make
C:\Users\kevin\scoop\shims\make.exe
C:\Users\kevin\...>make --version
GNU Make 4.3
Built for Windows32
```

nazar-pc:
Well, then it breaks other environments as you can see from the failed CI.

nazar-pc:
If you have time working on #701 would be more productive than work around various kinds of `make` available on Windows that all behave differently.

worker/build.rs
```diff
@@ -93,20 +92,6 @@ fn main() {
...
-                .arg("clean-all")
-                .spawn()
-                .expect("Failed to start")
-                .wait()
-                .expect("Wasn't running")
-                .success()
-            {
-                panic!("Failed to clean libmediasoup-worker")
-            }
-        }
```
nazar-pc, line 109: This is important for package publishing for instance. But yes, `MEDIASOUP_OUT_DIR` is missing here and needs to be added.

kcking:
> Well, then it breaks other environments as you can see from the failed CI.

I think this is just because I didn't convert a couple of other paths to backslashes. I pushed a commit that fixes those as well , but totally understand if you're not looking to accomodate native windows make.

> If you have time working on #701 would be more productive than work around various kinds of make available on Windows that all behave differently.

That's fair. Unfortunately I don't expect to have the time to do that any time soon.

Thanks again for looking at the PR! I can push a fix for the clean-all step in a separate PR.

nazar-pc:
> I can push a fix for the clean-all step in a separate PR.

Please do

kcking:
> > I can push a fix for the clean-all step in a separate PR.
> 
> Please do

Here you go https://github.com/versatica/mediasoup/pull/726

kcking:
https://github.com/versatica/mediasoup/runs/4412988116?check_suite_focus=true#step:8:147 🤦 thanks for trying haha

It's now more wrong but more consistent ✨

worker/Makefile
```diff
@@ -134,7 +134,7 @@ clean-all: clean-subprojects
 
 mediasoup-worker: setup
 ifeq ($(MEDIASOUP_WORKER_BIN),)
-	$(MESON) compile -j $(CORES) -C $(MEDIASOUP_OUT_DIR)/$(MEDIASOUP_BUILDTYPE) mediasoup-worker
+	$(MESON) compile -j $(CORES) $(MEDIASOUP_OUT_DIR)/$(MEDIASOUP_BUILDTYPE) mediasoup-worker
```
nazar-pc, line 137: Why? `-C` specified build directory, you shouldn't remove it.

kcking: Sorry I'm doing some local debugging -- the build is failing with git but working with a local path dependency. sorry for the noise ill close the PR for now

PR #729, opened 2021-12-07
feat: Add Send trait to future for transport trait

satoren:
It is inconvenient that only Transport cannot use the multi-threaded executor.


nazar-pc (approved):
There was a technical reason why `?Send` was used before, but with recent changes it is possible to have Send

PR #730, opened 2021-12-08
Create a buffer to process RTP packets.

gnauhtt:
Fix the problem about the method 'RtpPacket::SetExtensions' changed the content of the buffer beyond the rtp length.

### DIFF
--- a/worker/src/RTC/DirectTransport.cpp
+++ b/worker/src/RTC/DirectTransport.cpp
@@ -20,6 +20,8 @@ namespace RTC
 	DirectTransport::~DirectTransport()
 	{
 		MS_TRACE();
+
+		delete[] this->buffer;
 	}
 
 	void DirectTransport::FillJson(json& jsonObject) const
@@ -102,7 +104,14 @@ namespace RTC
 					return;
 				}
 
-				RTC::RtpPacket* packet = RTC::RtpPacket::Parse(data, len);
+				// If this is the first time to reveive a rtp packet then allocate the receiving buffer now.
+				if (!this->buffer)
+					this->buffer = new uint8_t[RTC::MtuSize + 100];
+				
+				// Copy the received packet into this buffer so it can be expanded later.
+				std::memcpy(this->buffer, data, static_cast<size_t>(len));
+
+				RTC::RtpPacket* packet = RTC::RtpPacket::Parse(this->buffer, len);
 
 				if (!packet)
 				{
--- a/worker/src/RTC/DirectTransport.cpp
+++ b/worker/src/RTC/DirectTransport.cpp
@@ -104,7 +104,7 @@ namespace RTC
 					return;
 				}
 
-				// If this is the first time to reveive a rtp packet then allocate the receiving buffer now.
+				// If this is the first time to reveive a RTP packet then allocate the receiving buffer now.
 				if (!this->buffer)
 					this->buffer = new uint8_t[RTC::MtuSize + 100];
 
--- a/worker/src/RTC/DirectTransport.cpp
+++ b/worker/src/RTC/DirectTransport.cpp
@@ -107,7 +107,7 @@ namespace RTC
 				// If this is the first time to reveive a rtp packet then allocate the receiving buffer now.
 				if (!this->buffer)
 					this->buffer = new uint8_t[RTC::MtuSize + 100];
-				
+
 				// Copy the received packet into this buffer so it can be expanded later.
 				std::memcpy(this->buffer, data, static_cast<size_t>(len));
 
--- a/worker/include/RTC/DirectTransport.hpp
+++ b/worker/include/RTC/DirectTransport.hpp
@@ -34,6 +34,10 @@ namespace RTC
 		void SendSctpData(const uint8_t* data, size_t len) override;
 		void RecvStreamClosed(uint32_t ssrc) override;
 		void SendStreamClosed(uint32_t ssrc) override;
+
+	private:
+		// Allocated by this.
+		uint8_t* buffer{ nullptr };
 	};
 } // namespace RTC
 


nazar-pc (commented):
Can you provide more details in which case this happens exactly? Also a test case would really help here.

ibc:
Which problem is this PR solving. Im pretty sure the issue you mean does not exist. This is tested.

gnauhtt:
I am trying to receive rtsp media in nodejs, and forward to mediasoup worker. The while loop of the following code will start working When there are more than two RTP packets in the UnixStreamSocket buffer.
```cpp
void ConsumerSocket::UserOnUnixStreamRead()
	{
		MS_TRACE();

		size_t msgStart{ 0 };

		// Be ready to parse more than a single message in a single chunk.
		while (true)
		{
			if (IsClosed())
				return;

			size_t readLen = this->bufferDataLen - msgStart;

			if (readLen < sizeof(uint32_t))
			{
				// Incomplete data.
				break;
			}

			uint32_t msgLen;
			// Read message length.
			std::memcpy(&msgLen, this->buffer + msgStart, sizeof(uint32_t));

			if (readLen < sizeof(uint32_t) + static_cast<size_t>(msgLen))
			{
				// Incomplete data.
				break;
			}

			this->listener->OnConsumerSocketMessage(
			  this,
			  reinterpret_cast<char*>(this->buffer + msgStart + sizeof(uint32_t)),
			  static_cast<size_t>(msgLen));

			msgStart += sizeof(uint32_t) + static_cast<size_t>(msgLen);
		}

		if (msgStart != 0)
		{
			this->bufferDataLen = this->bufferDataLen - msgStart;

			if (this->bufferDataLen != 0)
			{
				std::memmove(this->buffer, this->buffer + msgStart, this->bufferDataLen);
			}
		}
	}
```
The first time in the loop will process the first rtp packet. When running to this line of code of the method `RtpPacket::SetExtensions`. Will change the content beyond payloadLength
`std::memmove(this->payload + shift, this->payload, this->payloadLength + this->payloadPadding);`
So that second time in the loop, the 'msgLen' length information has been tampered with

gnauhtt:
> Which problem is this PR solving. Im pretty sure the issue you mean does not exist. This is tested.

This problem only happens when 'ConsumerSocket::UserOnUnixStreamRead' to parse more than 2 RTP packets in a single chunk.

nazar-pc (approved):
OK, I see, `RtpPacket::SetExtensions()` does `std::memmove()` internally and can override data it doesn't own if underlying buffer comes from `DirectTransport` as in this case we don't have pre-allocated buffer with extra space at the end.

It is unfortunate, but I do agree that we need to copy memory here.

nazar-pc:
BTW this can be worked around by sending `RTC::MtuSize + 100` sized message from Node.js even though only part of it will be covered by actual RTP packet. But that is an ugly API, so probably worth doing copy here even though it is less efficient.

@rtctt can you reformat code according to configured rules with `npm run format:worker` please?

nazar-pc:
GitHub Actions do not feel well today, will retry CI later (already did a few times).

gnauhtt:
> GitHub Actions do not feel well today, will retry CI later (already did a few times).

Okay, I'm honored to be able to contribute, if there is anything I need to do, please let me know.

ibc:
I fail to see the problem. When parsing a RTP packet in its original buffer, we don't write into the original buffer, in fact we cannot do it since in many cases (UDP) it's a read only buffer. Do I miss something?

nazar-pc:
When parsing we don't, but we do later in `Producer::MangleRtpPacket()` where `RtpPacket::SetExtensions()` is called from if I read it correctly.

nazar-pc:
The reason such problem doesn't exist in TCP/UDP is because there we have buffer from which we read and that buffer is filled with one chunk at a time. With DirectTransport though data might be buffered on Node.js side and multiple packets may fill the buffer. So while for TCP and UDP it is not a problem to write beyond the boundary of the data that belongs to the packet, in case of DirectTransport it becomes problematic.

I wish it was harder to do these kind of errors in C++, it is really not obvious that this happens in the first place until you have issues.

Actually, this PR fixes literal UB in when combined with Rust as in that case we send a piece of pre-allocated memory from Rust that is exactly the size of the packet, so writing past the edge of the initially allocated packet that is very, very bad idea.

ibc:
There are two concerns or steps here that I'd like to confirm:

1. When parsing a packet (no matter where it comes from) we don't mangle the original buffer. Pretty sure this is true. Is it?

2. Once parsed and the RtpPacket instance is created, it still references the original buffer. However in Producer.cpp we clone it into a per thread static buffer, we operate in the packet, write into that memory, and we release the original RtpPacket pointer.

So I still fail to see what is not true above so it behaves different when in DirectTransport.

nazar-pc:
1. Yes
2. It does reference original memory, but Producer.cpp DOES modify it in `Producer::ReceiveRtpPacket()`, it is not cloned before being modified, hence changes original memory directly.

It behaves identically with any transport, it just happens to be the case that with other transport it incidentally doesn't cause problems.

ibc:
I have to check bullet 2. BTW why doesn't it affect to TCP transport? Perhaps in TCP we memcpy before delivering the packet to the Router?

gnauhtt:
> I have to check bullet 2. BTW why doesn't it affect to TCP transport? Perhaps in TCP we memcpy before delivering the packet to the Router?

Yes, that why this problem just happens in DirctTransport.

ibc:
Then we should do in DirectTransport the same we do in TCP transport and memcpy the buffer to a separate buffer.

nazar-pc:
There is no copy in TCP and UDP that mediasoup does on the read side from what I can see though. Also in that case we'll copy things when they don't necessarily need to be copied.

Nawwf (approved):

worker/src/RTC/DirectTransport.cpp
```diff
@@ -102,7 +104,14 @@ namespace RTC
 					return;
 				}
 
-				RTC::RtpPacket* packet = RTC::RtpPacket::Parse(data, len);
+				// If this is the first time to reveive a rtp packet then allocate the receiving buffer now.
```
jmillan:
```suggestion
				// If this is the first time to reveive a RTP packet then allocate the receiving buffer now.
```

jmillan:
> There is no copy in TCP and UDP that mediasoup does on the read side from what I can see though.

For UDP we use a static read buffer, and for TCP we do allocate a buffer one time, when the first packet comes, as it is done in this PR.

jmillan (approved):

nazar-pc:
@ibc let's get this merged?

ibc:
I will review tomorrow, I was not convinced.

ibc (approved):

jmillan (approved):

PR #737, opened 2021-12-20
Clean everything except `mediasoup-worker` binary in postinstall step

### LINKED ISSUE #736: Mediasoup doesn't clean up after node postinstall build
## Bug Report

I noticed that the mediasoup node modules is over 55MB, which didn't seem to make sense.  Poking around I see most of that is Python files and compiled object files, which I don't believe are needed at run time.

### Your environment

- Operating system: MacOS 12.0.1,  M1 Max arm64
- Node version: 16.13.1
- npm version: 8.3.0
- gcc/clang version: Apple clang version 13.0.0 (clang-1300.0.29.3)
- mediasoup version: 3.9.2

### Issue description

I have a partial solution - in this script:

       case 'postinstall':
        {
                {
                        execute('node npm-scripts.js worker:build');
                        execute(`${MAKE} clean-pip -C worker`);
                        execute(`${MAKE} clean-subprojects -C worker`);
                }

                break;
        }

in this case "clean-subprojects" uses meson which needs Python, so the previous step "clean-pip" just gets undone as the following one re-installs pip.  Swapping the order of those should at least keep pip from being reinstalled.

After that there is still about 33MB left, though.

I went into the build output and recursively deleted all ".o" and ".a" files, along with some other large temp files (build.ninja, compile_commands.json, .ninja_*) and the whole module is down to about 9MB. (Removing everything from worker except mediasoup-worker gets the while thing down to about 5.5MB but I'd consider deleting the sources beyond what a build cleanup script would want to do and left to anyone really wanting to optimize for their own uses...)

Summary: low hanging fruit would be just to swap the clean commands so clean-pip is last.  But if you added another cleanup step just to remove the temporary object/build output from worker/out/Release that would make the whole project take less than 10MB, which is really nice when packaging it into a container, etc.

nazar-pc:
This makes `node_modules/mediasoup` smaller than ever before.

Fixes #736, cc @jcpage

### DIFF
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -121,8 +121,12 @@ switch (task)
 		if (!process.env.MEDIASOUP_WORKER_BIN)
 		{
 			execute('node npm-scripts.js worker:build');
-			execute(`${MAKE} clean-pip -C worker`);
+			// Clean build artifacts except `mediasoup-worker`.
+			execute(`${MAKE} clean-build -C worker`);
+			// Clean downloaded dependencies.
 			execute(`${MAKE} clean-subprojects -C worker`);
+			// Clean PIP/Meson/Ninja.
+			execute(`${MAKE} clean-pip -C worker`);
 		}
 
 		break;
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -94,6 +94,11 @@ Builds the `libmediasoup-worker` static library at `worker/out/Release/`.
 Cleans built objects and binaries.
 
 
+### `make clean-build`
+
+Cleans built objects and other artifacts, but keeps `mediasoup-worker` binary in place.
+
+
 ### `make clean-pip`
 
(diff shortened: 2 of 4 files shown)


jmillan (approved):
👍 nice!

npm-scripts.js
```diff
@@ -121,8 +121,12 @@ switch (task)
 		if (!process.env.MEDIASOUP_WORKER_BIN)
 		{
 			execute('node npm-scripts.js worker:build');
-			execute(`${MAKE} clean-pip -C worker`);
+			// Clean build artifacts except `mediasoup-worker`
+			execute(`${MAKE} clean-build -C worker`);
+			// Clean downloaded dependencies
 			execute(`${MAKE} clean-subprojects -C worker`);
```
ibc: Dot at the end of comments, please 😀

ibc (approved):

PR #741, opened 2021-12-24
Improve Windows support by not requiring MSVC configuration

nazar-pc:
Since Meson `0.59.0` there is a flag `--vsenv` that will configure MSVC automatically, so users will no longer have to run cryptic commands themselves, greatly improving development process. This is why additional setup instructions are no longer needed in CI either.

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -37,6 +37,13 @@ else
 	export PYTHONPATH := $(PIP_DIR):${PYTHONPATH}
 endif
 
+# Activate VS environment on Windows by default.
+ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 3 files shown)


worker/Makefile
```diff
@@ -37,6 +37,13 @@ else
 	export PYTHONPATH := $(PIP_DIR):${PYTHONPATH}
 endif
 
+# Default Windows activate VS environment
```
ibc: Dot please 😅

ibc (approved):
Lot of lines removed. I love it.

worker/Makefile
nazar-pc: I wish there was an automated check for this, I keep forgetting

PR #742, opened 2021-12-25
Make Tst header packed to fix Windows tests and unlock Windows worker CI

### LINKED ISSUE #702: Heap corruption on Windows in worker tests
### Your environment

- Operating system: Windows 10 19043.1288
- gcc/clang version: MSVC 19.29.30136
- mediasoup version: 3.9.0

### Issue description

I was checking why on Windows tests are crashing sometimes and it turns out it is reproducible with after at most a few attempts:
```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Error!

Program: ...nazar-pc\mediasoup\worker\out\Debug\mediasoup-worker-test.exe

HEAP CORRUPTION DETECTED: after Normal block (#27767) at 0x0000022817FFDC50.
CRT detected that the application wrote to memory after end of heap buffer.
```

Here is a call stack:
Знімок екрана з 2021-10-30 23-46-51

Apparently crashed on exit from "create FeedbackPsTstPacket" test.

nazar-pc:
I think this is correct, see https://github.com/versatica/mediasoup/issues/702#issuecomment-1000954034 for context.

Fixes #702

### DIFF
--- a/worker/include/RTC/RTCP/FeedbackPsTst.hpp
+++ b/worker/include/RTC/RTCP/FeedbackPsTst.hpp
@@ -24,6 +24,12 @@ namespace RTC
 		class FeedbackPsTstItem : public FeedbackItem
 		{
 		public:
+#ifdef _WIN32
+#pragma pack(push, 1)
+#define MEDIASOUP_PACKED
+#else
+#define MEDIASOUP_PACKED __attribute__((packed))
+#endif
 			struct Header
 			{
 				uint32_t ssrc;
@@ -36,7 +42,11 @@ namespace RTC
 				uint8_t reserved2 : 3;
 				uint8_t index : 5;
 #endif
-			};
+			} MEDIASOUP_PACKED;
+#ifdef _WIN32
+#pragma pack(pop)
+#endif
+#undef SCTP_PACKED
 
 		public:
 			static const size_t HeaderSize{ 8 };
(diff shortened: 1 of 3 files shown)


worker/include/RTC/RTCP/FeedbackPsTst.hpp
```diff
@@ -36,7 +42,11 @@ namespace RTC
 				uint8_t reserved2 : 3;
 				uint8_t index : 5;
 #endif
-			};
+			} MEDIASOUP_PACKED;
```
ibc, line 45: This is completely unknown for me, are we using the same somewhere else?

ibc (approved):

worker/include/RTC/RTCP/FeedbackPsTst.hpp
nazar-pc: The goal here is to have packed struct, such that compiler doesn't add extra alignment. However, as of C++11 standard there is no canonical way of doing it and different compilers have their own syntax.

MSVC:
```cpp
// Align to 1 byte
#pragma pack(push, 1)
struct X {};
// Revert back to previous behavior
#pragma pack(pop) 
```

GCC:
```cpp
struct X {} __attribute__((packed));
```

I took this pattern from usrsctp codebase that has a lot of packed data structures.

I have not seen any other place where packed structs are used and I after quick search through similar headers I have not seen any other cases where alignment would cause issues. I have not checked all structs throughout codebase though.

ibc (approved):

PR #744, opened 2021-12-27
Ice renomination

### LINKED ISSUE #734: Renomination support
Do mediasoup support ICE Renomination or suppose to support it ?

In verison 1.2.5, mediasoup added support for ICE renomination in sdp.
https://github.com/versatica/mediasoup/commit/994e3651cb6f14b18206bcf4546e12a5a67c72f8

But mediasoup-worker does not actully process NOMINATION attribute. 

When client truned on renomination, it will never sending a STUN packet with USE-CANDIDATE.
And mediasoup will use the address in first bind request as selected tuple.
But when client sending a STUN packet with NOMINATION later, mediasoup will not switch to it.

penguinol:
Implement https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination
According to https://webrtc.googlesource.com/src/+/refs/heads/main/p2p/base/connection.cc#592

### DIFF
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -69,7 +69,7 @@ namespace RTC
 			this->oldPassword = this->password;
 			this->password    = password;
 
-			this->remoteNomination = 0;
+			this->remoteNomination = 0u;
 		}
 		bool IsValidTuple(const RTC::TransportTuple* tuple) const;
 		void RemoveTuple(RTC::TransportTuple* tuple);
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -101,7 +101,7 @@ namespace RTC
 		std::string password;
 		std::string oldUsernameFragment;
 		std::string oldPassword;
-		uint32_t remoteNomination{ 0 };
+		uint32_t remoteNomination{ 0u };
 		IceState state{ IceState::NEW };
 		std::list<RTC::TransportTuple> tuples;
 		RTC::TransportTuple* selectedTuple{ nullptr };
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -61,15 +61,15 @@ namespace RTC
 		{
 			return this->selectedTuple;
 		}
-		void SetUsernameFragment(const std::string& usernameFragment)
+		void RestartIce(const std::string& usernameFragmnent, const std::string& password)
 		{
 			this->oldUsernameFragment = this->usernameFragment;
 			this->usernameFragment    = usernameFragment;
-		}
-		void SetPassword(const std::string& password)
-		{
+
 			this->oldPassword = this->password;
 			this->password    = password;
+
+			this->remoteNomination = 0;
 		}
 		bool IsValidTuple(const RTC::TransportTuple* tuple) const;
 		void RemoveTuple(RTC::TransportTuple* tuple);
@@ -78,7 +78,7 @@ namespace RTC
 		void ForceSelectedTuple(const RTC::TransportTuple* tuple);
 
 	private:
-		void HandleTuple(RTC::TransportTuple* tuple, bool hasUseCandidate);
+		void HandleTuple(RTC::TransportTuple* tuple, uint32_t nomination);
 		/**
 		 * Store the given tuple and return its stored address.
 		 */
@@ -101,6 +101,7 @@ namespace RTC
 		std::string password;
 		std::string oldUsernameFragment;
 		std::string oldPassword;
+		uint32_t remoteNomination{ 0 };
 		IceState state{ IceState::NEW };
 		std::list<RTC::TransportTuple> tuples;
 		RTC::TransportTuple* selectedTuple{ nullptr };
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -61,7 +61,7 @@ namespace RTC
 		{
 			return this->selectedTuple;
 		}
-		void RestartIce(const std::string& usernameFragmnent, const std::string& password)
+		void RestartIce(const std::string& usernameFragment, const std::string& password)
 		{
 			this->oldUsernameFragment = this->usernameFragment;
 			this->usernameFragment    = usernameFragment;
--- a/worker/src/RTC/IceServer.cpp
+++ b/worker/src/RTC/IceServer.cpp
@@ -344,7 +344,7 @@ namespace RTC
 					// Store the tuple.
 					auto* storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						MS_DEBUG_TAG(ice, "transition from state 'new' to 'completed'");
 
@@ -394,7 +394,7 @@ namespace RTC
 					// Store the tuple.
 					auto* storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						MS_DEBUG_TAG(ice, "transition from state 'disconnected' to 'completed'");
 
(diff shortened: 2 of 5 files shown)


nazar-pc:
Is this supposed to close #734?

penguinol:
> Is this supposed to close #734?

Yes, this pr fix #734 

With ice-renomination, mobile can switch form cell-networker to wifi-network smoothly.

nazar-pc:
Restart ICE tests are failing now

penguinol:
> Restart ICE tests are failing now

Can you provide more details?

make test is all passed.

nazar-pc:
This repo has GitHub Actions CI configured, you can see failed tests there.

penguinol:
> This repo has GitHub Actions CI configured, you can see failed tests there.

Ok, I see it, will check it tomorrow.

nazar-pc:
Thanks, similarly would be nice to have some tests for new behavior, so we can be sure it keeps working.

worker/include/RTC/IceServer.hpp
```diff
@@ -61,15 +61,15 @@ namespace RTC
 		{
 			return this->selectedTuple;
 		}
-		void SetUsernameFragment(const std::string& usernameFragment)
+		void RestartIce(const std::string& usernameFragment, const std::string& password)
```
ibc, line 64: Why is this method renamed?

worker/include/RTC/IceServer.hpp
```diff
@@ -61,15 +61,15 @@ namespace RTC
...
 		{
 			this->oldUsernameFragment = this->usernameFragment;
 			this->usernameFragment    = usernameFragment;
-		}
-		void SetPassword(const std::string& password)
-		{
+
 			this->oldPassword = this->password;
 			this->password    = password;
+
```
ibc, line 71: Please avoid this blank line.

worker/include/RTC/IceServer.hpp
penguinol: The only usage of `SetUsernamFragment ` and `SetPassword` is when restarting ice.
And remoteNomination also need to be set to 0 at the same time.
So instead of adding another function like `SetNomination`, i perferred to merge them into one.
`RestartIce` is an action and we can do more thing in it.

worker/include/RTC/IceServer.hpp
penguinol: The blank line here is to separate different assignment operation to make it more readable
```
void RestartIce(const std::string& usernameFragment, const std::string& password)
{
	this->oldUsernameFragment = this->usernameFragment;
	this->usernameFragment    = usernameFragment;

	this->oldPassword = this->password;
	this->password    = password;
        
        this->remoteNomination = 0;
}
```

penguinol:
> Thanks, similarly would be nice to have some tests for new behavior, so we can be sure it keeps working.

Good advice, but currently we don't have an ice-server test. And it's a little bit hard to make a valid STUN packet.

ibc (commented):
Thanks for this. Please let me do a proper review these days.

penguinol:
> Thanks for this. Please let me do a proper review these days.

worker/src/RTC/IceServer.cpp
```diff
@@ -198,8 +198,15 @@ namespace RTC
 
 				delete response;
 
+				uint32_t nomination = 0;
```
ibc: Can you type 0u instead for consistency?

worker/src/RTC/IceServer.cpp
```diff
@@ -370,22 +384,29 @@ namespace RTC
...
-					// Update state.
-					this->state = IceState::COMPLETED;
-					// Notify the listener.
-					this->listener->OnIceServerCompleted(this);
+					if (nomination > this->remoteNomination)
+					{
+						MS_DEBUG_TAG(ice, "transition from state 'disconnected' to 'completed'");
+
+						// Mark it as selected tuple.
+						SetSelectedTuple(storedTuple);
```
ibc, line 402: Wrong tabulation here right?

worker/src/RTC/IceServer.cpp
```diff
@@ -415,12 +436,17 @@ namespace RTC
 					if (!storedTuple)
 						storedTuple = AddTuple(tuple);
 
-					// Mark it as selected tuple.
-					SetSelectedTuple(storedTuple);
-					// Update state.
-					this->state = IceState::COMPLETED;
-					// Notify the listener.
-					this->listener->OnIceServerCompleted(this);
+					if (nomination > this->remoteNomination)
```
ibc: Here we are just updating the selected tuple if the STUN packet has a NOMINATION value higher than previous one. However, if the packet doesn't have that attribute and it has USE_CANDIDATE, we are not updating the selected tuple, so this is changing the previous behavior. I'm afraid of the side effects of this specially with Firefox which doesn't implement ICE Lite and I don't know if libwebrtc (Chrome) implements the NOMINATION attribute, does it?

What worries me is that, if the endpoint doesn't support NOMINATION attribute we are behaving differently than before:

- Before this PR any STUN packet with USE_CANDIDATE would update selected tuple.
- After this PR, the selected tuple is not updated if a previous tuple was already selected and a new STUN packet with USE_CANDIDATE is received.

I would like to keep previous behavior in case the STUN packet doesn't have any NOMINATION attribute. What do you think?

ibc:
> > Is this supposed to close #734?
> 
> 
> 
> Yes, this pr fix #734 

Can you add this sentence into the  PR description so, once merged, that issue will be automatically closed?

nazar-pc:
I already wired them together, you can see that in the right sidebar

worker/src/RTC/IceServer.cpp
penguinol: Yes, we can do this by change `if (nomination > this->remoteNomination)` to `if (nomination >= this->remoteNomination)` in COMPLETE state.

ibc: Then both cases/conditions (the current one in the PR) and the one you suggest (this one to keep previous behavior) should be needed, right?

worker/src/RTC/IceServer.cpp
penguinol: Sorry. what do you mean?

ibc: Sorry, wrong display in GH mobile. It's ok.

worker/src/RTC/StunPacket.cpp
```diff
@@ -209,6 +209,22 @@ namespace RTC
...
+					// Ensure attribute length is 4 bytes.
+					if (attrLength != 4)
+					{
+						MS_WARN_TAG(ice, "attribute NOMINATION must be 4 bytes length, packet discarded");
+
+						delete packet;
+						return nullptr;
+					}
+
+					packet->SetNomination(Utils::Byte::Get4Bytes(attrValuePos, 0));
```
ibc, line 223: Before this, I'd add a packet->SetHasNomination(true) which sets a new Packet.hasNomination Boolean  to true. This way, instead of checking nomination >= we can check (nomination > this.remoteNomination || !this.hasNomination). What do you think?

jmillan:
Thanks for this effort @penguinol

worker/src/RTC/StunPacket.cpp
penguinol: Here we set `nomination = 1` when the packet has USE-CANDIDATE attr. And chorme will never send STUN packet with NORMINATION attr and value = 0 or value equals previous tuple.
if we want to make it more readable, we need `hasCandidate`, `hasNomination`, `normination` as the args.
So `if(nomination == 0)` will be`if(!hasUseCandidate && !hasNomination)`
`if(nomination >= thisremoteNomination)` will be `if(hasUseCandidate || (hasNomination && normination > this.remoteNomination))`
Which one do you perferred?

ibc:
I'm gonna add missing things.

worker/src/RTC/StunPacket.cpp
ibc: I'm handling this in this branch directly, will push soon.

ibc, line 223: This is wrong. According to the draft the NOMINATION attribute is 24 bits long (so 3 bytes).

ibc: I am wrong. It looks that is 4bytes long even if the value is just 3 bytes. The draft spec sucks.

ibc:
Let's close this PR in favour of PR #756 in which I did some additions. Thanks a lot @penguinol

Issue #748, opened 2022-01-05
`reason` of C++ errors are not send to Node.js


piranna:
### Your environment

- Operating system: Ubuntu Linux 21.10
- Node version: v16.13.1
- npm version: 8.1.2
- gcc/clang version: 11.2.0-7ubuntu2
- mediasoup version: 3.9.3
- mediasoup-client version: (not applicable)

### Issue description

I'm calling to `AudioLevelObserver.removeProducer()` on Producer `close` event, so I'm getting an `Error:  [method:rtpObserver.removeProducer]` message and probably it's because the Producer is closed in C++ space and not accesible anymore, so an `Error` is being thrown at https://github.com/versatica/mediasoup/blob/34dcdcd157f8d6966d293e5b61272d1845688711/worker/src/RTC/Router.cpp#L528-L529, but not fully sure yet. Problem is, the `Producer not found` error message is not being printed anywhere (note the two spaces at `:  [` in the provided error message).

I've been investigating the C++ side source code trying to find the source of this problem, not sure if my conclussions are correct, please correct me if I'm wrong.

So, that error is thrown by the [MS_THROW_ERROR](https://github.com/versatica/mediasoup/blob/1e3bdac019f4507652581480cb6f9cdbdebbdb5f/worker/include/MediaSoupErrors.hpp#L29-L35) macro, that's using `std::snprintf()` function to write the `Producer not found` error description in a `thread_local` Exception variable, common to all `MediaSoupError` instances, that's used to define the exception `reason`.

This exception is being catch at https://github.com/versatica/mediasoup/blob/f0ebd8099a7bad12d55f9ff017a9a6b175e1820d/worker/src/Worker.cpp#L313-L316, that's throwing a new exception itself calling again to `MS_THROW_ERROR` macro to compose the error reason with the method name to know who was the one that throwed it, and later is finally catch and serialized as an error to the Node.js side at https://github.com/versatica/mediasoup/blob/f0ebd8099a7bad12d55f9ff017a9a6b175e1820d/worker/src/Channel/ChannelSocket.cpp#L251-L254.

[...]

PR #750, opened 2022-01-07
Fix bad printing of error messages from Worker

j1elo:
Directly passing the error buffer caused issues with the snprintf that MS_THROW_ERROR() does.

Possibly due to the fact that it might be overwriting the same memory buffer that is passed. This meant that incomplete messages would end up arriving to the client side.

## What was wrong

I was getting these logs from mediasoup:

```
mediasoup:Channel request() [method:transport.produce, id:5] +2ms
mediasoup:ERROR:Channel [pid:24259 RTC::Producer::Producer() | throwing MediaSoupTypeError: video/VP9 codec not supported for simulcast +0ms
mediasoup:ERROR:Channel [pid:24259 Worker::OnChannelRequest() | throwing MediaSoupTypeError: video/VP9 codec not supported for simulcastS[method:transport.produce] +0ms
mediasoup:WARN:Channel request failed [method:transport.produce, id:5]: S[method:transport.produce] +0ms
```

Note that I replaced the space with an `S` [here](https://github.com/versatica/mediasoup/blob/6df7e245b13d6cfe5f0ebd91aff2f009ec8950ff/worker/src/Worker.cpp#L311) (_only for my tests_), so it becomes more obvious that the original contents of the `buffer` were getting lost, _while exactly the new string is the only part that survives_.

And my call to `console.error("ERROR: " + error)` showed this:
```
ERROR: TypeError: S[method:transport.produce]
```

While I expected it to show the whole error string.

## How it's fixed

With this patch, the logs now look like this:

```
mediasoup:Channel request() [method:transport.produce, id:5] +1ms
mediasoup:ERROR:Channel [pid:27384 RTC::Producer::Producer() | throwing MediaSoupTypeError: video/VP9 codec not supported for simulcast +0ms
mediasoup:ERROR:Channel [pid:27384 Worker::OnChannelRequest() | throwing MediaSoupTypeError: video/VP9 codec not supported for simulcastS[method:transport.produce] +0ms
mediasoup:WARN:Channel request failed [method:transport.produce, id:5]: video/VP9 codec not supported for simulcastS[method:transport.produce] +0ms
```

And my call to `console.error("ERROR: " + error)` shows this:
```
ERROR: TypeError: video/VP9 codec not supported for simulcastS[method:transport.produce]
```

I.e. the complete error message is able to reach the Node.js client.

I've only tested the particular code path that happens in this case (a Worker error caused by attempting to use Simulcast with VP9) but it only makes sense that this memory overwriting error would happen everywhere there the `error.buffer` was passed directly instead of the class accessor method, `error.what()`.

### DIFF
--- a/worker/src/Worker.cpp
+++ b/worker/src/Worker.cpp
@@ -259,7 +259,7 @@ inline void Worker::OnChannelRequest(Channel::ChannelSocket* /*channel*/, Channe
 			}
 			catch (const MediaSoupError& error)
 			{
-				MS_THROW_ERROR("%s [method:%s]", error.buffer, request->method.c_str());
+				MS_THROW_ERROR("%s [method:%s]", error.what(), request->method.c_str());
 			}
 
 			auto* router = new RTC::Router(routerId);
@@ -283,7 +283,7 @@ inline void Worker::OnChannelRequest(Channel::ChannelSocket* /*channel*/, Channe
 			}
 			catch (const MediaSoupError& error)
 			{
-				MS_THROW_ERROR("%s [method:%s]", error.buffer, request->method.c_str());
+				MS_THROW_ERROR("%s [method:%s]", error.what(), request->method.c_str());
 			}
 
 			// Remove it from the map and delete it.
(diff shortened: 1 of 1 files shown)


ibc:
I fail to see how it never happened before to me but the fix makes sense. Just wondering: could you please run a grep looking for "error.buffer" and paste it here to see how many remaining occurrences should be fixed?

jmillan (approved):

ibc:
There are no more `error.buffer` so this is super fine. Merging, thanks.

ibc:
Released in 3.9.4

piranna:
Seems exactly like the same error I described in #748, thanks :-)

PR #756, opened 2022-01-13
ICE renomination support

### LINKED ISSUE #734: Renomination support

ibc:
PR based on https://github.com/versatica/mediasoup/pull/744 by @penguinol with some additions.

Fixes #734

References:
* https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination
* https://webrtc.googlesource.com/src/+/refs/heads/main/p2p/base/connection.cc#592

### DIFF
--- a/worker/src/RTC/IceServer.cpp
+++ b/worker/src/RTC/IceServer.cpp
@@ -344,7 +344,7 @@ namespace RTC
 					// Store the tuple.
 					auto* storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						MS_DEBUG_TAG(ice, "transition from state 'new' to 'completed'");
 
@@ -394,7 +394,7 @@ namespace RTC
 					// Store the tuple.
 					auto* storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						MS_DEBUG_TAG(ice, "transition from state 'disconnected' to 'completed'");
 
@@ -436,7 +436,7 @@ namespace RTC
 					if (!storedTuple)
 						storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						// Mark it as selected tuple.
 						SetSelectedTuple(storedTuple);
@@ -474,7 +474,7 @@ namespace RTC
 					if (!storedTuple)
 						storedTuple = AddTuple(tuple);
 
-					if(nomination > this->remoteNomination)
+					if (nomination > this->remoteNomination)
 					{
 						// Mark it as selected tuple.
 						SetSelectedTuple(storedTuple);
--- a/worker/src/RTC/IceServer.cpp
+++ b/worker/src/RTC/IceServer.cpp
@@ -198,12 +198,12 @@ namespace RTC
 
 				delete response;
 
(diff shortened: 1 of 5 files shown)


ibc:
@penguinol can you review, please? This is basically your PR with the additions that I mentioned in my review.

ibc:
@nazar-pc, mediasoup-rust CI is failing in this PR and I don't see any related reason, do I miss something? https://github.com/versatica/mediasoup/runs/4803528295?check_suite_focus=true

nazar-pc:
That is what PR from earlier today have fixed, try to merge `v3` or rebase on it

ibc:
Thanks, then I'll ignore it since this PR does not change things other than STUN related.

penguinol:
It seems there is some logic problems.
When ice-renomination is enabled, chrome will never send a packet with USE-CANDIDATE attr, it will only send packets with NOMINATION attr, and once we response the packet, the follow STUN packet will not has NOMINATION attr unless chrome wants to switch to another network.
So i think `if (!hasUseCandidate)` should be `if (!hasUsecandidate && !hasNomination)`
And we should switch to COMPLETE state once we received a STUN packet with USE-CANDIDATE or NOMINATION attr.

ibc:
I'll address later today. Thanks.

penguinol:
Here is a pcap file with nomination packet.
[nomination.zip](https://github.com/versatica/mediasoup/files/7864453/nomination.zip)

ibc:
Thanks @penguinol, feedback addressed here: https://github.com/versatica/mediasoup/pull/756/commits/e5777361991be2c5b6fc8fa71c9b7ccf6dbedfa4

ibc:
Hi @penguinol, do you approve it?

jmillan (approved):

ibc:
Merging because I'm pretty sure it's ok :)

penguinol:
I think it's ok now

worker/src/RTC/IceServer.cpp
```diff
@@ -375,17 +405,29 @@ namespace RTC
...
 					// Store the tuple.
 					auto* storedTuple = AddTuple(tuple);
 
-					// Mark it as selected tuple.
-					SetSelectedTuple(storedTuple);
-					// Update state.
-					this->state = IceState::COMPLETED;
-					// Notify the listener.
-					this->listener->OnIceServerCompleted(this);
+					if ((hasNomination && nomination > this->remoteNomination) || !hasNomination)
```
jmillan, line 411: Commenting here because I was re-reviewing this PR just know.

The condition previous to this `else` block assures that here `hasUseCandidate` is `true` and `hasNominamtion` is also `true`, so checking that again here is not needed, and makes it harder to read. I had to go to the previous block to make sure this conditions were repeated.

This pattern repeats over this PR.

ibc: Can you please comment this in the new ongoing PR? Otherwise I'm gonna get a bit crazy. I'll reply there, but ok, you are not right. In the else block it's not true that `hasUseCandidate` is true.

PR #761, opened 2022-01-26
Added support for RTP abs-capture-time header

oto313:
Based on other RTP headers, I added support also for http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time . This is useful for getting to know when exactly was RTP media frame captured.

We already talked about this modification: https://mediasoup.discourse.group/t/forwarding-ntp-timestamp-from-source/3427/3

I made some changes and I test it with wireshark that it passes through this RTP header. You told me that I should add test into TestRtpPacket.cpp file, But I did not make any changes to RtpPacket class.  So any additional test will be pointless. Should some other things be modificated?

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -12,7 +12,9 @@ jobs:
         os:
           - ubuntu-20.04
           - macos-10.15
+          - macos-11
           - windows-2019
+          - windows-2022
         rust:
           - stable
           # - nightly
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -35,6 +35,9 @@ jobs:
           - os: windows-2019
             cc: cl
             cxx: cl
+          - os: windows-2022
+            cc: cl
+            cxx: cl
         # A single Node.js version should be fine for C++.
         node:
           - 14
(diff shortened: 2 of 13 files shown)


ibc (approved):
Thanks.

ibc:
Running CI checks before merging.

ibc (changes requested):
clang linter complains: https://github.com/versatica/mediasoup/runs/4951301699?check_suite_focus=true

Please run `make format` in `worker/` folder, and ensure everything is ok by running `npm run lint` and `npm test`, and then push changes.

oto313:
Should be ok now.

nazar-pc:
Can you also make similar changes to Rust side, please?

oto313:
I tried to add support also for rust, but I never programmed in rust. In producer.rs I did not find any equivalent of MangleRtpPacket (like in Producer.cpp). So I dont know if the changes are sufficient. Also I did not tested it

ibc (approved):
Do you approve @nazar-pc?

ibc:
> I tried to add support also for rust, but I never programmed in rust. In producer.rs I did not find any equivalent of MangleRtpPacket (like in Producer.cpp). So I dont know if the changes are sufficient. Also I did not tested it

The Rust part is equivalent to the Node part. C++ code is just for the worker.

nazar-pc (approved):
Looks fine to me

nazar-pc:
Looks like tests need to be updated accordingly

ibc (changes requested):
Yup, tests failing. This is in both Node tests and Rust tests.

oto313:
Ok I fixed tests both node and rust (good to know that you must  run "npm run typescript:build" before running tests - my test was successful but with old code)

oto313:
Ok last error in rust: Failed to create Pipe transport: Response { reason: "uv_udp_bind() failed [transport:udp, ip:'127.0.0.1', port:60000]: permission denied. I dont know how to fix that.

nazar-pc:
I just restarted it. Port is hardcoded, maybe it happened to be used on that particular run (never happened before, but probable).

ibc:
Merging this. Thanks a lot.

ibc (approved):

ibc:
Released in mediasoup 3.9.6.

PR #764, opened 2022-02-07
Typing event emitters

unao:
This PR adds proper typing to event emitters for all classes documented in the official docs: https://mediasoup.org/documentation/v3/mediasoup/api/.

Listening to (and emitting) events is fully type checked.

The PR is compatible with current way of registering internal events (names starting with `@`).

Initially I wanted only to augment types, but finally very minor changes to `EnhancedEventEmitter` were introduced.

----

One mismatch between code and docs discovered is argument to `ActiveSpeakerObserver.on('dominantspeaker')`. According to docs it is `Producer` according to code `{ producer: Producer }`.


node/src/DataProducer.ts
```diff
@@ -48,9 +48,13 @@ export type DataProducerStat =
...
 export type DataProducerType = 'sctp' | 'direct';
 
+type ObserverEvents = {
+	close: [];
+}
+
 const logger = new Logger('DataProducer');
 
-export class DataProducer extends EnhancedEventEmitter
+export class DataProducer extends EnhancedEventEmitter<{ transportclose: [] }>
```
jmillan: No `Events` type constructed and used here?

node/src/Router.ts
```diff
@@ -104,9 +104,15 @@ type PipeTransportPair =
...
+type ObserverEvents = {
+	close: [];
+	newtransport: [Transport];
+	newrtpobserver: [RtpObserver];
+}
+
 const logger = new Logger('Router');
 
-export class Router extends EnhancedEventEmitter
+export class Router extends EnhancedEventEmitter<{ workerclose: [] }>
```
jmillan: `Events` type for consistency?

node/src/Worker.ts
```diff
@@ -176,7 +178,7 @@ const workerBin = process.env.MEDIASOUP_WORKER_BIN
 const logger = new Logger('Worker');
 const workerLogger = new Logger('Worker');
 
-export class Worker extends EnhancedEventEmitter
+export class Worker extends EnhancedEventEmitter<{ died: [Error]}>
```
jmillan: Same, defining `Events` makes it more consistent and allows for seamlessly adding new events in the future.

jmillan (commented):
Nice!

Added some comments

node/src/EnhancedEventEmitter.ts
```diff
@@ -47,4 +52,39 @@ export class EnhancedEventEmitter extends EventEmitter
...
 	}
+
+	on<K extends keyof E & string>(event: K, listener: (...args: E[K]) => void)
+	{
+		super.on(event, listener as (...args: any[]) => void);
+
+		return this;
+	}
+
+	off<K extends keyof E & string>(event: K, listener: (...args: E[K]) => void)
```
jmillan, line 63: IMHO there is no single place where we use `off()`

nazar-pc: But users of the library might want to unsubscribe from events

jmillan: That's right ⬆️ , and it does not do any harm, it's a mere alias. :+1:

node/src/Consumer.ts
```diff
@@ -154,9 +154,27 @@ export type ConsumerStat =
...
+type ObserverEvents = {
+	close: [];
+	pause: [];
+	resume: [];
+	score: [ConsumerScore];
+	layerschange: [ConsumerLayers?];
+	trace: [ConsumerTraceEventData];
+}
+
+type ConsumerEvents = Pick<ObserverEvents, 'score' | 'layerschange' | 'trace'>
```
ibc: Even if Consumer and Consumer.observer events sometimes match, can we repeat their definition instead of picking some of them from the other?

unao: fixed

node/src/EnhancedEventEmitter.ts
unao: For completeness I added `prepend(Once)Listener` as well.

Btw. there is `safeEmitAsPromise` which seems to be not used internally, but removing it would be a breaking change.

node/src/Worker.ts
unao: fixed

node/src/DataProducer.ts
unao: Agree: consistency is better than succinctness.

node/src/ActiveSpeakerObserver.ts
```diff
@@ -18,9 +19,18 @@ export interface ActiveSpeakerObserverActivity {
...
 }
 
+type ObserverEvents = RtpObserverEvents & {
+	dominantspeaker: [{ producer: Producer }];
+}
+
+type Events = {
+	routerclose: [];
+	dominantspeaker: [{ producer: Producer }];
+}
```
ibc: Some concerns here that affects all files and not just this one:

1. Can we please define first class events and then observer events?.
2. `routerclose` is not just an event in `ActiveSpeakerObserver` class. It's an event that all `RtpObserver` inherited classes.
3. The name of `RtpObserver` class in unfortunate since it also contains an `observer`. So events of the `observer` of `RtpObserver` should be named `RtpObserverObserverEvents`.
4. And events of the `RtpObserver` parent class should be named `RtpObserverEvents`.
5. Some classes (such as `Transport` and `RtpObserver` export they `XxxxxxEvents` and `XxxxxObserverEvents`. However, in child classes this PR defines `type Events` and `type ObserverEvents`. For consistency we should export these as well and prefix their name with the class name.
6. Not sure why ESLint doesn't complain but we mandate opening brackets to be in the next line.

So IMHO this should be as follows:

```suggestion
type ActiveSpeakerEvents = RtpObserverEvents &
{
	dominantspeaker: [{ producer: Producer }];
}

type ActiveSpeakerObserverEvents = RtpObserverObserverEvents &
{
	dominantspeaker: [{ producer: Producer }];
}
```

node/src/RtpObserver.ts
```diff
@@ -4,6 +4,18 @@ import { Channel } from './Channel';
...
+	close: [];
+	pause: [];
+	resume: [];
+	addproducer: [Producer];
+	removeproducer: [Producer];
+}
+
+type Events = {
+	routerclose: [];
+}
```
ibc:
```suggestion
export type RtpObserverEvents =
{
	routerclose: [];
}

export type RtpObserverObserverEvents =
{
	close: [];
	pause: [];
	resume: [];
	addproducer: [Producer];
	removeproducer: [Producer];
}
```

node/src/Transport.ts
```diff
@@ -81,9 +81,23 @@ export interface TransportTraceEventData
...
+
+type ObserverEvents = {
+	close: [];
+	newproducer: [Producer];
+	newconsumer: [Consumer];
+	newdataproducer: [DataProducer];
+	newdataconsumer: [DataConsumer];
+	trace: [TransportTraceEventData];
+
+}
```
ibc:
```suggestion
export type TransportObserverEvents =
{
	close: [];
	newproducer: [Producer];
	newconsumer: [Consumer];
	newdataproducer: [DataProducer];
	newdataconsumer: [DataConsumer];
	trace: [TransportTraceEventData];
}
```

node/src/Transport.ts
```diff
@@ -81,9 +81,23 @@ export interface TransportTraceEventData
 
 export type SctpState = 'new' | 'connecting' | 'connected' | 'failed' | 'closed';
 
+export type TransportBaseEvents = { routerclose: []; trace: [TransportTraceEventData] }
```
ibc: No need for "base" in the name:

```suggestion
export type TransportEvents = { routerclose: []; trace: [TransportTraceEventData] }
```

ibc (changes requested):
This is a great PR. I've requested some changes to get more consistency.

unao:
@ibc 
Updated the PR accordingly. I introduced bit more verbosity for the sake of decoupling.

> Not sure why ESLint doesn't complain but we mandate opening brackets to be in the next line.

According to their [docs](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/brace-style.md); you should have following rule to enable it for typescript:

```javascript
{
  // note you must disable the base rule as it can report incorrect errors
  "brace-style": "off",
  "@typescript-eslint/brace-style": ["error"]
}
```

Unfortunately (which I checked) it does not add support for `type`; only for _(...) enum, interface, namespace and module declarations_.

unao:
I opened [an issue for fixing the `brace-style` rule](https://github.com/typescript-eslint/typescript-eslint/issues/4537)

ibc:
I'm running CI checks on this PR. It looks good. I'll do my final review tomorrow and merge if no more changes are requested.

Thanks for this!

jmillan (approved):

ibc (approved):
Fantastic. Merging, THANKS

ibc:
Released in 3.9.7

PR #774, opened 2022-02-23
Export worker workflow artifacts

unao:
Having worker binaries it is much easier to start using mediasoup as one does not need to prepare the whole environment for building it. 

One caveat is that these artifacts expire in 30 days.

### DIFF
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -76,3 +76,9 @@ jobs:
       - run: npm run worker:build
       - run: npm run test:worker
         if: runner.os != 'Windows'
+
+      - name: Release artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: mediasoup-worker-${{ matrix.build.os }}-${{ matrix.build.cc }}
+          path: worker/out/Release


nazar-pc:
How would you use them though and why?

unao:
Maybe I am missing something important here, but having standalone ready to use binaries would be extremely useful for me and my project.

**how**
A path to a worker can be provided with `process.env.MEDIASOUP_WORKER_BIN`  https://github.com/versatica/mediasoup/blob/v3/node/src/Worker.ts#L178-L185

I tested it and it seems just fine. 

**why**

Running `npm install` will try to build the binaries for me. This process is environment and os specific, requires custom tooling present on local machine and takes substantial amount of time. I strongly prefer to run `npm install --ignore-scripts` as I do not trust `post-install` steps in general and always strive to have the setup as reproducible as possible.

Then for distributing my product I need to have standalone binaries anyway included as part of the distribution as building in the target hosts is not an option.

ibc:
So we are exporting here mediasoup-worker binaries that probably just can run on same OS as the CI machine?

nazar-pc:
Well, this is not a scenario supported and I don't think we want to support it, not yet for sure. Definitely not this way, that benefits your specific use cases and needs, not most users.

unao:
Oh I am not asking for official support and to marketing it as such.

I thought it would be nice to just export it from the GH actions and anyone who finds it and finds it useful could benefit from it.

What downsides do you see with this approach? Is the worker binary not standalone?

nazar-pc:
> Is the worker binary not standalone?

It is, but it is not a product of this project, it is an implementation detail. And it is not a static binary on most platforms, which means it will run on some distros and will not (or will crash in runtime) on others.

ibc:
I insist that we cannot export binaries that just work in machines with same OS and architecture than GH CI ones.

Said that we have (or used to have) a environment variable that makes "npm install mediasoup" NOT build the worker (useful in case the app is gonna be executed with that another env that tells it to use a binary somewhere else). I hope we have not lost that in the transition to Meson. Or maybe I'm wrong and it never existed since definitely we didn't document it in the Installation section in the website.

ibc:
This is and old friend feature request "provide binaries for X platforms". We don't want to do this because it means lot of work for us. And definitely we cannot go the way this PR proposed (which is just valid for machines that match OS and architecture of GH CI hosts).

I'm afraid we cannot accept this PR.

nazar-pc:
> I hope we have not lost that in the transition to Meson.

The feature is still there and used by some successfully.

ibc:
> The feature is still there and used by some successfully.

And can you tell me what the env was? 😀
I don't remember and I would like to document it.

ibc:
Ok, I see: `MEDIASOUP_WORKER_BIN` and it's documented in the `BUILDING.md` file in the repo. It's fine enough since there are more advanced options documented in there.

PR #777, opened 2022-02-27
Prebuild images

piranna:
This PR allows to create and use prebuild images of `mediasoup`, not needing to compile it at install on the target platform. This is done by compiling the Worker executables in advance for multiple platforms, and bundling them in the distributed package.

Prebuilds are being compiled using [Github Actions](https://github.com/features/actions). Currently it's being configured to be able to [launch the workflow by hand](https://docs.github.com/es/actions/managing-workflow-runs/manually-running-a-workflow), but also to do it when a new tag with the format `v*` (the same format used by `npm version` command) is being created. The tag-based dispatch has not been fully tested, so maybe it needs to be improved on the next release, but it's just a matter of configuration. Any suggestion on this topic is welcome.

To support the Worker executables for the multiple platforms inside the package, their location has been changed so their operating system and CPU architecture is included in their path, similar to how other similar solutions like [prebuildify](https://github.com/prebuild/prebuildify) works. In fact, I've used `prebuildify` as inspiration for this PR.

For simplicity, prebuilds are compiled using three different jobs running in parallel, one for each one of the operating systems supported by Github Actions (Ubuntu Linux, macOS and Windows), and [stored as artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). Alternatively I could have been able to use a build matrix so jobs steps are not duplicated, this can be a future improvement. Later, the artifacts with the Worker executables are being collected and placed in the path they would be located, if they would have been compiled at install time at their target platform. Finally, the package is published on Github Packages Registry or npmjs, depending of the provided credentials.

Currently Github Actions only support `x86_64` CPU architecture, so to compile for [Apple Silicon M1](https://en.wikipedia.org/wiki/Apple_M1) CPU, we would need to configure the job compiling for macOS to cross compile to `arm64` and have multiple builds, or generate an universal binary. This last solution is more elegant, but later we would need to support the selection of the universal binary when searching for the prebuilds, so it would make it more complex.

The other half is the install on the target platform. It just only checks what's the platform at install time, and check if there's a prebuild image for it. If it doesn't exists (like macOS on Apple Silicon M1, as said before), as a fallback it will compile the package as usual. Finally, at runtime checks for the running platform, and search again for Worker executable for that platform, being both a prebuild or a compiled one.


piranna:
By the way, this work has been sponsored by my employer, @dyte-in :-)

.github/workflows/mediasoup-pack.yaml
```diff
@@ -0,0 +1,125 @@
+name: mediasoup-pack
+
+on:
+  push:
+    tags:
+    - v*  # Only build for tags that match new versions of `npm` packages
```
nazar-pc, line 6: We're not using this format for tags, so it will definitely have to change.

.github/workflows/mediasoup-pack.yaml
```diff
@@ -0,0 +1,125 @@
...
+  # Manual dispatch, only for testing purposses
+  # See https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+  workflow_dispatch:
+
+jobs:
```
nazar-pc, line 11: These will not generally work unless you're on the same version of the same OS. Worker binary is not static in mediasoup. Unless you make it such, this approach will generate a lot of bug reports.

nazar-pc (commented):
Generally this is a viable approach for "shipping binaries" topic that was surfaced many times, but it is a bit more tricky than it seems initially.

Before any of this can be merged, either binaries need to be made fully static or we need to account for a particular OS version (namely, `ubuntu-20.04` is not necessarily the same as `alpine`).

piranna:
> Before any of this can be merged, either binaries need to be made fully static or we need to account for a particular OS version (namely, ubuntu-20.04 is not necessarily the same as alpine).

Good point. I don't know on Windows or macOS, but at least on Ubuntu if I understood correctly the output of `readelf`, there's no external dynamic linked libraries besides glibc, pthreads and alike, so there should not be problems. It's true that for better compatibility, instead of `ubuntu-latest` I've read we would need to use the oldest operating system (so it has the oldest glibc) possible. But it's true that there could be problems with musl based distros, so maybe we would use something like https://justine.lol/cosmopolitan/, what do you think?

.github/workflows/mediasoup-pack.yaml
piranna: No issues :-) My intention was to automate the process based on how `npm` works, but that can be changed.

nazar-pc:
> there's no external dynamic linked libraries besides glibc, pthreads and alike, so there should not be problems

Not every Linux is Ubuntu. There are different versions of glibc, there are distros without gibc at all (like Alpine Linux I already mentioned), so there will certainly be problems with current approach based on OS family.

> But it's true that there could be problems with musl based distros, so maybe we would use something like https://justine.lol/cosmopolitan/, what do you think?

Almost certainly not something we want to use and will likely require static builds anyway.

.github/workflows/mediasoup-pack.yaml
piranna: Could the solution of using the oldest OS work? What can be done regarding the different C libraries? I think there's no way to detect that on Node.js, although static builds (real ones, not glibc ones, that always has some dynamic dependencies on it) could work.

nazar-pc: I have seen binaries built on Windows Server 2019 (roughly Windows 10) not working on Windows 11, so you'll have to provide proof that it will work without issues.

Hartigan:
@piranna @nazar-pc I tried several times to run ```mediasoup``` on ```Alpine Linux```. It's unstable combination(```libuv``` + ```musl```). So, I think tests are needed for others ```libc``` version/implementations before use it.

.github/workflows/mediasoup-pack.yaml
piranna: So, how can we do that? Maybe using the oldest libc version is not enough? In theory, Windows is fully backwards binary compatible, in some cases for decades...

piranna:
> @piranna @nazar-pc I tried several times to run `mediasoup` on `Alpine Linux`. It's unstable combination(`libuv` + `musl`). So, I think tests are needed for others `libc` version/implementations before use it.

Maybe we can use https://www.npmjs.com/package/detect-libc to detect the actual libc being used, and multiple prebuilds also at libc level.

ibc:
I'll need some days before I can review all this.

piranna:
> I'll need some days before I can review all this.

Ok, no issues, take your time and thanks for notify us about that :-)

piranna:
By the way, I have this package published with already the prebuilds bundled in at https://github.com/dyte-in/mediasoup/packages/1275581, just only I changed the package name to includes my company org as scope to be able to publish it on Github Packages Registry. Except that change done in an independent commit, all the content is the same than the included in this PR.

piranna:
I've rebased my changes on top of `3.9.9` tag, and also cleaned a bit the commits about publish the packages.

piranna:
At https://github.com/dyte-in/mediasoup/packages/1275581?version=3.9.9 we have published an updated package. We have included a prebuild for Apple m1 in case you are interested, just only we needed to compile and add it ourselves by hand, we are thinking on ways to automate this until Github gets m1 runners.

snnz:
I believe there should be separate dedicated npm packages with the prebuild images. Either outside of this project, or at least that shouldn't mess up with the existing installation procedure.

piranna:
> I believe there should be separate dedicated npm packages with the prebuild images. Either outside of this project, or at least that shouldn't mess up with the existing installation procedure.

Why do you think so? I've done the install procedure of the prebuild images to be as smooth and transparent as possible, based on how [prebuildify](https://github.com/prebuild/prebuildify) works, that's the gold standard to have prebuild native Node.js modules in npm...

snnz:
> Why do you think so?

Because I do not want to download all these images every time I am installing or upgrading Mediasoup. And even less I want them to preceed the compilation from the sources. Just make a separate package for those who is willing to use the gold standard.

nazar-pc:
First of all, I agree this does seem to be a useful feature in general, but, this is not something we're ready to maintain in mediasoup itself right now.

You can already achieve similar result by creating a separate NPM package which will download pre-built binaries and instruct mediasoup package to not build anything and then use those binaries by specifying `MEDIASOUP_WORKER_BIN` environment variable with a path to pre-built binary from that package.

So thanks for contribution, but we'll not accept this right now.

ibc:
The idea behind `MEDIASOUP_WORKER_BIN` environment variable is that others can provide a prebuilt image to keep `npm install mediasoup` from compiling the worker from sources. How they do is up to them. However, if finally we have to produce those binaries and depend on GH capabilities to build binaries for each architecture, this is gonna be more maintenance work for us. As @nazar-pc said, a separate project makes much more sense IMHO. Thanks for you effort. I hope it can be used to build a separate package.

piranna:
I think maintenance is minimal, since it's fully automated, and can have multiple advantages for final users. How would it be done as a separate package to by-pass the actual install of this one, and use the other one instead without needing to change the code?

nazar-pc:
If `MEDIASOUP_WORKER_BIN` environment variable is set, installation will be skipped automatically:
https://github.com/versatica/mediasoup/blob/092ebee3b39605ef541b1f540697113ff8c401ba/npm-scripts.js#L119-L133

piranna:
> If `MEDIASOUP_WORKER_BIN` environment variable is set, installation will be skipped automatically:

I know, but that would only work if you already have your own bundle in advance. How to do it as a dependency? Not sure if we can trust the prebuild package to be run before `mediasoup` one and define an environment variable itself...

nazar-pc:
You need to have binaries before you use mediasoup, not before you install it. Thus no issue if it is downloaded as part of installation of another package.

piranna:
> You need to have binaries before you use mediasoup, not before you install it. Thus no issue if it is downloaded as part of installation of another package.

Not fully true... main problem is not about using custom builds or not, it's about install step is really slow because currently it's needed to compile mediasoup, that's what we are trying to bypass.

ibc:
> > If `MEDIASOUP_WORKER_BIN` environment variable is set, installation will be skipped automatically:
> 
> I know, but that would only work if you already have your own bundle in advance. How to do it as a dependency? Not sure if we can trust the prebuild package to be run before `mediasoup` one and define an environment variable itself...

In the worst case, the user should install that new "mediasoup-prebuild" package somewhere else and it may include all binaries or a command to get the desired one in the desired location. Then the user should set the `MEDIASOUP_WORKER_BIN` environment variable before running `npm install mediasoup` in the project and every time it runs `npm install`. How the user manages this and the env variable is up to him (in Docker this is easy). The thing is that we don't want to maintain per arch binaries just to avoid that work to the user.



> Not fully true... main problem is not about using custom builds or not, it's about install step is really slow because currently it's needed to compile mediasoup, that's what we are trying to bypass.

As said above, this command **will skip** worker compilation:

```bash
MEDIASOUP_WORKER_BIN=/home/foo/binaries/mediasoup-worker npm install mediasoup
```

ibc:
NOTE: Finally we are doing this taking into account all concerns that happened here: https://github.com/versatica/mediasoup/pull/1087

PR #778, opened 2022-02-28
Fix VP9 kSVC forwarding logic to not forward lower unneded layers

ggarber:
Description:

When using VP9 in kSVC model (current supported mode in WebRTC) the higher spatial layers are independent of the lower spatial layers and need to be filtered.  Otherwise we are consuming much more bandwidth tan needed (30%?).  

image
https://www.w3.org/TR/webrtc-svc/

To detect if it is a keyframe packet that needs to be forwarded even if it is a lower spatial layer this change checks the P bit in the VP9 header that distinguishes frames that are encoded independently of other frames (key frames).

Test Plan:
Ran producer and consumer locally and changed MaxOutgoingBitrate of the consumer to allocate different bitrates and check the video can always be decoded by the consumer.

### DIFF
--- a/worker/src/RTC/Codecs/VP9.cpp
+++ b/worker/src/RTC/Codecs/VP9.cpp
@@ -281,9 +281,19 @@ namespace RTC
 				}
 			}
 
-			// Filter spatial layers higher than current one (unless old packet).
-			if (packetSpatialLayer > tmpSpatialLayer && !isOldPacket)
-				return false;
+			// Unless old packet filter spatial layers that are either
+			// * higher than current one
+			// * different than the current one when KSVC is enabled and this is not a keyframe
+			// (interframe p bit = 1)
+			if (!isOldPacket)
+			{
+				if (
+				  packetSpatialLayer > tmpSpatialLayer ||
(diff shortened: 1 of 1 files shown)


worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -281,9 +281,19 @@ namespace RTC
...
-				return false;
+			// Unless old packet filter spatial layers that are either
+			// * higher than current one
+			// * different than the current one when KSVC is enabled and this is not a keyframe
+			// (interframe p bit = 1)
+			if (!isOldPacket)
+			{
+				if (
+				  packetSpatialLayer > tmpSpatialLayer ||
+				  (context->IsKSvc() && this->payloadDescriptor->p && packetSpatialLayer != tmpSpatialLayer))
```
ibc: Please merge both conditions into one. No reason for having a if inside another if.

ibc (changes requested):

ibc (approved):

ibc:
Will merge later today

jmillan (approved):

ibc:
Release in 3.9.8

PR #779, opened 2022-02-28
Fix update bandwidth estimation configuration and available bitrate when updating max outgoing bitrate

ggarber:
Description:

When updating the max outgoing bitrate we need to update it also in the webrtc bandwidth estimator configuration or it won't be applied until the desired bitrate changes.   Also the availableBitrate needs to be updated or the SFU will keep sending a bitrate larger than maxOutgoingBitrate for an undetermined amount of time.

Test Plan:
Ran producer and consumer locally with SVC/Simulcast enabled and configure different maxOutgoingBitrates checking that the forwarded layers change accordingly.


ibc (approved):

ibc:
Thanks. Will merge once @jmillan approves.

jmillan (approved):

ibc:
Release in 3.9.8

PR #782, opened 2022-03-01
Use more efficient collections

jmillan:
Extracted from `memory-optimizations` branch to make it more readable and concise.

### DIFF
--- a/worker/include/Channel/ChannelRequest.hpp
+++ b/worker/include/Channel/ChannelRequest.hpp
@@ -2,9 +2,9 @@
 #define MS_CHANNEL_REQUEST_HPP
 
 #include "common.hpp"
+#include <absl/container/flat_hash_map.h>
 #include <nlohmann/json.hpp>
 #include <string>
-#include <unordered_map>
 
 using json = nlohmann::json;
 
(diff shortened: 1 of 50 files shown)


ibc (approved):
If it works I assume this is better

nazar-pc:
I had first 5 commits from that PR extracted into https://github.com/nazar-pc/mediasoup/tree/more-efficient-data-structures

jmillan:
> I had first 5 commits from that PR extracted

IMHO the smaller and more specific PRs the better.  Otherwise reviewing them is too complicated. These two commits do just modify the containers. 

Do you suggest doing this any other way @nazar-pc ?

nazar-pc:
My branch just replaces more containers, that is the only difference. I have nothing against landing this one.

jmillan:
> My branch just replaces more containers, that is the only difference.

These 2 commits do not touch the logic. One of those 5 commits deletes unused unrelated code, another one touches the one byte RTP extensions code. Splitting them into different PRs will facilitate the review, make it easier adding a specific test if needed,  and git bisect in case there is need for troubleshoot in the future.

I would be grateful if you could create different PRs out of that branch, or I can also do it myself if you prefer.

nazar-pc (approved):

PR #784, opened 2022-03-03
Remove unused code

jmillan:
Extracted from `memory-optimizations` branch.


jmillan:
mmm, CI failing for Rust.

ibc:
Please no :(

ibc (commented):
Why are we removing this? It's also in UDP and Unix socket

jmillan:
> Please no :(

It's unused code. Is there any plan to use it?

nazar-pc:
I fixed Rust in the other PR. It was caused by use of unstable crates in examples and lack of lock file in the repo

nazar-pc (approved):

ibc:
> Why are we removing this? It's also in UDP and Unix socket 

What about this?

nazar-pc:
> What about this?

In that place it is actually used

ibc:
> > What about this?
> 
> 
> 
> In that place it is actually used

Sorry, what do you mean?

nazar-pc:
I mean the code was probably copy pasted from UDP into TCP. In UDP those methods are functional, in TCP they are just dead code.

ibc:
> I mean the code was probably copy pasted from UDP into TCP. In UDP those methods are functional, in TCP they are just dead code.

Are we really using the `std::string` arg based send method in UDP? I'm not aware of it.

nazar-pc:
> Are we really using the std::string arg based send method in UDP? I'm not aware of it.

I mean whatever the method looks like there. The point is the code removed by this PR is not used anywhere.

ibc (approved):
Sorry, my fault. Indeed we never use these tcp methods since we always write two datas due to the framing mechanism of RTC TCP.

PR #786, opened 2022-03-04
Optimize RTP header extension handling

jmillan:
Yet another extract from `memory-optimizations` branch.

Use `std::array` instead of a  map for one byte extensions.

### DIFF
--- a/worker/include/RTC/RtpPacket.hpp
+++ b/worker/include/RTC/RtpPacket.hpp
@@ -5,6 +5,7 @@
 #include "Utils.hpp"
 #include "RTC/Codecs/PayloadDescriptorHandler.hpp"
 #include <absl/container/flat_hash_map.h>
+#include <array>
 #include <nlohmann/json.hpp>
 #include <string>
 #include <vector>
@@ -477,7 +478,7 @@ namespace RTC
 					return false;
 
 				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
-				return this->mapOneByteExtensions[id - 1] != nullptr;
+				return this->oneByteExtensions[id - 1] != nullptr;
 			}
 			else if (HasTwoBytesExtensions())
 			{
@@ -514,7 +515,7 @@ namespace RTC
 					return nullptr;
 
 				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
-				auto* extension = this->mapOneByteExtensions[id - 1];
+				auto* extension = this->oneByteExtensions[id - 1];
 
 				if (!extension)
 					return nullptr;
@@ -616,9 +617,8 @@ namespace RTC
 		uint8_t* csrcList{ nullptr };
 		HeaderExtension* headerExtension{ nullptr };
 		// There might be up to one-byte header extensions
-		// (https://datatracker.ietf.org/doc/html/rfc5285#section-4.2), which is why we'll just allocate
-		// a static array for them
-		OneByteExtension* mapOneByteExtensions[14];
+		// (https://datatracker.ietf.org/doc/html/rfc5285#section-4.2), use std::array.
+		std::array<OneByteExtension*, 14> oneByteExtensions;
 		absl::flat_hash_map<uint8_t, TwoBytesExtension*> mapTwoBytesExtensions;
 		uint8_t midExtensionId{ 0u };
 		uint8_t ridExtensionId{ 0u };
--- a/worker/include/RTC/RtpPacket.hpp
+++ b/worker/include/RTC/RtpPacket.hpp
@@ -616,7 +616,7 @@ namespace RTC
 		Header* header{ nullptr };
 		uint8_t* csrcList{ nullptr };
 		HeaderExtension* headerExtension{ nullptr };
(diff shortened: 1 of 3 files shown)


worker/include/RTC/RtpPacket.hpp
```diff
@@ -473,9 +474,11 @@ namespace RTC
 			}
 			else if (HasOneByteExtensions())
 			{
-				auto it = this->mapOneByteExtensions.find(id);
+				if (id > 14)
+					return false;
 
-				return it != this->mapOneByteExtensions.end();
+				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
```
ibc:
```suggestion
				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14.
```

BTW what does `0..=13` notation men?

worker/include/RTC/RtpPacket.hpp
```diff
@@ -508,12 +511,14 @@ namespace RTC
...
 			else if (HasOneByteExtensions())
 			{
-				auto it = this->mapOneByteExtensions.find(id);
-
-				if (it == this->mapOneByteExtensions.end())
+				if (id > 14)
 					return nullptr;
 
-				auto* extension = it->second;
+				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
```
ibc: Same.

worker/include/RTC/SenderBandwidthEstimator.hpp
```diff
@@ -6,6 +6,7 @@
 #include "RTC/RateCalculator.hpp"
 #include "RTC/SeqManager.hpp"
 #include "RTC/TrendCalculator.hpp"
+#include <absl/container/btree_map.h>
```
ibc, line 9: Why is this included her if not used in this .hpp file? If just used in the .cpp file then it must be included there.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -491,8 +489,9 @@ namespace RTC
 				if (extension.id == 0 || extension.id > 14 || extension.len == 0 || extension.len > 16)
 					continue;
 
-				// Store the One-Byte extension element in the map.
-				this->mapOneByteExtensions[extension.id] = reinterpret_cast<OneByteExtension*>(ptr);
+				// Store the One-Byte extension element in an array.
+				// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
```
ibc: Same.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -575,12 +574,12 @@ namespace RTC
 		}
 		else if (HasOneByteExtensions())
 		{
-			auto it = this->mapOneByteExtensions.find(id);
+			// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
```
ibc: Same

worker/src/RTC/RtpPacket.cpp
```diff
@@ -885,8 +884,9 @@ namespace RTC
 						break;
 					}
 
-					// Store the One-Byte extension element in the map.
-					this->mapOneByteExtensions[id] = reinterpret_cast<OneByteExtension*>(ptr);
+					// Store the One-Byte extension element in an array.
+					// `-1` because we have 14 elements total 0..=13 and `id` is in the range 1..=14
```
ibc: Same

worker/include/RTC/SenderBandwidthEstimator.hpp
jmillan: It is used in the .hpp file, but it was taken from `RtpPacket.hpp` before.

ibc (approved):
I don really know how that lib works but since tests pass I assume it's ok

worker/include/RTC/RtpPacket.hpp
nazar-pc: This is Rust syntax for ranges, `=` before second number means that that number is included in the range. This is just to avoid ambiguity.

ibc: 👍🏼

nazar-pc (approved):
The first observation here is that for 1-byte extensions there is a limit of 14, so it is way cheaper to just allocate contiguous space for 14 bytes as a static array rather than using red-black tree-based map from standard library. This provides huge savings on trivial temporary heap allocations.

For two-byte header extensions just more efficient data structure is used.

worker/include/RTC/RtpPacket.hpp
```diff
@@ -611,8 +616,10 @@ namespace RTC
 		Header* header{ nullptr };
 		uint8_t* csrcList{ nullptr };
 		HeaderExtension* headerExtension{ nullptr };
-		absl::btree_map<uint8_t, OneByteExtension*> mapOneByteExtensions;
-		absl::btree_map<uint8_t, TwoBytesExtension*> mapTwoBytesExtensions;
+		// There might be up to one-byte header extensions
```
nazar-pc:
```suggestion
		// There might be up to 14 one-byte header extensions
```

PR #788, opened 2022-03-07
Fix frozen video due to double call to Consumer::UserOnTransportDisconnected()

ibc:
Rationale given here: https://github.com/versatica/mediasoup/pull/787#issuecomment-1060891193

We always check flags and return fast if the flag was already set (or unset, it depends), then we set or unset the flag and run code. Here we failed and didn't honor this pattern.


ibc:
If merged we must close PR #787.

ibc:
Published in 3.9.9

jmillan (commented):
:+1:

PR #792, opened 2022-03-12
Download and use MSYS/make locally for Windows postinstall

snnz:
This makes unnecessary to install `MSYS` manually and avoids the problem of the incorrect `make` versions in the `PATH`.


npm-scripts.js
```diff
@@ -120,13 +120,34 @@ switch (task)
 	{
 		if (!process.env.MEDIASOUP_WORKER_BIN)
 		{
+			if (isWindows)
```
nazar-pc: This is needed not only in `postinstall`, but every time worker is built, hence should likely be in `worker:build` instead, as well as in `test:worker`. Cleanup should stay here though.

worker/scripts/getmake.py
```diff
@@ -0,0 +1,15 @@
+import io, tarfile, urllib.request
+
+def get(url):
+  tar = tarfile.open(fileobj=io.BytesIO(urllib.request.urlopen(url).read()))
+  tar.extractall('worker/out/msys')
+  tar.close()
+
+get('https://sourceforge.net/projects/mingw/files/MSYS/Base/msys-core/msys-1.0.19-1/msysCORE-1.0.19-1-msys-1.0.19-bin.tar.xz/download')
```
nazar-pc: We need to verify hashes of these downloads, otherwise this is an attack vector or just something that will potentially break our CI in unexplainable way.

nazar-pc (changes requested):

worker/scripts/getmake.py
snnz: I've add a hash check.

worker/scripts/getmake.py
```diff
@@ -1,15 +1,17 @@
-import io, tarfile, urllib.request
+import io, hashlib, tarfile, urllib.request
 
-def get(url):
-  tar = tarfile.open(fileobj=io.BytesIO(urllib.request.urlopen(url).read()))
+def get(url, sha1):
```
nazar-pc: I'd strongly prefer sha256 though, sha1 shouldn't really be used for anything security-related anymore

npm-scripts.js
snnz: I've added a separate task `install-msys-make` for make installation (since I doubt that those who need to run `test:worker` or `worker:build` outside of `postinstall` would like to download everything each time they run the tasks.)

worker/scripts/getmake.py
snnz: No problem, sha1 is just what happens to be in the files info at Sourceforge. Changed the check to sha256.

npm-scripts.js
```diff
@@ -211,3 +213,23 @@ function execute(command)
...
+			console.log('Cannot find Python executable.');
+			process.exit(1);
+		}
+	}
+	execute(String(res.stdout).trim() + ' worker\\scripts\\getmake.py');
+}
+
+function addMsysToPath()
+{
+	process.env['PATH'] = process.cwd() + '\\worker\\out\\msys\\bin;' + process.env['PATH'];
```
nazar-pc: `process.cwd()` is probably not a good idea, since it can be called from anywhere. `__dirname` should be a better alternative

nazar-pc (commented):
Makes sense to me overall. Did you test it as a git dependency on Windows already?

npm-scripts.js
snnz: But the paths in all commands are relative to the cwd.

snnz:
I tested it from a local tar obtained by `npm pack`.

npm-scripts.js
nazar-pc: Hm... yes, other paths are implicitly relative to cwd, but that isn't necessarily what they should be.

nazar-pc:
You need to make linter happy

npm-scripts.js
snnz: Yet this is what they are. :slightly_smiling_face: Currently it looks like the script (with package.json) can be anywhere, but everything else should be in the current directory, rather than vice versa.

npm-scripts.js
```diff
@@ -17,9 +17,19 @@ const MAKE = process.env.MAKE || (isFreeBSD ? 'gmake' : 'make');
...
 {
-	addMsysToPath();
+	if (!fs.existsSync('worker/out/msys/bin/make.exe'))
+	{
+		installMsysMake();
+	}
+
+	const msysPath = `${process.cwd()}\\worker\\out\\msys\\bin`;
+
+	if (process.env['PATH'].indexOf(msysPath) < 0)
```
nazar-pc: Better this:
```suggestion
	if (!process.env['PATH'].includes(msysPath))
```

However I'm not sure the check is even necessary, it doesn't hurt even if included multiple times.

nazar-pc (approved):

nazar-pc (approved):

ibc (approved):

nazar-pc:
@snnz would you be so kind porting this to Rust as well, please?

snnz:
Unfortunately, I do not know Rust at all. But since it is not a big task, I've tried to write something similar to the Node version looking at the examples. Probably it is clumsy, but it works. However, I didn't manage to complete the installation, because it failed with an error "Filename longer than 260 characters". There are some really long names like "`subprojects/usrsctp-9d6b99b10a70f7a63d21cd80d03c353da9ac19d3/libusrsctp.a.p/C__Users_sergeyn_.cargo_registry_src_github.com-1ecc6299db9ec823_mediasoup-sys-0.3.3_subprojects_usrsctp-9d6b99b10a70f7a63d21cd80d03c353da9ac19d3_usrsctplib_netinet_sctp_cc_functions.c.obj`". How do people deal with this?

I also noticed that the script `cpu_cores.sh` tries to run `nproc` in MinGW environment and cannot find it. I wonder what MSYS package contains it (if any)? It is not installed with `msys-base`.

nazar-pc:
`cargo fmt` is necessary to run to fix formatting and let CI pass that step.

`nproc` must be present in one of the MinGW packages, I don't have access to Windows VM right now to check which one it came from.

`C__Users_sergeyn_.cargo_registry_src_github.com-1ecc6299db9ec823_mediasoup-sys-0.3.3_subprojects_usrsctp-9d6b99b10a70f7a63d21cd80d03c353da9ac19d3_usrsctplib_netinet_sctp_cc_functions.c.obj` is a ridiculous file name, I'll see what can be done about that.

snnz:
`cargo fmt` is done.

But as to the `nproc`, I believe, `NUMBER_OF_PROCESSORS` environment variable should be used instead in Windows as more reliable (`NUM_CORES=$NUMBER_OF_PROCESSORS` in `cpu_cores.sh`)

nazar-pc:
If it works in both native Windows and MinGW environment - sure, let's use that.

snnz:
I have added it to the PR, since whether `nproc` exists somewhere in MinGW or not, this subset of MSYS definitely does not install it.

nazar-pc (approved):

npm-scripts.js
```diff
@@ -190,3 +210,21 @@ function execute(command)
...
+{
+	let res = spawnSync('where', [ 'python3.exe' ]);
+
+	if (res.status !== 0)
+	{
+		res = spawnSync('where', [ 'python.exe' ]);
+		if (res.status !== 0)
+		{
+			// eslint-disable-next-line no-console
+			console.error('Cannot find Python executable.');
```
ibc: Can you verify if we add dot at the end of other console logs in this file and whether we add a blank line after it?

ibc (changes requested):

npm-scripts.js
snnz: This seems to be the only error report like this in the file. I've removed the point and added a blank line, just in case.

npm-scripts.js
```diff
@@ -190,3 +210,22 @@ function execute(command)
...
 }
+
+function installMsysMake()
+{
+	let res = spawnSync('where', [ 'python3.exe' ]);
+
+	if (res.status !== 0)
+	{
+		res = spawnSync('where', [ 'python.exe' ]);
+		if (res.status !== 0)
```
ibc:
```suggestion
		res = spawnSync('where', [ 'python.exe' ]);

		if (res.status !== 0)
```

npm-scripts.js
```diff
@@ -190,3 +210,22 @@ function execute(command)
...
+		{
+			// eslint-disable-next-line no-console
+			console.error('Cannot find Python executable');
```
ibc: Please honor the syntax of other console calls in this file.

```suggestion
			console.error('`npm-scripts.js [ERROR] cannot find Python executable');
```

npm-scripts.js
```diff
@@ -17,6 +17,21 @@ const MAKE = process.env.MAKE || (isFreeBSD ? 'gmake' : 'make');
...
+		installMsysMake();
+	}
+
+	const msysPath = `${process.cwd()}\\worker\\out\\msys\\bin`;
+
+	if (!process.env['PATH'].includes(msysPath))
+	{
+		process.env['PATH'] = `${msysPath};${process.env['PATH']}`;
+	}
+}
```
ibc: This block is being executed no matter which task is given as argument to `npm-scripts.js`. I don't think this is acceptable. Can we out this into a function and just call it by tasks that really require this?

ibc (changes requested):

npm-scripts.js
```diff
@@ -323,3 +343,23 @@ function executeCmd(command, exitOnError = true)
 		}
 	}
 }
+
+function installMsysMake()
+{
+	let res = spawnSync('where', [ 'python3.exe' ]);
```
ibc:
```suggestion
{
	console.log('npm-scripts.js [INFO] installMsysMake()');

	let res = spawnSync('where', [ 'python3.exe' ]);
```

ibc (commented):
Thanks a lot.
Is this ready? Tested?

ibc:
So? Is this ready?

snnz:
Yes, I suppose so. Github's ci runs without problems, and I also use it successfully to upgrade mediasoup in my dev system.

ibc:
Running GH CI here. Merging once it passes. Thanks!

npm-scripts.js
```diff
@@ -323,3 +343,25 @@ function executeCmd(command, exitOnError = true)
...
+
+	if (res.status !== 0)
+	{
+		res = spawnSync('where', [ 'python.exe' ]);
+
+		if (res.status !== 0)
+		{
+			// eslint-disable-next-line no-console
+			console.error('`npm-scripts.js [ERROR] cannot find Python executable');
```
ibc:
```suggestion
			console.error('`npm-scripts.js [ERROR] installMsysMake() cannot find Python executable');
```

ibc: Let me commit it.

ibc (approved):

ibc:
Released in 3.11.10. Thanks!

PR #804, opened 2022-03-30
Modify SimulcastConsumer to keep using layers without good scores

ggarber:
We have cases were some simulcast consumers get all the video layers disabled very often when the producers have a high amount of packet loss.    In those cases we see up to 5-6 transitions per minute from "no layers" <-> "first layer" and we end up with frozen video or black video depending on when it happens.

I think mediasoup doesn't need to have the logic to estimate the quality of the producer streams because of these reasons:
- In many use cases it can be better to have bad quality video that no video at all.
- The current logic is creating many transitions (5-6 constantly per minute in some of our user reports) that makes the experience not very stable.
- With this change an app listening for the 'layerschange' event from the consumer will know that if layers have changed is because of the bandwidth estimation available instead of for a combination of multiple factors (bandwidth estimation and score of the producer).   So it is easier to present that information in the UI.
- The current logic can be implemented by the app checking the producers scores if it is important.
- This way the logic is simpler.
- I haven't seen this logic in other SFUs like Janus/Signal/Jitsi.... (I could be wrong and it is not a strong reason per se, but just fyc).

If it is better to open this as an issue or forum post instead of a PR let me know.   Thx for taking a look.

### DIFF
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -13,7 +13,6 @@ namespace RTC
 {
 	/* Static. */
 
-	static constexpr uint8_t StreamGoodScore{ 5u };
 	static constexpr uint64_t StreamMinActiveMs{ 2000u };           // In ms.
 	static constexpr uint64_t BweDowngradeConservativeMs{ 10000u }; // In ms.
 	static constexpr uint64_t BweDowngradeMinActiveMs{ 8000u };     // In ms.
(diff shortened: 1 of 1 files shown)


jmillan (approved):
LGTM.

Makes sense to me.

ibc:
I assume this should be also changed in `SvcConsumer`, right?

ggarber:
@ibc This logic was only available in SimulcastConsumer and not SvcConsumer.   Probably another reason apart from the ones mentioned in the description to remove it from SimulcastConsumer.

ibc:
I think this is ok. Will merge and release today.

ibc:
Published in 3.9.11

PR #808, opened 2022-04-03
Meson dependencies updates

nazar-pc:
All of the wraps are upstreamed. hence `packagefiles` is no longer necessary.

One of the major upgrades is OpenSSL 3.0.2, I updated usage of deprecated functions to modern ones, which may be a breaking change for those who are not using bundled OpenSSL and customize it somehow downstream (which to be fair we never advertised).

libuv 1.44.1 is out, but has regressions (https://github.com/libuv/libuv/issues/3586), will wait for 1.42.2 and update it then.
abseil-cpp had an update as well, but wrap was broken in the process of update, waiting for https://github.com/mesonbuild/wrapdb/pull/368 to be merged before it can be updated.


ibc (approved):
Please add also an entry at the top of main CHANGELOG file within a WIP version

nazar-pc:
I think his will require switching to 3.10.0 due to significance of OpenSSL upgrade

PR #809, opened 2022-04-03
Rust SVC/Simulcast example

nazar-pc:
Simple example based on echo with Up/Down buttons that circles loops over range of layers between S0T0 and S2T2.

Credit for most of the work goes to @untaggable (in squashed commit, originally https://github.com/nazar-pc/mediasoup/pull/3), I just cleaned up history to make it easier to review and fixed a few things on top.

Reviewing commit by commit should make sense.


ibc (approved):

jmillan:
:+1:

PR #810, opened 2022-04-04
Handle soft errors gracefully

nazar-pc:
In Rust all of the communication with worker is heavily asynchronous and even multi-threaded, there is no direct way to guarantee order of messages except creating an explicit queue. This sometimes leads to situations where consumer is closed after transport (since everything is dropped and multiple close requests are racing for completion on worker with each other and might be delivered in any order) for instance.

Such situations produce unhelpful error messages that these changes will suppress and turn into success. If something tries to close consumer that doesn't exist on worker anyway, it is now treated as successful completion (similarly to idempotent HTTP method `DELETE`).

### DIFF
--- a/rust/src/messages.rs
+++ b/rust/src/messages.rs
@@ -90,10 +90,18 @@ pub(crate) struct DataConsumerInternal {
 pub(crate) trait Request: Debug + Serialize {
     type Response: DeserializeOwned;
 
+    /// Request method to call on worker.
     fn as_method(&self) -> &'static str;
+
+    /// Default response to return in case of soft error, such as channel already closed, entity
+    /// doesn't exist on worker during closing.
+    fn default_for_soft_error() -> Option<Self::Response> {
+        None
+    }
 }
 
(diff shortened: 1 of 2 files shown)


Hartigan:
@nazar-pc  ```Consumer::get_stats``` generates many errors too, even with ```Consumer::closed``` checks. Will all such ```consumer not found``` be fixed?

nazar-pc:
I don't think so. Racing during closing is something that user of the library doesn't control. The rest is user-controlled, so should be preventable. I don't recall immediately why producer isn't owned by consumer so it is not closed automatically, maybe I'll add that later. For something like `get_stats` there isn't really a reasonable default value to return either.

Having heavier typing for errors would be nice, but that might be a lot of work.

ibc (approved):

nazar-pc:
Merging as is for now, I'll see what other improvements can be done about similar situations

jmillan:
:+1:

PR #811, opened 2022-04-04
Fix marker bit not reseted after mangling in each consumer

ggarber:
The marker bit is set in some cases when processing the payload for VP9 SVC consumers and that bit is not reseted so a consumer affects to the next ones.

The effect is that some SVC consumers receive more frames with marker bit than expected and libwebrtc is not able to decode them properly.    From a user point of view the effect is having very low framerate because it is only able to decode the framerates and the browser keeps sending PLIs forever.

### DIFF
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -822,6 +822,8 @@ namespace RTC
 			this->keyFrameForTsOffsetRequested = false;
 		}
 
+		bool marker{ false };
+
 		if (shouldSwitchCurrentSpatialLayer)
 		{
 			// Update current spatial layer.
@@ -841,14 +843,14 @@ namespace RTC
 			EmitScore();
 
 			// Rewrite payload if needed.
-			packet->ProcessPayload(this->encodingContext.get());
+			packet->ProcessPayload(this->encodingContext.get(), marker);
 		}
 		else
 		{
 			auto previousTemporalLayer = this->encodingContext->GetCurrentTemporalLayer();
 
 			// Rewrite payload if needed. Drop packet if necessary.
-			if (!packet->ProcessPayload(this->encodingContext.get()))
+			if (!packet->ProcessPayload(this->encodingContext.get(), marker))
 			{
 				this->rtpSeqManager.Drop(packet->GetSequenceNumber());
 
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -581,7 +581,9 @@ namespace RTC
 		auto previousSpatialLayer  = this->encodingContext->GetCurrentSpatialLayer();
 		auto previousTemporalLayer = this->encodingContext->GetCurrentTemporalLayer();
 
-		if (!packet->ProcessPayload(this->encodingContext.get()))
+		bool marker{ false };
+		bool origMarker = packet->HasMarker();
+		if (!packet->ProcessPayload(this->encodingContext.get(), marker))
 		{
 			this->rtpSeqManager.Drop(packet->GetSequenceNumber());
 
@@ -611,6 +613,10 @@ namespace RTC
 		// Rewrite packet.
 		packet->SetSsrc(this->rtpParameters.encodings[0].ssrc);
 		packet->SetSequenceNumber(seq);
+		if (marker)
+		{
+			packet->SetMarker(true);
+		}
 
 		if (isSyncPacket)
 		{
@@ -649,6 +655,7 @@ namespace RTC
 		// Restore packet fields.
 		packet->SetSsrc(origSsrc);
 		packet->SetSequenceNumber(origSeq);
+		packet->SetMarker(origMarker);
 
 		// Restore the original payload if needed.
 		packet->RestorePayload();
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -583,6 +583,7 @@ namespace RTC
 
 		bool marker{ false };
 		bool origMarker = packet->HasMarker();
+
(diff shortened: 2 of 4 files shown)


worker/src/RTC/SvcConsumer.cpp
```diff
@@ -581,7 +581,9 @@ namespace RTC
 		auto previousSpatialLayer  = this->encodingContext->GetCurrentSpatialLayer();
 		auto previousTemporalLayer = this->encodingContext->GetCurrentTemporalLayer();
 
-		if (!packet->ProcessPayload(this->encodingContext.get()))
+		bool marker{ false };
+		bool origMarker = packet->HasMarker();
+		if (!packet->ProcessPayload(this->encodingContext.get(), marker))
```
ibc, line 587: Blank line after if please

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -611,6 +613,10 @@ namespace RTC
 		// Rewrite packet.
 		packet->SetSsrc(this->rtpParameters.encodings[0].ssrc);
 		packet->SetSequenceNumber(seq);
+		if (marker)
```
ibc, line 618: Same

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -822,6 +822,8 @@ namespace RTC
 			this->keyFrameForTsOffsetRequested = false;
 		}
 
+		bool marker{ false };
```
ibc, line 825: This is always false and it's never set to true, so it would be better if we call `packet->ProcessPayload(this->encodingContext.get(), /*marker*/ false);` in lines below, right?

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -581,7 +581,10 @@ namespace RTC
 		auto previousSpatialLayer  = this->encodingContext->GetCurrentSpatialLayer();
 		auto previousTemporalLayer = this->encodingContext->GetCurrentTemporalLayer();
 
-		if (!packet->ProcessPayload(this->encodingContext.get()))
+		bool marker{ false };
```
ibc, line 584: Same here.

worker/src/RTC/SimulcastConsumer.cpp
SteveMcFarlin: EDIT: VP8.cpp does not use the marker parameter, so yes this will always be false

worker/src/RTC/SvcConsumer.cpp
SteveMcFarlin: With VP9 the marker may be set in `packet->ProcessPayload(...)`

https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/Codecs/VP9.cpp#L353

worker/src/RTC/SimulcastConsumer.cpp
ggarber: @ibc  ProcessPayload needs a boolean& and not a boolean

ggarber:
@ibc How does the latest version work?

For your consideration for v4 I think the way mediasoup work now is:
```
Consumer::onPacket(rtp) {
    // Modify Packet (sequence numbers, hdrext....)
    // Make a copy of the packet
    // Encrypt and send packet
    // Reset packet
   }
}
```

But it could be simpler and avoid issues like the one in this PR if using this different approach:
```
Consumer::onPacket(rtp) {
    // Make a copy of the packet
    // Modify Packet (sequence numbers, hdrext....)
    // Encrypt and send packet
   }
}
```

worker/src/RTC/SimulcastConsumer.cpp
ibc: > EDIT: VP8.cpp does not use the marker parameter, so yes this will always be false

Simulcast is not just about VP8 (we also support H264) and SinulcastConsumer is supposed to be agnostic. I don't understand if this is then ok or not.

ibc: > @ibc  ProcessPayload needs a boolean& and not a boolean

Can you clarify what you mean please?

ibc:
The latter would make it more expensive since each pocket should be cloned for every consumer. Yes, we do it anyway in RtpSender now but there is an ongoing PR already that changes that.

ggarber:
What is that PR? I thought libsrtp was only supporting encryption in place.

ibc:
I mean the one in mediasoup about memory optimizations.

ggarber:
As far as I can see that PR doesn't modify SrtpSession.cpp that is where the additional memcpy happens.   That's the memcpy that could be moved to the Consumer to avoid modifications in the original packet buffer shared by different consumers as mentioned in the pseudocode above: https://github.com/versatica/mediasoup/pull/811#issuecomment-1101232232

Anyway it is a design decision you have made / have to make and this is just a suggestion.

ibc:
Not all the packets go to a SRTP endpoint so such a memcpy is not always needed. Anyway I fail to see why we are discussing this in the context of this PR.

ggarber:
> Not all the packets go to a SRTP endpoint so such a memcpy is not always needed. Anyway I fail to see why we are discussing this in the context of this PR.

Because I think that alternative design (make one single copy of the buffer at the beginning and avoid the need to reset changes done during the rtp processing) avoids bugs like the one fixed in this PR.      But feel free to ignore the suggestion.

worker/src/RTC/SimulcastConsumer.cpp
SteveMcFarlin: The function signature requires a reference to a boolean value. This value will be set by the implementation if a market bit is observed to be true. Not all codec operations will require using this bit. In the case of VP9/SVC we need to know if this bit is set.

jmillan (approved):
:+1:

ibc (approved):
Thanks, merging and releasing 3.9.10

PR #815, opened 2022-05-05
Fix DtlsTransport timeout related things

### LINKED ISSUE #814: (slightly) improved DTLS timeout handling
while looking at libwebrtc's DTLS stuff relating to timeouts last week (@lgrahl made me!) I spotted some minor things that could be improved in mediasoup (and Janus so tagging @lminiero):

https://github.com/versatica/mediasoup/blob/c559b95848a17a5ab070b040f5733132596ece1e/worker/src/RTC/DtlsTransport.cpp#L1002

- there is no reason to do that. 0 means "now".
I assume that currently this will just run into the next timeout current and still works.

https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/DtlsTransport.cpp#L1422
- if that returns an error, the connection shall fail. Typically that only happens after DTLS runs into the 12-retries timeout which takes several minutes.

ibc:
Fixes #814

CC @fippo

### DIFF
--- a/worker/src/RTC/DtlsTransport.cpp
+++ b/worker/src/RTC/DtlsTransport.cpp
@@ -988,17 +988,22 @@ namespace RTC
 		uv_timeval_t dtlsTimeout{ 0, 0 };
 		uint64_t timeoutMs;
 
-		// NOTE: If ret == 0 then ignore the value in dtlsTimeout.
 		// NOTE: No DTLSv_1_2_get_timeout() or DTLS_get_timeout() in OpenSSL 1.1.0-dev.
 		ret = DTLSv1_get_timeout(this->ssl, static_cast<void*>(&dtlsTimeout)); // NOLINT
 
 		if (ret == 0)
+		{
+			OnTimer(this->timer);
+
 			return true;
+		}
 
 		timeoutMs = (dtlsTimeout.tv_sec * static_cast<uint64_t>(1000)) + (dtlsTimeout.tv_usec / 1000);
 
 		if (timeoutMs == 0)
 		{
+			OnTimer(this->timer);
+
 			return true;
 		}
 		else if (timeoutMs < 30000)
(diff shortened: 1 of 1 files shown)


jmillan (approved):

ibc:
@lminiero just in case you are interested in this. It works. In fact `ret = 0` most of the times.

lminiero:
Thanks for the heads up, but I did make a similar change last week already: https://github.com/meetecho/janus-gateway/commit/b63e422972d3f48e51d0c426eb69508342b64b04
I also had a pre-existing hard timeout of 20s, so in my case DTLS may abort sooner than the 12-retries (which can happen with OpenSSL's more loose 1s based exponential backoff) when -1 is never returned instead.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -988,17 +988,22 @@ namespace RTC
...
+			OnTimer(this->timer);
+
 			return true;
+		}
 
 		timeoutMs = (dtlsTimeout.tv_sec * static_cast<uint64_t>(1000)) + (dtlsTimeout.tv_usec / 1000);
 
 		if (timeoutMs == 0)
 		{
+			OnTimer(this->timer);
```
jmillan, line 1005: Here we are calling `OnTimer()` explicitly without considering whether the timer is already armed.

`OnTimer()` may call `Reset()` in which case `this->ssl` get's invalidated. Then the legitimately armed timer expires and `OnTimer()` get's called again, calling `DTLSv1_handle_timeout(this->ssl);`  with the invalidated `this->ssl`

ibc: Don't we stop the timer in the Reset() method?

jmillan: well noted, yes.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -988,17 +988,22 @@ namespace RTC
 		uv_timeval_t dtlsTimeout{ 0, 0 };
 		uint64_t timeoutMs;
 
-		// NOTE: If ret == 0 then ignore the value in dtlsTimeout.
 		// NOTE: No DTLSv_1_2_get_timeout() or DTLS_get_timeout() in OpenSSL 1.1.0-dev.
 		ret = DTLSv1_get_timeout(this->ssl, static_cast<void*>(&dtlsTimeout)); // NOLINT
 
 		if (ret == 0)
+		{
+			OnTimer(this->timer);
```
jmillan, line 996: This is the only suspicious addition I see, considering that the doc says that when it returns 0 it means that there is no internal timer, so probably we should not be calling this.


```
/* DTLSv1_get_timeout queries the next DTLS handshake timeout. If there is a
--
  | * timeout in progress, it sets \|*out\| to the time remaining and returns one.
  | * Otherwise, it returns zero.
```

jmillan: BTW, janus also does not call `DTLSv1_handle_timeout` if   `DTLSv1_get_timeout` returns 0. It does as we did before this commit.

https://github.com/meetecho/janus-gateway/blob/master/src/dtls.c#L1088

PR #817, opened 2022-05-10
Do not discard TL0PICIDX from Temporal Layers higher than 0

jcague:
We saw scattered video artifacts when using VP8 Simulcast in a few cases. Most (if not all) cases happened when consumers were limited to the lower temporal layers and with packet losses. 

**How I reproduced the issue**
I reproduced the same issue consistently by introducing packet losses on the producer side and setting the temporal layer 0 to the consumer side. The packet losses in the producer result in unordered packets, and having the consumer tied to the TL0l quickly leads to the following sequence of packets:

```
Packet 1: TL: 0, PID: 0, TL0PICIDX: 0 --> **Forwarded** as TL:0, PID: 0, TL0PICIDX: 0
Packet 2: TL: 0, PID: 1, TL0PICIDX: 1 --> Packet lost before reaching Mediasoup
Packet 3: TL: 1, PID: 2, TL0PICIDX: 1 --> Packet dropped as it belongs to TL1, TL0PICIDX is marked as dropped.
Packet 2 (RTX): TL: 0, PID: 1, TL0PICIDX: 1 --> Packet dropped (TL0PICIDX was marked as dropped previously).
Packet 4: TL: 0, PID: 3, TL0PICIDX: 2 --> **Forwarded** as TL: 0, PID: 1, TL0PICIDX: 1
```
This scenario produces artifacts in the decoder as the following packets keep being decoded.

The expected behavior would be:
```

Packet 1: TL: 0, PID: 0, TL0PICIDX: 0 --> **Forwarded** as TL:0, PID: 0, TL0PICIDX: 0
Packet 2: TL: 0, PID: 1, TL0PICIDX: 1 --> Packet lost before reaching Mediasoup
Packet 3: TL: 1, PID: 2, TL0PICIDX: 1 --> Packet dropped as it belongs to TL1, TL0PICIDX is marked as dropped.
Packet 2 (rtx): TL: 0, PID: 1, TL0PICIDX: 1 --> **Forwarded** as TL:0, PID: 1, TL0PICIDX: 1
Packet 4: TL: 0, PID: 3, TL0PICIDX: 2 --> **Forwarded** as TL: 0, PID: 2, TL0PICIDX: 2
```
The unit test might help to understand the issue better.


worker/src/RTC/Codecs/VP8.cpp
```diff
@@ -340,7 +344,6 @@ namespace RTC
 			{
 				this->payloadDescriptor->Encode(data, pictureId, tl0PictureIndex);
 			}
-
```
ibc: Please don't remove this blank line.

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -210,3 +210,80 @@ SCENARIO("parse VP8 payload descriptor", "[codecs][vp8]")
...
+	uint16_t pictureId,
+	uint8_t tl0PictureIndex,
+	uint8_t tlIndex,
+	bool layerSync = true
+)
+{
+	uint16_t netPictureId = htons(pictureId);
+	std::memcpy(buffer + 2, &netPictureId, 2);
+	buffer[2] |= 0x80;
+
```
ibc: Remove this blank line

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -210,3 +210,80 @@ SCENARIO("parse VP8 payload descriptor", "[codecs][vp8]")
...
+	buffer[4] = tl0PictureIndex;
+	buffer[5] = tlIndex << 6;
+	if (layerSync)
+		buffer[5] |= 0x20; // y bit
+
+	auto* payloadDescriptor = Codecs::VP8::Parse(buffer, bufferLen);
+
+	REQUIRE(payloadDescriptor);
+	return payloadDescriptor;
```
ibc, line 235: Blank line before return.

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -210,3 +210,80 @@ SCENARIO("parse VP8 payload descriptor", "[codecs][vp8]")
...
+	bool marker;
+	auto* payloadDescriptor = CreatePacket(buffer, sizeof(buffer), pictureId, tl0PictureIndex, tlIndex, layerSync);
+	std::unique_ptr<Codecs::VP8::PayloadDescriptorHandler> payloadDescriptorHandler(
+		new Codecs::VP8::PayloadDescriptorHandler(payloadDescriptor));
+
+	if (payloadDescriptorHandler->Process(&context, buffer, marker))
+	{
+		return std::unique_ptr<Codecs::VP8::PayloadDescriptor>(Codecs::VP8::Parse(buffer, sizeof(buffer)));
+	}
+	return nullptr;
```
ibc, line 262: Same

ibc (changes requested):
Other than requested cosmetic changes, could you explain the rationale of these changes based on vp8 specs?

worker/src/RTC/Codecs/VP8.cpp
```diff
@@ -267,7 +267,10 @@ namespace RTC
 				if (this->payloadDescriptor->tlIndex > context->GetTargetTemporalLayer())
```
ggarber, line 267: Does anybody know why we check "GetTargetTemporalLayer()" here and in the next condition (5 lines below) we use "GetCurrentTemporalLayer()" instead?  Shouldn't both check for Current?

ibc: Ufff to hard to remember now but of course that's on purpose. Can we rephrase the question? Why would it be wrong as it is?

worker/test/src/RTC/Codecs/TestVP8.cpp
ibc: This blank line is here yet

worker/src/RTC/Codecs/VP8.cpp
ggarber: NVM, After rereading it I think the existing code is correct and I understand why those conditions use a different GetXXXXTemporalLayer().   Thx

jcague:
Thanks for the quick cosmetic review! 

According to the spec: [TL0PICIDX](https://datatracker.ietf.org/doc/html/rfc7741#section-4.2) is a running index for the temporal base layer frames, i.e., the frames with TID set to 0. TL0PICIDX MUST be incremented when TID is 0.

In the example I explained above, the producer encodes the VP8 stream with this sequence of frames for Temporal Layer 0:

```
Frame 1 (PID: 0, TL0PICIDX: 0, keyframe) ---> Frame 2 (PID: 1, TL0PICIDX: 1) --> Frame 3 (PID: 3, TL0PICIDX: 2)
```

But VP8.cpp drops the second packet (with the sequence I explain in the description or the unit test). The consumer will decode this sequence:

```
Frame 1 (PIC: 0, TL0PICIDX: 0, keyframe) --> Frame 3 (PID: 1, TL0PICIDX: 1)
```

Everything meets the specs, and the decoder "thinks" Frame 3 comes just after Frame 1 (PID, TL0PICIDX, and everything are ok). Thus, we make the consumer's decoder think there was no Frame 2, breaking the video decoding and generating video artifacts.

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -210,3 +210,85 @@ SCENARIO("parse VP8 payload descriptor", "[codecs][vp8]")
...
+  uint8_t tl0PictureIndex,
+  uint8_t tlIndex,
+  bool layerSync = true)
+{
+	uint16_t netPictureId = htons(pictureId);
+	std::memcpy(buffer + 2, &netPictureId, 2);
+	buffer[2] |= 0x80;
+	buffer[4] = tl0PictureIndex;
+	buffer[5] = tlIndex << 6;
+	if (layerSync)
```
ibc, line 228: Please blank line before if condition.

ibc (changes requested):

ibc:
> TL0PICIDX MUST be incremented when TID is 0.

Do you mean TID or PID?

ibc:
AFAIS this PR doesn't solve the same issue in case we are forwarding TL 0 and TL 1, we drop a TL 3 packet and later a retransmission of a TL 2 packet with same TL0PICIDX as the one in the dropped TL 3 packet arrives. Am I wrong?

ggarber:
> > TL0PICIDX MUST be incremented when TID is 0.
> 
> Do you mean TID or PID?

TID

ibc:
But the trace below that text displays PID rather than TID.

ggarber:
> AFAIS this PR doesn't solve the same issue in case we are forwarding TL 0 and TL 1, we drop a TL 3 packet and later a retransmission of a TL 2 packet with same TL0PICIDX as the one in the dropped TL 3 packet arrives. Am I wrong?

With this change from Javier when you drop that TL 3 packet it won't call `context->tl0PictureIndexManager.Drop()` so everything should work fine.   The tl0Idx won't be rewritten and that's correct because you are not dropping any packet in the temporal layer 0 that is the only one that should affect the tl0Idx.

jmillan (approved):
Good find!

ggarber:
> But the trace below that text displays PID rather than TID.

That trace is including just the frames for the base temporal layer (TID = 0) and I'm assuming that's why Javier is not including that information.

You can look at the full trace in the PR description.   It is the same example.

ibc:
I'll merge in a few minutes and release. THANKS

ibc (approved):

ibc (approved):

ibc:
3.9.13 released

jcague:
thanks a lot for reviewing and merging!

PR #823, opened 2022-05-17
Fix spatial layer switch with unordered packets

jcague:
I've seen sporadic corrupted frames when debugging VP8 Simulcast locally. I recorded a couple of cases here:
https://user-images.githubusercontent.com/557698/168805998-dada1fc0-a47b-4383-b46d-d0a0052fc342.mov
and here
https://user-images.githubusercontent.com/557698/168805983-95f2942c-fd92-4a6c-b89b-d2315ff0f41d.mov

In these cases I saw the same pattern:

We forward frames belonging to layer 0 following their sequence number, but there is a sequence number gap in the target spatial layer (due to packet losses or reorder) just before the keyframe that acts as the sync packet. In that case, if we receive the packets that were previously lost we will forward them to the consumer, and the browser will render a broken frame. 

The impact is quite limited as it usually ends up showing just a corrupted frame.

This is what happens to the traffic:

``` txt
<-- Current Spatial Layer 0, Target Spatial Layer 0
| SN 0   | SL 0 | M |    | --> forwarded as SN 0
| SN 100 | SL 1 | M |    | --> dropped (different SL)
| SN 1   | SL 0 | M |    | --> forwarded as SN 1
| SN 101 | SL 1 | M |    | --> lost
<-- Current Spatial Layer 0, Target Spatial Layer 1
| SN 2   | SL 0 |   | KF | --> forwarded as SN 2
| SN 102 | SL 1 |   | KF | --> forwarded as SN 4 (previous packet did not include marker)
| SN 101 | SL 1 |   |    | --> forwarded as SN 3 and this is usually decoded as it meets all requirements (SN, PID, etc.)

SN = Sequence Number, SL = Spatial Layer, KF = keyframe, and M = marker bit set
```

As a result, the browser will decode:

```raw
Frame from packet SN 0, SL0
Frame from packet SN 1, SL0
Frame from packet SN 2, SL0, start of a keyframe
       and packet SN 3, SL1, that was retransmitted and belonged to a different layer, possibly causing a corrupted frame
Frame from packet SN 4, SL1, keyframe
``` 

Here I propose a solution to avoid sending packets with sequence numbers lower than the first keyframe which was the sync packet.


worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -821,13 +822,31 @@ namespace RTC
 			this->keyFrameForTsOffsetRequested = false;
 		}
 
+		if (!shouldSwitchCurrentSpatialLayer && this->checkingForOldPacketsInSpatialLayer)
```
ggarber, line 827: Does this logic work in case the SimulcastConsumer has been inactive for a while?

I'm worried about this sequence:
- ShouldSwitchCurrentSpatialLayer is true so we set checkingForOldPacketsInSpatialLayer
- After that the consumer is disabled (maybe because not enough bandwidth) for a while
- The consumer is resumed and the first sequence number after that is snReferenceSpatialLayer - 1 (because there was a rollover of sequence numbers).

jcague: Yep, you were right. I'm now resetting `checkingForOldPacketsInSpatialLayer` to `false` when somebody call `UserOnResumed()`

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -823,13 +824,31 @@ namespace RTC
...
+		{
+			// If this is a packet previous to the spatial layer switch, ignore the packet
```
ibc: Please add dot at the end of the comments.

ibc:
@jmillan any other concern here? Ready to merge?

jmillan:
I'll check it tomorrow.

jmillan (approved):
:+1:

ibc (approved):

ibc:
Thanks. Released in mediasoup 3.9.16.

PR #833, opened 2022-06-13
TransportCongestionControlClient: allow setting maxOutgoingBitrate before created

### LINKED ISSUE #831: SetMaxOutgoingBitrate does not work if set before first consumer created
https://github.com/versatica/mediasoup/blob/f147f77efe29d2ef1fadec91b277025b47d0ee99/worker/src/RTC/Transport.cpp#L633-L646
According to current code, the max outgoing bitrate will be set only when tccClient is created.
And set max incoming bitrate has not this limitation
https://github.com/versatica/mediasoup/blob/f147f77efe29d2ef1fadec91b277025b47d0ee99/worker/src/RTC/Transport.cpp#L593-L617
I think it's because we want to validate the max bitrate in tccClient->SetMaxOutgoingBitrate.

ibc:
- Fixes #831
- Remove no longer used `RtpBufferSize` in `RtpPacket.hpp`.

### DIFF
--- a/worker/include/RTC/RtpPacket.hpp
+++ b/worker/include/RTC/RtpPacket.hpp
@@ -14,8 +14,6 @@ using json = nlohmann::json;
 
 namespace RTC
 {
-	// Max RTP length.
-	constexpr size_t RtpBufferSize{ 65536u };
 	// Max MTU size.
 	constexpr size_t MtuSize{ 1500u };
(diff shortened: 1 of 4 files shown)


ibc:
@ggarber any thoughts about this?

jmillan (approved):
:+1:

PR #834, opened 2022-06-17
WebRtcServer

ibc:
### Details

So instead of having each `WebRtcTransport` listening into its own local ports, here we introduce a new `WebRtcServer` class that listens in given protocol, IPs and ports and we extend the `router.createWebRtcTransport()` API so now it accepts a `webRtcServer` as option (if so, `listenIps` are ignored).

Usage example in mediasoup-demo project: https://github.com/versatica/mediasoup-demo/pull/110

### TODO

- Rust layer. Hi @nazar-pc :)


ibc:
Everything works without having to change anything at C++ level XDDD

CleanShot-2022-06-17-at-20 26 57@2x

CleanShot-2022-06-17-at-20 28 14@2x

ibc:
@nazar-pc Rust tests fail because worker.dump() now contains a new webRtcServerIds array.

nazar-pc:
I'll try to look at this during this weekend and will push necessary changes to the Rust side of things

ibc:
There is a bug somewhere. I'll explain tomorrow.

ibc:
And of course I know where the bug is because I introduced it right now...

ibc:
Fixed

ibc:
Ready for review.

worker/src/RTC/IceServer.cpp
```diff
@@ -564,6 +565,43 @@ namespace RTC
 		// Notify the listener.
 		this->listener->OnIceServerTupleAdded(this, storedTuple);
 
+		// Don't allow more than MaxTuples.
+		if (this->tuples.size() > MaxTuples)
+		{
+			MS_WARN_TAG(ice, "too many tuples, removing the oldest one that is not the selected one");
```
jmillan: Cosmetic:

""too many tuples, removing the oldest non selected one""

worker/src/RTC/IceServer.cpp
```diff
@@ -564,6 +565,43 @@ namespace RTC
...
+
+			// Find the older tuple which is neither the added one nor the selected
```
jmillan: Cosmetic: /older/oldest/

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 ### WIP
 
+* `WebRtcServer`: A new class that brings the ability to listen in a single UDP/TCP port for in `WebRtcTransports (PR #834).
```
jmillan: /listen in/listen on/

jmillan: /for in/for/

node/src/Transport.ts
```diff
@@ -385,6 +385,63 @@ export class Transport<Events extends TransportEvents = TransportEvents,
 		this.#observer.safeEmit('close');
 	}
 
+	/**
+	 * Just called in WebRtcTransports when closing the associated WebRtcServer.
+	 *
+	 * @private
+	 * @virtual
+	 */
+	mustClose(): void
```
jmillan: Comment:

`close()`, `routerClosed` and `mustClose()` methods are identical with the exception that `close` sends a `transport.close` request through the channel. There is place for some code reduction here. Ie:

All of three calling an internal private `clear( )` that clears Producers, Consumers, etc, etc.

If `reason === undefined` then send the 'transport.closed' request. Otherwise the code is identical for the three cases.

worker/include/RTC/TransportTuple.hpp
```diff
@@ -28,23 +28,43 @@ namespace RTC
...
 		{
+			GenerateHash();
 		}
 
 		explicit TransportTuple(const TransportTuple* tuple)
 		  : udpSocket(tuple->udpSocket), udpRemoteAddr(tuple->udpRemoteAddr),
 		    tcpConnection(tuple->tcpConnection), localAnnouncedIp(tuple->localAnnouncedIp),
 		    protocol(tuple->protocol)
 		{
+			this->hash = tuple->hash;
```
jmillan: Member initialisation as done with the rest? `hash(tuple->hash)`?

worker/src/RTC/IceServer.cpp
```diff
@@ -12,13 +12,40 @@ namespace RTC
...
+		// Notify the listener.
+		this->listener->OnIceServerLocalUsernameFragmentAdded(this, usernameFragment);
+	}
+
+	IceServer::~IceServer()
+	{
+		MS_TRACE();
+
+		// Here we must notify to the listener about the removal of all handled
+		// tuples and usernameFragment (including the old one if any).
```
jmillan: Utterly cosmetic: Order is comment is not the same as order in code:

* Comment: notify about the removal of *tuples* and *usernameFragment*
* Code: notifies *usernameFragment* first and *tuples*

jmillan (approved):
Absolutely perfect

ibc:
As Philipp commented in Twitter, we should document that the new WebRtcServer feature requires a long UDP/TCP send/write buffer (kernel stuff).

node/src/Transport.ts
ibc: If you carefully compare `close()` and `routerClosed()` you'll see that they are not that similar. The former emits many more events. I'd prefer to refactor this topic (if needed) in the future. Right now it's true that we have two methods that do **mostly** the same (but not even the same, see how `Transport.routerClosed()` emits 'routerclose' event while `Transport.mustClose()` doesn't, however `WebRtcTransport.mustClose()` calls parent `Transport.mustClose()` and then emits 'webrtcserverclose'). In any other cases this is not even true. Honestly I think that trying to unify those code fragments into a single method that receives arguments would make things harded to read.

node/src/Router.ts
```diff
@@ -345,6 +345,7 @@ export class Router extends EnhancedEventEmitter<RouterEvents>
 	 */
 	async createWebRtcTransport(
```
nazar-pc, line 351: I really dislike that everything is pushed into the same method. Maybe worth separating these into two slightly different methods just like they are separated on the worker side?

node/src/Worker.ts
```diff
@@ -561,6 +583,41 @@ export class Worker extends EnhancedEventEmitter<WorkerEvents>
 		await this.#channel.request('worker.updateSettings', undefined, reqData);
 	}
 
+	/**
```
nazar-pc, line 582: I understand why you did this, but API is a bit awkward. If this was on router level, then we could have nice abstraction
```
Worker -> Router -> WebRtcTransport
                 or WebRtcServer -> WebRtcTransport
```

And it wouldn't be necessary to push two different invariants of WebRtcTransport configurations into the same method on Router.

node/src/Worker.ts
```diff
@@ -467,13 +472,23 @@ export class Worker extends EnhancedEventEmitter<WorkerEvents>
...
 	 * @emits close
+	 * @emits newwebrtcserver: - (webRtcServer: WebRtcServer)
 	 * @emits newrouter - (router: Router)
 	 */
 	get observer(): EnhancedEventEmitter<WorkerObserverEvents>
 	{
 		return this.#observer;
 	}
 
+	/**
```
nazar-pc, line 479: Might be convenient in tests, but you can already store all instances returned by `createWebRtcServer` without this method and you have dump information too.

worker/include/RTC/TransportTuple.hpp
```diff
@@ -125,6 +130,28 @@ namespace RTC
 				return this->tcpConnection->GetSentBytes();
 		}
 
+	private:
+		void GenerateHash()
```
nazar-pc: Name doesn't reflect what is happening clearly:
```suggestion
		void UpdateOwnHash()
```

node/src/Transport.ts
nazar-pc: Hm, I understand why we have `close` and why `routerClosed` has a distinct use case of transport being closed involuntarily due to router above it closing. Should this one be called `webRtcServerClosed` then?

worker/include/RTC/TransportTuple.hpp
```diff
@@ -125,6 +130,28 @@ namespace RTC
...
+			std::string remoteIp;
+			uint16_t remotePort;
+
+			Utils::IP::GetAddressInfo(GetLocalAddress(), family, localIp, localPort);
+			Utils::IP::GetAddressInfo(GetRemoteAddress(), family, remoteIp, remotePort);
+
+			std::string id = remoteIp + std::to_string(remotePort) + localIp + std::to_string(localPort) +
+			                 std::to_string(family) + protocol;
+
+			this->hash = std::hash<std::string>{}(id);
```
nazar-pc: Does it need to be `std::string` or can be something non-heap allocated?
Also why not comparing socket addresses directly alongside with protocol?

nazar-pc (commented):
Didn't dig too much into C++ implementation, just some high-level architectural comments.

In Rust there is a clear dependency tree:
```
WorkerManager
  > Worker
    > Router
      > *Transport
        > Producer
        > Consumer
        > Data consumer
        > Data producer
      > *RtpObserver
```

Things down in dependency tree keep things up in dependency tree alive, meaning it is not possible to destroy Transport that has consumers or producers attached to it.

There is a special case with `pipe_to_router` where things are interlinked in less straightforward way to keep API nice and clean, but that is about it.

With this we add another special case with transports depending on Router and WebRtcServer at the same time. Which is not a big issue, I'm just wondering if that is a correct level in the hierarchy as opposed to adding one more path like this:
```
WorkerManager
  > Worker
    > Router
      > WebRtcServer
        > WebRtcTransport
```

node/src/Transport.ts
ibc: We cannot have a method called `webRtcServerClosed()` in the `Transport` parent class :(

I'll explain it better below in response to your feedback.

node/src/Router.ts
ibc: I wouldn't like to have these two methods:

* `router.createWebRtcTransport(options: WebRtcTransportOptions)`;
* `router.createWebRtcTransportWithServer(options: WebRtcTransportWithServerOptions)`;

It's an overkill. It's extremely easy to check whether `listenIps` or `webRtcServer` param is given in TS side. But if you prefer to split it into 2 methods in Rust side it's ok. In TypeScript there is no need.

node/src/Worker.ts
ibc: dump information returns the internal state of the C++ side. Here I'm checking the TS side. In fact we test both.

worker/include/RTC/TransportTuple.hpp
ibc: Why "update"? It's not updating the hash but generating it for first time (and it will never change).

worker/include/RTC/TransportTuple.hpp
ibc: We need a `key` to be used in C++ maps. In `WebRtcServer.hpp`:

```c++
// Map of WebRtcTransports indexed by local ICE usernameFragment.
absl::flat_hash_map<std::string, RTC::WebRtcTransport*> mapLocalIceUsernameFragmentWebRtcTransport;
// Map of WebRtcTransports indexed by TransportTuple.hash.
absl::flat_hash_map<size_t, RTC::WebRtcTransport*> mapTupleWebRtcTransport;
```

This is not about comparing tuples but about having an unique key/hash/id for each tuple so we can use `map.find(key)`.

Note that `TransportTuple.hash` is a `size_t`. Yes, it would be nicer if it could avoid having to generate an intermediate string but we don't know how to do it. @jmillan tried it hard and gave up.

To clarify: given a `TransportTuple` (which has protocol "udp" or "tcp", IP family, local IP, local port, remote IP and remote port) we have to generate a number that it's unique and deterministic for any `TransportTuple` with same internal information.

ibc:
> Didn't dig too much into C++ implementation, just some high-level architectural comments.
> 
> In Rust there is a clear dependency tree:
> 
> ```
> WorkerManager
>   > Worker
>     > Router
>       > *Transport
>         > Producer
>         > Consumer
>         > Data consumer
>         > Data producer
>       > *RtpObserver
> ```
> 
> Things down in dependency tree keep things up in dependency tree alive, meaning it is not possible to destroy Transport that has consumers or producers attached to it.
> 
> There is a special case with `pipe_to_router` where things are interlinked in less straightforward way to keep API nice and clean, but that is about it.
> 
> With this we add another special case with transports depending on Router and WebRtcServer at the same time. Which is not a big issue, I'm just wondering if that is a correct level in the hierarchy as opposed to adding one more path like this:
> 
> ```
> WorkerManager
>   > Worker
>     > Router
>       > WebRtcServer
>         > WebRtcTransport
> ```

There is N misunderstandings here:

- `WebRtcServer` belongs to a `Worker` rather than to a `Router`. If you run 15 "rooms" (each room has a `Router`) in your app you probably want that clients of all those "rooms" connect to the **same** port.
- Of course, if you run 4 `Workers` in your app, you cannot have the port listening for all them.

Regarding ownership: there is no issue. I admit that the TS layer is confusing but you should look at the C++ side:

- `WebRtcServer` does **not** manage `WebRtcTransports`, it just holds them. 
- The parent of a `WebRtcTransport` is still the `Router` as it was before.
- When use calls `webRtcServer.close()` in Node we have to close all `WebRtcTransports` that are using that `WebRtcServer`. Why? Because they become completely useless.
- The `~WebRtcServer()` destructor in C++ iterates all handled `WebRtcTransports` and call `MustClose()` on them.
- `WebRtcTransport.cpp` doesn't have any `MustClose()` method. It's directly implemented in the `Transport` parent.
- `Transport::MustClose()` just notifies its `Router` parent via `this->listener->OnTransportMustClose(this);`.
- `Router::OnTransportMustClose()` does **exactly the same** as it would do in case `CLOSE_TRANSPORT` IPC request is received from Node or Rust layer.
- This is: `WebRtcServer` does **not** close `WebRtcTransports` but ask them to die, and those ask their `Router` parent to do it nicely.

So this is definitely good at C++ side and doesn't break any hierarchy in mediasoup-worker. However I agree that it's not that way in the Node layer.

In Node layer:

- `WebRtcServer.close()` iterates all handles `WebRtcTransports` and call `mustClose()` on them.
- `WebRtcTransport.mustClose()` cleans some internal stuff, calls parent `super.mustClose()` and then emits 'webrtcserverclose' event.
- Parent `Transport.mustClose()` closes everything internally (without boring the C++ side since things are already clear in C++ side, which is a common guideline in mediasoup design) and emits `this.observer.emit("close")`. Nothing else.

Yes, it's not the same flow than in C++, it's an optimized one. The thing is that we cannot have a Node `Router` listening for any "webrtcserverclose" event in any **generic** `transport` since such an event wouldn't make any sense in a `PlainTransport`. Anyway I agree this could be better but I don't know how.

ibc:
I have an idea: we can add a generic "serverclose" event into parent `Transport` in Node layer. It would just happen in `WebRtcTransports` since other transports cannot be provided with any "server" but instead must listen into their own ports.

ibc:
@nazar-pc @jmillan following above rationale I've made things more generic:

https://github.com/versatica/mediasoup/pull/834/commits/1bce071d7eb9b9a61473760661219647f71db5e8

worker/include/RTC/TransportTuple.hpp
nazar-pc: `GenerateHash` has a name that implies that the hash will be returning. I found it confusing that it reads internal state and updated object's property.

nazar-pc:
> WebRtcServer belongs to a Worker rather than to a Router

I understand that, but it could belong to the router. Just like transports could have belonged to the worker, but the decision was made to have a router abstraction in front of transports for convenience.

> This is: WebRtcServer does not close WebRtcTransports but ask them to die, and those ask their Router parent to do it nicely.

Which we could have avoided if WebRtcServer belonged to router and then all WebRtcTransport instances belonged to it.

> The thing is that we cannot have a Node Router listening for any "webrtcserverclose" event in any generic transport since such an event wouldn't make any sense in a PlainTransport. Anyway I agree this could be better but I don't know how.

That is only an issue if you want router to close transports. If they belong to WebRtcServer then problem doesn't exist in the first place.

worker/include/RTC/TransportTuple.hpp
ibc: What about `SetHash()`?

nazar-pc: I expect `SetHash()` to take an argument with the hash, but it is a better name than `GenerateHash()`

nazar-pc:
What I'll probably do in Rust is making `WebRtcTransport` have two parents: router and optional server. Then when either closes, transport closes as well. In the first case `router_closed` callback is called, in second `webrtc_server_closed` callback is closed. I really don't see why server case is any special here. This is in case server stays where it is in the hierarchy.

ibc:
> > WebRtcServer belongs to a Worker rather than to a Router
> 
> I understand that, but it could belong to the router. Just like transports could have belonged to the worker, but the decision was made to have a router abstraction in front of transports for convenience.

It cannot belong to Router as explained above. You want to share the same listening port in all Routers that you create in a Worker.

> > This is: WebRtcServer does not close WebRtcTransports but ask them to die, and those ask their Router parent to do it nicely.
> 
> Which we could have avoided if WebRtcServer belonged to router and then all WebRtcTransport instances belonged to it.

All Transports belong to Router (PlainTransport, PipeTransport, DirectTransport and of course WebRtcTransport). It doesn't make sense that WebRtcTransports belong to WebRtcServer given that it's an optional feature.

 
> > The thing is that we cannot have a Node Router listening for any "webrtcserverclose" event in any generic transport since such an event wouldn't make any sense in a PlainTransport. Anyway I agree this could be better but I don't know how.
> 
> That is only an issue if you want router to close transports. If they belong to WebRtcServer then problem doesn't exist in the first place.

I hope the commit I referenced https://github.com/versatica/mediasoup/commit/1bce071d7eb9b9a61473760661219647f71db5e8 makes the design more consistent. Definitely a WebRtcServer is something separate from Routers. It's just given when creating a WebRtcTransport so tell it to not open its own ports but use those already open by the WebRtcServer. This should not imply any architectural/hierarchical change and current design honours it.

worker/include/RTC/TransportTuple.hpp
ibc: If `SetHash()` signature doesn't contain any argument it's clear that it sets something by its own :)
I'll change this.

nazar-pc:
> I hope the commit I referenced https://github.com/versatica/mediasoup/commit/1bce071d7eb9b9a61473760661219647f71db5e8 makes the design more consistent.

It does, but now I'm wondering why `listenServerClosed` should be on `Transport` and not just `WebRtcTransport` if only `WebRtcTransport` can have it? If you move it you can as well call it `webRtcServerClosed`.

ibc:
> What I'll probably do in Rust is making `WebRtcTransport` have two parents: router and optional server. Then when either closes, transport closes as well. In the first case `router_closed` callback is called, in second `webrtc_server_closed` callback is closed. I really don't see why server case is any special here. This is in case server stays where it is in the hierarchy.

I suggest doing what's done in C++ layer:

- WebRtcServer does **not** close WebRtcTransports.
- It nicely asks them to tell their Router parent to close them.

nazar-pc:
> WebRtcServer does not close WebRtcTransports

But why not? It certainly could because transport is useless when server is closed.

ibc:
> > I hope the commit I referenced [1bce071](https://github.com/versatica/mediasoup/commit/1bce071d7eb9b9a61473760661219647f71db5e8) makes the design more consistent.
> 
> It does, but now I'm wondering why `listenServerClosed` should be on `Transport` and not just `WebRtcTransport` if only `WebRtcTransport` can have it? If you move it you can as well call it `webRtcServerClosed`.

Because `this->listener` is `private` member in `Transport.hpp` and child classes don't have access to it by design. So this must be handled in a generic way.

Now, imagine that tomorrow we have a different transport or we enhance PlainTransport with the ability to share the same listening port (for example by matching the client via SRTP decrypt info). We would like to have a generic "listen server" API/events as we have now in this branch.

ibc:
> > WebRtcServer does not close WebRtcTransports
> 
> But why not? It certainly could because transport is useless when server is closed.

Because that would break an important guideline in mediasoup: just parents close children.

nazar-pc:
> child classes don't have access to it by design

There are friend classes in C++, so this should be achievable.

> Because that would break an important guideline in mediasoup: just parents close children.

I mean technically yes, but if they're doing on behalf of something else, is it really the case anymore?

ibc:
> > child classes don't have access to it by design
> 
> There are friend classes in C++, so this should be achievable.
> 
> > Because that would break an important guideline in mediasoup: just parents close children.
> 
> I mean technically yes, but if they're doing on behalf of something else, is it really the case anymore?

I already tried this and didn't like how it looks. If a WebRtcServer closes a WebRtcTransport then we would need a new callback in Router telling that the transport has been closed by someone else so it can clean its maps, etc. I just wouldn't like to introduce that flow.

But anyway, not sure what's the problem here. Parents close children. We keep that design. WebRtcServer asks children to be closed, children ask it to their parent, parent does it and cleans its maps of children and so on.

I don't understand why we should instead consider another different approach that breaks the "parent closes child" convention everywhere in the code.

To be honest, we may not even close transports when WebRtcServer is closed. We may just emit `transport.on("listenserverclose")` and let the app close the transport by itself. So things would be easier. But IMHO there is a benefit in being nice here since there is zero use cases for a transport whose WebRtcServer has been closed.

nazar-pc:
> transport has been closed by someone else

Isn't it sufficient to just know that transport has closed, without knowing why exactly? If that is the case, current `close` event is already sufficient for router to clean its maps.

ibc:
> > transport has been closed by someone else
> 
> Isn't it sufficient to just know that transport has closed, without knowing why exactly? If that is the case, current `close` event is already sufficient for router to clean its maps.

There is no "close" event in `Transport` telling its parent `Router` that it has been closed. It's never been that way. The router closes the transport so it never needs to listen to any "close" event. Of course I'm talking about C++ side.

nazar-pc:
> There is no "close" event in Transport telling its parent Router that it has been closed. It's never been that way. The router closes the transport so it never needs to listen to any "close" event. Of course I'm talking about C++ side.

Hm, that is strange then. We do have `transport.close` message, which I would have expected triggers closing of the transport and everything related to it cleans up automatically.

ibc:
> Hm, that is strange then. We do have `transport.close` message, which I would have expected triggers closing of the transport and everything related to it cleans up automatically.

Of course we have, but it's the `Router` the one that intercepts it and closes the transport. I insist, in C++ parent closes the child, always :)

CleanShot-2022-06-18-at-18 57 24@2x

nazar-pc:
> Of course we have, but it's the Router the one that intercepts it and closes the transport. I insist, in C++ parent closes the child, always :)

Hm, if it was like that before then I guess it is fine, but I do find C++ less elegant and quite convoluted comparing to TypeScript.

ibc:
Take into account that we don't expose any public C++ API. If we did then indeed we should do something more similar to what is done in TS.

fippo:
I wonder if it is possible to measure if recv_mmsg in libuv gets more efficient with this since the chance that you pull multiple packets in a single read increases. Maybe @saghul knows

worker/src/RTC/WebRtcServer.cpp
```diff
@@ -0,0 +1,517 @@
...
+		size_t colonPos = username.find(":");
+
+		// If no colon is found just return the whole USERNAME attribute anyway.
+		if (colonPos == std::string::npos)
+			return username;
+
+		return username.substr(0, colonPos);
+	}
+
+	inline void WebRtcServer::OnPacketReceived(RTC::TransportTuple* tuple, const uint8_t* data, size_t len)
```
fippo, line 349: You search for a ufrag in mapLocalIceUsernameFragmentWebRtcTransport first. Isn't it more efficient to search mapTupleWebRtcTransport for the tuple hash first and only fall back to stun (mainly for new connections) when that lookup fails?

ibc:
> I wonder if it is possible to measure if recv_mmsg in libuv gets more efficient with this since the chance that you pull multiple packets in a single read increases. Maybe @saghul knows

We already use it for long.

worker/src/RTC/WebRtcServer.cpp
ibc: Actually computing the tuple hash (see TransportTuple.hpp SetHash() method) is less efficient than doing a lookup of a string in a C++ map. Basically because in order to match the tuple we must also do lookup of the computed tuple.hash in a map.

fippo: Did you look into something specialized like libre's [sa_hash](https://github.com/baresip/re/blob/62f755ce05bd2597f75009a2a6ff4baa2719e568/src/sa/sa.c#L507)? Which might also allow using ordered map.

fippo:
> We already use it for long.

I know but if you have more packets coming into the single port you get more efficiency here which would be :tada:

worker/src/RTC/WebRtcServer.cpp
ibc: @jmillan 👆🏻👆🏻👆🏻👆🏻👆🏻👆🏻🌽

jmillan: yes, that's what we're looking for

jmillan:
> I know but if you have more packets coming into the single port you get more efficiency here which would be 🎉

That's my understanding too.

worker/src/RTC/WebRtcServer.cpp
ibc: @jmillan take into account that such `sa_hash` function just reads a `struct sa` while we have to read two (local and remote). So we should use `uint64_t` instead of `uint32_t` and "concatenate" the resulting numbers:

```
// Pseudo code:
uint64_t hash = (localSaHash << 32) + remoteSaHash);
```

ibc: Working on https://github.com/versatica/mediasoup/pull/839

ibc:
MERGING

nazar-pc:
Working on Rust support for this and wondering if `port` should be optional in `WebRtcServerListenInfo`. I think it would make sense to be able to allocate port from the pool of ports available to the worker just like in case of individual transports.

Robokishan:
This is most amazing thing @ibc Thanks for mediasoup :). This is totally magic. as soon as I saw this and I did bring down my another ec2 workers. and it is very cost effective

truanguyenvan:
hi @all, is there any way we can use webRtcServer in pipeTransport?

nazar-pc:
Questions on the forum, don't abuse this PR

truanguyenvan:
> 
oh, so sorry. thank @nazar-pc

PR #836, opened 2022-06-18
Improve EnhancedEventEmitter

ibc:
- Do not allow any event name starting with '@' (those were not TS checked).
- Instead, properly include private `@xxxx` events in `XxxxEvents` types.
- Add missing `EventEmitter` methods.
- Remove useless `@emits xxxx` notation in classes.


jmillan (approved):

lgrahl:
Correct me if I'm wrong but it seems this introduced a bug: `removeAllListeners` now requires an `eventName`. However, it is useful to remove all listeners from all events in some scenarios.

ibc:
> Correct me if I'm wrong but it seems this introduced a bug: `removeAllListeners` now requires an `eventName`. However, it is useful to remove all listeners from all events in some scenarios.

I've fixed it in v3 branch right now.

lgrahl:
Awesome. Faster than light. :zap:

PR #838, opened 2022-06-19
Upgrade Node.js

nazar-pc:
Node.js 12 is EOL for some time now and Node.js 18 is both current and will become LTS in a few months.

No code changes to intentionally break older versions yet, but doesn't make sense to test on older versions in CI either.


ibc (approved):
Thanks. I noticed this when CI failed yesterday due to usage of webRtcServer?.id.

ibc:
Merge please

PR #840, opened 2022-06-22
RtpStreamSend: Memory optimizations

jmillan:
Essentially extracted from https://github.com/versatica/mediasoup/pull/675

* RtpStreamSend does not clone each RTP packet coming from Producer.
  - A `std::shared_ptr` holding the original packet is used by all RtpStreamSend
   instances.

* RtpStreamSend does not clone each RTP packet when sending DTX.
  - The original RTP packet is DTX encoded/decoded when needed.

* RtpStreamSend does not store up to 65536 RTP packets in a `std::map`.
  - Only a limited range of RTP packets are stored in a `std::deque` (up to 2000 ms of media).

Memory usage differences using mediasoup-demo with 6 attendees in a room:

* v3: 60MB
* rtpstreamsend-optimizations: 17.5MB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* `RtpStreamSend`: Memory optimizations. Extracted from #675.
+
+
 ### 3.10.0
 
 * `WebRtcServer`: A new class that brings to `WebRtcTransports` the ability to listen on a single UDP/TCP port (PR #834).
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -10,6 +10,10 @@ namespace RTC
 	class RtpStreamSend : public RTC::RtpStream
 	{
 	public:
+		// Don't retransmit packets older than this (ms).
+		static constexpr uint32_t MaxRetransmissionDelay{ 2000 };
+
+	public:
 		class Listener : public RTC::RtpStream::Listener
 		{
 		public:
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -20,6 +20,7 @@ namespace RTC
 	public:
 		struct StorageItem
 		{
+			// TODO: Remove after testing.
 			void Dump() const;
 			void Reset();
 
@@ -54,6 +55,7 @@ namespace RTC
 			void Insert(uint16_t seq, StorageItem* storageItem);
 			void RemoveFirst();
 			void Clear();
+			// TODO: Remove after testing.
 			void Dump();
 
 		private:
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -11,9 +11,9 @@ namespace RTC
 	{
 	public:
 		// Minimum retransmission buffer size (ms).
-		constexpr static uint32_t MinRetransmissionDelay{ 200 };
+		const static uint32_t MinRetransmissionDelay;
 		// Maximum retransmission buffer size (ms).
-		constexpr static uint32_t MaxRetransmissionDelay{ 2000 };
+		const static uint32_t MaxRetransmissionDelay;
 
 	public:
 		class Listener : public RTC::RtpStream::Listener
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -38,9 +38,11 @@ namespace RTC
 		};
 
 	private:
-		// Special container that can store `StorageItem*` elements addressable by their `uint16_t`
-		// sequence number, while only taking as little memory as necessary to store the range between
-		// minimum and maximum sequence number instead all 65536 potential elements.
+		// Special container that stores `StorageItem*` elements addressable by
+		// their `uint16_t` sequence number, while only taking as little memory as
+		// necessary to store the range covering a maximum of
+		// `MaxRetransmissionDelay` milliseconds.
+		// NOTE: See `MaxRetransmissionDelay` in implementation file.
 		class StorageItemBuffer
 		{
 		public:
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -26,8 +26,6 @@ namespace RTC
 	public:
 		struct StorageItem
 		{
-			// TODO: Remove after testing.
-			void Dump() const;
 			void Reset();
 
 			// Original packet.
@@ -61,8 +59,6 @@ namespace RTC
 			void Insert(uint16_t seq, StorageItem* storageItem);
 			void RemoveFirst();
 			void Clear();
-			// TODO: Remove after testing.
-			void Dump();
 
 		private:
 			uint16_t startSeq{ 0 };
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -49,7 +49,7 @@ namespace RTC
 			StorageItem* GetFirst() const;
 			StorageItem* Get(uint16_t seq) const;
 			size_t GetBufferSize() const;
-			bool Insert(uint16_t seq, StorageItem* storageItem);
+			void Insert(uint16_t seq, StorageItem* storageItem);
 			void RemoveFirst();
 			void Clear();
 			void Dump();
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -21,6 +21,7 @@ namespace RTC
 		struct StorageItem
 		{
 			void Dump() const;
+			void Reset();
 
 			// Original packet.
 			std::shared_ptr<RTC::RtpPacket> packet{ nullptr };
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -47,7 +47,6 @@ namespace RTC
 		// their `uint16_t` sequence number, while only taking as little memory as
 		// necessary to store the range covering a maximum of
 		// `MaxRetransmissionDelay` milliseconds.
-		// NOTE: See `MaxRetransmissionDelay` in implementation file.
 		class StorageItemBuffer
 		{
 		public:
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -10,7 +10,9 @@ namespace RTC
 	class RtpStreamSend : public RTC::RtpStream
 	{
 	public:
-		// Don't retransmit packets older than this (ms).
+		// Minimum retransmission buffer size (ms).
+		static constexpr uint32_t MinRetransmissionDelay{ 200 };
+		// Maximum retransmission buffer size (ms).
 		static constexpr uint32_t MaxRetransmissionDelay{ 2000 };
 
 	public:
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -11,9 +11,9 @@ namespace RTC
 	{
 	public:
 		// Minimum retransmission buffer size (ms).
-		static constexpr uint32_t MinRetransmissionDelay{ 200 };
+		constexpr static uint32_t MinRetransmissionDelay{ 200 };
 		// Maximum retransmission buffer size (ms).
-		static constexpr uint32_t MaxRetransmissionDelay{ 2000 };
+		constexpr static uint32_t MaxRetransmissionDelay{ 2000 };
 
 	public:
 		class Listener : public RTC::RtpStream::Listener
--- a/worker/include/RTC/RtpStreamSend.hpp
+++ b/worker/include/RTC/RtpStreamSend.hpp
@@ -3,7 +3,7 @@
 
 #include "RTC/RateCalculator.hpp"
 #include "RTC/RtpStream.hpp"
-#include <vector>
+#include <deque>
 
 namespace RTC
 {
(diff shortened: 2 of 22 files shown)


jmillan:
@ibc, @nazar-pc,

https://github.com/versatica/mediasoup/pull/675 is extremely big and difficult to review. This is the Nth split I make out of that PR. Let's never do such big PR again with many different kind of changes, it's too time consuming and it takes too much time to get them merged.

AFAIK, after this only the object pool stuff would be missing.

Please, do dedicate some time reviewing. I've tested it and the memory savings big.

I have made some extra enhancements such as not cloning the RTP packet not even for sending DTX.

jmillan:
Also @nazar-pc , removed the `useNack` argument from `RtpStreamSend` constructor since it's present in the already provided `params` argument.

jmillan:
Note: I'll write more tests.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -14,30 +14,174 @@ namespace RTC
...
+		}
+
+		if (RTC::SeqManager<uint16_t>::IsSeqHigherThan(seq, this->startSeq))
+		{
+			auto idx{ static_cast<uint16_t>(seq - this->startSeq) };
+
+			// Packet arrived out of order, so we already have a slot allocated for it.
+			if (idx <= static_cast<uint16_t>(this->buffer.size() - 1))
+			{
+				MS_ASSERT(this->buffer[idx] == nullptr, "Must insert into empty slot");
```
ibc, line 77: Line break after assert

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -14,30 +14,174 @@ namespace RTC
...
+			for (uint16_t i{ 1 }; i < addToBack; ++i)
+				this->buffer.push_back(nullptr);
+
+			this->buffer.push_back(storageItem);
+		}
+		else
+		{
+			// Calculate how many elements would it be necessary to add when pushing new item
+			// to the front of the deque.
+			auto addToFront{ static_cast<uint16_t>(this->startSeq - seq) };
```
ibc: is addToFront an instance method? If so it must be AddToFrom

jmillan: It's a `uint16_t` variable.

jmillan: Initialized with `=` instead.

worker/src/RTC/PipeConsumer.cpp
```diff
@@ -573,9 +573,7 @@ namespace RTC
 				}
 			}
 
-			// Create a RtpStreamSend for sending a single media stream.
-			size_t bufferSize = params.useNack ? 600u : 0u;
-			auto* rtpStream   = new RTC::RtpStreamSend(this, params, bufferSize);
+			auto* rtpStream = new RTC::RtpStreamSend(this, params, this->rtpParameters.mid);
```
ibc, line 576: Why are we passing MID here? are we taking into consideration that MID could be empty (not given, not used)?

worker/src/RTC/Router.cpp
```diff
@@ -782,6 +782,8 @@ namespace RTC
 
 		auto& consumers = this->mapProducerConsumers.at(producer);
 
+		std::shared_ptr<RTC::RtpPacket> shared(packet);
```
ibc: Can we rename `shared` to `sharedPacket`?

worker/test/src/RTC/TestRtpStreamSend.cpp
```diff
@@ -26,85 +51,58 @@ SCENARIO("NACK and RTP packets retransmission", "[rtp][rtcp]")
...
 		RtpStream::Params params;
 
-		params.ssrc      = packet1->GetSsrc();
+		params.ssrc      = 1111;
 		params.clockRate = 90000;
 		params.useNack   = true;
 
-		// Create a RtpStreamSend.
-		RtpStreamSend* stream = new RtpStreamSend(&testRtpStreamListener, params, 4);
+		std::string mid{ "" };
```
ibc: We don't need to initialize a string to "", it's that by default plus we are not consistent here (in other files we don't initialize it this way).

worker/test/src/RTC/TestRtpStreamSend.cpp
```diff
@@ -117,37 +115,196 @@ SCENARIO("NACK and RTP packets retransmission", "[rtp][rtcp]")
...
-		delete packet4;
-		delete packet5;
-		delete stream;
+		RtpStream::Params params1;
+
+		params1.ssrc      = 1111;
+		params1.clockRate = 90000;
+		params1.useNack   = true;
+
+		std::string mid{ "" };
```
ibc: Same

worker/test/src/RTC/TestRtpStreamSend.cpp
```diff
@@ -117,37 +115,196 @@ SCENARIO("NACK and RTP packets retransmission", "[rtp][rtcp]")
...
+		// Create a RtpStreamSend instance.
+		TestRtpStreamListener testRtpStreamListener1;
+
+		RtpStream::Params params1;
+
+		params1.ssrc      = 1111;
+		params1.clockRate = clockRate;
+		params1.useNack   = true;
+
+		std::string mid{ "" };
```
ibc: Same

worker/test/src/RTC/TestRtpStreamSend.cpp
```diff
@@ -117,37 +115,196 @@ SCENARIO("NACK and RTP packets retransmission", "[rtp][rtcp]")
...
```
ibc: Same

worker/src/RTC/PipeConsumer.cpp
ibc: Ok, I understand it now. Resolved.

jmillan: > Why are we passing MID here

For retransmissions.

> are we taking into consideration that MID could be empty (not given, not used)?

Yes

jmillan:
@ibc, @nazar-pc,

If you can think of any other test to be added, let me know.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -14,30 +14,141 @@ namespace RTC
...
+	}
+
+	RtpStreamSend::StorageItem* RtpStreamSend::StorageItemBuffer::GetFirst() const
+	{
+		return this->Get(this->startSeq);
+	}
+
+	RtpStreamSend::StorageItem* RtpStreamSend::StorageItemBuffer::Get(uint16_t seq) const
+	{
+		auto idx{ static_cast<uint16_t>(seq - this->startSeq) };
```
ibc, line 49: Could this produce a negative value that would then be cashed to a false positive?

ibc (commented):
Review not done (I'm mobile only now) so just a basic question: where are RTP packets stored? Let me guess: each received packet is wrapped into a shared pointer then given to consumers and those store the pointer in their RtpStreamSend, so the packet will exist in memory until the las consumer holding it gets rid of its reference to it, right?

worker/src/RTC/RtpStreamSend.cpp
jmillan: Handled. We now check is `seq` is lower than` this->startSeq` and return `nullptr` in that case.

jmillan:
> Review not done (I'm mobile only now) so just a basic question: where are RTP packets stored? Let me guess: each received packet is wrapped into a shared pointer then given to consumers and those store the pointer in their RtpStreamSend, so the packet will exist in memory until the las consumer holding it gets rid of its reference to it, right?

Yes. In essence `Packet::Clone()` is not used in the codebase at the moment. The same packet buffer used to store the original incoming `RtpPacket` is the one used to deliver such packet to every remote `Consumer`.

ibc (approved):
LGTM, great work.

BTW all tests assume default `sendNackDelayMs = 0` (which is not the value we use then in the real code). Indeed it's hard to test with a value != 0.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -14,30 +14,144 @@ namespace RTC
...
 
 	/* Instance methods. */
 
 	RtpStreamSend::RtpStreamSend(
-	  RTC::RtpStreamSend::Listener* listener, RTC::RtpStream::Params& params, size_t bufferSize)
-	  : RTC::RtpStream::RtpStream(listener, params, 10), buffer(bufferSize > 0 ? 65536 : 0, nullptr),
-	    storage(bufferSize)
+	  RTC::RtpStreamSend::Listener* listener, RTC::RtpStream::Params& params, std::string& mid)
+	  : RTC::RtpStream::RtpStream(listener, params, 10), mid(mid),
+	    retransmissionBufferSize(MaxRetransmissionDelay)
```
ibc:
```suggestion
	    retransmissionBufferSize(RtpStreamSend::MaxRetransmissionDelay)
```

?

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -177,7 +297,17 @@ namespace RTC
...
 			this->hasRtt = true;
+		}
+
+		// Smoothly change retransmission buffer size towards RTT + 100ms, but not more than
+		// `MaxRetransmissionDelay`.
+		auto newRetransmissionBufferSize = static_cast<uint32_t>(this->rtt + 100.0);
+		auto avgRetransmissionBufferSize =
+		  (this->retransmissionBufferSize * 7 + newRetransmissionBufferSize) / 8;
+		this->retransmissionBufferSize = std::max(
+		  std::min(avgRetransmissionBufferSize, MaxRetransmissionDelay), MinRetransmissionDelay);
```
ibc:
```suggestion
		  std::min(avgRetransmissionBufferSize, RtpStreamSend::MaxRetransmissionDelay), RtpStreamSend::MinRetransmissionDelay);
```

jmillan:
@nazar-pc, could you please look at the Rust CI error?

jmillan:
> @nazar-pc, could you please look at the Rust CI error?

I'm checking

jmillan:
Rust is failing to link `MaxRetransmissionDelay` and `MinRetransmissionDelay` in the GH workflows.

`cargo test` is properly working for me locally. Any idea?

Screenshot 2022-06-24 at 21 47 19

worker/include/RTC/RtpStreamSend.hpp
```diff
@@ -3,12 +3,18 @@
...
-#include <vector>
+#include <deque>
 
 namespace RTC
 {
 	class RtpStreamSend : public RTC::RtpStream
 	{
+	public:
+		// Minimum retransmission buffer size (ms).
+		constexpr static uint32_t MinRetransmissionDelay{ 200 };
```
ibc: 200u

worker/include/RTC/RtpStreamSend.hpp
```diff
@@ -3,12 +3,18 @@
...
+		// Maximum retransmission buffer size (ms).
+		constexpr static uint32_t MaxRetransmissionDelay{ 2000 };
```
ibc: 2000u

ibc:
OMG YES. static members initialization must be done in cpp files.

jmillan:
> OMG YES. static members initialization must be done in cpp files.

Not required for numeric variables as per the spec. Indeed `cargo test` was working in localhost.

ibc:
Prove that we initialize any static class numeric variable the way it was done in this PR before.

jmillan:
> Prove that we initialize any static class numeric variable the way it was done in this PR before.

I didn't say otherwise.

jmillan:
Great problem I've found with the approach in this PR.

A `shared_ptr` is created in `Router` out of the original `RtpPacket`. The `payload` buffer of it points to a memory pointer of `ReadBuffer` in `UdpSocketHandler` or the corresponding in `TcpConnectionHandler` if we are using TCP.

This means that such payload pointer will get invalidated once the next packet arrives.

This is the reason why the original PR did a `RtpPacket::Clone()` on the first `RtpStreamSend`. Instead of that I'll clone the packet in `Router` , before passing it to the `Consumer` s, so they have an `RtpPacket` with it's own memory buffer that can be used safely in the future whenever a retransmission is needed.

ibc:
> A `shared_ptr` is created in `Router` out of the original `RtpPacket`. The `payload` buffer of it points to a memory pointer of `ReadBuffer` in `UdpSocketHandler` or the corresponding in `TcpConnectionHandler` if we are using TCP.
> 
> This means that such payload pointer will get invalidated once the next packet arrives.

Indeed. Just wondering, didn't this bug produce terrible issues already? It shouldn't be hard to reproduce AFAIU.

> This is the reason why the original PR did a `RtpPacket::Clone()` on the first `RtpStreamSend`. Instead of that I'll clone the packet in `Router` , before passing it to the `Consumer` s, so they have an `RtpPacket` with it's own memory buffer that can be used safely in the future whenever a retransmission is needed.

Why cloning it in Router before passing it to Consumers? It should be cloned when generated in the Producer, right? or even in RtcStreamRecv.

jmillan:
> Why cloning it in Router before passing it to Consumers? It should be cloned when generated in the Producer, right? or even in RtcStreamRecv.

If there is no `Consumer` there is even no need to clone the packet.

ibc:
We use RtpObservers as well. Please clone before.

jmillan:
> We use RtpObservers as well. Please clone before.

And do they need to access the RtpPacket once after this libuv iteration?

They process the current packet and don't need to access the packet in the future.

jmillan:
We can clone the packet before, it's just not needed if there are no Consumers.

ibc:
> > We use RtpObservers as well. Please clone before.
> 
> 
> 
> And do they need to access the RtpPacket once after this libuv iteration?
> 
> 
> 
> They process the current packet and don't need to access the packet in the future.

This is true for now. Will it be forever?

jmillan:
> This is true for now. Will it be forever?

`v3` does only clone in `RtpStreamSend`, thus only if `Consumer`s exists.

I've set an explicit comment in `Router`, before cloning:

```
			// Clone the packet so it holds its own buffer, usable for future
			// retransmissions.
```

If you have a preference to always  clone a `RtpPacket` now I won't oppose. We just can save those cycles when no needed.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -704,6 +711,10 @@ namespace RTC
 		packet->frameMarkingExtensionId      = this->frameMarkingExtensionId;
 		packet->ssrcAudioLevelExtensionId    = this->ssrcAudioLevelExtensionId;
 		packet->videoOrientationExtensionId  = this->videoOrientationExtensionId;
+		// Clone payload descriptor handler.
```
ibc: This is not a clone. Is it safe by just copying it? It's not a pointer so this is ok right? And why didn't we do it before?

jmillan: It was a `unique_ptr` before, it is a `shared_ptr` now. It's safe.

Today there is no place in the code where we need the `PayloadDescriptorHandler`  in cloned packets, but if we did it would fail. I consider this a fix.

jmillan: I'm double checking.

jmillan: Correct, we don't hold any reference in `PayloadDescirptorHandler` to the original buffer, we are OK.

jmillan: I should have added a comment here before since this is unrelated to this PR. Yet it's a fix IMHO.

ibc: But comment should not say "clone" right?

ibc (approved):
Other than pending comment this is great

worker/src/RTC/RtpPacket.cpp
jmillan: Very true. Comment fixed.

nazar-pc:
:tada:

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* `RtpStreamSend`: Memory optimizations. Extracted from #675.
```
ibc, line 6: This should reference the exact PR as in other items below.

jmillan: Done

PR #844, opened 2022-06-24
Increase DTLS key size

nazar-pc:
I was getting `error:0A00018F:SSL routines::ee key too small` error when running tests on Ubuntu 22.04 and turns out in this version OpenSSL config (that library reads) changed security parameters and doesn't consider 1024-bit key secure anymore.

So I have re-generated 4096-bit key so it doesn't complain and tests pass again.

https://github.com/debauchee/barrier/issues/126#issuecomment-505933083 is where I got this hint.


ibc (approved):

ibc:
Should we document this in the worker dtlsXxxxx options?

nazar-pc:
That on some operating systems depending on OpenSSL configuration they might need larger key size? I mean I can add that, but not sure if it is necessary.

ibc:
This would just happen if folks pass a 1024 bits cert which is unusual nowadays so I think we are good.

jmillan (approved):

PR #845, opened 2022-06-26
Add `WebRtcServer` support for Rust

nazar-pc:
A bit different API, but overall almost the same as TypeScript


nazar-pc:
Interestingly, listening on `127.0.0.1:PORT` and then `0.0.0.0:PORT` doesn't fail on Windows as expected (added Node runs in CI for macOS and Windows because of that).

Had to change the tests to make it fail consistently across all platforms.

ibc:
Driving long now so I'll review tonight. BTW it's true that we should allow random port in WebRtcServer. If you want to make such a change in this PR I'll do in Node side hopefully tonight.

nazar-pc:
I'd prefer to land this first as it works, then land changes that make port optional

ibc:
Changed my mind. We have a design problem with worker port range. This should be different but it's too late for it in v3. Typically it's hard to open (make public) such a huge range of ports, however in mediasoup we may need to provide the worker with a huge range of ports for INTERNAL (local host) usage. Unfortunately we just have a global port range. So we may need to provide mediasoup worker with a port range like 10000-12000 for internal usage while we just need to expose ports 5000-5031 for public WebRTC servers. So in most common deployment scenarios it doesn't make sense to have WebRTC servers using random ports within the ports range given to the worker and hence I suggest we don't make the WebRTC server port optional.

rust/examples/multiopus.rs
```diff
@@ -225,7 +225,7 @@ impl EchoConnection {
         // this at a different time and/or in different order.
         let consumer_transport = router
             .create_webrtc_transport(WebRtcTransportOptions::new(TransportListenIps::new(
-                TransportListenIp {
+                ListenIp {
```
ibc, line 228: Why has this changed?

rust/src/messages.rs
```diff
@@ -39,11 +40,22 @@ pub(crate) struct RouterInternal {
...
+    #[serde(rename = "webRtcServerId")]
+    pub(crate) webrtc_server_id: WebRtcServerId,
+}
+
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub(crate) struct TransportInternal {
     pub(crate) router_id: RouterId,
     pub(crate) transport_id: TransportId,
+    // TODO: This field should have really been in the `data`, not in `internal` data structure
```
ibc, line 55: Why? When we call consume() the producerId is also included in internal object rather than in data. Everything that references internal/public ids is always given in the internal object.

nazar-pc: `internal` was always the thing that is included in every single request and is never optional, `webRtcServerId` seemed like something that only matters during creation of the transport and later can be inferred. There are no other methods except creation that require `webRtcServerId` field to be specified and for such cases we were putting one-time use fields into the `data` field.

So every single method before seemed to follow the same convention, while this one is an outlier. On that note, I think `internal` should be replaced completely with just an ID of corresponding entity (for transports it'll be transport ID). The reason for that is that worker already knows all the relationships and sending all the fields every time is a API that can potentially be used incorrectly, you can send one set of values in one method call and the other in the other method call, ideally we wouldn't even have an option of such possibility, making internal API more misuse-resistant.

rust/examples/multiopus.rs
nazar-pc: `TransportListenIp` was previously only used in `WebRtcTransport` for specifying which IPs to listen on, now we have a similar use case in `WebRtcServer`, so even though TypeScript uses `ip` and `announcedIp` as separate fields in a `WebRtcServerListenInfo` data structure, I decided to reuse the same data structure for API consistency (`WebRtcServerListenInfo.listen_ip` field, but during JSON serialization/deserialization it is flattened just like in TypeScript), hence renaming to more generic name.

rust/src/messages.rs
ibc: Channel messages are received by the Worker. If the Worker needs to know which Router a producerId belongs it should own a mapProduderToRouter (and so on). I think it's not that bad that we signal the entire tree in internal to avoid having those maps, but I see your point.

Whether webRtcTransportId should be part of data... I kinda agree. This is not for internal routing. I'm fine if w change this but that must also be done in other cases (I insist: when you call consume() the producerId is given in internal, it should also be given in data). Also in rtpObserver.addProducer() (producerId should be in the request data rather than internal).

ibc (approved):

jmillan (approved):

rust/src/router/webrtc_transport.rs
```diff
@@ -80,33 +81,50 @@ impl TryFrom<Vec<TransportListenIp>> for TransportListenIps {
...
+        enable_udp: bool,
+        /// Listen in TCP.
+        /// Default false.
+        enable_tcp: bool,
+        /// Prefer UDP.
+        /// Default false.
+        prefer_udp: bool,
+        /// Prefer TCP.
+        /// Default false.
+        prefer_tcp: bool,
```
nazar-pc, line 109: @ibc I meant to ask earlier, but want to confirm whether this separation is actually correct or some of those flags are still applicable to WebRtcServer scenario.

ibc: OF COURSE! And they work as intended. See:

- https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/Router.cpp#L270
- https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/WebRtcServer.cpp#L283

nazar-pc: I'll have to partially revert this change then so they are usable again, thanks!

ibc: Sorry, I didn't realize you removed them. The thing is that, based on those flags, resulting ICE candidates may vary (based on whether the WebRtcServer was provided or not with UDP and TCP ports).

nazar-pc: Yes, makes sense. They're not removed, they're just not specifiable with WebRtcServer specifically. I'll bring them back.

ibc: Just to clarify, nothing changed at API level. Those flags still exist when you call `router.createWebRtcTransport()` despite you pass a `WebRtcServer` or not into the options.

PR #847, opened 2022-06-29
Channel: move some ids from internal object to data object

ibc:
For instance, when calling `transport.consume(options)`, the user must pass `producerId` within the given options. When it comes to route the channel request to the corresponding `Transport` in C++, such a `producerId` is not used for "internal routing" and hence it must be in the `data` object instead.

In fact, we were already doing it properly in `rtpObserver.addProducer(options)` in which `options.producerId` was sent to the worker within the `request.data` object rather than `request.internal`.

Changes in Node layer:

- `router.createWebRtcTransportWithServer()`: Move `webRtcServerId` from `internal` to `data`.
- `transport.consume()`: Move `producerId` from `internal` to `data`. Also in `PipeTransport.consume()` method.
- `transport.consumeData()`: Move `dataProducerId` from `internal` to `data`.

Changes in Rust layer:

- In theory similar changes (including the `webRtcServerId` one).
- Tests pass so ¯\_(ツ)_/¯


ibc:
@nazar-pc help 🥲

jmillan (approved):

ibc:
@nazar-pc I fixed a bug in Rust :)

https://github.com/versatica/mediasoup/pull/847/commits/5fe5ac53205809ff748ca946c61afc4984f19628

ibc:
This is ready for review. Tests pass.

nazar-pc:
Good catch, sorry about that :upside_down_face:

ibc:
> Good catch, sorry about that 🙃

It worked before because `ConsumerEnableTraceEventRequest` included `producerId` :-X

nazar-pc (approved):
Makes sense to me. Glad to see you start touching Rust codebase a bit in recent PRs :slightly_smiling_face:

jiyeyuran:
[mediasoup-go](https://github.com/jiyeyuran/mediasoup-go) will break if the mediasoup-worker protocol is incompatible. Maybe the golang wrapper also needs to be placed in this repository like rust.

ibc:
We run out of energy and time to manage another wrapper in the repo at this time. We may do it in the future if mediasoup-go remains maintained, but right now it would be too much for us.

jiyeyuran:
I understand. I also prefer to maintain the golang wrapper as a standalone repo. We just need to realize that the protocol compatibility is important to other wrappers.

ibc:
Yup, agreed. Not sure how to address this. I suggest you to subscribe to all issues/PRs in mediasoup and we'll take care of properly document or indicate whether there are changes in the public API or internal C++ classes. Does it sound good enough?

nazar-pc:
As long as mediasoup-go points to specific revision of mediasoup upstream it'll not break, but if you upgrade that revision then there is no way to avoid checking if API has changed as it is not considered public (at least not right now).

jiyeyuran:
> Yup, agreed. Not sure how to address this. I suggest you to subscribe to all issues/PRs in mediasoup and we'll take care of properly document or indicate whether there are changes in the public API or internal C++ classes. Does it sound good enough?

That is good enough, thanks.

jiyeyuran:
> As long as mediasoup-go points to specific revision of mediasoup upstream it'll not break, but if you upgrade that revision then there is no way to avoid checking if API has changed as it is not considered public (at least not right now).

mediasoup-go aims to be as backward compatible with mediasoup-worker as possible. I don't want to make the old mediasoup-worker unavailable by upgrading mediasoup-go. Previously due to protocol format issues (from netstring to lv), mediasoup-go had to introduce a WorkerVersion option to control which format to use.

nazar-pc:
Then you are kind of prisoner of that decision. I really don't see why would anyone want to use older worker with newer version of the library, you just have more code to maintain.

jiyeyuran:
Some companies have customized mediasoup-worker, and the cost of upgrading is relatively high, but the wrapper has fixed bugs and needs to be upgraded to the latest version. I think wrappers and mediasoup-worker should be upgraded independently and compatible like applications and operating systems.

nazar-pc:
Well, the idea of Open Source software is to contribute improvements upstream. If someone changes internal things, it is on them to maintain it, not on mediasoup authors. Worker right now is not a standalone product, it is an implementation detail.

jiyeyuran:
rust, node, and mediasoup-worker are put together in one repo and can be kept in sync, for them mediasoup-worker is an internal implementation detail, but not for the other wrappers.

ibc:
@jiyeyuran lets talk after summer (in 2 months) about this. I'd love to have mediasoup-go integrated in mediasoup repo.

PR #848, opened 2022-06-30
Do not clone audio RTP packets

jmillan:
We are currently cloning all RTP packets and creating a shared_ptr out
of them.

* Do not clone audio RTP packets, saving the corresponding RtpPacket
  instance creation, memory allocation, memory copy, etc.

### DIFF
--- a/worker/src/RTC/Producer.cpp
+++ b/worker/src/RTC/Producer.cpp
@@ -609,6 +609,9 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		bool isRtx{ false };
+		ReceiveRtpPacketResult result;
+
 		// Reset current packet.
 		this->currentRtpPacket = nullptr;
 
@@ -621,14 +624,13 @@ namespace RTC
 		{
 			MS_WARN_TAG(rtp, "no stream found for received packet [ssrc:%" PRIu32 "]", packet->GetSsrc());
 
-			return ReceiveRtpPacketResult::DISCARDED;
+			result = ReceiveRtpPacketResult::DISCARDED;
+
+			goto done;
 		}
 
 		// Pre-process the packet.
 		PreProcessRtpPacket(packet);
-
-		ReceiveRtpPacketResult result;
-		bool isRtx{ false };
 
 		// Media packet.
 		if (packet->GetSsrc() == rtpStream->GetSsrc())
@@ -642,7 +644,7 @@ namespace RTC
 				if (this->mapSsrcRtpStream.size() > numRtpStreamsBefore)
 					NotifyNewRtpStream(rtpStream);
 
-				return result;
+				goto done;
 			}
 		}
 		// RTX packet.
@@ -653,7 +655,7 @@ namespace RTC
 
 			// Process the packet.
 			if (!rtpStream->ReceiveRtxPacket(packet))
-				return result;
+				goto done;
 		}
 		// Should not happen.
 		else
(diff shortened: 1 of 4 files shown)


worker/src/RTC/Router.cpp
```diff
@@ -780,24 +780,27 @@ namespace RTC
...
 
-		auto& consumers = this->mapProducerConsumers.at(producer);
+		std::shared_ptr<RTC::RtpPacket> sharedPacket{ nullptr };
 
-		if (!consumers.empty())
-		{
+		// The packet does not need to be accessible in the future, do not clone it.
+		if (producer->GetKind() == RTC::Media::Kind::AUDIO)
+			std::shared_ptr<RTC::RtpPacket> sharedPacket(packet);
+		else
```
ibc, line 790: Brackets in if and else please

ibc (commented):
Looking good

worker/src/RTC/Router.cpp
```diff
@@ -780,24 +780,31 @@ namespace RTC
...
 		MS_TRACE();
 
-		auto& consumers = this->mapProducerConsumers.at(producer);
+		std::shared_ptr<RTC::RtpPacket> sharedPacket{ nullptr };
 
-		if (!consumers.empty())
+		// The packet does not need to be accessible in the future, do not clone it.
+		if (producer->GetKind() == RTC::Media::Kind::AUDIO)
+		{
+			std::shared_ptr<RTC::RtpPacket> sharedPacket(packet);
```
nazar-pc: I don't think it does what you think it does, this is probably more correct:
```suggestion
			sharedPacket.reset(packet);
```

But it shouldn't compile because it is just a pointer and can't be de-allocated.

worker/src/RTC/Router.cpp
```diff
@@ -780,24 +780,31 @@ namespace RTC
...
+		}
+		else
 		{
 			// Clone the packet so it holds its own buffer, usable for future
 			// retransmissions.
 			std::shared_ptr<RTC::RtpPacket> sharedPacket(packet->Clone());
```
nazar-pc:
```suggestion
			sharedPacket.reset(packet->Clone());
```

worker/src/RTC/Transport.cpp
```diff
@@ -1726,7 +1727,9 @@ namespace RTC
 			default:;
 		}
 
-		delete packet;
+		// Media forwarded, the packet lifetime is already handled.
+		if (result != RTC::Producer::ReceiveRtpPacketResult::MEDIA_FORWARDED)
+			delete packet;
```
nazar-pc: This is super fragile, let's try to avoid such things when de-allocation of something depends on return value somewhere in completely different place.

nazar-pc (commented):
I don't think this actually works for audio at all, if I read it correctly audio packets with this change are always `nullptr`, which in Rust wouldn't compile or would crash immediately. This needs a different approach, probably a different `SendRtpPacket`, one for cloned packet in smart pointer and another with just raw pointer or something like that.

Though generally one clone per packet wasn't a bottleneck from my testing, but I agree it might be handy to avoid it if possible.

worker/src/RTC/Router.cpp
jmillan: Yep, it's been fixed in the last commit.

worker/src/RTC/Transport.cpp
jmillan: Handled in the last commit. Sorry @nazar-pc, indeed even though I marked it as PR I already told @ibc to wait for review until a missing commit was pushed..., I forgot to label this as draft again..

jmillan:
> This needs a different approach, probably a different SendRtpPacket, one for cloned packet in smart pointer and another with just raw pointer or something like that.

Yes, this will be a next action.

jmillan:
Note for reviewers. This is a very small PR. Please review it in bulk rather than commit by commit since some things were re-done.

jmillan:
Converted to draft. Please, do not review yet.

jmillan:
> SendRtpPacket, one for cloned packet in smart pointer and another with just raw pointer or something like that.

This is what we had in mind for a second iteration. I'm giving it a try indeed.

worker/src/RTC/Producer.cpp
```diff
@@ -621,15 +624,13 @@ namespace RTC
 		{
 			MS_WARN_TAG(rtp, "no stream found for received packet [ssrc:%" PRIu32 "]", packet->GetSsrc());
 
-			return ReceiveRtpPacketResult::DISCARDED;
+			result = ReceiveRtpPacketResult::DISCARDED;
+			goto done;
```
ibc, line 628: OMG really? Pretty sure we can avoid this...

I strongly prefer early return result.

ibc:
THIS PR HAS BEEN CLOSED
???

jmillan:
> THIS PR HAS BEEN CLOSED

This, this PR has been closed in favour of https://github.com/versatica/mediasoup/pull/850

worker/src/RTC/Producer.cpp
jmillan: not used in new PR

PR #852, opened 2022-07-02
Improve `WebRtcTransportOptions` type to not allow WebRTC server and listen IPs/port at the same time

nazar-pc:
Just more type safety. It only allows to either specify `listenIps` and port or WebRTC server, but not both at the same time because it doesn't make sense.


node/src/Router.ts
```diff
@@ -363,21 +365,24 @@ export class Router extends EnhancedEventEmitter<RouterEvents>
 			maxSctpMessageSize = 262144,
 			sctpSendBufferSize = 262144,
 			appData
-		}: WebRtcTransportOptions
-	): Promise<WebRtcTransport>
-	{
-		logger.debug('createWebRtcTransport()');
+		} = webRtcTransportOptions;
```
ibc: 🙁

I don't like this. We use fields directly rather than having a named object. This change is making a difference with no much reason.

node/src/WebRtcTransport.ts
```diff
@@ -12,24 +12,34 @@ import {
 import { WebRtcServer } from './WebRtcServer';
 import { SctpParameters, NumSctpStreams } from './SctpParameters';
 
-export type WebRtcTransportOptions =
+export interface WebRtcTransportListenIndividual
```
ibc: Why interface? I know it's ok but we use type everywhere else.

node/src/WebRtcTransport.ts
```diff
@@ -12,24 +12,34 @@ import {
...
-	/**
-	 * Instance of WebRtcServer. Mandatory unless listenIps is given.
-	 */
-	webRtcServer?: WebRtcServer;
 	/**
 	 * Listening IP address or addresses in order of preference (first one is the
 	 * preferred one). Mandatory unless webRtcServer is given.
 	 */
-	listenIps?: (TransportListenIp | string)[];
+	listenIps: (TransportListenIp | string)[];
```
ibc, line 22: This is now mandatory but it shouldn't (if webrtcserver is given).

node/src/WebRtcTransport.ts
nazar-pc: Hm, there are actually a few other places where interfaces are used, I can change it back if you'd like. Personally though I find it strange to use `type` for anything other than type aliases.

Changed all to `type` for consistency in this file at least.

node/src/WebRtcTransport.ts
nazar-pc: It was incomplete indeed, now it works properly, please check updated version.

node/src/Router.ts
nazar-pc: Older revision of the code didn't compile otherwise, changed back now.

node/src/Router.ts
```diff
@@ -373,12 +373,6 @@ export class Router extends EnhancedEventEmitter<RouterEvents>
 		else if (appData && typeof appData !== 'object')
 			throw new TypeError('if given, appData must be an object');
 
-		if (webRtcServer)
-		{
-			listenIps = undefined;
-			port = undefined;
-		}
```
nazar-pc, line 380: This is no longer necessary and is guaranteed at a type level instead

ibc:
How are these changes affect TypeScript usage?

nazar-pc:
If someone specified both `listenIps` and `webRtcServer` at the same time code will stop compiling. Otherwise no change, just more specific/strict types.

UPD: Also if someone specified neither, code will also stop compiling because you need to specify something, right?

ibc (approved):

node/src/WebRtcTransport.ts
```diff
@@ -11,25 +11,36 @@ import {
...
+export type WebRtcTransportListenServer =
+{
+	/**
+	 * Instance of WebRtcServer. Mandatory unless listenIps is given.
+	 */
+	webRtcServer: WebRtcServer;
+}
+
+export type WebRtcTransportListen =
+	Either<WebRtcTransportListenIndividual, WebRtcTransportListenServer>;
```
jmillan, line 40: ?

```suggestion
	WebRtcTransportListenIndividual | WebRtcTransportListenServer;
```

ibc: OMG true. We don't need any new Either.

nazar-pc: We do, otherwise fields of both will be combined. We don't want that.
We want either fields from one and none from the other or the opposite.

ibc: Don't wrap listenIps and webRtcServer into objects but just add a listen: listenIps | webRrcServer; and it should work, although we are changing the api if so...

So your Either doesn't imply any api change right? If so let's go

nazar-pc: Yes, Either ensures API stays the same. `listen: listenIps | webRtcServer` isn't a full solution because there is also a port, which doesn't make sense in case of `webRtcServer`.

Just apply these changes and see what compiles and what no longer does.

jmillan (approved):

ibc:
Let's merge.

PR #857, opened 2022-07-04
Fix libuv build

vpalmisano:
Fixes a build error:
```
subprojects/libuv-v1.43.0/libuv.a.p/src_unix_stream.c.o.d -o subprojects/libuv-v1.43.0/libuv.a.p/src_unix_stream.c.o -c ../../../subprojects/libuv-v1.43.0/src/unix/stream.c
../../../subprojects/libuv-v1.43.0/src/unix/stream.c: In function ‘uv__write’:
../../../subprojects/libuv-v1.43.0/src/unix/stream.c:929:3: error: C++ style comments are not allowed in ISO C90
```

### DIFF
--- a/worker/meson.build
+++ b/worker/meson.build
@@ -163,7 +163,6 @@ libuv_proj = subproject(
     'warning_level=0',
     'build_tests=false',
     'build_benchmarks=false',
-    'c_std=c99',
   ],
 )
 libsrtp2_proj = subproject(
--- a/worker/meson.build
+++ b/worker/meson.build
@@ -163,6 +163,7 @@ libuv_proj = subproject(
     'warning_level=0',
     'build_tests=false',
     'build_benchmarks=false',
+    'c_std=c99',
   ],
 )
 libsrtp2_proj = subproject(
--- a/worker/subprojects/libuv.wrap
+++ b/worker/subprojects/libuv.wrap
@@ -1,12 +1,11 @@
 [wrap-file]
-directory = libuv-v1.43.0
-source_url = https://dist.libuv.org/dist/v1.43.0/libuv-v1.43.0.tar.gz
-source_filename = libuv-v1.43.0.tar.gz
-source_hash = 90d72bb7ae18de2519d0cac70eb89c319351146b90cd3f91303a492707e693a4
-patch_filename = libuv_1.43.0-1_patch.zip
-patch_url = https://wrapdb.mesonbuild.com/v2/libuv_1.43.0-1/get_patch
-patch_hash = 9bd241c95110c67f9486f1a77a988216bf5ec192dc3002651a6ccefb2e6aba5a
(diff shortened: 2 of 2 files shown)


nazar-pc:
Hm... this must be regression in Meson, `c_std=c99` is already specified in libuv wrap: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/libuv/meson.build#L4

Can you try bumping libuv wrap to 1.44.1 instead? It switched from `c89` to `c99`, maybe that is related somehow...
https://github.com/mesonbuild/wrapdb/pull/418

vpalmisano:
> Can you try bumping libuv wrap to 1.44.1 instead? It switched from `c89` to `c99`, maybe that is related somehow...
> [mesonbuild/wrapdb#418](https://github.com/mesonbuild/wrapdb/pull/418)

It works :)

nazar-pc (approved):
Looks like something wasn't accounted for in older versions of Meson, but now it is properly checked

ibc (approved):

ibc:
I forgot to say that PRs must include an entry in `CHANGELOG.md`. I'll add it in v3 directly.

dsdolzhenko:
> Hm... this must be regression in Meson

@nazar-pc Am I right that the regression is in Meson version 0.63.0 and the build started failing right after it was released?

What do you think about specifying exact version in [the Makefile](https://github.com/versatica/mediasoup/blob/c6dd18dc62ed175286e8596f8ca1ad8154fc9353/worker/Makefile#L77) to make builds more reproducible?

nazar-pc:
The idea was that newer versions of Meson should not regress and bring improvements, but I tend to agree that we can lock the version until Meson reached 1.0 to prevent such breakage in the future.

dsdolzhenko:
> I tend to agree that we can lock the version until Meson reached 1.0

Personally, I would've prefer to keep it locked even after 1.0 release and update manually whenever it's needed. Strictly speaking no one promises that after 1.0 there won't be any regressions ;)

Shall I create a PR?

nazar-pc:
> Shall I create a PR?

Yes, please do

eli-schwartz:
This isn't really a Meson bug, it's a bug in the libuv wrap itself which was fixed at the end of April. :p 

The WrapDB does get sanity-checked with new prereleases of Meson (available in advance of any new feature release, you can install this in CI to catch errors early), so updating wraps should likely be the first thing to check (and do for general maintenance?)

In this case, Meson used to totally ignore the std of the subproject and use the one from the main project, which for mediasoup was a lot newer than c89. The new Meson release actually permits subprojects to specify their own std: https://mesonbuild.com/Release-notes-for-0-63-0.html#compiler-options-can-be-set-per-subproject

libuv just happened to specify one that did not actually work. :(

PR #863, opened 2022-07-04
Fix release contents by including `meson_options.txt`

nazar-pc:
I think this might affect both Rust and TypeScript.

Cargo actually creates a package and then tries to build it, so it fails to publish if required files are not included.

Without this I'm getting following error:
```
  meson.build:29:3: ERROR: Tried to access unknown option 'ms_log_trace'.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "worker/test/src",
     "worker/Makefile",
     "worker/meson.build",
+    "worker/meson_options.txt",
     "npm-scripts.js"
   ],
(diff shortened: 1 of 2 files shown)


skavish:
do you plan to make new release with this fix included?
building 3.10.1 on M1 fails now

nazar-pc:
@ibc please release 3.10.2 whenever you can

ibc:
3.10.2 released.

Issue #861, opened 2022-07-04
Segmentation fault in DTLS procedures


prlanzarin:
### Your environment

- Operating system: Ubuntu 20.04.4 LTS
- Node version: v16.15.0
- npm version: 8.5.5
- gcc/clang version: 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1) 
- mediasoup version: 3.10.0
- mediasoup-client version: n/a

### Issue description

We're seeing sporadic worker crashes (SIGSEGV) after trying mediasoup 3.9.12 or later that seem related to DTLS.
 - I've tried it with 3.9.12, 3.9.13 and 3.10.0 in some of our demo servers and could see crashes pop up in monitoring with all of them.
 - It doesn't seem to be happening in another demo server with similar user load and control plane that's still running mediasoup `3.9.10`.
 
I've set workers to be "long lived" in those demo servers (ie I usually rotate them every ~3 to ~5 days) and the crash seems to occur every ~two days (with slight variations to that  - ie it's not _that_ frequent).

For the sake of consistency, the info I'm attaching is related to a `3.10.0` occurrence.

<details><summaryDetails</summary>
<p>

### Core dump

The attached tarball includes a core dump, logs from the worker and the worker binary file.
  * Since not every log is annotated with the `pid` and the server is running multiple workers, some log entries (eg transport creation etc) may be missing because I couldn't isolate them reliably.
  * [ms3100c.tar.gz](https://github.com/versatica/mediasoup/files/9041123/ms3100c.tar.gz)
  
### Backtrace snippet

```
0  0x000055d775fdbc6a in dtls1_get_timeout ()
No symbol table info available.
#1  0x000055d775fdbd80 in dtls1_is_timer_expired ()
No symbol table info available.
#2  0x000055d775fdbf22 in dtls1_handle_timeout ()
No symbol table info available.
#3  0x000055d775fdbff5 in dtls1_ctrl ()
No symbol table info available.
#4  0x000055d775c9e962 in RTC::DtlsTransport::SetTimeout() ()
No symbol table info available.
#5  0x000055d775c9e9d4 in RTC::DtlsTransport::SetTimeout() ()
No symbol table info available.
#6  0x000055d775c9e9d4 in RTC::DtlsTransport::SetTimeout() ()
No symbol table info available.
[... goes on with thousands of similar entries ...]
#74693 0x000055d775c9e9d4 in RTC::DtlsTransport::SetTimeout() ()
[...]

PR #867, opened 2022-07-05
Fix DTLS related crash (issue #861)

ibc:
It may fix issue #861, which was potentially introduced by PR #815.

WIP: no idea what's going on here.

### DIFF
--- a/worker/src/RTC/DtlsTransport.cpp
+++ b/worker/src/RTC/DtlsTransport.cpp
@@ -980,28 +980,28 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		MS_ASSERT(this->ssl, "this->ssl is not set");
 		MS_ASSERT(
 		  this->state == DtlsState::CONNECTING || this->state == DtlsState::CONNECTED,
 		  "invalid DTLS state");
 
-		int64_t ret;
 		uv_timeval_t dtlsTimeout{ 0, 0 };
 		uint64_t timeoutMs;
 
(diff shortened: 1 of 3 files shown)


ibc:
Asking for @fippo's and @ggarber thoughts.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1432,19 +1431,18 @@ namespace RTC
 		{
 			MS_WARN_TAG(dtls, "DTLSv1_handle_timeout() failed");
 
-			Reset();
+			// NOTE: Don't do anything here (such as calling Reset() because the error
```
ggarber: who is going to call CheckStatus in this case?

fippo:
would it be possible for the timer to survive the object? That is the only thing I can think of...

worker/src/RTC/DtlsTransport.cpp
ibc: Right, it shouldn't...

ibc:
> would it be possible for the timer to survive the object? That is the only thing I can think of...

Nope, of course the timer is properly destroyed in the `DtlsTransport` destructor.

worker/src/RTC/DtlsTransport.cpp
ibc: Reverted.

ibc:
Released in 3.10.4

Issue #832, opened 2022-06-13
Allow static typing of appData


mango-martin:
## Feature Request
Hey guys. This is my first created issue please be gentle :)

A while back I ran into trouble typing my appData object because of "any" type incompatibilities. That's why I ended up not typing it.
Latest changes to replace `any` were an improvement for sure. And it is possible to extract appData and do a type conversion when you need it, like
```
const appData = router.appData as RouterAppData;
```
But it is not ideal in my opinion.

I played around with generic types in the definition files and it is now possible for me to declare an AppDataType that is directly accessible. As an example I'll take a Router instance.

```
declare interface MyRouterAppData {
  existing: string
}

async function exampleWithType() {
  const worker = await createWorker();

  const router = await worker.createRouter<MyRouterAppData>({
    appData: { existing: 'TEST' },
  });

  const existing = router.appData.existing; // No type error => OK
  const nonExisting = router.appData.nonExisting; // Type error => OK
}
```

Changes look like this:
`Worker.d.ts`
```
createRouter<T = Record<string, unknown>>({ mediaCodecs, appData }?: RouterOptions): Promise<Router<T>>;
```
`Router.d.ts`
```
export declare class Router <T = Record<string, unknown>> extends EnhancedEventEmitter<RouterEvents> {
...
constructor(
...
        appData?: T;
...
get appData(): T;
set appData(appData: T);
```
It has the current type Record<string, unknown> as a fallback value so it is a non-breaking change (but I think with these changes the standard type can be removed and the fallback type can be set to `any`. If you want to use a single number as your appData then I don't see any reason why it shouldn't be possible). And `T` is just a placeholder here. I think something like `AppDataType` would be better to use.

These changes would need to be made in many places depending on the scope of the refactoring. It does look a bit unnecessary to have the AppDataType variable at the top level like this but I don't really see how else this would be possible. Maybe someone who is more proficient in Typescript has a better approach? With these changes you would also need to add some boilerplate to your code. Wherever you use the `Router` type (to stay with the example) you would then need to write `Router<MyRouterAppData>` to opt-in and use your own types.

But all in all I still think it's desirable to have some way of statically typing the appData. Any thoughts on this? Cheers

**Edit: Regarding #821**
I can see how it might be a wanted solution to keep application logic completely separate from Mediasoup entities. But is a huge convenience for me as a developer to keep related extra data directly in the instance.

[...]

PR #869, opened 2022-07-15
Add generic types for appData

mango-martin:
Regarding #832 

I've been pretty swamped lately but I finally got around to do a barebones prototype of the suggested changes. I'll have to preface that I am not very fluent in TypeScript. But from what VSCode tells me it's been successful. I'll attach some screenshots to show the impact of the changes. Because of the fallbacktype `= AppData` `({ [key: string]: unknown })` this is a non-breaking change.

To be clear:
This only covers the Router class as an example. To bring these changes to all classes where appData is used, this would need to be implemented across many more files. I just want to bring this issue into an actionable state. If you like, I can also extend this to cover all of the core code but it will take some time.
Update: I've added another commit that covers all of mediasoup core code.

typedecl
correctinf
random

mustbeobject

router

The only real downside with this, as far as I can tell, is that this does not correctly alarm of a TypeError if there was no appData provided - but I think it's an edge case. If you don't use appData, you also shouldn't be accessing it. 

Cheers


node/src/Router.ts
```diff
@@ -120,7 +120,12 @@ export type RouterObserverEvents =
 
 const logger = new Logger('Router');
 
-export class Router extends EnhancedEventEmitter<RouterEvents>
+export type AppData = {
```
mango-martin: I put this here temporarily because I didn't really know where else to keep it

mango-martin:
I've updated the PR already to cover all of the code - it was much more straightforward than I thought.
Few questions remain about naming the types individually (in some classes a general name would conflict with another type that was declared at the class level - so I went with naming them individually). This might be a bit hard to maintain.
Another thing is the location of my helper type `AppData`. Basically you could replace every occurence of AppData with Record<string, unknown> but it just seemed reasonable to have a bit of a shorter name for that.
Lastly about formatting. Because The added types would often exceed the linter limit of characters per line, I had to put in a few line-breaks.

Also there are no fallback types for any Options. It just takes a generic. This might lead to confusing result where you can type an option like
```
const routerOptions: RouterOptions<number> = { // OK 
  appData: 10,
}

const router = await worker.createRouter(routerOptions) // Typeerror
```
Can add the same structure for options as well so it will already throw a typeerror when declaring them, it just seemed a little bit of an overkill but looking at it now, it should not be like this.


Let me know how you want to proceed with this. Or not, up to you :D

ibc:
We don't forget about this. It's just that we have been very busy.

node/src/index.ts
```diff
@@ -87,3 +87,9 @@ export function getSupportedRtpCapabilities(): RtpCapabilities
 {
 	return utils.clone(supportedRtpCapabilities) as RtpCapabilities;
 }
+
+/* Helper type */
+
+export type AppData = {
+	[key: string]: unknown;
+}
```
ibc, line 95: Please let's not do this here. Add this into `types.ts` instead.

ibc (changes requested):

ibc:
@mango-martin I've replaced your PR with this one https://github.com/versatica/mediasoup/pull/1046 due to missing additions that shown up after we migrated tests to TypeScript. Credits to you. Thanks!

PR #870, opened 2022-07-20
Do not use JSON for internal routing

jmillan:
JSON encoding/decoding is very CPU intensive.

Use character separated string for `internal` routing representation.

`data` info from node to worker for `Channel` remains being JSON.
`data` info from node to worker for `PayloadChannel` is a now string to be decoded by the target:
  * At this moment only DataChannel ppid is used as `data`, and we are NOT sending it as JSON but as plain string.


worker/src/Channel/ChannelRequest.cpp
```diff
@@ -133,12 +136,12 @@ namespace Channel
 
 		this->replied = true;
 
-		json jsonResponse = json::object();
+		std::string response("{\"id\":");
```
jmillan, line 139: Manually create this basic JSON.

jmillan:
Canceling Rust GH actions because that changes are missing.

node/src/Consumer.ts
```diff
@@ -195,6 +195,7 @@ export class Consumer extends EnhancedEventEmitter<ConsumerEvents>
 		routerId: string;
```
ibc: I don't see why we have to keep this internal object rather than just a single string. We just need consumerId and we can store it in a separate id private property.

worker/src/Channel/ChannelRequest.cpp
```diff
@@ -197,4 +200,16 @@ namespace Channel
 
 		this->channel->Send(jsonResponse);
 	}
+
+	const std::string& ChannelRequest::GetNextInternalRoutingId()
```
ibc, line 204: Perhaps "Pop" instead of "Get"? A method called "GetXxxx()" should not modify internal state and should be idempotent. "Pop" instead means remove and give me the last item.

jmillan: We are not doing `pop`. We maintain an internal `nextRoutingLevel` member that we increase everytime this function is called.

Yes, "GetXxx" should generally not modify the state.

node/src/Consumer.ts
jmillan: `#internal` as an object is widely used along the classes, accessing any of their members.

worker/src/Channel/ChannelRequest.cpp
jmillan: initially we commented about keeping the routing IDs within a `std::queue`, and `pop()` as we get the next id. But after implementation it seemed more convoluted than re-using the original `std::vector` returned from splitting the message, plus this way we can avoid creating a string instance for every call to the function.

node/src/Consumer.ts
jmillan: You are right @ibc, I'll check the other TS classes.

jmillan: Yep, I'm going to change this, good catch.

jmillan: Done.., I've added type definitions for `internal` constructor arguments too.

Hartigan:
What you think about replacing json by protobuf? This way we can save CPU/RAM and generate types for typescript/c++/rust.

jmillan:
We are completely removing any data structure for `internal`.

Data structures for `data` will be reviewed afterwards.

ibc:
Closing in favor of https://github.com/versatica/mediasoup/pull/893

PR #875, opened 2022-07-22
Log error detais happen in channel socket.

mstyura:
I've observed occasional error messages in the `stdout` of service which uses `mediasoup`:
```
Channel::ChannelSocket::OnConsumerSocketMessage() | discarding wrong Channel request
```
This PR extend information logged in such cases, so it should reveal more details of what happen.


ibc:
Thanks

PR #876, opened 2022-07-22
Make consumers, dataProducers and dataConsumers private.

CosmosisT:


CosmosisT:
A little new, unsure how to update previous "case" but perhaps this is what is wanted so we don't use them via API.

ibc:
Did you check TypeScript? This is not gonna work.

CosmosisT:
I don't use the TS version sorry, but that's why I wondered why producers was made private. Sorry, haven't explored fully.

Can dismiss this.

PR #878, opened 2022-07-24
Refactor request and notification handlers to conform to the same API

nazar-pc:
This is the first of a few PRs for `internal` refactoring to remove JSON. Here nothing really changes, except most worker entities that process requests or notifications implement the same parent class. But method names and even most of arguments are the same, most of cpp files are unchanged too.

This will allow us to put those similarly-looking entities into collections (maps), where key is ID of corresponding entity. After that we'll be able to do something like this (pseudo-code): `entities.get(id).HandleRequest(request)`. What entity it is wouldn't matter, we wouldn't need to send `internal` with a series of IDs and traverse multiple layers down, just one ID would be enough. As the result we can also avoid JSON encoding/decoding, avoid coma-separated values handling, just a simple string ID and a single map lookup is all we'll need.

I have some more refactoring in my local branch, just trying to send non-breaking simple changes separately for your convenience.

### DIFF
--- a/worker/include/Channel/ChannelSocket.hpp
+++ b/worker/include/Channel/ChannelSocket.hpp
@@ -51,15 +51,22 @@ namespace Channel
 	class ChannelSocket : public ConsumerSocket::Listener
 	{
 	public:
-		class Listener
+		class RequestHandler
+		{
+		public:
+			virtual ~RequestHandler() = default;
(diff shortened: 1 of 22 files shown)


jmillan (approved):
👍 Looking good @nazar-pc

PR #881, opened 2022-07-25
Refactor message handling

nazar-pc:
This is another refactoring that essentially changes nothing, just moves message handling logic from `Router` and `*Transport` to entities that are actually responsible for handling them.

Reviewing commit by commit should make sense.

I think is the last PR that does refactoring, next PR will actually deal with `internal`, but I had to prepare things for that PR to be managable.

Hm, this PR is based on #880, ignore the first commit here.

### DIFF
--- a/worker/include/RTC/Producer.hpp
+++ b/worker/include/RTC/Producer.hpp
@@ -196,7 +196,8 @@ namespace RTC
 		bool videoOrientationDetected{ false };
 		struct VideoOrientation videoOrientation;
 		struct TraceEventTypes traceEventTypes;
-		uint8_t* buffer{ nullptr };
+		// Static buffer.
+		thread_local static uint8_t* buffer;
 	};
 } // namespace RTC
 
--- a/worker/include/RTC/Producer.hpp
+++ b/worker/include/RTC/Producer.hpp
@@ -4,6 +4,7 @@
 #include "common.hpp"
 #include "Channel/ChannelRequest.hpp"
 #include "Channel/ChannelSocket.hpp"
+#include "PayloadChannel/PayloadChannelSocket.hpp"
 #include "RTC/KeyFrameRequestManager.hpp"
 #include "RTC/RTCP/CompoundPacket.hpp"
 #include "RTC/RTCP/Packet.hpp"
@@ -23,7 +24,8 @@ namespace RTC
 {
 	class Producer : public RTC::RtpStreamRecv::Listener,
 	                 public RTC::KeyFrameRequestManager::Listener,
-	                 public Channel::ChannelSocket::RequestHandler
+	                 public Channel::ChannelSocket::RequestHandler,
+	                 public PayloadChannel::PayloadChannelSocket::NotificationHandler
 	{
 	public:
 		class Listener
@@ -32,8 +34,10 @@ namespace RTC
 			virtual ~Listener() = default;
 
 		public:
-			virtual void OnProducerPaused(RTC::Producer* producer)  = 0;
-			virtual void OnProducerResumed(RTC::Producer* producer) = 0;
+			virtual void OnProducerReceiveData(RTC::Producer* producer, size_t len)                  = 0;
+			virtual void OnProducerReceiveRtpPacket(RTC::Producer* producer, RTC::RtpPacket* packet) = 0;
+			virtual void OnProducerPaused(RTC::Producer* producer)                                   = 0;
+			virtual void OnProducerResumed(RTC::Producer* producer)                                  = 0;
 			virtual void OnProducerNewRtpStream(
 			  RTC::Producer* producer, RTC::RtpStream* rtpStream, uint32_t mappedSsrc) = 0;
 			virtual void OnProducerRtpStreamScore(
@@ -132,6 +136,10 @@ namespace RTC
 	public:
 		void HandleRequest(Channel::ChannelRequest* request) override;
 
+		/* Methods inherited from PayloadChannel::PayloadChannelSocket::NotificationHandler. */
+	public:
+		void HandleNotification(PayloadChannel::Notification* notification) override;
+
 	private:
 		RTC::RtpStreamRecv* GetRtpStream(RTC::RtpPacket* packet);
 		RTC::RtpStreamRecv* CreateRtpStream(
@@ -188,6 +196,7 @@ namespace RTC
 		bool videoOrientationDetected{ false };
 		struct VideoOrientation videoOrientation;
 		struct TraceEventTypes traceEventTypes;
+		uint8_t* buffer{ nullptr };
 	};
 } // namespace RTC
 
--- a/worker/include/RTC/RtpObserver.hpp
+++ b/worker/include/RTC/RtpObserver.hpp
@@ -11,7 +11,21 @@ namespace RTC
 	class RtpObserver : public Channel::ChannelSocket::RequestHandler
 	{
 	public:
-		RtpObserver(const std::string& id);
+		class Listener
+		{
+		public:
+			virtual ~Listener() = default;
+
+		public:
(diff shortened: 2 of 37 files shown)


worker/include/RTC/Producer.hpp
```diff
@@ -32,8 +34,10 @@ namespace RTC
 			virtual ~Listener() = default;
 
 		public:
-			virtual void OnProducerPaused(RTC::Producer* producer)  = 0;
-			virtual void OnProducerResumed(RTC::Producer* producer) = 0;
+			virtual void OnDataReceived(RTC::Producer* producer, size_t len)                 = 0;
+			virtual void OnReceiveRtpPacket(RTC::Producer* producer, RTC::RtpPacket* packet) = 0;
```
jmillan: For consistency.

```suggestion
			virtual void OnRtpPacketReceived(RTC::Producer* producer, RTC::RtpPacket* packet) = 0;
```

jmillan: These two callbacks could IMHO have a better naming.., not a show stopper for now but a TODO.

jmillan (approved):
Just a cosmetic comment made.

worker/include/RTC/Producer.hpp
nazar-pc: They were called similarly to the names of the methods in the transport that are called from these `On` methods, renamed for consistency as suggested.

worker/include/RTC/Producer.hpp
```diff
@@ -32,8 +34,10 @@ namespace RTC
 			virtual ~Listener() = default;
 
 		public:
-			virtual void OnProducerPaused(RTC::Producer* producer)  = 0;
-			virtual void OnProducerResumed(RTC::Producer* producer) = 0;
+			virtual void OnDataReceived(RTC::Producer* producer, size_t len)                  = 0;
+			virtual void OnRtpPacketReceived(RTC::Producer* producer, RTC::RtpPacket* packet) = 0;
```
jmillan: To be consistent with other callback methods like: `OnProducerRtcpSenderReport` or `OnProducerRtpStreamScore` these should be named:

`OnProducerData` and `OnProducerRtpPacket` respectively.

nazar-pc: Well, previous refactoring already made it worse IMHO because it created a confusion. If I was adding `OnProducer` prefix everywhere then I should have called method `OnProducerRtpPacketReceived`, but turns out such method already exists and serves a different purpose entirely. So what about me renaming it like `OnProducerReceiveData` and `OnProducerReceiveRtpPacket` to make them look different?

jmillan: Works for me 👍

nazar-pc:
Refactored methods both in producer and data producer

jmillan (approved):

worker/include/RTC/Transport.hpp
```diff
@@ -191,6 +191,14 @@ namespace RTC
 
 		/* Pure virtual methods inherited from RTC::Producer::Listener. */
 	public:
+		void OnProducerReceiveData(RTC::Producer* /* producer */, size_t len) override
```
ibc: The rest of the code uses `/*producer*/` (without spaces).

worker/src/RTC/Producer.cpp
```diff
@@ -605,6 +607,56 @@ namespace RTC
...
+				auto len         = notification->payloadLen;
+
+				// Increase receive transmission.
+				this->listener->OnProducerReceiveData(this, len);
+
+				if (len > RTC::MtuSize + 100)
+				{
+					MS_WARN_TAG(rtp, "given RTP packet exceeds maximum size [len:%zu]", len);
+
+					return;
```
ibc: Please use `break` instead of `return` within `case` everywhere (there are more like this).

worker/src/RTC/DataProducer.cpp
```diff
@@ -145,6 +146,51 @@ namespace RTC
...
+
+				if (len > this->maxMessageSize)
+				{
+					MS_WARN_TAG(
+					  message,
+					  "given message exceeds maxMessageSize value [maxMessageSize:%zu, len:%zu]",
+					  len,
+					  this->maxMessageSize);
+
+					return;
```
ibc: break instead

worker/src/RTC/Producer.cpp
```diff
@@ -605,6 +607,56 @@ namespace RTC
...
+				// Copy the received packet into this buffer so it can be expanded later.
+				std::memcpy(this->buffer, data, static_cast<size_t>(len));
+
+				RTC::RtpPacket* packet = RTC::RtpPacket::Parse(this->buffer, len);
+
+				if (!packet)
+				{
+					MS_WARN_TAG(rtp, "received data is not a valid RTP packet");
+
+					return;
```
ibc: break

ibc (changes requested):

worker/src/RTC/Producer.cpp
jmillan: @nazar-pc, the `buffer` belonged, prior to this PR, to DirectTransport, hence all Producers would use the same. Within this PR every Producer will own one, with the corresponding memory usage increase if many Producers in the same DirectTransport need it. In fact it could be a static buffer that is used by every instance. cc @ibc.

worker/src/RTC/Producer.cpp
nazar-pc: It was like that before (this code was moved), but changed as suggested

worker/src/RTC/Producer.cpp
nazar-pc: Agree, applied corresponding change in latest commit

worker/include/RTC/Producer.hpp
```diff
@@ -188,6 +196,8 @@ namespace RTC
 		bool videoOrientationDetected{ false };
 		struct VideoOrientation videoOrientation;
 		struct TraceEventTypes traceEventTypes;
+		// Static buffer
```
ibc: Dot please ☺️

worker/include/RTC/RtpObserver.hpp
```diff
@@ -8,12 +8,29 @@
...
+			virtual void OnRtpObserverAddProducer(RTC::RtpObserver* rtpObserver, RTC::Producer* producer) = 0;
+			virtual void OnRtpObserverRemoveProducer(
+			  RTC::RtpObserver* rtpObserver, RTC::Producer* producer) = 0;
+		};
+
+	public:
+		RtpObserver(const std::string& id, RTC::RtpObserver::Listener* listener);
 		virtual ~RtpObserver();
 
+	private:
```
ibc: Private methods must be after public ones.

ibc (changes requested):

nazar-pc:
Had to fix clippy warnings introduced in latest Rust release

worker/src/RTC/Router.cpp
```diff
@@ -1121,4 +1079,28 @@ namespace RTC
 		// Delete it.
 		delete transport;
 	}
+
+	void Router::OnRtpObserverAddProducer(RTC::RtpObserver* rtpObserver, RTC::Producer* producer)
+	{
+		// Add to the map.
+		this->mapProducerRtpObservers[producer].insert(rtpObserver);
+	}
+	void Router::OnRtpObserverRemoveProducer(RTC::RtpObserver* rtpObserver, RTC::Producer* producer)
```
ibc:
```suggestion
	}

	void Router::OnRtpObserverRemoveProducer(RTC::RtpObserver* rtpObserver, RTC::Producer* producer)
```

ibc (approved):
I pushed a cosmetic commit. LGTM

PR #884, opened 2022-07-27
Update Dockerfile for fuzzing

jmillan:


worker/Dockerfile
```diff
@@ -1,23 +1,21 @@
-FROM ubuntu:16.04
+# Get latest Ubuntu LTS.
+FROM ubuntu:latest
```
nazar-pc: Let's use `22.04` or `20.04` explicitly, would be annoying for it to break later. `latest` is not LTS, it will switch to 22.10 as soon as it is out.
```suggestion
FROM ubuntu:22.04
```

nazar-pc:
CI failure is fine, I'll update it in a separate PR, old macOS workers are being shut down by GitHub

worker/Dockerfile
jmillan: Indeed it seems `latest` is LTS:

https://hub.docker.com/_/ubuntu/

`The ubuntu:latest tag points to the "latest LTS", since that's the version recommended for general use. The ubuntu:rolling tag points to the latest release (regardless of LTS status).`

nazar-pc: Didn't read that far. Still, I think for long-term reliability it would be beneficial to specify the version explicitly.

jmillan: yes, makes sense.

nazar-pc (approved):

PR #886, opened 2022-07-29
Update OSs used in CI

nazar-pc:
macOS and Windows are not used for production (I think), so testing just latest version is fine. Testing 3 latest Ubuntu versions is important though.

This both reduces number of jobs we'll have and makes them more relevant.

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -11,9 +11,8 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - macos-10.15
-          - macos-11
-          - windows-2019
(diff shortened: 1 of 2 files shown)


BlueMagnificent:
> macOS and Windows are not used for production (I think)

Do you mean production environment or production in the context of the CI setup?

jmillan (approved):

ibc:
I'm not sure Windows is not being used in production.

nazar-pc:
By production I mean deployment that users use. Surely users don't connect to instance running in CI pipeline :shrug: 

Still, we don't drop support for anything, just reducing number of jobs for things that are less relevant. We can fix issues for older OSs if they ever happen and worth fixing.

PR #891, opened 2022-08-23
Node: add types for internal and data objects

ibc:
...and avoid tons of `any` types.


nazar-pc (approved):
Certainly in favor of heavier typing :+1:

nazar-pc (approved):
Much nicer :slightly_smiling_face:

PR #894, opened 2022-08-24
New ChannelMessageHandlers class (fixes issue #892)

### LINKED ISSUE #892: Leak in Worker::mapChannelRequestHandlers and other maps
### Your environment

- mediasoup version: 3.10.5

### Issue 1 description

By looking at where we call `OnChannelRequestHandlerRemoved()` I'm pretty sure that we are leaking an entry in `Worker::mapChannelRequestHandlers` when we close a `Router` via Node `router.close()` method and such a  `Router` contains `RtpObservers`.

`this->listener->OnChannelRequestHandlerRemoved(rtpObserver->id);` is just called in `Router` when receiving a `Channel::ChannelRequest::MethodId::ROUTER_CLOSE_RTP_OBSERVER` request. **However** we do **not** send that channel request when we close the `Router` directly via `router.close()`, so here a leak.

Regarding other entities (for instance, `Consumer`) we rely on `Router::OnTransportConsumerClosed()` to clean the entry in the map when the `Consumer` is closed, but right now I do not remember if such a `Router::OnTransportConsumerClosed` is called in **every case**. For example, if we call `transport.close()`:
- `OnTransportConsumerClosed()` is called in `Transport::CloseProducersAndConsumers()`, so the same: I'm not sure if `Transport::CloseProducersAndConsumers()` is always called when closing the `Transport` directly or when closing the `Router` directly`.

Some applies to `Producer` and `DataXxxxxxxx` classes, and not sure about transports.

### Issue 2 description

Now a different potential issue. Imagine this scenario:

- Worker is creating an entity (`Producer` or whatever).
- It calls `OnChannelRequestHandlerRemoved(producerId)` so it gets into the map.
- Next steps in `Producer` creation fail so the code throws.
- We have that `producerId` in the map forever.

### Proposal that would fix all issues

Entities inherit from `ChannelSocket` and so on. Such a class may implement a constructor and a destructor so the constructor is the only one that calls `OnChannelRequestHandlerAdded(this)` and the destructor calls `OnChannelRequestHandlerRemoved(this)`. We do something similar somewhere else to register and unregister things. I think this mechanism would be much safer and reliable than current one.

### LINKED ISSUE #895: `make fuzzer` no longer works (missing target in meson.build)
Basically that. `mediasoup-worker-fuzzer` target was not included in `meson.build`.

```bash
$ make fuzzer

ERROR: Can't invoke target `mediasoup-worker-fuzzer`: target not found
make: *** [fuzzer] Error 1
```

ibc:
- Fixes #892
- Bonus track: fixes #895

### Details
 
Introduce a new `ChannelMessageHandlers` in worker that handles the `Channel` and `PayloadChannel` request/notification handlers.

Entities that wish to receive requests/notifications from Node/Rust must register themselves in their class constructor and unregister in their destructors.

```c++
class ChannelMessageHandlers
{
public:
	static void RegisterHandler(
	  const std::string& id,
	  Channel::ChannelSocket::RequestHandler* channelRequestHandler,
	  PayloadChannel::PayloadChannelSocket::RequestHandler* payloadChannelRequestHandler,
	  PayloadChannel::PayloadChannelSocket::NotificationHandler* payloadChannelNotificationHandler);
	static void UnregisterHandler(const std::string& id);
```

The `RegisterHandler()` function is smart and unregisters the entity from all maps in case its `id` was already present in any of them.

### TODO

- Adapt tests in Rust so they take into account the new `channelRequestHandlers` object field in the `Worker` dump output (I was not able to figure out how to compare the new obtained objects in Rust land).


ibc:
@jmillan @nazar-pc this is ready for review. Unfortunately I need it merged today :(

nazar-pc (commented):
Makes sense to me, though I'm not a big fan of introduction of even more global data structures.

ibc:
> Makes sense to me, though I'm not a big fan of introduction of even more global data structures.

I would like to avoid those static maps and use a class instance instead, but that requires much more work and huge changes everywhere. But I'll think about it and will try to change it in the future. I won't forget.

ibc:
@nazar-pc approved? XD

nazar-pc (approved):
With a heavy :heart: (sorry :crab:, I'll fix you, I promise)

jmillan (commented):
:+1:

PR #893, opened 2022-08-23
Channel and PayloadChannel: Optimize message format and JSON generation

ibc:
- Replaces PR #870.

### TODO

- Rust layer.

PR #898, opened 2022-08-28
Fix Rust support after recent changes

nazar-pc:
First commit restores code unnecessarily removed in #893 without which Rust can't send anything to worker.

Second commit fixes Rust support after #893, there I had to change how requests and notifications look like a bit, as the result we have less code and it is a bit cleaner.

Third commit updates dump data structure and tests according to changes in #894.

Last commit bumps versions, dependencies and updates changelog so we can publish it.

### DIFF
--- a/worker/src/Channel/ChannelSocket.cpp
+++ b/worker/src/Channel/ChannelSocket.cpp
@@ -188,9 +188,13 @@ namespace Channel
 		uint32_t messageLen;
 		size_t messageCtx;
 
+		// Try to read next message using `channelReadFn`, message, its length and context will be
+		// stored in provided arguments.
 		auto free = this->channelReadFn(
 		  &message, &messageLen, &messageCtx, this->uvReadHandle, this->channelReadCtx);
 
+		// Non-null free function pointer means message was successfully read above and will need to be
+		// freed later.
 		if (free)
 		{
 			try
@@ -225,9 +229,11 @@ namespace Channel
 				MS_ERROR_STD("discarding wrong Channel request: %s", error.what());
 			}
 
+			// Message needs to be freed using stored function pointer.
 			free(message, messageLen, messageCtx);
 		}
 
+		// Return `true` if something was processed.
 		return free != nullptr;
 	}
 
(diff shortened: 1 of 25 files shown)


ibc (approved):
Thanks.

worker/src/Channel/ChannelSocket.cpp
```diff
@@ -191,6 +191,43 @@ namespace Channel
 		auto free = this->channelReadFn(
 		  &message, &messageLen, &messageCtx, this->uvReadHandle, this->channelReadCtx);
 
+		if (free)
```
ibc, line 198: Please add some inline doc here telling what it does.

ibc: Also, is this code executed when using Node? If so, could it be disabled via some define passed when building the worker?

nazar-pc: `ChannelSocket::CallbackRead` is not called by Node, this is just for direct communication with worker when we can make a function call. Will add some comments.

jmillan (approved):

ibc:
Any chance to have this merged tomorrow morning? I'd like to make a release for Node and I assume we need another one for Rust.

nazar-pc:
It is just the comments that are left. I'll try to write some within next few hours.

worker/src/Channel/ChannelSocket.cpp
nazar-pc: Do added comments make sense?

ibc: Yes, but I miss some comments telling that the block is only used in Rust due to how message channel works in Rust (within some process) compared to how it works in Node (inter-process comm).

nazar-pc: Hm, if you go to the place from which this function is called, it makes total sense

nazar-pc: I described the whole workflow in https://github.com/versatica/mediasoup/pull/666 that introduced this code

ibc: Merge please 😀

ibc:
I'll merge. Thanks for this.

ibc:
3.6.10 released

nazar-pc:
Rust 0.11.0 is out too

PR #902, opened 2022-09-07
Do not create a std::vector in Consumer::GetRtpStreams()

jmillan:
No need to allocate that memory every time the method is called, which is done every time a RTCP Compound packet is sent out. Return a reference to the original vector instead.

It was seen certain CPU usage on the allocation of this vector.


ibc (approved):
Are you sure the returned array is never modified by the parent? Note that making it const is not gonna prevent that.

jmillan:
> Are you sure the returned array is never modified by the parent? Note that making it const is not gonna prevent that.

The returned Vector is constant and cannot be modified.

The returned Vector's content (the `RtpStreamSend` pointers) are returned as they were.

The only change this PR does is avoiding constructing a new Vector each time this method is called, and returns a reference to the private one instead.

ibc:
Ok, already approved

PR #903, opened 2022-09-07
ObjectPoolAllocator

jmillan:
This is an extract from a bigger "Memory Optimizations" PR #675

It contains only the `ObjectPoolAllocator` part with a few exceptions:

* `RtpPacket::Parse()` does return a `RtpPacket` instance. In #675 it returned a shared pointer but since shared pointers are just needed for cloned packets (for retransmission), the former method does not need to create one.
* `RtcpCompoundPacket` is left as it was. This is, a **new** one is created when needed (this will be enhanced in a separate PR).

As per my tests I'm not seeing much difference with this and v3. There is maybe some noticeable CPU enhancement.

Please @nazar-pc , @Hartigan give it a try and instrument this if possible to see how it enhances comparing to v3. I'll do some more testing just to make sure this addition is worth it.

Performed performance unit tests:

Parsed RtpPackets | ObjectPoolAllocator (sec)  | new/delete (sec) |
|-------------------| -------------------------- | ----------------- |
| 10K                         |  0.0028                                | 0.0039                  |
| 100K                       |  0.027                                  | 0.038                    |
| 1M                           |  0.26                                    | 0.33                      |

Cloned RtpPackets | ObjectPoolAllocator (sec)  | new/delete (sec) |
|-------------------| -------------------------- | ----------------- |
| 10K                         | 0.0034                                |   0.0049               |
| 100K                       | 0.033                                  | 0.049                   |
| 1M                           | 0.33                                   | 0.48                      |

ObjectPoolAllocator is definitely more performant as it reuses memory instead of freeing & re-allocating it.


worker/include/RTC/RtpPacket.hpp
```diff
@@ -133,7 +142,7 @@ namespace RTC
 
 		static RtpPacket* Parse(const uint8_t* data, size_t len);
 
-	private:
+	public:
```
ibc, line 146: Why is this public now?

worker/include/Utils.hpp
```diff
@@ -1,7 +1,14 @@
 #ifndef MS_UTILS_HPP
 #define MS_UTILS_HPP
 
+#ifndef MS_CLASS
+#define MS_CLASS "Utils"
```
ibc: We cannot define this macro in a hpp file but just in cpp files so it's local to that cpp file. Otherwise every cpp file importing útiles has its MS_CLASS macro overrided. This breaks everything.  Please move these utils that need logging to the corresponding Utils/Xxxxx.cpp file.

ibc (changes requested):
Just a few things reviewed yet

worker/include/Utils.hpp
nazar-pc: BTW, why not using `#pragma once`? It is supported by all compilers we likely care about and from what I read can even help with compilation speed a bit.

Hartigan: @nazar-pc  `#pragma once` is non-standard feature and there is no guarantee that the compiler(gcc, clang, microsoft c++) will still support this feature.

nazar-pc: Theoretically yes, practically it was there for a long time and is not going anywhere. It is even present in files generated by Microsoft's IDEs if what I read is to be believed. So I think the risk of using it is extremely close to 0.

ibc: What is #pragma?

jmillan: Guys, this does not belong to the PR, sorry. It is a leftover and will be removed.

jmillan:
NOTE: Currently measuring the CPU time differences with pools VS new/delete.

worker/include/RTC/RtpPacket.hpp
jmillan: It's needed for `std::allocate_shared`

jmillan:
Converted to draft while re-testing re-profiling. Feel fee to try it out and review.

worker/include/RTC/RtpPacket.hpp
jmillan: ⬆️ Check `RtpPacket::Clone()`

worker/src/RTC/RtpPacket.cpp
```diff
@@ -9,6 +9,15 @@
 
 namespace RTC
 {
+	/* Static Class methods. */
+
+	void RtpPacket::Deallocate(RtpPacket* packet)
+	{
+		// Destroy and deallocate the RtpPacket.
+		packet->~RtpPacket();
```
Hartigan: @jmillan  I think [destroy](https://en.cppreference.com/w/cpp/memory/allocator_traits/destroy) is preferable than direct destructor call.
```
RtpPacket::AllocatorTraits::destroy(RtpPacket::Allocator::Pool, packet);
```

jmillan: Indeed, thanks @Hartigan

worker/src/RTC/RtpPacket.cpp
```diff
@@ -145,9 +165,12 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		// This is a cloned RtpPacket.
 		if (this->buffer)
 		{
-			delete[] this->buffer;
+			this->buffer->~array();
```
Hartigan: ```std::allocator_traits<Alloc>::destroy```

worker/src/RTC/Transport.cpp
```diff
@@ -30,6 +30,30 @@ namespace RTC
...
+		if (sent)
+		{
+			ctx->tccClient->PacketSent(ctx->packetInfo, DepLibUV::GetTimeMsInt64());
+
+			ctx->sentInfo.sentAtMs = DepLibUV::GetTimeMs();
+
+			ctx->senderBwe->RtpPacketSent(ctx->sentInfo);
+		}
+
+		OnSendCallbackCtx::Allocator::Pool.deallocate(ctx, 1);
```
Hartigan, line 46: @jmillan  looks like ```destroy``` of ```OnSendCallbackCtx``` is missing

worker/src/RTC/Transport.cpp
```diff
@@ -30,6 +30,30 @@ namespace RTC
...
+	}
+#else
+	void Transport::OnSendCallback(bool sent, OnSendCallbackCtx* ctx)
+	{
+		if (sent)
+			ctx->tccClient->PacketSent(ctx->packetInfo, DepLibUV::GetTimeMsInt64());
+
+		OnSendCallbackCtx::Allocator::Pool.deallocate(ctx, 1);
```
Hartigan, line 55: @jmillan missing ```destroy```

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -123,7 +123,10 @@ namespace RTC
 
 		auto* storageItem = this->buffer[0];
 
-		delete storageItem;
+		// Reset (free RTP packet) the old storage item.
+		storageItem->Reset();
```
Hartigan: @jmillan I think default destructor do same things, and we can remove ```Reset``` and use ```std::allocator_traits<Alloc>::destroy````

nazar-pc:
@Hartigan please send multiple comments as a review together rather than individually, it will reduce number of GitHub notifications everyone receives.

worker/src/RTC/RtpPacket.cpp
jmillan: Right :+1:

worker/src/RTC/Transport.cpp
jmillan: :+1:

worker/src/RTC/Transport.cpp
jmillan: yes :+1:

worker/src/RTC/RtpStreamSend.cpp
jmillan: Correct, Reset has the same body as the default destructor should have.

worker/include/ObjectPoolAllocator.hpp
```diff
@@ -0,0 +1,89 @@
+#ifndef MS_OBJECT_POOL_ALLOCATOR_HPP
+#define MS_OBJECT_POOL_ALLOCATOR_HPP
+
+// #define MS_ALLOCATOR_FREE_ON_RETURN 1
```
ibc, line 4: Is this ready or not? Shouldn't this be uncommented and the condition below removed and always do this instead?

jmillan: The flag, if activated will `free` the objects rather than returning them to the pool. This is, if this define is uncommented then the ObjectProolAllocator will just call `new` and `delete` as if it was not used.

ibc: 👍

worker/src/RTC/RtpPacket.cpp
```diff
@@ -117,7 +126,18 @@ namespace RTC
 		           payloadLength + size_t{ payloadPadding },
 		  "packet's computed size does not match received size");
 
-		return new RtpPacket(header, headerExtension, payload, payloadLength, payloadPadding, len);
+		auto* rtpPacket = RtpPacket::Allocator::Pool.allocate(1);
+		RtpPacket::AllocatorTraits::construct(
```
ibc, line 131: Blank line here.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -633,12 +657,14 @@ namespace RTC
...
 
-	RtpPacket* RtpPacket::Clone() const
+	std::shared_ptr<RtpPacket> RtpPacket::Clone() const
 	{
 		MS_TRACE();
 
-		auto* buffer = new uint8_t[MtuSize + 100];
-		auto* ptr    = const_cast<uint8_t*>(buffer);
+		auto* buffer = RtpPacket::BufferAllocator::Pool.allocate(1);
+		RtpPacket::BufferAllocatorTraits::construct(RtpPacket::BufferAllocator::Pool, buffer);
```
ibc, line 674: Blank line

worker/src/RTC/Transport.cpp
```diff
@@ -2945,6 +2943,8 @@ namespace RTC
 			// Indicate the pacer (and prober) that a packet is to be sent.
 			this->tccClient->InsertPacket(packetInfo);
 
+			auto* ctx = OnSendCallbackCtx::Allocator::Pool.allocate(1);
+			OnSendCallbackCtx::AllocatorTraits::construct(OnSendCallbackCtx::Allocator::Pool, ctx);
```
ibc, line 2948: Blank line

worker/test/src/RTC/TestRtpPacket.cpp
```diff
@@ -127,7 +129,7 @@ SCENARIO("parse RTP packets", "[parser][rtp]")
 		REQUIRE(packet->ReadAbsSendTime(absSendTime) == true);
 		REQUIRE(absSendTime == 0x65341e);
 
-		auto* clonedPacket = packet->Clone();
+		auto clonedPacket = packet->Clone();
```
ibc, line 132: Why not auto* anymore?

ibc (changes requested):
Some cosmetic changes requested and a question. Other than that it looks very good however I'm not in a good shape to properly review these huge changes, but I trust you guys.

Hartigan (approved):

worker/test/src/RTC/TestRtpPacket.cpp
jmillan: before this was representing a raw pointer, now it's a shared pointer instance. I believe is makes more sense since the shared pointer itself is not a pointer but it contains one.

ibc:
Which is the status of this PR?

jmillan:
It's ready to merge. It's working properly. Even though we may want to enrich he pool allocator in the future is a good starting point.

ibc:
Ok?

ibc:
Reminder: this PR needs love.

jmillan:
Yes, it's on my TODO

jmillan:
I'll revive this PR for the new year.

jmillan:
Guys, back to this topic.

I see the benefits of using an object allocator in favour of `new()` and `delete()`. On the other side I'm not 100% confortable with the given implementation here as we are also loosing track of used objects in case there is a memory leak anywhere. 

My opinion here is keeping the branch for future consultation, and spend the proper time on a solution with which we feel more confortable when the need for this performance enhancement comes.

My personal experience here is that using a different allocator such us [jemalloc](https://github.com/jemalloc/jemalloc) increases substantially the performance out of the box. I know these are two different topics, but anyone could use an alternative allocator transparently, according to their needs.

What do you think @ibc, @nazar-pc?

Hartigan:
Hi @jmillan 

I agree, I have good experience with `mimalloc` for example. We may create separate MR only with smart pointers for memory safety

Thanks

ibc:
Let's turn this PR into draft because it's obviously super outdated and should be done from scratch.

PR #911, opened 2022-09-27
Forward abs-capture-time RTP extension also in audio packets

ibc:
For some unknown reason (bug) we were not doing it...

### DIFF
--- a/worker/src/RTC/Producer.cpp
+++ b/worker/src/RTC/Producer.cpp
@@ -1303,21 +1303,6 @@ namespace RTC
 				bufferPtr += extenLen;
 			}
 
-			// Add http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time.
-			{
-				extenLen = 3u;
-
-				// NOTE: Add value 0. The sending Transport will update it.
-				uint32_t absSendTime{ 0u };
-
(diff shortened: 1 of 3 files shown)


worker/src/RTC/Producer.cpp
```diff
@@ -1303,6 +1303,37 @@ namespace RTC
...
+			{
+				std::memcpy(bufferPtr, extenValue, extenLen);
+
+				extensions.emplace_back(
+				  static_cast<uint8_t>(RTC::RtpHeaderExtensionUri::Type::ABS_CAPTURE_TIME),
+				  extenLen,
+				  bufferPtr);
+
+				// Not needed since this is the latest added extension.
+				// bufferPtr += extenLen;
```
nazar-pc: Looks like this was accidentally copy-pasted from below
```suggestion
				bufferPtr += extenLen;
```

jmillan (approved):

worker/src/RTC/Producer.cpp
nazar-pc: Edited above suggestion just now, it should move cursor forward

nazar-pc (approved):

ibc:
I removed abs-send-time exten in audio packets since such a header is just used for REMB which just works with video packets.

jmillan (approved):

PR #912, opened 2022-09-28
Fix consumer NACK with the specified sequence number packet but the retransmission fails

### LINKED ISSUE #906: consumer NACK with the specified sequence number packet but the retransmission fails
Hello, I want to inquire about a problem that NACK is sent but no retransmission succeed. 

The scenario is as follows: SDK pushes the stream to mediasoup server1, and mediasoup server2 receives this stream from mediasoup server1 and then restores the packet to a frame (mediasoup server2 as consumer and receive data through pipetransport).

When I set the packet loss rate to 10% (the rtt is small) in test enviroment, I found that mediasoup server2 would sometimes request a packet with a certain sequence number until it finally reached the maximum number of times, and mediasoup server1 received the NACK of this sequence number packet message, but the packet buffer of this sequence number was not found in the send buffer of the consumer. 

I continued to debug it and found that the problem is:

Suppose mediasoup server1 received RTP packets with sequence numbers 1, 2, 3, 4, and then received RTX packets with original sequence numbers 5 (the RTP packets with sequence numbers greater than 5 have not been received now), and later received RTP packet with sequence number 7. Because the sequence is out of order, nackGenerator will generate the NACK message of the seq=6 packet (the packet with the sequence number 5 will not be required to retransmit because it has been marked in the recoverList), but when the RTX packet with the original sequence number 5 is received, this sequence number is not in the In the nackList, so the packet with sequence number 5 was discarded and was not forwarded to the consumer (mediasoup server2). The seq=6 packet was finally retransmitted to the mediasoup server2 successfully, but the seq=5 packet was discarded, even if the mediasoup server2 kept requesting retransmission, it could not be retransmitted. 

image

I compared the processing logic of NackModule::OnReceivedPacket in nack_module.cc in webrtc, and found that in this case, webrtc just inserts the packet into the recoverList, does not require retransmission, but it does not discard the packet, and It is to continue to stuff data into the packetBuffer in RtpVideoStreamReceiver.

image

I would like to ask, what is the purpose of not forwarding this packet to consumers here? Is it better to return true here, thanks

penguinol:
- Fixes #906

libwebrtc may send duplicate rtx packets to estimate bandwidth. When there is packet loss, we may receive rtx packet with seq higher than rtp packet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### NEXT
+
+- C++: Update libsrtp to v3.0-alpha.
+
 ### 3.13.16
 
 - Node: Add new `worker.on('subprocessclose')` event ([PR #1307](https://github.com/versatica/mediasoup/pull/1307)).
--- a/worker/include/RTC/SrtpSession.hpp
+++ b/worker/include/RTC/SrtpSession.hpp
@@ -38,13 +38,13 @@ namespace RTC
 		~SrtpSession();
 
 	public:
-		bool EncryptRtp(const uint8_t** data, int* len);
-		bool DecryptSrtp(uint8_t* data, int* len);
-		bool EncryptRtcp(const uint8_t** data, int* len);
-		bool DecryptSrtcp(uint8_t* data, int* len);
+		bool EncryptRtp(const uint8_t** data, size_t* len);
+		bool DecryptSrtp(uint8_t* data, size_t* len);
+		bool EncryptRtcp(const uint8_t** data, size_t* len);
+		bool DecryptSrtcp(uint8_t* data, size_t* len);
 		void RemoveStream(uint32_t ssrc)
 		{
-			srtp_remove_stream(this->session, uint32_t{ htonl(ssrc) });
+			srtp_stream_remove(this->session, uint32_t{ htonl(ssrc) });
 		}
 
 	private:
--- a/worker/src/RTC/PipeTransport.cpp
+++ b/worker/src/RTC/PipeTransport.cpp
@@ -484,9 +484,9 @@ namespace RTC
 		}
 
 		const uint8_t* data = packet->GetData();
-		auto intLen         = static_cast<int>(packet->GetSize());
+		auto len            = packet->GetSize();
 
-		if (HasSrtp() && !this->srtpSendSession->EncryptRtp(&data, &intLen))
+		if (HasSrtp() && !this->srtpSendSession->EncryptRtp(&data, &len))
 		{
 			if (cb)
 			{
@@ -496,8 +496,6 @@ namespace RTC
 
 			return;
 		}
-
-		auto len = static_cast<size_t>(intLen);
 
 		this->tuple->Send(data, len, cb);
 
@@ -515,14 +513,12 @@ namespace RTC
 		}
 
 		const uint8_t* data = packet->GetData();
-		auto intLen         = static_cast<int>(packet->GetSize());
+		auto len            = packet->GetSize();
 
-		if (HasSrtp() && !this->srtpSendSession->EncryptRtcp(&data, &intLen))
+		if (HasSrtp() && !this->srtpSendSession->EncryptRtcp(&data, &len))
 		{
 			return;
 		}
-
-		auto len = static_cast<size_t>(intLen);
 
 		this->tuple->Send(data, len);
 
@@ -542,14 +538,12 @@ namespace RTC
 		packet->Serialize(RTC::RTCP::Buffer);
 
 		const uint8_t* data = packet->GetData();
-		auto intLen         = static_cast<int>(packet->GetSize());
+		auto len            = packet->GetSize();
 
-		if (HasSrtp() && !this->srtpSendSession->EncryptRtcp(&data, &intLen))
+		if (HasSrtp() && !this->srtpSendSession->EncryptRtcp(&data, &len))
 		{
 			return;
 		}
-
-		auto len = static_cast<size_t>(intLen);
 
 		this->tuple->Send(data, len);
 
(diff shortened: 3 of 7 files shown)


worker/src/RTC/NackGenerator.cpp
```diff
@@ -121,9 +120,8 @@ namespace RTC
 			if (it != this->recoveredList.begin())
 				this->recoveredList.erase(this->recoveredList.begin(), it);
 
-			// Do not let a packet pass if it's newer than last seen seq and came via
-			// RTX.
-			return false;
+			// Let the packet pass if it's newer than last seen seq and came via RTX.
```
jmillan: Please add a comment indicating that some implementations use RTX for probation and such packets may arrive before the corresponding RTP packet.

jmillan (approved):

jmillan:
@penguinol,

Would you please add a couple of TEST in the NackGenerator test file to confirm the fixes? 

https://github.com/versatica/mediasoup/blob/master/worker/test/src/RTC/TestNackGenerator.cpp

ibc:
I am a bit worried given that these changes make our `NackGenerator.cpp` file diverge from the one in `libwebrtc`. Are we sure these changes don't come with other implications such as wrong packet loss stats?

penguinol:
> I am a bit worried given that these changes make our `NackGenerator.cpp` file diverge from the one in `libwebrtc`. Are we sure these changes don't come with other implications such as wrong packet loss stats?

In fact libwebrtc will pass all rtx packets to decoder not matter whether it's ahead of rtp packet, as follow:
https://webrtc.googlesource.com/src/+/refs/heads/main/video/rtp_video_stream_receiver2.cc#645

But we need to drop the duplicate packets.

The RTX packets arrivered ahead will increase the number of repaired packet in statistic. But i think it will not effect the packet loss in RR.

ibc (commented):
I don't understand. If we let packet RTX which, once decoded, has real seq 1000, and later real packet seq 1000 arrives, it's gonna be discarded by the remote decoder. Do I miss something?

penguinol:
Yes, libwebrtc will ignore dulpicate rtp packet.
In fact, clients may also receive dulpicate packets when rtx is disabled, because we have to send rtp packet with same seq for retransmitting.

ibc:
What I mean is that mediasoup can not leave RTC with internal seq greater than highest seen pass because those packets do not contain a real video frame. If we let them pass them real packets with same seq sent later will be ignored by the receiver.

penguinol:
@ibc libwebrtc will resend rtp packet in rtx for probing before we sending nack, so there is valid video frame in these packets.
Libwebrtc may also send padding rtx packet, but the seqnum is unique, the seq num of rtp packet sending after padding rtx will +1.

If the situation you described does exist, discard rtx greater than highest seen also can not avoid it, for example:
Client send rtp(100), rtx(101, without valid rtp data), rtp(101), rtp(102)
We may received rtp(100), rtp(102), then we send nack(101), and then received rtx(101 without valid rtp data).

ibc:
> @ibc libwebrtc will resend rtp packet in rtx for probing before we sending nack, so there is valid video frame in these packets. Libwebrtc may also send padding rtx packet, but the seqnum is unique, the seq num of rtp packet sending after padding rtx will +1.
> 
> If the situation you described does exist, discard rtx greater than highest seen also can not avoid it, for example: Client send rtp(100), rtx(101, without valid rtp data), rtp(101), rtp(102) We may received rtp(100), rtp(102), then we send nack(101), and then received rtx(101 without valid rtp data).

So in summary:

1. In mediasoup we must be ready to receive RTX packets that contain a real seq **higher** than highest seen.
2. If those RTX-decoded packets contain payload (not just padding) they **must** be forwarded to consumers.
3. If those RTX-decoded packets do **NOT** contain payload (but just padding) then:
  3.1 they **must** also be forwarded to consumers, or
  3.2 or they can be discarded by mediasoup as far as such a seq number is properly excluded (`SeqManager` task) in `Consumer` side.

Is it correct?

BTW @ggarber, weren't you working in a PR that makes mediasoup **discard** packets with no payload? How could that affect here?

ibc:
@penguinol I've updated the title of this PR and the description by adding this:

- Fixes https://github.com/versatica/mediasoup/issues/906
- Fixes https://github.com/versatica/mediasoup/issues/1059

Is it correct?

ggarber:
> > @ibc libwebrtc will resend rtp packet in rtx for probing before we sending nack, so there is valid video frame in these packets. Libwebrtc may also send padding rtx packet, but the seqnum is unique, the seq num of rtp packet sending after padding rtx will +1.
> > If the situation you described does exist, discard rtx greater than highest seen also can not avoid it, for example: Client send rtp(100), rtx(101, without valid rtp data), rtp(101), rtp(102) We may received rtp(100), rtp(102), then we send nack(101), and then received rtx(101 without valid rtp data).
> 
> So in summary:
> 
> 1. In mediasoup we must be ready to receive RTX packets that contain a real seq **higher** than highest seen.
> 2. If those RTX-decoded packets contain payload (not just padding) they **must** be forwarded to consumers.
> 3. If those RTX-decoded packets do **NOT** contain payload (but just padding) then:
>    3.1 they **must** also be forwarded to consumers, or
>    3.2 or they can be discarded by mediasoup as far as such a seq number is properly excluded (`SeqManager` task) in `Consumer` side.
> 
> Is it correct?
> 
> BTW @ggarber, weren't you working in a PR that makes mediasoup **discard** packets with no payload? How could that affect here?

* The description of the behaviour you mention looks reasonable and correct to me.
* This change is in the producer side and my change in the consumer side so they should be totally compatible and it is great that they are orthogonal.   Don't check if it is padding or not in the producer and discard those in the consumer if needed.

penguinol:
This PR does not fix #1059, I will create another PR for #1059 later.

penguinol:
Update: Filter duplicate rtp packet.
If we received rtx packet first, and received rtp packet later, we may send duplicate rtp packet with same rtp seq and different tcc seq.
The duplicate rtp packet will be dropped by srtp of client(srtp_err_status_replay_fail) , and this will lead to tcc feedbacking wrong loss rate.
So we should filter duplicate rtp packet.

jmillan:
@penguinol, what's the status of this PR? Could you please update the issue description with a detailed description?

penguinol:
@jmillan Tried to merge newest code to this branch, but failed. I'll create another PR later.

PR #919, opened 2022-09-30
Consumer: use a bitset instead of a set for supported payload types

jmillan:
This is way more performant.
* Before the change, set.find() could take as much as half of the CPU time taken by RtpStreamSend::ReceivePacket().
* After the change, bitset[] CPU time is negligible.

Extra: It takes less space than the set.


ibc (approved):
Nice

PR #922, opened 2022-10-10
Backport changes to congestion control from libwebrtc

### LINKED ISSUE #849: min_pending_time should not be negative
https://github.com/versatica/mediasoup/blob/f23eceadc524c223b7bc4661cb8736e28f779820/worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/goog_cc_network_control.cc#L434

Should it be like following?
```
TimeDelta min_pending_time = max_recv_time - feedback.receive_time;
```

### LINKED ISSUE #872: signed-to-unsigned overflow in send_side_bandwidth_estimation.cc
### Issue description
From libwebrtc upstream:
https://bugs.chromium.org/p/webrtc/issues/detail?id=14272

mediasoup has the similar (but not exactly the same, probably an old version) code [here](https://github.com/versatica/mediasoup/blame/v3/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/send_side_bandwidth_estimation.cc#L394), the [fix](https://webrtc-review.googlesource.com/c/src/+/269002/2/modules/congestion_controller/goog_cc/send_side_bandwidth_estimation.cc) should still apply

sarumjanuch:
This MR will track progress of sync libwebrtc congestion control in mediasoup to catchup with all the changes for the last 2 years, and most importantly- loss based estimator v2


worker/deps/libwebrtc/libwebrtc/api/transport/network_types.h
```diff
@@ -34,9 +47,9 @@ struct StreamsConfig {
   Timestamp at_time = Timestamp::PlusInfinity();
   absl::optional<bool> requests_alr_probing;
   absl::optional<double> pacing_factor;
-  union {
-    absl::optional<DataRate> min_total_allocated_bitrate = absl::nullopt;
-  };
+
+  // TODO(srte): Use BitrateAllocationLimits here.
```
ibc: Can you please rewrite those TODO comments to "TODO: (srte) xxxx"? IDEs can detect and highlight TODO comments (via add-ons or natively) but for that they need to follow some syntax rules such as starting with "TODO: ".

ibc:
Amazing. Just to clarify, this PR:

- Updates libwebrtc code related to TCC and so on.
- Adds new loss based estimator v2 but doesn't use it yet.

Is it correct?

sarumjanuch:
That is 100% correct. This is very WIP, but first commits that compiles. I will continue working on this.

ibc:
Wondering if we should mark the PR as draft to avoid wrong expectations for other readers. Not critical at all.

sarumjanuch:
Feel free to modify as needed, my laptop already died. Dunno when i ll
return:(

jmillan:
Converted to `draft`

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+namespace webrtc {
+
+constexpr char RobustThroughputEstimatorSettings::kKey[];
+
+RobustThroughputEstimatorSettings::RobustThroughputEstimatorSettings(
+    const WebRtcKeyValueConfig* key_value_config) {
+  Parser()->Parse(
+      key_value_config->Lookup(RobustThroughputEstimatorSettings::kKey));
+  if (window_packets < 10 || 1000 < window_packets) {
+    MS_WARN_TAG(bwe, "Window size must be between 10 and 1000 packets");
```
ibc: Let's start debugs with lowcase please.

```suggestion
    MS_WARN_TAG(bwe, "window size must be between 10 and 1000 packets");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+    window_packets = 20;
+  }
+  if (max_window_packets < 10 || 1000 < max_window_packets) {
+    MS_WARN_TAG(bwe, "Max window size must be between 10 and 1000 packets");
```
ibc:
```suggestion
    MS_WARN_TAG(bwe, "max window size must be between 10 and 1000 packets");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+    max_window_packets = 500;
+  }
+  max_window_packets = std::max(max_window_packets, window_packets);
+
+  if (required_packets < 10 || 1000 < required_packets) {
+    MS_WARN_TAG(bwe, "Required number of initial packets must be between "
```
ibc:
```suggestion
    MS_WARN_TAG(bwe, "required number of initial packets must be between "
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+                           "10 and 1000 packets");
+    required_packets = 10;
+  }
+  required_packets = std::min(required_packets, window_packets);
+
+  if (min_window_duration < TimeDelta::Millis<100>() ||
+      TimeDelta::Millis<3000>() < min_window_duration) {
+    MS_WARN_TAG(bwe, "Window duration must be between 100 and 3000 ms");
```
ibc:
```suggestion
    MS_WARN_TAG(bwe, "window duration must be between 100 and 3000 ms");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+    min_window_duration = TimeDelta::Millis<750>();
+  }
+  if (max_window_duration < TimeDelta::Seconds<1>() ||
+      TimeDelta::Seconds<15>() < max_window_duration) {
+    MS_WARN_TAG(bwe, "Max window duration must be between 1 and 15 s");
```
ibc:
```suggestion
    MS_WARN_TAG(bwe, "max window duration must be between 1 and 15 seconds");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/acknowledged_bitrate_estimator_interface.cc
```diff
@@ -0,0 +1,93 @@
...
+    max_window_duration = TimeDelta::Seconds<5>();
+  }
+  min_window_duration = std::min(min_window_duration, max_window_duration);
+
+  if (unacked_weight < 0.0 || 1.0 < unacked_weight) {
+    MS_WARN_TAG(bwe, "Weight for prior unacked size must be between 0 and 1.");
```
ibc: No dot at the end of comments.

```suggestion
    MS_WARN_TAG(bwe, "weight for prior unacked size must be between 0 and 1");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/alr_detector.cc
```diff
@@ -83,25 +90,30 @@ void AlrDetector::OnBytesSent(size_t bytes_sent, int64_t send_time_ms) {
...
+  } else if (alr_budget_.budget_ratio() < conf_.stop_budget_level_ratio &&
              alr_started_time_ms_) {
     state_changed = true;
     alr_started_time_ms_.reset();
   }
-
-  if (state_changed)
-    MS_DEBUG_DEV("state changed");
+  if (state_changed) {
+		MS_DEBUG_TAG(bwe, "ALR State change");
```
ibc:
```suggestion
		MS_DEBUG_TAG(bwe, "ALR state change");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/alr_detector.cc
```diff
@@ -83,25 +90,30 @@ void AlrDetector::OnBytesSent(size_t bytes_sent, int64_t send_time_ms) {
...
 
 void AlrDetector::SetEstimatedBitrate(int bitrate_bps) {
   //RTC_DCHECK(bitrate_bps);
-  int target_rate_kbps =
-      static_cast<double>(bitrate_bps) * bandwidth_usage_ratio_ / 1000;
-  alr_budget_.set_target_rate_kbps(target_rate_kbps);
+
+	if (last_estimated_bitrate_ != bitrate_bps) {
+		last_estimated_bitrate_ = bitrate_bps;
+		MS_DEBUG_TAG(bwe, "Setting ALR bitrate to %d", bitrate_bps);
```
ibc:
```suggestion
		MS_DEBUG_TAG(bwe, "setting ALR bitrate to %d bps", bitrate_bps);
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/inter_arrival_delta.cc
```diff
@@ -0,0 +1,139 @@
...
+          current_timestamp_group_.send_time - prev_timestamp_group_.send_time;
+      *arrival_time_delta = current_timestamp_group_.complete_time -
+                            prev_timestamp_group_.complete_time;
+
+      TimeDelta system_time_delta = current_timestamp_group_.last_system_time -
+                                    prev_timestamp_group_.last_system_time;
+
+      if (*arrival_time_delta - system_time_delta >=
+          kArrivalTimeOffsetThreshold) {
+        MS_WARN_TAG(bwe, "The arrival time clock offset has changed (diff = %lld ms), resetting.",
```
ibc: No dots at the end of logs.

```suggestion
        MS_WARN_TAG(bwe, "the arrival time clock offset has changed (diff = %lld ms), resetting",
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/inter_arrival_delta.cc
```diff
@@ -0,0 +1,139 @@
...
+					          arrival_time_delta->ms() - system_time_delta.ms());
+        Reset();
+        return false;
+      }
+      if (*arrival_time_delta < TimeDelta::Zero()) {
+        // The group of packets has been reordered since receiving its local
+        // arrival timestamp.
+        ++num_consecutive_reordered_packets_;
+        if (num_consecutive_reordered_packets_ >= kReorderedResetThreshold) {
+          MS_WARN_TAG(bwe, "Packets between send burst arrived out of order, resetting. arrival_time_delta %lld send time delta %lld",
```
ibc:
```suggestion
          MS_WARN_TAG(bwe, "packets between send burst arrived out of order, resetting. arrival_time_delta %lld send time delta %lld",
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
```diff
@@ -49,60 +47,119 @@ size_t ReadTrendlineFilterWindowSize(
...
+constexpr char TrendlineEstimatorSettings::kKey[];
+
+TrendlineEstimatorSettings::TrendlineEstimatorSettings(
+    const WebRtcKeyValueConfig* key_value_config) {
+  if (key_value_config->Lookup(kBweWindowSizeInPacketsExperiment).find("Enabled") == 0) {
+    window_size = ReadTrendlineFilterWindowSize(key_value_config);
+  }
+  Parser()->Parse(key_value_config->Lookup(TrendlineEstimatorSettings::kKey));
+  if (window_size < 10 || 200 < window_size) {
+		MS_WARN_TAG(bwe, "Window size must be between 10 and 200 packets");
```
ibc:
```suggestion
		MS_WARN_TAG(bwe, "window size must be between 10 and 200 packets");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
```diff
@@ -49,60 +47,119 @@ size_t ReadTrendlineFilterWindowSize(
...
+    window_size = kDefaultTrendlineWindowSize;
+  }
+  if (enable_cap) {
+    if (beginning_packets < 1 || end_packets < 1 ||
+        beginning_packets > window_size || end_packets > window_size) {
+      MS_WARN_TAG(bwe, "Size of beginning and end must be between 1 and %d", window_size);
```
ibc:
```suggestion
      MS_WARN_TAG(bwe, "size of beginning and end must be between 1 and %d", window_size);
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
```diff
@@ -49,60 +47,119 @@ size_t ReadTrendlineFilterWindowSize(
...
+      enable_cap = false;
+      beginning_packets = end_packets = 0;
+      cap_uncertainty = 0.0;
+    }
+    if (beginning_packets + end_packets > window_size) {
+      MS_WARN_TAG(bwe, "Size of beginning plus end can't exceed the window size");
```
ibc:
```suggestion
      MS_WARN_TAG(bwe, "size of beginning plus end can't exceed the window size");
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
```diff
@@ -49,60 +47,119 @@ size_t ReadTrendlineFilterWindowSize(
...
+      enable_cap = false;
+      beginning_packets = end_packets = 0;
+      cap_uncertainty = 0.0;
+    }
+    if (cap_uncertainty < 0.0 || 0.025 < cap_uncertainty) {
+      MS_WARN_TAG(bwe, "Cap uncertainty must be between 0 and 0.025");
```
ibc:
```suggestion
      MS_WARN_TAG(bwe, "cap uncertainty must be between 0 and 0.025");
```

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.h
```diff
@@ -0,0 +1,183 @@
...
+  Timestamp last_time_estimate_reduced_ = Timestamp::MinusInfinity();
+  absl::optional<DataRate> cached_instant_upper_bound_;
+  std::vector<double> instant_upper_bound_temporal_weights_;
+  std::vector<double> temporal_weights_;
+  std::deque<BandwidthUsage> delay_detector_states_;
+  Timestamp recovering_after_loss_timestamp_ = Timestamp::MinusInfinity();
+  DataRate bandwidth_limit_in_current_window_ = DataRate::PlusInfinity();
+  bool limited_due_to_loss_candidate_ = false;
+	// MS_NOTE changed min bitrate from 1 to 100, this allows faster recover
+	// from huge drop.
```
ibc: Use spaces to indent in libwebrtc files.

```suggestion
  // NOTE: changed min bitrate from 1 to 100, this allows faster recover
  // from huge drop.
```

ibc (changes requested):
Cosmetic changes in logs (all diff provided).

ibc:
@sarumjanuch I'll commit above cosmetic suggestions 😀

worker/deps/libwebrtc/libwebrtc/api/transport/network_types.h
ibc:
```suggestion
  // TODO: (srte) Use BitrateAllocationLimits here.
```

ibc:
@sarumjanuch `libwebrtcFieldTrials` PR is merged in v3. Please, when appropriate follow steps at https://github.com/versatica/mediasoup/pull/960#issuecomment-1332517890

ggarber:
Apart from upgrading the libwebrtc version maybe you also want to include this bugfix in this PR: https://github.com/versatica/mediasoup/pull/962

sarumjanuch:
Hello @ggarber! The goal of back-porting was not just to simply upgrade libwebrtc version, but use Loss V2 estimator, multiple bugfixes was a bonus. When loss v2 estimator is enabled RR packets loss is ignored, loss v2 operates purely on twcc feedback. But, we also did tests with feedback_only flag month ago, and it was producing worse results. This was my graphs: 
image

I will try it with new loss estimator. Because right now we have beautiful graphs when loss limited, and still bad when delay limited.

sarumjanuch:
@ggarber You are more than welcome to checkout this branch and do testing and playing with params, if you want, you can contact me on telegram, viber, whatsapp: +380630593550. I would explain you the road we went so far, and some params of new loss estimator. The more eyes we will have here, the better will be the result.

sarumjanuch:
> @sarumjanuch `libwebrtcFieldTrials` PR is merged in v3. Please, when appropriate follow steps at [#960 (comment)](https://github.com/versatica/mediasoup/pull/960#issuecomment-1332517890)

@ibc Done.

vpalmisano:
> Because right now we have beautiful graphs when loss limited, and still bad when delay limited

Hi @sarumjanuch, can you briefly describe the scenario you are using for testing? Thanks

sarumjanuch:
Hello @vpalmisano! So initially we started testing with artificial congestion created by traffic shapers, but they are mostly useless, because they do not reflect how the real network queues behaves. Google CC algorithm relies on two main characteristics: packet loss and delay gradient values linear regression slope.  It is very hard to  simulate those with traffic shapers as main assumption is that either delay or packet loss start increasing near the top of our link capacity, and therefore turned into a signal to hold or decrease BW estimate. Traffic shapers create sudden bursts on top of the link capacity, in form of loss, because mostly they are simple leaky bucket implementation. GCC will just produce sawtooth signal for such circumstances, because it would be limited only by burst loss. The only useful scenario in such cases is to observer how GCC behaves in random loss environment. That said, our main goal to test GCC is to create real congestion and see how it behaves under this circumstances. To easily reproduce this scenario Jose created special branch https://github.com/versatica/mediasoup-demo/tree/jmillan/consumerReplicas. Which allow you to run a single  page with producer and then run consumer page, where you would be able to specify enough consumer replicas to create congestion in your downlink. I personally have a separate server on the wild internet where i run mentioned mediasoup demo  branch, and I am connecting to it from a different networks all over the place. So for example in my home network BW is limited by loss, on top of link capacity small loss noise start emerging, and los v2 estimator catches it very beautifully. On the other network that I am testing BW is Delay limited, and it is where i am having issues right now that i am trying to fix. So in short in my case I run two browser tabs on a separate machines, where one is pure isolated producer, and second is pure isolated consumer, that are connected to external server, and are sending traffic over the real network, then i pick consumerReplicas value big enough for GCC being continuously congested, and start sampling values for my charts. 
So two tabs url params might look like this:
Consumer: https://YOUR_SERVER/?roomId=test&produce=false&consumerReplicas=19
Producer: https://YOUR_SERVER/?roomId=test&consumer=false&consumerReplicas=19

vpalmisano:
Thanks for the explanation. It I've correctly understood, you are running:
- 1 single producer at HD resolution on machine A;
- 19 consumers running on machine B, consuming the same producer sent from A;
- both machines A and B run under your home network.
Is that correct?

sarumjanuch:
Now few words about importance of Loss v2 estimator. Loss v2 is a giant leap forward in comparison what GCC was doing before. Prior it loss adjustments to BW estimate were calculated here: https://github.com/sarumjanuch/mediasoup/blob/bwe_backport/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/send_side_bandwidth_estimation.cc#L522. Very simple and rudimentary formulas. This is how chart looked like prior Loss v2 when link capacity is loss limited:
image
This is because loss is not differentiated. Main assumption for Loss v2 is that loss can be separated in two categories: Instant Loss and Inherent Loss. Instant loss is basically bursts and Inherent loss is the one that is Inherent for a given link at certain period of time, both has limits, both by default 5% so over 5% of those limits  means we should decrease, and when near top of it we should hold. Loss v2 uses maximum likelihood estimation approach inside. Also it has a meaning candidates and their factors to pick BW estimate from. It is integrated with Delay estimator, Trendline estimator, and Probe estimator (not yet in this branch). And it produces [the minimum of three values](https://github.com/sarumjanuch/mediasoup/blob/bwe_backport/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.cc#L174).  Loss v2 estimator is currently active experiment in chrome, you can check it by taking a look at your [filed trials](https://webrtccourse.com/course/webrtc-codelab/module/fiddle-of-the-month/lesson/identifying-chrome-field-trials/). And is still evolving. Unfortunately it is not described at any papers, nor has documentation, so we are kinda in a wild west here. With loss v2 estimator, loss limited downlink estimation looks like:
image
And here Is the mix of loss and delay estimate, you can clearly see straight lines where there is loss present, still delay based estimation is far from being perfect:
image

sarumjanuch:
> Thanks for the explanation. It I've correctly understood, you are running:
> 
> * 1 single producer at HD resolution on machine A;
> * 19 consumers running on machine B, consuming the same producer sent from A;
> * both machines A and B runs under your home network.
>   Is that correct?

Generally yes. But they are connected to the outside server (I am in Ukraine, server in Frankfurt). But it is very different range of networks all over the place, and they all behave slightly different. As you may know we are having issues with electricity in Ukraine, because of war with Russia. So am testing this all over the place where I land with my laptop (different flavours of Wifi, 4G, 3G, Starlink, fiber, copper).

vpalmisano:
What are the minimum, maximum, average and total loss rates for both cases? It seems to me that the loss rates using the new version are higher.
Did you made a comparison running a concurrent TCP download?

sarumjanuch:
You can see loss percent (blue line) as a second column on y axis on a first chart, at every screenshot. Avarage loss rates might be higher, in case of loss limited BW estimate as loss v2 estimator allow certian amoun of random loss (Inherent loss). You could tune that param by modifying fieldtrial string, InherentLossUpperBoundOffset - default 0.05, meaning that estiamtor allows up to 5% random loss. There is an API now to pass and override default fieldTrial string right now. No I havent tested this yet in compeeting scenario.

balazskreith:
@sarumjanuch awesome work, I think it will improve the overall quality a lot! We will try to deploy it to a relatively separate grid next week and test it with various scenarios. Can you give an example of how to manipulate the `fieldTrial` string so we can have multiple scenarios?

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bandwidth_estimation.cc
```diff
@@ -173,8 +190,15 @@ void LossBasedBandwidthEstimation::Update(Timestamp at_time,
       !has_decreased_since_last_loss_report_ &&
       (at_time - time_last_decrease_ >=
        last_round_trip_time + config_.decrease_interval);
+  // If packet lost reports are too old, dont increase bitrate.
+  const bool loss_report_valid =
+      at_time - last_loss_packet_report_ < 1.2 * kMaxRtcpFeedbackInterval;
```
balazskreith, line 195: how the RTCP feedback calculated in mediaosup? If I remember RFC3550 calculate based on assumed bw. If we calculate based on that assumption, and the bw suddenly drop because of congestion collapse than the interval will increase massively which makes this loss_report_valid false. at this moment the interpretation collides with the behaviour of the rest of the code, so right now it does not cause any problem and only beauty suggestion is to add a debug log if loss_report_valid is false, so if we search something related to it in the future we might have a chance to find it easier.

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.cc
```diff
@@ -0,0 +1,1067 @@
...
+  }
+
+  return packet_results_summary;
+}
+
+double GetLossProbability(double inherent_loss,
+                          DataRate loss_limited_bandwidth,
+                          DataRate sending_rate) {
+  if (inherent_loss < 0.0 || inherent_loss > 1.0) {
+		MS_WARN_TAG(bwe, "Terent loss must be in [0,1]: %f", inherent_loss);
```
balazskreith, line 89: Suggestion: 
```
MS_WARN_TAG(bwe, "Inherent loss must be in [0,1]: %f", inherent_loss);
```

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/alr_detector.h
```diff
@@ -23,6 +24,19 @@
...
 
+class RtcEventLog;
+
+struct AlrDetectorConfig {
+  // Sent traffic ratio as a function of network capacity used to determine
+  // application-limited region. ALR region start when bandwidth usage drops
+  // below kAlrStartUsageRatio and ends when it raises above
+  // kAlrEndUsageRatio. NOTE: This is intentionally conservative at the moment
+  // until BW adjustments of application limited region is fine tuned.
+  double bandwidth_usage_ratio = 0.65;
```
balazskreith, line 35: might be worth adding a defined scalar value for the drop as well to trigger the detector.
for example, if the throughput is 90 and it drops to 65, it is a congestion collapse, but it will not trigger the detector, because the drop is not enough, but in reality its a significant drop.
on the other hand with higher throughput rate (like 4k) 35% drop is awfully lot, so to say if the drop is around 500k we should trigger this detector.

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/robust_throughput_estimator.cc
```diff
@@ -0,0 +1,189 @@
...
+  // RTC_DCHECK(std::is_sorted(packet_feedback_vector.begin(),
+  //                          packet_feedback_vector.end(),
+  //                          PacketResult::ReceiveTimeOrder()));
+  for (const auto& packet : packet_feedback_vector) {
+    // Ignore packets without valid send or receive times.
+    // (This should not happen in production since lost packets are filtered
+    // out before passing the feedback vector to the throughput estimator.
+    // However, explicitly handling this case makes the estimator more robust
+    // and avoids a hard-to-detect bad state.)
+    if (packet.receive_time.IsInfinite() ||
```
balazskreith, line 62: in case it happens, a warn log is helpful becasue it means somewhere else in the code there is a problem!

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/robust_throughput_estimator.cc
```diff
@@ -0,0 +1,189 @@
...
+
+    // Insert the new packet.
+    window_.push_back(packet);
+    window_.back().sent_packet.prior_unacked_data =
+        window_.back().sent_packet.prior_unacked_data *
+        settings_.unacked_weight;
+    // In most cases, receive timestamps should already be in order, but in the
+    // rare case where feedback packets have been reordered, we do some swaps to
+    // ensure that the window is sorted.
+    for (size_t i = window_.size() - 1;
```
balazskreith, line 75: I think the compiler optimizes it, but `--i` instead of `i--`. such a nuance that it might not be worth mentioning.

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
```diff
@@ -127,56 +184,74 @@ TrendlineEstimator::TrendlineEstimator(
...
+  ++num_of_deltas_;
+  num_of_deltas_ = std::min(num_of_deltas_, kDeltaCounterMax);
+  if (first_arrival_time_ms_ == -1)
+    first_arrival_time_ms_ = arrival_time_ms;
+
+  // Exponential backoff filter.
+  accumulated_delay_ += delta_ms;
+  // BWE_TEST_LOGGING_PLOT(1, "accumulated_delay_ms", arrival_time_ms,
+  //                      accumulated_delay_);
+  smoothed_delay_ = smoothing_coef_ * smoothed_delay_ +
```
balazskreith, line 229: Worth to note here, It is a very strong / simplified / practical assumption here that the delay change follows linear trend, as near to the bw the changes spurious, and more like a second order regression could fit. However it could give worse results than this, but what I try to point out here is the error rate of the regression should be surface in order to use it in decisions at higher level. (the less it fits to a linear regression the more we can be sure there is a problem)

worker/include/Settings.hpp
```diff
@@ -38,7 +38,9 @@ class Settings
 		uint16_t rtcMaxPort{ 59999u };
 		std::string dtlsCertificateFile;
 		std::string dtlsPrivateKeyFile;
-		std::string libwebrtcFieldTrials{ "WebRTC-Bwe-AlrLimitedBackoff/Enabled/" };
+		std::string libwebrtcFieldTrials{
+			"WebRTC-Bwe-AlrLimitedBackoff/Enabled/WebRTC-Bwe-LossBasedBweV2/Enabled:true,CandidateFactors:1.02|1.0|0.95,DelayBasedCandidate:true,HigherBwBiasFactor:0.0002,HigherLogBwBiasFactor:0.02,ObservationDurationLowerBound:250ms,InstantUpperBoundBwBalance:75kbps,BwRampupUpperBoundFactor:1000000.0,InstantUpperBoundTemporalWeightFactor:0.9,TemporalWeightFactor:0.9,MaxIncreaseFactor:1.3,NewtonStepSize:0.75,InherentLossUpperBoundBwBalance:75kbps,LossThresholdOfHighBandwidthPreference:0.15,NotIncreaseIfInherentLossLessThanAverageLoss:true,TrendlineIntegrationEnabled:true,SendingRateSmoothingFactor:0.6/"
```
balazskreith: This string has more importance than at this moment it seems. Is it possible to separate the key,values with a newline instead of a `|`? Though its beauty, but its really hard to see what is the config default from the code at this point.

nazar-pc:
@balazskreith please leave multiple comments as review instead of individual comments next time, it helps with email notifications a lot.

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bandwidth_estimation.cc
sarumjanuch: This is old attempt of loss based BW estimate. It is disabled by default and is not used. Can be turned on by field trial. So we have:
- Loss v1 BW estimator:
    I have tested it and it was producing bad values.
- Loss v2 BW estimator:
    The one we are trying to use. It does not use RR RTC feedback. I relies only on TWCC RTCP feedback. When enabled, default estimator is ignored.
[- Default Loss estimation:](https://github.com/sarumjanuch/mediasoup/blob/bwe_backport/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/send_side_bandwidth_estimation.cc#L522)
    Too simple to be used, as it odes not differentiate between random and burst, and has not very meaningful boundaries.

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/robust_throughput_estimator.cc
sarumjanuch: Robust throughput estimator is meant to be a better alternative to acknowledged bitrate estimator. But is disabled by default in favour of acknowledged bitrate estimator. Controlled by filed trial. But I cant see any ongoing Chrome expiriment for it, and have not tested to see the difference.

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
sarumjanuch: Right now I want to build a dashboard which will allow to observe all the estimators values, and trendline estimator is my first candidate to observe because delay based estimator and loss v2 in some sense relies on it's values, and by only observing the log of how it's state changes I didn't liked it, but need visual representation to confirm it..

worker/include/Settings.hpp
sarumjanuch: We will need to modify libwebrtc parser for this, but i will take a look.

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/robust_throughput_estimator.cc
sarumjanuch: Read this for a reference:
https://bugs.chromium.org/p/webrtc/issues/detail?id=10274
https://bugs.chromium.org/p/webrtc/issues/detail?id=13402

sarumjanuch:
> Can you give an example of how to manipulate the `fieldTrial` string so we can have multiple scenarios?

createWorker has a new param now, which allow you to override: https://github.com/versatica/mediasoup/blob/b9a11610c7703f8ddf94058c8d1dbc07962f9487/node/src/Worker.ts#L75

The format is libwebrtc dictated ${FLAG}${DELIMITER}${FLAG_PARAMS}${DELIMITER}.

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bandwidth_estimation.cc
balazskreith: Okay, so I am lost in the code a bit. Does this mean it won't have any effect because another estimator is used after this PR? In this case should this code marked somehow as deprecated?

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/robust_throughput_estimator.cc
balazskreith: Right! Only suggested a warn log in case it is invoked especially with a packet does not have a receive_time. If this part of the code is not invoked for the new estimator, should it be marked as deprecated?

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
balazskreith: I see, and makes sense. For validation purposes still make sense to measure the error rate, so how well the points fits to the.trend if the points are perfectly aligned then the error is 0, if they are very fuzzy around the assumed slope the error rate is high, which makes sense in validation too because it gives some information.

worker/include/Settings.hpp
balazskreith: Might not need to change the libwebrtc parse, though it is ported, might just make a function replace every newline character to a |. But the this is stupidly minor, if a documentation surfaces these metrics with a default value that's the best! and then we don't need to change here at all.

worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.cc
balazskreith: > > Can you give an example of how to manipulate the `fieldTrial` string so we can have multiple scenarios?
> 
> createWorker has a new param now, which allow you to override:
> 
> https://github.com/versatica/mediasoup/blob/b9a11610c7703f8ddf94058c8d1dbc07962f9487/node/src/Worker.ts#L75
> 
> The format is libwebrtc dictated FLAG{DELIMITER}${FLAG_PARAMS}${DELIMITER}.

Thanks!

worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc
sarumjanuch: So let me rephrase, in order to be sure that I understand:  You are telling that we should also calculate R squared or Standart Error here, and the farer it is from 0, the higher chances of we are overusing?

balazskreith: No, not necessary. it is only makes sense together with the calculated trend. Higher error rate means higher fluctuation. if we don't have trend, then we probably have in this case high jitter and the playout should solve the problem. What might be a useful information from error rate for congestion control if the trend is not high is the presence of a concurrent flow. 

if we have trend and the error rate is low in this case you either have incipient congestion when the delay is being increased quasi in a linear fashion. I think majority of the cases the error rate will be high but what could be interesting to see if is there a correlation between the severity of the congestion and the error rate. We know for sure trend is high in case of congestion, but can we have an information how bad is the situation. Just a thought.

vsviatetskyi:
@sarumjanuch hi, thanks for the summary of Loss-Based BWE findings. Current WebRTC's estimator will collapse to 100 kbps if there's a 10% loss for a long period of time, so it's great to see that with Loss-Based BWE estimator can maintain 40 mbps value and not drop. Looking forward to Google making this experiment a default behavior.

50ed9689-2091-40c7-b1b0-345cb9d599bd

sarumjanuch:
Hello @vsviatetskyi, I believe i have seen your questions on google webrtc groups. New loss estimator can be adjusted to tolerate this by changing this flags over filed trial setting when creating new worker worker. Same you can do by adjusting field trial string when running chrome.
This one controls inherent loss: https://github.com/sarumjanuch/mediasoup/blob/bwe_backport/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.cc#L310
This one control instant bursts (average loss over observation period window): https://github.com/sarumjanuch/mediasoup/blob/bwe_backport/worker/deps/libwebrtc/libwebrtc/modules/bitrate_controller/loss_based_bwe_v2.cc#L329
 Both by default is 5% You can play with this params to see how the suits your use case, please share you experience.

node/src/Worker.ts
```diff
@@ -63,7 +63,7 @@ export type WorkerSettings =
 	 * certificate is dynamically created.
 	 */
 	dtlsPrivateKeyFile?: string;
-
+	/* eslint-disable max-len */
```
ibc, line 66: Please don't use this but `eslint-disable-next-line` instead and put it in the line above the problematic one.

Using `eslint-disable xxxx`  means that `xxx` rule gets disabled for the **whole file**.

Alternatively you can also use `eslint-disable xxxx` but then you MUST write `eslint-enable xxxx` after the problematic code.

ibc (dismissed):

node/src/Worker.ts
sarumjanuch: that was by mistake in latest commit it is fine. 
https://github.com/versatica/mediasoup/blob/becaef171cc512fe46523c2f2280a918d4601a59/node/src/Worker.ts#L66

sarumjanuch:
@ggarber After additional testing and this branch includes fix from #962.

sarumjanuch:
For the curious: Loss v2 Estimator successfully passed and will be enabled by default in some next chrome releases:
https://webrtc.googlesource.com/src/+/e04726281c3110718f5df9d4e8780229a23234e6%5E%21/

umeshc:
Is there a timeline for when this PR can be merged into the mainline code? For the past 3 weeks, not had much activity, Is anything that I can be of any help with?

sarumjanuch:
@umeshc That is github showing wrong info, i was committing code yesterday, and will commit today. We are very actively testing everything to pick right params, don't worry, it will be merged as soon as we will be confident in stability and satisfied with overall results.

sarumjanuch:
As of right now, this is blocked by #989

ibc:
May want to see this https://github.com/versatica/mediasoup/pull/1031

vpalmisano:
Do you have any update/plan for this PR? Thanks!

ibc:
> Do you have any update/plan for this PR? Thanks!

We are in an offsite this week, let's come to this next week after Monday.

ibc:
> Do you have any update/plan for this PR? Thanks!

So current state is on hold. This PR indeee modernized libwebrtc but it also comes with too many custom changes due to libwebrtc assumptions and hardcoded values, making it harder to maintain if we merge it. So we need to discuss how to proceed with this. Will update as soon as there is news.

ggarber:
@sarumjanuch FYC Given how awesome (and risky) is this change I think it would be great if there is a simple doc / readme with this info:
* What is the exact version that is being used in this PR (is the one from libwebrtc 8mo ago?) and what files have been upgraded vs non-upgraded.
* What are the high level differences compared with the previous version (f.e. loss based estimator) and the impact (pros and cons).  
* What changes have been made in this PR compared with the official libwebrtc branch, both in terms of code and non default configuration.   Including the reason of those changes as well as the testing done and implications.
* What is the testing that has been done in this branch (what network scenarios) and what are the results of those tests compared with the same tests in the v3 branch.

That info would be great for people like me to understand the implications of the change and maybe could help to move it forward.

Just an idea

jmillan:
> That info would be great for people like me to understand the implications of the change and maybe could help to move it forward.

@ggarber, @vpalmisano  and anyone interested. Here the report for all the research and changes @sarumjanuch did on this PR.

https://teamaround.notion.site/BWE-research-report-4e3a371d04fd4b7889c64c0321fa950c

Let's leave this PR as a space for any question or comment about it.

jmillan:
Closing this as there's an ongoing work porting/adapting the latest libwebrtc BWE estimation into mediasoup.

PR #931, opened 2022-10-25
Transport: remove duplicate call to method

jmillan:
We were calling twice Producer->GetRtcp() and Consumer->GetRtcp().

Those calls were not having any effect because the above methods just return if they are called before than expected, so the second call had no effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### NEXT
 
+* Transport: remove duplicate call to method (PR #931)
+
 
 ### 3.10.12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### NEXT
 
-* Transport: remove duplicate call to method (PR #931)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT RELEASE
+
+* Transport: remove duplicate call to method (PR #931)
```
ibc: Dot and capital letter :)

ibc (approved):

CHANGELOG.md
jmillan: Done, thanks.

ibc:
Conflicts

jmillan:
Fixed, I've force pushed this branch.

ibc (approved):

CosmosisT:
Would the be the culprit to sometimes having delayed frames discussed in forums?

ibc:
As the description of the PR says:

> Those calls were not having any effect because the above methods just return if they are called before than expected, so the second call had no effect.

PR #934, opened 2022-10-27
RTCP: Adjust maximum compound packet size

jmillan:
Consider IP and transport header sizes on the MTU in order to avoid creating bigger packets than the MTU.


ibc (approved):

ibc:
Does this deserve a new release?

PR #936, opened 2022-10-29
Fix type of `bufferedAmount` value in JSON

### LINKED ISSUE #930: Failed to parse notification error with webrtc datachannel
When using a webrtc datachannel I noticed with 0.11.1 rust crate each datachannel message causes an error in the log:
ERROR    mediasoup::router::data_consumer    - Failed to parse notification: invalid type: string "0", expected u32 at line 1 column 107

It looks to me like the Notification enum BufferedAmountLow member buffered_amount which is an i32 can no longer be deserialized from json containing a string instead of a number. 

It looks like this error wasn't there with the 0.10.0 rust crate

nazar-pc:
Fixes #930, which is a regression introduced in #893 where value changed from being a number to a string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * Transport: Remove duplicate call to method (PR #931).
 * RTCP: Adjust maximum compound packet size (PR #934).
+* Fix `bufferedAmount` type to be a number again (PR #936).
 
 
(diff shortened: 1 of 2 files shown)


ibc (approved):
Good thing that we are gonna get rid of JSON entirely soon...

PR #940, opened 2022-11-03
Fix clippy lints from the latest Rust release

nazar-pc:
Those in example apps are just suppressed


ibc (approved):
Thanks.

ibc:
Why do these things happen?
CleanShot-2022-11-03-at-20 05 08@2x

nazar-pc:
macOS was still failing, it is conditionally compiling slightly different code depending on platform. Since I'm on Linux I have not noticed one more place. Fixed and pushed, should be good now.

When one job fails, others get cancelled to not waste time.

PR #941, opened 2022-11-03
ActiveSpeakerObserver: Fix 'dominantspeaker' event

ibc:
- It must have a single `Producer` as argument rather than an array with a single `Producer` into it as per API [documentation](https://mediasoup.org/documentation/v3/mediasoup/api/#activeSpeakerObserver-on-producer):
  CleanShot-2022-11-03-at-19 40 30@2x
- Changes in Rust also done.


ibc:
CC @ggarber

ibc:
Mmmm, I think I was wrong and the Rust version is also emitting the 'dominantspeaker' event with an **object containing a Producer** as argument, am I right @nazar-pc?

CleanShot-2022-11-03-at-21 43 47@2x
CleanShot-2022-11-03-at-21 45 01@2x

ibc:
@nazar-pc I've applied the required changes in Rust 😀
https://github.com/versatica/mediasoup/pull/941/commits/8c68d2eeeef969ad73a2f153f7f9f46d7b1afaba

ibc:
I fail to understand how this can happen...

https://github.com/versatica/mediasoup/actions/runs/3390014781/jobs/5633764048

```
Downloading nlohmann_json source from https://github.com/nlohmann/json/releases/download/v3.10.5/include.zip
  Download size: 260679
  Downloading: ...
  A fallback URL could be specified using source_fallback_url key in the wrap file

  meson.build:169:0: ERROR: Incorrect hash for source:
   b94997df68856753b72f0d7a3703b7d484d4745c567f3584ef97c96c25a5798e expected
   eab65462e22c0c85f924b950bf69fc86adad4f078ff85da8594b6ffac3deaf0c actual.
```

nazar-pc:
> I fail to understand how this can happen...

GitHub is buggy sometimes, thankfully there is a hash check, try to re-run the job, it may pass on second attempt.

ggarber (approved):

ibc:
@nazar-pc like this PR? Approval? 😊

jmillan (approved):

rust/src/router/active_speaker_observer.rs
```diff
@@ -40,20 +40,10 @@ impl Default for ActiveSpeakerObserverOptions {
...
-}
-
 #[derive(Default)]
 #[allow(clippy::type_complexity)]
 struct Handlers {
-    dominant_speaker: Bag<
-        Arc<dyn Fn(&ActiveSpeakerObserverDominantSpeaker) + Send + Sync>,
-        ActiveSpeakerObserverDominantSpeaker,
-    >,
+    dominant_speaker: Bag<Arc<dyn Fn(&Producer) + Send + Sync>, Producer>,
```
nazar-pc, line 46: This feels like unnecessary breaking change to be honest.

node/src/ActiveSpeakerObserver.ts
```diff
@@ -28,12 +28,12 @@ export interface ActiveSpeakerObserverActivity
 
 export type ActiveSpeakerObserverEvents = RtpObserverEvents &
 {
-	dominantspeaker: [{ producer: Producer }];
+	dominantspeaker: [Producer];
```
nazar-pc, line 31: This is still an array despite changelog claims, what is the point of these changes?

ibc: It's not an array, it's the syntax of the arguments of the emitted events. Example:

CleanShot-2022-11-05-at-01 13 08@2x

rust/src/router/active_speaker_observer.rs
ibc: The API documentation said this since day 0. There is no reason to emit an `Object` with a `producer` key instead of a simple `Producer`. Nobody has complained and I've told about these changes to the authors of `ActiveSpeakerObserver`.

node/src/ActiveSpeakerObserver.ts
nazar-pc: I see, so the comment about array is misleading.

Well, I'd argue that it was better to fix the documentation rather than breaking API for everyone. People relied on `.producer` field of an object, now their code will stop working.

rust/src/router/active_speaker_observer.rs
nazar-pc: This is a breaking change though. The API looked one way, now it looks the other way. I'll release this as 0.12.0 instead of 0.11.x, but on TS side this will be shipped as supposedly backwards compatible change while it isn't, this is why I think we have an issue here.

nazar-pc: Also this is Rust code, it was always documented exactly the way code is written because documentation is auto-generated: https://docs.rs/mediasoup/latest/mediasoup/prelude/struct.ActiveSpeakerObserver.html#method.on_dominant_speaker

ibc: I see your point, but API documentation states that the event is emitted with a single `Producer` argument since the beginning (and that's what I expected when I wrongly reviewed the feature PR). If the code behaves different than the API docs IMHO it's a "bug".

node/src/ActiveSpeakerObserver.ts
nazar-pc: For TypeScript users (in strict mode) this will stop compiling, for JS users this will potentially break their apps.

rust/src/router/active_speaker_observer.rs
nazar-pc: It is a bug, but in the docs, not in the code. In Rust case there was no bug to begin with.

node/src/ActiveSpeakerObserver.ts
ibc: The Rust code mimics the corresponding TS code, but the TS code didn't match the expectation (API docs).

rust/src/router/active_speaker_observer.rs
ibc: ok, I'll revert it next week.

nazar-pc: I have a better option: on Rust side there is no issue in releasing breaking version of the crate. On TypeScript you can set `Producer.producer` field before firing an event. That way code will match docs, but old assumptions will still be maintained for those who relied on it. We can have a TODO to remove this workaround in v4. WDYT?

ibc: That's too complex and I'd just hate if we add a `producer` field into a `Producer`. These changes have been released less than 6 hours ago, IMHO it's much better to revert them and change/fix the API docs. I'll do tomorrow probably.

nazar-pc: NPM might have a feature to remove a version of the package

ibc: Things reverted in https://github.com/versatica/mediasoup/pull/947

node/src/ActiveSpeakerObserver.ts
ibc: Things reverted in https://github.com/versatica/mediasoup/pull/947

PR #942, opened 2022-11-04
ActiveSpeakerObserver: fix memory leak and more improvements

ibc:
- 🚨 When deleted, `ActiveSpeakerObserver` must iterate all entries in its `mapProducerSpeakers` and free them. This is because `ActiveSpeakerObserver::RemoveProducer()` **won't** be called for each existing `Producer` if the `ActiveSpeakerObserver` or its parent `Router` are directly closed. So fix a memory leak here.

- Make `mapProducerSpeakers` contain pointers to `ProducerSpeaker` as done everywhere else.

- If `AddProducer()` is called with an already paused `Producer`, consider it (because this can happen).

- Code cleanup such as moving class `static` members to the top, initialize numeric members, make `ProducerSpeaker` be a class instead of a struct, use `producerSpeaker` to name values in `this->mapProducerSpeakers` instead of naming them `rtpObserver`, etc.

### DIFF
--- a/worker/include/RTC/ActiveSpeakerObserver.hpp
+++ b/worker/include/RTC/ActiveSpeakerObserver.hpp
@@ -95,7 +95,7 @@ namespace RTC
 		std::string dominantId;
 		Timer* periodicTimer{ nullptr };
 		uint16_t interval{ 300u };
-		// Map of ProducerSpeakers indeed by Producer id.
+		// Map of ProducerSpeakers indexed by Producer id.
 		absl::flat_hash_map<std::string, ProducerSpeaker*> mapProducerSpeakers;
 		uint64_t lastLevelIdleTime{ 0u };
 	};
--- a/worker/include/RTC/ActiveSpeakerObserver.hpp
+++ b/worker/include/RTC/ActiveSpeakerObserver.hpp
@@ -90,7 +90,7 @@ namespace RTC
 		std::string dominantId;
 		Timer* periodicTimer{ nullptr };
 		uint16_t interval{ 300u };
-		// Map of ProducerSpeakers indeed by Producer id.
+		// Map of ProducerSpeakers indexed by Producer id.
 		absl::flat_hash_map<std::string, struct ProducerSpeaker> mapProducerSpeakers;
 		uint64_t lastLevelIdleTime{ 0u };
 	};
--- a/worker/include/RTC/ActiveSpeakerObserver.hpp
+++ b/worker/include/RTC/ActiveSpeakerObserver.hpp
@@ -54,8 +54,13 @@ namespace RTC
 			size_t nextLevelIndex{ 0u };
 		};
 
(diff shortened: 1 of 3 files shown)


worker/include/RTC/ActiveSpeakerObserver.hpp
```diff
@@ -84,13 +86,13 @@ namespace RTC
...
-		static constexpr int relativeSpeachActivitiesLen{ 3 };
-		double relativeSpeachActivities[relativeSpeachActivitiesLen];
-		std::string dominantId{ "" };
+		double relativeSpeachActivities[RelativeSpeachActivitiesLen];
+		std::string dominantId;
 		Timer* periodicTimer{ nullptr };
 		uint16_t interval{ 300u };
-		absl::flat_hash_map<std::string, struct ProducerSpeaker> mapProducerSpeaker;
-		uint64_t lastLevelIdleTime{ 0 };
+		// Map of ProducerSpeakers indeed by Producer id.
```
jmillan:
```suggestion
		// Map of ProducerSpeakers indexed by Producer id.
```

worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -308,10 +325,11 @@ namespace RTC
 			dominantSpeaker->EvalActivityScores();
 			double newDominantC2 = C2;
 
-			for (const auto& it : this->mapProducerSpeaker)
+			for (auto& kv : this->mapProducerSpeakers)
```
jmillan, line 325: can't it be `const` ?

worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -351,10 +371,11 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		for (const auto& it : this->mapProducerSpeaker)
+		for (auto& kv : this->mapProducerSpeakers)
```
jmillan, line 371: same, can't it be `const`?

jmillan (approved):
Just few cosmetic changes requested :+1:

worker/src/RTC/ActiveSpeakerObserver.cpp
ibc: This is for consistency with the rest of the code. If we change this we should change it everywhere in tons of files.

worker/src/RTC/ActiveSpeakerObserver.cpp
ibc: Same

ibc:
In fact I'll do more changes.

- Why are we allocating `ProducerSpeaker` instances directly in `mapProducerSpeakers` instead of storing pointers to `ProducerSpeaker`as we do everywhere?
- And why do we need a pointer to `Speaker` in `ProducerSpeaker` class instead of directly having a `Speaker`?

PR #945, opened 2022-11-04
tests: normalize hexadecimal data representation

jmillan:
Each byte is represented as 0x%02x separated with comma and space within each other.

We were mixing many styles: 0x, b, d...


ibc (approved):

nazar-pc:
It wasn't necessarily bad before. If something in the spec defines how bits should be arranged, it should be in binary in the codebase as well so it is readable. When you replace it with hex, you can't see with your eyes immediately what individual bits are. I guess whoever was writing it before was going from the spec and that is why it was the way it was.

jmillan:
While I understand your rationale, and I share it to some extent, there were up to three different ways to represent the content in the same buffer (hex, binary, decimal), which does not help readability. Less if there is no comment supporting what the value is representing.

I think it's sometimes desirable to see the binary value of a buffer position for the reasons you give, and we can add it if/when needed. On the other hand it's common for editors to show you a representation of a given value in different encodings, so I can see the binary, octal or decimal value of the hex value under the cursor.

PR #948, opened 2022-11-05
Meson: Pin ninja version to 1.10.2.4

ibc:
- Avoid issue in Linux using `ninja` latest version 1.11.1 which returns error code 245 instead of 0 when running `ninja --version`, which makes Meson fail.
- Thanks Vittorio Palmisano @vpalmisano.

More info:

- https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer
- https://github.com/ninja-build/ninja/issues/2211
- https://github.com/ninja-build/ninja/issues/2212

### DIFF
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -34,7 +34,6 @@ MESON_ARGS ?= ""
 PIP_BUILD_BINARIES = $(shell [ -f /etc/NIXOS -o -d /etc/guix ] && echo "--no-binary :all:")
 # Let's use a specific version of ninja to avoid issues like this one:
 # https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer/
-# https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer/
 NINJA_VERSION ?= 1.10.2.4
 
 # Disable `*.pyc` files creation.
--- a/worker/Makefile
+++ b/worker/Makefile
@@ -32,8 +32,10 @@ MESON_ARGS ?= ""
 # Workaround for NixOS and Guix that don't work with pre-built binaries, see:
 # https://github.com/NixOS/nixpkgs/issues/142383.
 PIP_BUILD_BINARIES = $(shell [ -f /etc/NIXOS -o -d /etc/guix ] && echo "--no-binary :all:")
-# Let's use a specific version of ninja to avoid issues like this one:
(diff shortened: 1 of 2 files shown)


nazar-pc (approved):
From Meson chat:
> it's because ninja returns "1.11.1.git.kitware.jobserver-1" for "ninja --version" and meson fails to parse that

worker/Makefile
```diff
@@ -32,6 +32,10 @@ MESON_ARGS ?= ""
 # Workaround for NixOS and Guix that don't work with pre-built binaries, see:
 # https://github.com/NixOS/nixpkgs/issues/142383.
 PIP_BUILD_BINARIES = $(shell [ -f /etc/NIXOS -o -d /etc/guix ] && echo "--no-binary :all:")
+# Let's use a specific version of ninja to avoid issues like this one:
+# https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer/
+# https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer/
```
nazar-pc:
```suggestion
# https://mediasoup.discourse.group/t/partly-solved-could-not-detect-ninja-v1-8-2-or-newer/
```

ibc: I wanted to paste the ninja GH issue... will fix

ibc:
Released in mediasoup node 3.11.2.

@nazar-pc I assume a mediasoup Rust version is also required.

nazar-pc:
Upstream issue: https://github.com/scikit-build/ninja-python-distributions/issues/157
Looks like in the end basic commands return non-zero status code :upside_down_face:

ibc:
Yup.

henryiii:
Could you retry? I think ninja may no longer support manylinux1 (centos 5).

henryiii:
This is fixed, I believe. manylinux2010 images are working elsewhere. The manylinux1 images might get removed at some point, but they actually do work on manylinux1 I believe (since cibuildwheel's tests passed). I've also been able to reproduce the failure and the fix in CI tests.

Also, please do not pin if there's a broken release & it's not an intentional breakage; instead just block out that release (`!=1.11.1` in this case). However, in this case, 1.11.1 itself was fixed, since it was not a code problem.

nazar-pc:
Thanks @henryiii, we've just being overly cautious by pinning to last known good version.

PR #958, opened 2022-11-20
Migrate Node tests to TypeScript

ibc:
*NOTE:* Had to move `node/tests` folder to `node/src/tests` since otherwise compiled JS would be in `node/lib/src` which is ugly.


jmillan (approved):

node/.eslintrc.js
```diff
@@ -59,7 +64,7 @@ const eslintConfig =
 			}
 		],
 		'keyword-spacing'      : 2,
-		'linebreak-style'      : [ 2, 'unix' ],
+		'linebreak-style'      : [ 2, isWindows ? 'windows' : 'unix' ],
```
nazar-pc, line 67: Why would it depend on OS thought? This will mean people contributing from different operating systems having different code style that either passes Windows CI or *nix CI.

ibc: This was for CI lint pass on Windows.

nazar-pc: Then it doesn't really make sense to me. Why would we want to check for different line breaks depending on OS check is running on?

ibc: Pretty sure that line endings change in place from LF to CRLF when fetching the git repo on Windows. AFAIR that could be a feature of git. Probably it can be disabled.

nazar-pc: Hm, I recall something like that too... Okay, thanks for clarifying.

ibc: Here: https://stackoverflow.com/questions/21822650/disable-git-eol-conversions

AFAIU it's a local setting so we cannot force the Windows machine in GH CI to behave as we would wish.

nazar-pc: I see, it is fine then, thanks!

PR #960, opened 2022-11-30
createWorker(): add libwebrtcFieldTrials option

ibc:
Will be mostly useful in https://github.com/versatica/mediasoup/pull/922 once merged.

Default value of `libwebrtcFieldTrials` is `"WebRTC-Bwe-AlrLimitedBackoff/Enabled/"`. If changed, it must be modified in:

- `worker/include/Settings.hpp` (here is where the real default value is set).
- `node/src/Worker.ts` (online doc).
- `rust/src/worker.ts` (online doc).

### Bonus Tracks

- Use `snprintf` and `vsnprintf` instead of `sprintf` and `vsprintf` because the latter are deprecated.

### DIFF
--- a/worker/src/DepUsrSCTP.cpp
+++ b/worker/src/DepUsrSCTP.cpp
@@ -5,6 +5,7 @@
 #include "DepLibUV.hpp"
 #include "Logger.hpp"
 #include <usrsctp.h>
+#include <cstdio> // std::vsnprintf()
 #include <mutex>
 
 /* Static. */
@@ -40,7 +41,7 @@ inline static void sctpDebug(const char* format, ...)
 	va_list ap;
 
 	va_start(ap, format);
-	vsprintf(buffer, format, ap);
+	vsnprintf(buffer, 10000, format, ap);
 
 	// Remove the artificial carriage return set by usrsctp.
 	buffer[std::strlen(buffer) - 1] = '\0';
--- a/worker/src/DepUsrSCTP.cpp
+++ b/worker/src/DepUsrSCTP.cpp
@@ -41,7 +41,7 @@ inline static void sctpDebug(const char* format, ...)
 	va_list ap;
 
 	va_start(ap, format);
-	vsnprintf(buffer, 10000, format, ap);
+	vsnprintf(buffer, sizeof(buffer), format, ap);
 
 	// Remove the artificial carriage return set by usrsctp.
 	buffer[std::strlen(buffer) - 1] = '\0';
--- a/worker/src/Settings.cpp
+++ b/worker/src/Settings.cpp
@@ -53,12 +53,13 @@ void Settings::SetConfiguration(int argc, char* argv[])
 	// clang-format off
 	struct option options[] =
 	{
-		{ "logLevel",            optional_argument, nullptr, 'l' },
-		{ "logTags",             optional_argument, nullptr, 't' },
-		{ "rtcMinPort",          optional_argument, nullptr, 'm' },
-		{ "rtcMaxPort",          optional_argument, nullptr, 'M' },
-		{ "dtlsCertificateFile", optional_argument, nullptr, 'c' },
-		{ "dtlsPrivateKeyFile",  optional_argument, nullptr, 'p' },
+		{ "logLevel",             optional_argument, nullptr, 'l' },
+		{ "logTags",              optional_argument, nullptr, 't' },
+		{ "rtcMinPort",           optional_argument, nullptr, 'm' },
+		{ "rtcMaxPort",           optional_argument, nullptr, 'M' },
+		{ "dtlsCertificateFile",  optional_argument, nullptr, 'c' },
+		{ "dtlsPrivateKeyFile",   optional_argument, nullptr, 'p' },
+		{ "libwebrtcFieldTrials", optional_argument, nullptr, 'W' },
 		{ nullptr, 0, nullptr, 0 }
 	};
 	// clang-format on
@@ -135,6 +136,14 @@ void Settings::SetConfiguration(int argc, char* argv[])
 			{
 				stringValue                                = std::string(optarg);
 				Settings::configuration.dtlsPrivateKeyFile = stringValue;
+
+				break;
+			}
+
+			case 'W':
+			{
+				stringValue                                  = std::string(optarg);
+				Settings::configuration.libwebrtcFieldTrials = stringValue;
 
 				break;
 			}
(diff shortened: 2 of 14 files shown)


worker/src/DepUsrSCTP.cpp
```diff
@@ -40,7 +41,7 @@ inline static void sctpDebug(const char* format, ...)
 	va_list ap;
 
 	va_start(ap, format);
-	vsprintf(buffer, format, ap);
+	vsnprintf(buffer, 10000, format, ap);
```
jmillan: 10000 characters? what's the rationale for this change?

jmillan (approved):

worker/src/DepUsrSCTP.cpp
ibc: pr-offtopic-changes

ibc: No change here other than `vsnprintf` instead of `vsprintf`. Buffer was already 10000 chars for whatever reason:
CleanShot-2022-11-30-at-18 36 39@2x

ibc:
@jmillan please when this is merged in v3, merge v3 into flatbuffers branch and ensure changes here are not "discarded" or whatever.

@sarumjanuch when this is merged in v3, merge v3 into your branch and ensure that default `libwebrtcFieldTrials` value is modified in all files as explained in the PR description.

worker/src/DepUsrSCTP.cpp
jmillan: The meme should say:

"While Making a PR I make unrelated changes just because, and reviewers must expand the diffs to understand my rationale and :0 "

jmillan:
> @jmillan please when this is merged in v3, merge v3 into flatbuffers branch and ensure changes here are not "discarded" or whatever.

They will not be discarded.

worker/src/DepUsrSCTP.cpp
ibc: 🤣🤣🤣🤣🤣

jmillan: BTW, that change is error prone. It should be `vsnprintf(buffer, sizeof(buffer), format, ap)`
 
Because  otherwise tomorrow, we change the size of `buffer` and forget to change `the second argument of vsnprintf` and we have a bug.

jmillan (commented):
Please change:

`vsnprintf(buffer, 10000, format, ap);`

to

`vsnprintf(buffer, sizeof(buffer), format, ap);`

ibc:
> Please change:
> 
> `vsnprintf(buffer, 10000, format, ap);`
> 
> to
> 
> `vsnprintf(buffer, sizeof(buffer), format, ap);`

Done: https://github.com/versatica/mediasoup/pull/960/commits/40195d5a2c23c85d3339948d5fa8b934056f6d65

worker/src/DepUsrSCTP.cpp
ibc: Done: https://github.com/versatica/mediasoup/pull/960/commits/40195d5a2c23c85d3339948d5fa8b934056f6d65

nazar-pc (commented):
This is not good in Rust (and TypeScript to be honest):
> NOTE: For advanced users only. An invalid value will make the worker crash.

Can't we have enums that list all applicable options instead of making developers guess what is valid and what isn't?

ibc:
> Can't we have enums that list all applicable options instead of making developers guess what is valid and what isn't?

This was my first attempt but it's completely impossible. Possible values, default values and syntax of many of these fields are completely undocumented in libwebrtc. Being honest, this feature is purely intended for testing the new BWE branch with different settings without having to rebuild mediasoup. Eugene (coworker of Jose and me at Around) has spent long months on it (see the referenced PR in the description of this PR). Here we are supporting exactly the same syntax that chrome supports in command line to set these field trials. We may not event announce/document this feature and it could be removed once we get the proper values.

nazar-pc:
Hm... but we're vendoring part of libwebrtc, surely we must be able to look into the code we've copied and create corresponding enum variants or am I missing something?
I're really prefer to have specific enums rather than saying "use this, but it will crash whenever it wants).

In particular, if we update libwebrtc version in the future and drop some field trial as the result, code will stop compiling rather than crashing in runtime.

nazar-pc:
If this is an internal feature we should mark it as such. In Rust it is done with `#[doc(hidden)]`, so it is not visible in IDE and not appearing in docs.

ibc:
> Hm... but we're vendoring part of libwebrtc, surely we must be able to look into the code we've copied and create corresponding enum variants or am I missing something? I're really prefer to have specific enums rather than saying "use this, but it will crash whenever it wants).

Please don't request this. We had yesterday a long meeting with Eugene (the author of https://github.com/versatica/mediasoup/pull/922) and, based on his knowledge and experience dealing with libwebrtc code, doing that is basically impossible. Honestly I don't want that mediasoup documents all libwebrtc field trials and validates them. Those are supposed to have a fixed/hardcoded chosen values (some as Chrome does). This feature just makes it easy for us, developers, play with different libwebrtc settings to decide which ones are the proper ones, and this is a long term effort.


> In particular, if we update libwebrtc version in the future and drop some field trial as the result, code will stop compiling rather than crashing in runtime.

"Fortunately" the crash doesn't happen "in real runtime" but when launching the worker. I'll add a warning log in case the app/user overrides default `libwebrtcFieldTrials` value.

ibc:
> If this is an internal feature we should mark it as such. In Rust it is done with `#[doc(hidden)]`, so it is not visible in IDE and not appearing in docs.

Done in https://github.com/versatica/mediasoup/pull/960/commits/d330313dc6075d9f33c827be8f6d5240eaf77146

worker/src/Settings.cpp
```diff
@@ -142,8 +142,16 @@ void Settings::SetConfiguration(int argc, char* argv[])
...
 			{
-				stringValue                                  = std::string(optarg);
-				Settings::configuration.libwebrtcFieldTrials = stringValue;
+				stringValue = std::string(optarg);
+
+				if (stringValue != Settings::configuration.libwebrtcFieldTrials)
+				{
+					MS_WARN_TAG(
+					  info,
+					  "overriding default value of libwebrtcFieldTrials may generate crashes in medoasoup-worker");
```
ibc: typo fixed in next pushed commit

sarumjanuch:
Hello @nazar-pc. We initially thought to do so, but based on libwebrtc code structure, it would require tons of efforts, and only few people out there would know what are the possible flags, and what are the possible params that are hiding those flags. Plus each param of the flag is then validated. When flag is removed it would not be an issue, the same is true regarding entering wrong values for params. Just take a look at this single flag and all the possible params for it https://webrtc.googlesource.com/src/+/refs/heads/main/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc#346, and validation of them https://webrtc.googlesource.com/src/+/refs/heads/main/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc#519, many type conversions and tricky parsing, imagine porting that to C++ and Rust to mediasoup land. But not only that, we would need to heavily modify libwebrtc code to do so. Most of them have meaningful defaults, and mediasoup lived with default settings for entire life. And there are many other files that has tuneable params with respective flags. Very few out there would be able to find out what they want to change and why, and now we at least provide them an ability to do so, without having their own branch. What we are doing right now, is we allow experienced users to adjust field trial string at runtime, because prior to that, it was only possible at compile time.

ibc:
I'll merge this. I satisfied given feedback as much as possible, but we cannot create an official and proper API in behalf of libwebrtc.

PR #964, opened 2022-12-05
Update OpenSSL to 3.0.7

### LINKED ISSUE #939: Update OpenSSL to (at least) 3.0.7
We use OpenSSL 3.0.2. We need 3.0.7 as it fixes this recent [CVE-2022-3786 and CVE-2022-3602](https://www.openssl.org/blog/blog/2022/11/01/email-address-overflows/). However last version of OpenSSL in https://mesonbuild.com/Wrapdb-projects.html is still 3.0.2-1, which is really outdated.

It's hard to create an openssl 3.0.7 wrap file since it depends on a patch that does very complex things. This is our current `openssl.wrap` file in `mediasoup/worker/subprojects/`:

```ini
[wrap-file]
directory = openssl-3.0.2
source_url = https://www.openssl.org/source/openssl-3.0.2.tar.gz
source_filename = openssl-3.0.2.tar.gz
source_hash = 98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63
patch_filename = openssl_3.0.2-1_patch.zip
patch_url = https://wrapdb.mesonbuild.com/v2/openssl_3.0.2-1/get_patch
patch_hash = 762ab4ea94d02178d6a1d3eb63409c2c4d61315d358391cdac62df15211174d4

[provide]
libcrypto = libcrypto_dep
libssl = libssl_dep
openssl = openssl_dep
```

It depends on a patch file fetched from https://wrapdb.mesonbuild.com/v2/openssl_3.0.2-1/get_patch which does too many things.  It was written by @nazar-pc: https://github.com/mesonbuild/wrapdb/commits/master/subprojects/openssl.wrap

Comment by @nazar-pc:

> As to OpenSSL update you'll need to update it in Meson wraps first unless someone else did it. According to https://mesonbuild.com/Wrapdb-projects.html version there is quite old and needs to be updated, you can see one of the PRs I did and follow readme as you already noticed. For instance here: https://github.com/mesonbuild/wrapdb/pull/281

The README of the patch:

<details>

# OpenSSL for Meson

## How this works?
TL;DR: this wrap abuses OpenSSL build system within Node.js.

Node.js has OpenSSL built-in with additional scripting around it to generate configs for GYP build system and thus bypass OpenSSL's native build system.

This wrap abuses that feature by replacing bundled OpenSSL with upstream version, patching mentioned mechanism to also generate a bunch of `meson.build` files for different platforms and uses top-level `meson.build` to wire everything together.

During installation unmodified Node.js tarball will be downloaded, its bundled OpenSSL will be replaced with upstream version and patched with `meson.build` files, enabling ability to build OpenSSL with Meson 🎉.

## How to update to newer release
Unless Node.js changes the mechanism we abuse above (unlikely, but possible, please check the diff between corresponding versions), `generator.sh` file can be used.

Just update OpenSSL version in wrap file, update Node.js version in `generator.sh` file to such that contains matching OpenSSL version bundled with it and run `generator.sh` from the root of the repository:
```
subprojects/packagefiles/openssl/generator.sh
```

Generated files in `generated-config` directory, after which you can try to build it. `create_release.py` will run it as part of the release process, so it doesn't need to be included in Git.

## Acknowledgement
This OpenSSL port wouldn't be possible without [Node.js project](https://github.com/nodejs/node) under [MIT license](https://github.com/nodejs/node/blob/master/LICENSE), whose OpenSSL build system was decomposed and heavily refactored.

</details>

So basically it depends on OpenSSL of Node project, which for now uses 3.0.5: https://github.com/nodejs/node/blob/main/deps/openssl/openssl/NEWS.md

**Summary:** We cannot even do this again since the version of OpenSSL in Node.js is not >= 3.0.7 yet so it doesn't fix the CVEs above.

More stuff: https://mesonbuild.com/Adding-new-projects-to-wrapdb.html

nazar-pc:
They have a new command in recent Meson version, `meson wrap update openssl` is what can be used now.

Resolves #939

### DIFF
--- a/worker/subprojects/openssl.wrap
+++ b/worker/subprojects/openssl.wrap
@@ -1,14 +1,14 @@
 [wrap-file]
-directory = openssl-3.0.2
-source_url = https://www.openssl.org/source/openssl-3.0.2.tar.gz
-source_filename = openssl-3.0.2.tar.gz
-source_hash = 98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63
-patch_filename = openssl_3.0.2-1_patch.zip
-patch_url = https://wrapdb.mesonbuild.com/v2/openssl_3.0.2-1/get_patch
-patch_hash = 762ab4ea94d02178d6a1d3eb63409c2c4d61315d358391cdac62df15211174d4
+directory = openssl-3.0.7
+source_url = https://www.openssl.org/source/openssl-3.0.7.tar.gz
+source_filename = openssl-3.0.72.tar.gz
+source_hash = 83049d042a260e696f62406ac5c08bf706fd84383f945cf21bd61e9ed95c396e
+patch_filename = openssl_3.0.7-1_patch.zip
+patch_url = https://wrapdb.mesonbuild.com/v2/openssl_3.0.7-1/get_patch
+patch_hash = 8f04d911dc22d1dddc6a192ab27d6d8275976a252bd9c73e09f95f1f927e42b5
+wrapdb_version = 3.0.7-1
 
 [provide]
 libcrypto = libcrypto_dep
 libssl = libssl_dep
 openssl = openssl_dep
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Node: Migrate tests to TypeScript (PR #958).
 * Node: Remove compiled JavaScript from repository and compile TypeScript code on NPM `prepare` script on demand when installed via git (PR #954).
 * `Worker`: Add `RTC::Shared` singleton for RTC entities (PR #953).
+* Update OpenSSL to 3.0.7.
 
 
 ### 3.11.3


worker/subprojects/openssl.wrap
```diff
@@ -1,14 +1,14 @@
...
-source_filename = openssl-3.0.2.tar.gz
-source_hash = 98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63
-patch_filename = openssl_3.0.2-1_patch.zip
-patch_url = https://wrapdb.mesonbuild.com/v2/openssl_3.0.2-1/get_patch
-patch_hash = 762ab4ea94d02178d6a1d3eb63409c2c4d61315d358391cdac62df15211174d4
+directory = openssl-3.0.7
+source_url = https://www.openssl.org/source/openssl-3.0.7.tar.gz
+source_filename = openssl-3.0.72.tar.gz
+source_hash = 83049d042a260e696f62406ac5c08bf706fd84383f945cf21bd61e9ed95c396e
+patch_filename = openssl_3.0.7-1_patch.zip
```
ibc, line 6: Just in case, shouldn't path version match opensssl version?

ibc (approved):

worker/subprojects/openssl.wrap
nazar-pc: Not sure what you mean exactly, can you elaborate?

nazar-pc:
Looks like we'll want to skip `subprojects` from CodeQL or at least `openssl` in there.

worker/subprojects/openssl.wrap
ibc: source_filename has version 3.0.72.
path_filename has version 3.0.7-1. I would expect 3.0.72-1 but it maybe perfectly ok, just in case.

nazar-pc: Patch name is only derived from version specified in wrapdb and has no direct connection to source code, source code can come from anywhere.

Also see https://github.com/mesonbuild/wrapdb/pull/806/files/308f3fc80531983f049087bf7efde4f1a223ba5e#diff-b0cd4975c76ae72ddd8217c684fedf270ef7c0b43df7a923367c913024c828e2

ibc:
> Looks like we'll want to skip `subprojects` from CodeQL or at least `openssl` in there.

AFAIR I did, but will check later today.

worker/subprojects/openssl.wrap
eli-schwartz: The source_filename has a typo, "72" instead of "7" -- as you can see, it doesn't match the remote download url.

Fortunately, the only actual effect of that is that the cache filename will be weird. It's still downloaded from the right place, checksummed with the right checksum, extracted to the right directory, and the wrapped subproject builds correctly.

The cache filename doesn't much matter IMHO.

eli-schwartz:
You can use e.g.

```yml
      with:
        config-file: .github/codeql/codeql-config.yml
```

And then specify:
```yml
paths-ignore:
  - 'worker/subprojects'
```
in the config file.

ibc:
> Looks like we'll want to skip `subprojects` from CodeQL or at least `openssl` in there.

I tried but I have no idea. I've created an issue and will investigate it some other day:

https://github.com/versatica/mediasoup/issues/966

PR #969, opened 2022-12-09
Fix crash when using WebrtcServer

penguinol:
When using WebrtcServer, the lifecycle of udpsocket maybe longer than WebrtcTransport.
If the rtp data is sent asynchronously, the WebrtcTransport may desotryed before the data has been sent.
It' will crash in the data sent callback, we should use weak_ptr to avoid this.


ibc:
Awesome! Thanks!

I've done some cosmetic changes so had to clone your branch into https://github.com/versatica/mediasoup/pull/970. Will merge that instead of this one.

PR #970, opened 2022-12-09
Fix crash when using WebRtcServer

ibc:
Copied with cosmetic changes from https://github.com/versatica/mediasoup/pull/969 by @penguinol

### DIFF
--- a/worker/include/RTC/Transport.hpp
+++ b/worker/include/RTC/Transport.hpp
@@ -307,10 +307,10 @@ namespace RTC
 		absl::flat_hash_map<uint32_t, RTC::Consumer*> mapSsrcConsumer;
 		absl::flat_hash_map<uint32_t, RTC::Consumer*> mapRtxSsrcConsumer;
 		Timer* rtcpTimer{ nullptr };
-		RTC::TransportCongestionControlClient* tccClient{ nullptr };
-		RTC::TransportCongestionControlServer* tccServer{ nullptr };
+		std::shared_ptr<RTC::TransportCongestionControlClient> tccClient{ nullptr };
+		std::shared_ptr<RTC::TransportCongestionControlServer> tccServer{ nullptr };
 #ifdef ENABLE_RTC_SENDER_BANDWIDTH_ESTIMATOR
-		RTC::SenderBandwidthEstimator* senderBwe{ nullptr };
+		std::shared_ptr<RTC::SenderBandwidthEstimator> senderBwe{ nullptr };
 #endif
 		// Others.
 		bool direct{ false }; // Whether this Transport allows PayloadChannel comm.
--- a/worker/src/RTC/Transport.cpp
+++ b/worker/src/RTC/Transport.cpp
@@ -2527,6 +2527,11 @@ namespace RTC
 			// Indicate the pacer (and prober) that a packet is to be sent.
 			this->tccClient->InsertPacket(packetInfo);
 
+			// When using WebRtcServer, the lifecycle of a RTC::UdpSocket maybe longer
+			// than WebRtcTransport so there is a chance for the send callback to be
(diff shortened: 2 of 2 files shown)


ibc:
CC @penguinol

nazar-pc:
FYI you don't need to close original PRs and open a new one, you can typically push commits into the original PR instead to make things cleaner. At least unless author of the PR explicitly disabled that.

nazar-pc:
Maybe, just maybe, one day worker will be written in Rust and we'll finally stop having memory issues of this kind. They seem to never end no matter how careful you are.

ibc:
> FYI you don't need to close original PRs and open a new one, you can typically push commits into the original PR instead to make things cleaner. At least unless author of the PR explicitly disabled that.

I use [`gh` command](https://cli.github.com/) to checkout the branch:

```bash
$ gh pr checkout 969
```

If I make changes and push them, somehow I get a copy of the branch created into versatica org. I think this is due to lack of permissions to modify the original branch.

nazar-pc:
I don't use GH CLI, what I did in the past was to manually add remote with a fork from which pull request was created, then I can just push commits into that remote branch with my SSH credentials.

It is also possible Junlong Zou disabled pushing into his fork intentionally, IDK, but that is not the default.

ibc:
> Maybe, just maybe, one day worker will be written in Rust and we'll finally stop having memory issues of this kind. They seem to never end no matter how careful you are.

Blaming the language is not always the correct way to go. In fact, in this case the issue is not related to the language. Same would have happened if the Worker is written in Rust or Node. The thing here is that we have a callback that is asynchronously called and it may be invoked after the `Transport` instance was closed. While closed it would call methods in private members of the closed `Transport` instance.

In C++ this means bad memory access because the `Transport` instance was decallocated/freed so their private members as well. In Rust or Node it would mean that we are calling methods on instances that are closed as well, and that's buggy despite there may not be bad memory access.

In Node or Rust (I hope) we can fix it by checking a `this.closed` flag within the callback (since the `Transport` instance wouldn't be garbage collected while there is a reference to it or a pending callback accessing its members). In C++ we cannot use a `this->closed` flag because such a flag would also be deallocated so it would point to random memory, and that's why weak ptr is needed. It's a different approach to solve the same problem, but it's not a language related problem.

nazar-pc:
> Blaming the language is not always the correct way to go. In

Absolutely true, though this is a classical example of an error that Rust memory prevents on fundamental level.

> Same would have happened if the Worker is written in Rust

No, safe Rust doesn't allow you to have pointers to invalid (deallocated) memory in the first place, if there is an issue of this kind, it will not compile at all.

> In Rust or Node it would mean that we are calling methods on instances that are closed as well, and that's buggy despite there may not be bad memory access.

Depends on how you design the API. In Rust when you write idiomatic code, you will likely physically have no way to call a method in that case. You'll be forced to check the invariant in this case and either print an error and exit gracefully or opt-in to crashing the whole app with a panic, but it will not happen implicitly on its own like in C++.

> In Node or Rust (I hope) we can fix it by checking a this.closed flag within the callback (since the Transport instance wouldn't be garbage collected while there is a reference to it or a pending callback accessing its members)

The way I designed Rust side is that if callback is running, Transport is not closed. When transport starts closing, you no longer have access to it by definition. In order to close transport you HAVE TO drop last reference to it. So API is basically designed to make it impossible to use incorrectly as a user. This approach is also largely applicable to internal code as well.

ibc:
> The way I designed Rust side is that if callback is running, Transport is not closed. When transport starts closing, you no longer have access to it by definition

What does it mean "by definition"?

nazar-pc:
> What does it mean "by definition"?

In order to close transport you HAVE TO drop last reference to it. If you don't have a reference, you can't call methods either :slightly_smiling_face:

PR #984, opened 2023-01-17
Fix wrong PictureID rolling over in VP9 and VP8

jcague:
VP9 and VP8 payloads include the PictureID. This field has a length of 15 bits, but VP8.cpp and VP9.cpp were using a SeqManager with uint16_t, which is 16 bits. 

This may cause different issues. I've reproduced very low framerates in VP9/SVC, but there could also be video artifacts, frozen videos, etc. VP8 also translates PictureIDs so that it could have an additional impact. The problem I saw is that we were marking all VP9/VP8 packets as `isOldPacket` after rolling over the PictureID. So we started to send higher temporal and spatial layers, and it broke VP9/SVC decoding.

I'm adding a bit length parameter when creating the SeqManager (`SeqManager<uint16_t, 15>`) to handle PictureID properly. I've added some tests for this case too.

### DIFF
--- a/worker/include/RTC/SeqManager.hpp
+++ b/worker/include/RTC/SeqManager.hpp
@@ -7,11 +7,11 @@
 
 namespace RTC
 {
-	template<typename T>
+	template<typename T, unsigned int N = 0>
 	class SeqManager
 	{
 	public:
-		static constexpr T MaxValue = std::numeric_limits<T>::max();
+		static constexpr T MaxValue = (N == 0) ? std::numeric_limits<T>::max() : ((1 << N) - 1);
 
 	public:
 		struct SeqLowerThan
--- a/worker/include/RTC/SeqManager.hpp
+++ b/worker/include/RTC/SeqManager.hpp
@@ -7,7 +7,9 @@
 
 namespace RTC
 {
-	template<typename T, unsigned int N = 0>
+	// T is the base type (uint16_t, uint32_t, ...).
+	// N is the max number of bits used in T.
+	template<typename T, uint8_t N = 0>
 	class SeqManager
 	{
 	public:
@@ -27,6 +29,7 @@ namespace RTC
 	private:
 		static const SeqLowerThan isSeqLowerThan;
 		static const SeqHigherThan isSeqHigherThan;
+		static T Delta(const T lhs, const T rhs);
 
 	public:
 		static bool IsSeqLowerThan(const T lhs, const T rhs);
--- a/worker/src/RTC/SeqManager.cpp
+++ b/worker/src/RTC/SeqManager.cpp
@@ -43,6 +43,7 @@ namespace RTC
 	T SeqManager<T, N>::Delta(const T lhs, const T rhs)
 	{
 		T value = (lhs > rhs) ? (lhs - rhs) : (MaxValue - rhs + lhs);
+
 		return value & MaxValue;
 	}
 
--- a/worker/src/RTC/SeqManager.cpp
+++ b/worker/src/RTC/SeqManager.cpp
@@ -7,39 +7,46 @@
 
 namespace RTC
 {
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	bool SeqManager<T, N>::SeqLowerThan::operator()(const T lhs, const T rhs) const
 	{
 		return ((rhs > lhs) && (rhs - lhs <= MaxValue / 2)) ||
 		       ((lhs > rhs) && (lhs - rhs > MaxValue / 2));
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	bool SeqManager<T, N>::SeqHigherThan::operator()(const T lhs, const T rhs) const
 	{
 		return ((lhs > rhs) && (lhs - rhs <= MaxValue / 2)) ||
 		       ((rhs > lhs) && (rhs - lhs > MaxValue / 2));
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	const typename SeqManager<T, N>::SeqLowerThan SeqManager<T, N>::isSeqLowerThan{};
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	const typename SeqManager<T, N>::SeqHigherThan SeqManager<T, N>::isSeqHigherThan{};
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	bool SeqManager<T, N>::IsSeqLowerThan(const T lhs, const T rhs)
 	{
 		return isSeqLowerThan(lhs, rhs);
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	bool SeqManager<T, N>::IsSeqHigherThan(const T lhs, const T rhs)
 	{
 		return isSeqHigherThan(lhs, rhs);
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
+	T SeqManager<T, N>::Delta(const T lhs, const T rhs)
+	{
+		T value = (lhs > rhs) ? (lhs - rhs) : (MaxValue - rhs + lhs);
+		return value & MaxValue;
+	}
+
+	template<typename T, uint8_t N>
 	void SeqManager<T, N>::Sync(T input)
 	{
 		// Update base.
@@ -52,7 +59,7 @@ namespace RTC
 		this->dropped.clear();
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	void SeqManager<T, N>::Drop(T input)
 	{
 		// Mark as dropped if 'input' is higher than anyone already processed.
@@ -62,13 +69,13 @@ namespace RTC
 		}
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	void SeqManager<T, N>::Offset(T offset)
 	{
 		this->base = (this->base + offset) & MaxValue;
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	bool SeqManager<T, N>::Input(const T input, T& output)
 	{
 		auto base = this->base;
@@ -105,8 +112,8 @@ namespace RTC
 
 		output = (input + base) & MaxValue;
 
-		T idelta = input - this->maxInput;
-		T odelta = output - this->maxOutput;
+		T idelta = SeqManager<T, N>::Delta(input, this->maxInput);
+		T odelta = SeqManager<T, N>::Delta(output, this->maxOutput);
 
 		// New input is higher than the maximum seen. But less than acceptable units higher.
 		// Keep it as the maximum seen. See Drop().
@@ -121,13 +128,13 @@ namespace RTC
 		return true;
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	T SeqManager<T, N>::GetMaxInput() const
 	{
 		return this->maxInput;
 	}
 
-	template<typename T, unsigned int N>
+	template<typename T, uint8_t N>
 	T SeqManager<T, N>::GetMaxOutput() const
 	{
 		return this->maxOutput;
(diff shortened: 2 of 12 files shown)


ibc:
Amazing. Let's review this week ASAP.

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -0,0 +1,76 @@
...
+		Codecs::VP9::EncodingContext context(params);
+		context.SyncRequired();
+		context.SetCurrentTemporalLayer(0);
+		context.SetTargetTemporalLayer(0);
+
+		context.SetCurrentSpatialLayer(0);
+		context.SetTargetSpatialLayer(0);
+
+		// Frame 1
+		auto forwarded = ProcessVP9Packet(context, 32767, 0);
```
jmillan: just for readability, can you use this representation for `pictureId`?

```c++
auto pictureId = (1 << 15) - 1;
```

Same for VP9 test.

worker/include/RTC/SeqManager.hpp
```diff
@@ -7,11 +7,11 @@
 
 namespace RTC
 {
-	template<typename T>
+	template<typename T, unsigned int N = 0>
```
jmillan: Using `uint8_t` type looks more concise IMHO and it's more than enough space to hold all the possible values.

```suggestion
	template<typename T, uint8_t N = 0>
```

worker/src/RTC/SeqManager.cpp
```diff
@@ -121,21 +121,22 @@ namespace RTC
...
+	template<typename T, unsigned int N>
+	T SeqManager<T, N>::GetMaxOutput() const
 	{
 		return this->maxOutput;
 	}
 
 	// Explicit instantiation to have all SeqManager definitions in this file.
 	template class SeqManager<uint8_t>;
 	template class SeqManager<uint16_t>;
+	template class SeqManager<uint16_t, 15>; // For PictureID (15bits).
```
jmillan:
```suggestion
	template class SeqManager<uint16_t, 15>; // For PictureID (15 bits).
```

jmillan (commented):
Very nice!

Could you please add a test for SeqManager too?

worker/include/RTC/SeqManager.hpp
jmillan: Also please add a little comment indicating that `N` represents the number of bits to be used in `T`.

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -586,6 +586,7 @@ namespace RTC
 				MS_DEBUG_TAG(rtp, "sync key frame received");
 
 			this->rtpSeqManager.Sync(packet->GetSequenceNumber() - 1);
+			this->encodingContext->SyncRequired();
```
ibc, line 589: Why is this added? It doesn't seem really related to changes in this PR. Is it a missing call?

@jmillan can you confirm please?

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -292,4 +292,32 @@ SCENARIO("process VP8 payload descriptor", "[codecs][vp8]")
...
+		RTC::Codecs::EncodingContext::Params params;
+		params.spatialLayers  = 0;
+		params.temporalLayers = 2;
+		Codecs::VP8::EncodingContext context(params);
+		context.SyncRequired();
+
+		context.SetCurrentTemporalLayer(0);
+		context.SetTargetTemporalLayer(0);
+
+		// Frame 1
```
ibc:
```suggestion
		// Frame 1.
```

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -292,4 +292,32 @@ SCENARIO("process VP8 payload descriptor", "[codecs][vp8]")
...
+		auto forwarded = ProcessPacket(context, 32767, 0, 0);
+		REQUIRE(forwarded);
+		REQUIRE(forwarded->pictureId == 1);
+		REQUIRE(forwarded->tl0PictureIndex == 1);
+
+		// Frame 2
```
ibc:
```suggestion
		// Frame 2.
```

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -292,4 +292,32 @@ SCENARIO("process VP8 payload descriptor", "[codecs][vp8]")
...
+		forwarded = ProcessPacket(context, 0, 0, 0);
+		REQUIRE(forwarded);
+		REQUIRE(forwarded->pictureId == 2);
+		REQUIRE(forwarded->tl0PictureIndex == 1);
+
+		// Frame 3
```
ibc:
```suggestion
		// Frame 3.
```

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+		Codecs::VP9::EncodingContext context(params);
+		context.SyncRequired();
+		context.SetCurrentTemporalLayer(0);
+		context.SetTargetTemporalLayer(0);
+
+		context.SetCurrentSpatialLayer(0);
+		context.SetTargetSpatialLayer(0);
+
+		// Frame 1
```
ibc:
```suggestion
		// Frame 1.
```

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -0,0 +1,76 @@
...
+		auto forwarded = ProcessVP9Packet(context, 32767, 0);
+		REQUIRE(forwarded);
+		REQUIRE(forwarded->pictureId == 32767);
+
+		// Frame 2
```
ibc:
```suggestion
		// Frame 2.
```

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -0,0 +1,76 @@
...
+		forwarded = ProcessVP9Packet(context, 0, 0);
+		REQUIRE(forwarded);
+		REQUIRE(forwarded->pictureId == 0);
+
+		// Frame 3
```
ibc:
```suggestion
		// Frame 3.
```

worker/src/RTC/SvcConsumer.cpp
jmillan: I believe this is OK, so we don't re-use any previous PictureId and completely start from a new one.

Is that the intention @jcague?

jcague: It's not exactly like that, but you are close. SvcConsumer uses the SeqManager only to know if a PictureID is old (it doesn't translate it as SimulcastConsumer does). The isOldPacket gets calculated with the MaxInput value inside the SeqManager. And it looks like the right way to initialize MaxInpu is by calling SeqManager::Sync().

This is what I inferred from the code yesterday. Do you think it makes sense?

worker/test/src/RTC/Codecs/TestVP8.cpp
jcague: Should I update the same comments in TestVP9?

jcague:
I've updated how we calculate deltas inside SeqManager to fix an issue I found executing long-running tests locally. And added more unit tests to cover that case properly.

worker/test/src/RTC/Codecs/TestVP8.cpp
ibc: Yes please, dot at the end in all comments. ☺️

ibc:
Just a question: somewhere we shift the payload or VP8 and VP9 (if the original received packet uses 8 bits for pictureId and the new computed pictureId value in the Consumer doesn't fit into 8 bits):
- Here we should shift the field if the new value doesn't fit into 7 bits and convert it into 16 bits and set first one to 1 to indicate that it's a 16 bits field, am I right?
- Or should we respect the size of the pictureId field in the original packet and just use 7 bits?

worker/src/RTC/SvcConsumer.cpp
jmillan: It makes sense to me

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -5,6 +5,8 @@
 
 using namespace RTC;
 
+constexpr uint16_t kMaxPictureId = (1 << 15) - 1;
```
jmillan: why the `k` prefix? :-). We don't use it in the codebase.

ibc: Please replace with `MaxPictureId`.

jcague: My fault. I did a quick search and found some of them. But they were inside the libwebrtc folder.

jcague:
SimulcastConsumer shifts the payload in VP8 because there is a translation. But SvcConsumer doesn't translate pictureIds in VP9/SVC, so I think there's no need to shift the payload and force the two-byte header.

ibc:
I've been told by @lminiero that "the only problem with 7-bit PictureID is that browsers often freeze video when they're used, because of a known bug, but no one uses 7-bit anymore apparently, older versions of ffmpeg were the only ones".

So I'd like to ensure what exactly we are doing when it comes to deal with 7bit pictureID.

CleanShot-2023-01-19-at-10 40 23@2x

So in VP8, we always force pictureId to be 15-bits and we properly set the first bit to 1 (so it indicates that 15 bits are used for the value). So IMHO we are good.

worker/src/RTC/SeqManager.cpp
```diff
@@ -7,43 +7,50 @@
...
 		return isSeqHigherThan(lhs, rhs);
 	}
 
-	template<typename T>
-	void SeqManager<T>::Sync(T input)
+	template<typename T, uint8_t N>
+	T SeqManager<T, N>::Delta(const T lhs, const T rhs)
+	{
+		T value = (lhs > rhs) ? (lhs - rhs) : (MaxValue - rhs + lhs);
+		return value & MaxValue;
```
ibc: Let's please always add a blank line before `return`.
```suggestion
		T value = (lhs > rhs) ? (lhs - rhs) : (MaxValue - rhs + lhs);

		return value & MaxValue;
```

ibc:
Please apply this diff in `CHANGELOG.md` file and IMHO we are done:

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index a632a5a4e..42076219a 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### Next
+
+* Fix wrong `PictureID` rolling over in VP9 and VP8 ([PR #984(https://github.com/versatica/mediasoup/pull/984) by @jcague).
+
+
 ### 3.11.5
 
 * Require Node.js >= 16 ([PR #973(https://github.com/versatica/mediasoup/pull/973)).
```

ibc (approved):
IMHO there is nothing missing in this PR and we are good. @jmillan can we merge?

lminiero:
> So in VP8, we always force pictureId to be 15-bits and we properly set the first bit to 1 (so it indicates that 15 bits are used for the value). So IMHO we are good.

That depends on what you're doing. If you're leaving the packet as it is, setting the M bit to 1 with a PictureID that is 7 bit, means the recipient will read the following byte as well thinking it's part of the PictureID, when it isn't. This will cause errors reading the payload too, as everything will be read shifted by a byte, which will cause ugly artifacts. The only way to really force 15-bit PictureID also when using 7-bit PictureID is to shift the whole payload after that point by 1 byte, so that you reserve room for a 15-bit PictureID.

ibc:
> That depends on what you're doing. If you're leaving the packet as it is, setting the M bit to 1 with a PictureID that is 7 bit, means the recipient will read the following byte as well thinking it's part of the PictureID, when it isn't. 

As per attached code screenshot above we don't do that. We always convert to 15-bits and set first bit to 1.

I mean, we never let a vp8 payload with 7-bits pictureId pass, we always convert it to 15-bits.

lminiero:
> As per attached code screenshot above we don't do that. We always convert to 15-bits and set first bit to 1.
> 
> I mean, we never let a vp8 payload with 7-bits pictureId pass, we always convert it to 15-bits.

Ack, apologies, I didn't look at the code :grin:

ibc:
> Ack, apologies, I didn't look at the code 😁

Ok, you didn't look at the text either 😂

> So in VP8, we always force pictureId to be 15-bits and we properly set the first bit to 1 (so it indicates that 15 bits are used for the value). So IMHO we are good.

jmillan (approved):

ibc:
Merging this. Thanks a lot @jcague

ibc:
Released in 3.11.6.

fippo:
> but no one uses 7-bit anymore apparently

Not after @ggarber made libwebrtc emit the two-byte version unconditionally a couple of years back

PR #985, opened 2023-01-23
Fix crash with invalid arrival_time in libwebrtc bwe

ggarber:
Tentative fix for #357

According to the assert `complete_time_ms` is <0 and it cannot be -1 because there is a check before for that.   So the value is probably int64.min().

This fix try to avoid ever setting `complete_time_ms` to any negative value what should prevent the crash from happening.


jmillan (approved):

ibc (approved):

nazar-pc:
What concerns me is the root cause of this. It does prevent crash, but why was it set to invalid value in the first place? Maybe #357 should not be closed yet?

ibc:
Since there is an ongoing effort to update libwebrtc, IMHO it's better to forget about this bug. If it happens once libwebrtc is updated in mediasoup, we'll have to investigate it from scratch anyway.

Issue #989, opened 2023-01-25
Inconsistency of Sending Bitate with Available Outgoing Bitrate


sarumjanuch:
During my work on BWE backport branch, I have noticed that mediasoup sometimes sends at higher bitrate than that which is reported by BWE algorithm. I tried to make this reproducible as much as possible so the setup is next: mediasoup-demo consumerReplicas branch running on the localhost, with two browser tabs opened on the same machine. Tab1: produce=false&consumerReplicas=19. Tab2: consumer=false&consumerReplicas=19. So we have single producer that produces simulcast stream, and 20 consumers replicas on the other side. Inside Trasnport.cpp I have hardcoded available bitrate to be 60 Mbps, it is never changing. Despite that this is what can be seen on graphs:
<img width="1256" alt="image" src="https://user-images.githubusercontent.com/4701803/214628046-82aa8a89-c674-452c-904c-f7a7f36f204c.png">

It seems that it correlates with packet loss:
<img width="1275" alt="image" src="https://user-images.githubusercontent.com/4701803/214628458-4ae4ba3e-bd4b-426e-8279-1c0ab7a25326.png">

In the same time RTX bitrate is relatively small to explain such a big gap:
<img width="1273" alt="image" src="https://user-images.githubusercontent.com/4701803/214628845-b1312a70-3c31-4429-8585-0f1188f5e737.png">

webrtc-internals send side:
<img width="639" alt="image" src="https://user-images.githubusercontent.com/4701803/214629380-1be0c543-5b0d-4043-a7b2-6e470150b913.png">

webrtc-internals receive side:
<img width="645" alt="image" src="https://user-images.githubusercontent.com/4701803/214629227-6a50d956-e65f-42d6-9bb4-7c0ba4638590.png">

Distribution of available bitrate seems to be fine per logs: https://pastebin.com/inTDYp2Q, but actual sending bitrate is not.

chrome://webrtc-intenals for one of this cases:  https://drive.google.com/file/d/1nvciwm9SW3KAUjAPHCQtMbQN9fs2YKar/view?usp=sharing

ibc:
Even if mic is muted it's still generating bitrate. Can this be repeated after closing the micProducer? CC.disableMic() in console in sender.

ibc:
More things to be asked by us:

1. Are we considering the RTX traffic sent from sender to mediasoup when we count the Producer bitrate?
[...]

PR #991, opened 2023-01-26
Fix issue #989 to properly allocate bitrate in simulcast and SVC.

sarumjanuch:
Mediasoup was wrongfully allocating bitrate for temporal layers, as it was getting bitrate for a specific temporal layer and not the sum of all temporal layers. Before fix, if we would look at required bitrate, we can see that layer 2:1 has lower bitrate that 2:0 which is wrong. :

`mediasoup:Channel [pid:14937] RTC::SimulcastConsumer::IncreaseLayer() | setting provisional layers to 2:0 [virtual bitrate:4405928, required bitrate:2261974] +0ms

  mediasoup:Channel [pid:14937] RTC::SimulcastConsumer::IncreaseLayer() | setting provisional layers to 2:1 [virtual bitrate:2143954, required bitrate:1951808] +0ms`
  
  After the fix we now see the proper values:
  ` mediasoup:Channel [pid:33559] RTC::SimulcastConsumer::IncreaseLayer() | setting provisional layers to 2:0 [virtual bitrate:11346749, required bitrate:1329069] +0ms
  
  mediasoup:Channel [pid:33559] RTC::SimulcastConsumer::IncreaseLayer() | setting provisional layers to 2:1 [virtual bitrate:10017680, required bitrate:3262380] +0ms
`
For VP9 it was even worse.

### DIFF
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -424,7 +424,7 @@ namespace RTC
 				}
 
 				requiredBitrate =
-				  this->producerRtpStream->GetLayerBitrate(nowMs, spatialLayer, temporalLayer);
+				  this->producerRtpStream->GetBitrate(nowMs, spatialLayer, temporalLayer);
 
 				MS_DEBUG_DEV(
 				  "testing layers %" PRIi16 ":%" PRIi16 " [virtual bitrate:%" PRIu32
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -423,8 +423,7 @@ namespace RTC
 					continue;
 				}
 
-				requiredBitrate =
-				  this->producerRtpStream->GetBitrate(nowMs, spatialLayer, temporalLayer);
+				requiredBitrate = this->producerRtpStream->GetBitrate(nowMs, spatialLayer, temporalLayer);
 
 				MS_DEBUG_DEV(
 				  "testing layers %" PRIi16 ":%" PRIi16 " [virtual bitrate:%" PRIu32
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -500,7 +500,7 @@ namespace RTC
 					continue;
 				}
 
-				requiredBitrate = producerRtpStream->GetLayerBitrate(nowMs, 0, temporalLayer);
+				requiredBitrate = producerRtpStream->GetBitrate(nowMs, 0, temporalLayer);
 
 				// This is simulcast so we must substract the bitrate of the current temporal
 				// spatial layer if this is the temporal layer 0 of a higher spatial layer.


jmillan (approved):

ibc (commented):
Let me please review this tomorrow morning (sick today). I really want to understand the wrong rationale fixed by this PR.

Good catch BTW

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -423,8 +423,7 @@ namespace RTC
 					continue;
 				}
 
-				requiredBitrate =
-				  this->producerRtpStream->GetLayerBitrate(nowMs, spatialLayer, temporalLayer);
+				requiredBitrate = this->producerRtpStream->GetBitrate(nowMs, spatialLayer, temporalLayer);
```
ibc, line 426: I'm looking at this and I don't understand the change. Here we need to compute the bitrate needed to upgrade to next layer so what we need to obtain is the bitrate of the next layer (the desired layer) rather than the total bitrate of the desired layer plus all layers below. This is the IncreaseLayer() method which means "hey take this available bitrate and use it to upgrade one layer and tell me how much bitrate you needed to upgrade to that layer".

NOTE: Not 100% sure of what I'm saying.

sarumjanuch: We have 3 spatial and 2 temporal layers.
so 2:1 is the highest. It must be a sum. i.e. 2:0 bitrate is 2.9 Mbps and 2:1 is 1.9 Mbps, but when switching to 2:1 we need to account for 4.8Mbps but mediasoup was accounting for 1.9 Mbps only which was leading to this issue

sarumjanuch: NOTE: I am also already not 100% sure about that.

ibc: > 2:0 bitrate is 2.9 Mbps and 2:1 is 1.9 Mbps, but when switching to 2:1 we need to account for 4.8Mbps 

This is what the change in the PR does but this is exactly what I mean is wrong. This consumer already used the 2.9MB of the 2:0 layer in the previous call to IncreaseLayer() so now that it upgrades to 2:1 it needs to use 1.9MB and that's what the method has to return now. In the previous call it already used and returned 2.9MB. I don't think this was wrong at all (note: I've not yet taken a look yet to changes in SimulcastConsumer).

sarumjanuch: @ibc Seems like you are right, more over, i did tests, and that is not fixing an issue((( Only makes it worse.

jmillan: Yep, definitely this PR is not OK. We are correctly increasing layer by layer already.

ibc:
I've reviewed the `IncreaseLayer()` methods in `SimulcastConsumer` and `SvcConsumer`. Both were ok and changes in this PR are wrong as explained in comments above. I'm not saying that there is no some issue somewhere, but it's not here. Let's close this PR.

PR #992, opened 2023-01-27
SimulcastConsumer::GetDesiredBitrate(): Choose the highest bitrate among all Producer streams

### LINKED ISSUE #989: Inconsistency of Sending Bitate with Available Outgoing Bitrate

ibc:
Instead of assuming that the highest simulcast stream is the one consuming highest bitrate, iterate all them and pick up the highest value.

This is because when the sender enables a higher stream, initially the bitrate of it could be less than the bitrate of a lower stream. In a few seconds it will become higher. However don't report its irrelevant current bitrate in the meantime.

*NOTE:* This PR is not intended to fix https://github.com/versatica/mediasoup/issues/989

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -14,9 +14,6 @@ jobs:
           - ubuntu-22.04
           - macos-12
           - windows-2022
-        rust:
-          - stable
-          # - nightly
 
     runs-on: ${{ matrix.os }}
 
@@ -24,20 +21,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          override: true
-          components: rustfmt, clippy
-
       - name: Configure cache
         uses: actions/cache@v2
         with:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ matrix.os }}-cargo-${{ matrix.rust }}-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ matrix.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
 
       - name: cargo fmt
         uses: actions-rs/cargo@v1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,178 +1,183 @@
 # Changelog
 
 
+### NEXT
+
+* `SimulcastConsumer::GetDesiredBitrate()`: Choose the highest bitrate among all Producer streams ([PR #992](https://github.com/versatica/mediasoup/pull/992)).
+
+
 ### 3.11.7
 
-* libwebrtc: Fix crash due to invalid `arrival_time` value ([PR #985(https://github.com/versatica/mediasoup/pull/985) by @ggarber).
-* libwebrtc: Replace `MS_ASSERT()` with `MS_ERROR()` ([PR #988(https://github.com/versatica/mediasoup/pull/988)).
+* libwebrtc: Fix crash due to invalid `arrival_time` value ([PR #985](https://github.com/versatica/mediasoup/pull/985) by @ggarber).
+* libwebrtc: Replace `MS_ASSERT()` with `MS_ERROR()` ([PR #988](https://github.com/versatica/mediasoup/pull/988)).
 * Update NPM deps.
 
 
 ### 3.11.6
 
-* Fix wrong `PictureID` rolling over in VP9 and VP8 ([PR #984(https://github.com/versatica/mediasoup/pull/984) by @jcague).
+* Fix wrong `PictureID` rolling over in VP9 and VP8 ([PR #984](https://github.com/versatica/mediasoup/pull/984) by @jcague).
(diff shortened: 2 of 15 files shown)


ibc:
@nazar-pc I strongly think we should freeze Rust version in mediasoup-rust to avoid these unexpected and totally unrelated CI errors: https://github.com/versatica/mediasoup/actions/runs/4023390642/jobs/6914204443

jmillan (approved):
Makes sense.

nazar-pc (approved):
Sure, I'll add config that locks the version so we can update it manually. These changes are easy to fix though, just run `cargo clippy --all-targets` and fix things until it stops complaining.

The changes here make sense to me.

ibc:
Problem is that cargo clippy works for me in my computer. How should I update Rust? Via brew install rust? Or brew install cargo?

nazar-pc:
New version was out yesterday with new lints enabled by default: https://blog.rust-lang.org/2023/01/26/Rust-1.67.0.html
Updating can be done easily with `rustup update`.

ibc:
@nazar-pc I need help here:
CleanShot-2023-01-27-at-12 25 50@2x

ibc:
> @nazar-pc I need help here: CleanShot-2023-01-27-at-12 25 50@2x

Trying to apply this change:

```diff
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -500,8 +500,7 @@ impl Inner {
                     Err(error) => Err(io::Error::new(
                         io::ErrorKind::Other,
                         format!(
-                            "unexpected first notification from worker [id:{}]: {:?}; error = {}",
-                            id, notification, error
+                            "unexpected first notification from worker [id:{id}]: {notification}; error = {error}"
                         ),
                     )),
                 };
```

ibc:
I've tried this(as per output help):
CleanShot-2023-01-27-at-12 30 09@2x
then it complains again with:
CleanShot-2023-01-27-at-12 30 46@2x

nazar-pc:
It was `{:?}`, which is print debug version of the data, not regular (`{}`), so it should be `{notification:?}`:
```rust
                        format!(
                            "unexpected first notification from worker [id:{id}]: {notification:?}; \
                            error = {error}"
                        ),
```

When strings are longer than configured length formatter gives up formatting other code around it sometimes, so better splitting it into multiple lines explicitly in those cases.

UPD: Error message actually suggested `:?`, I guess you overlooked it due to 2 other variables.

ibc:
Thanks @nazar-pc, pushed. Should be green now.

nazar-pc:
Rust is not happy with your skills today, sir. Use `cargo clippy --all-targets` at the root of the repo in case you run it under `rust` only

ibc:
Did `cargo clippy --all-targets` ignore `rust/tests` folder because I executed it in `rust` folder? WHAT? How can be this language legal?

nazar-pc:
I don't think that was the case, tests are covered by that command.
CI just runs that command in the root of the repo and refuses to accept any warnings.
I think you're close though.

ibc:
CleanShot-2023-01-27-at-14 11 35@2x

CleanShot-2023-01-27-at-14 12 00@2x

Really???

nazar-pc:
Only plain variables can be used, not expressions or property accesses like `self.id`. Let me fix those for you and push a commit into this branch.

ibc:
All fixed in last commit.

ibc:
This error didn't show locally: https://github.com/versatica/mediasoup/actions/runs/4024845039/jobs/6917332591

nazar-pc:
> This error didn't show locally: https://github.com/versatica/mediasoup/actions/runs/4024845039/jobs/6917332591

Because it is platform-specific code. Unless you're on Windows, it is not compiled.

`rust-toolchain.toml` will make sure `cargo` downloads and uses desired version of Rust toolchain and components necessary for operation, including for CI to run successfully. There should be no guessing or mismatched tooling versions anymore.

ibc:
> `rust-toolchain.toml` will make sure `cargo` downloads and uses desired version of Rust toolchain and components necessary for operation, including for CI to run successfully. There should be no guessing or mismatched tooling versions anymore.

amazing

nazar-pc:
All :green_circle:

PR #1000, opened 2023-02-01
New Rust release 0.11.3

nazar-pc:
Just to release latest fixes for people to use

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 0.11.3
+
+* Updates from mediasoup TypeScript `3.10.13..=3.11.8`.
+
 # 0.11.2
 
 * Updates from mediasoup TypeScript `3.10.11..=3.11.2`.
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mediasoup"
-version = "0.11.2"
+version = "0.11.3"
 description = "Cutting Edge WebRTC Video Conferencing in Rust"
 categories = ["api-bindings", "multimedia", "network-programming"]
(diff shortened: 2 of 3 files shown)


rust/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# 0.11.3
+
+* Updates from mediasoup TypeScript `3.10.13..=3.11.8`.
```
ibc, line 5: I don't understand these changes. Could you explain each of them please?

nazar-pc: This is to avoid copy-pasting changelog, like previously only Rust-specific changes are mentioned explicitly.

ibc (approved):

nazar-pc:
Looks like stats are only printed when you or Jose create PRs.
Maybe add check to the pipeline to skip and not fail when other people submit PRs?

ibc:
> Looks like stats are only printed when you or Jose create PRs.
> 
> Maybe add check to the pipeline to skip and not fail when other people submit PRs?

I don't know how to do that plus I don't understand why just Jose and me. Maybe because you are not member of Versatica org? Aren't you?

If you check the stats yaml CI task, it is supposed to also generate them every Monday but honestly no idea where those generated stats are supposed to be sent or printed.

nazar-pc:
Yeah, I think it is because you're members of the org.

There are two conditions there: once a week AND on every opened PR (those conditions are acting like OR, not AND). Did you mean just once a week?

I can take a look into it later, never used it before.

ibc:
> There are two conditions there: once a week AND on every opened PR (those conditions are acting like OR, not AND). Did you mean just once a week?

I meant both. Will remove the "on every opened PR" and let's see what happens.

ibc:
Let's see what happens: https://github.com/versatica/mediasoup/commit/64e4d89bd8385dc664cecb231a1338c6acbe04b3

PR #1004, opened 2023-02-01
Worker: More const correctness and qualified auto enhancements

jmillan:


worker/src/handles/TcpConnectionHandler.cpp
```diff
@@ -135,7 +135,7 @@ void TcpConnectionHandler::Setup(
 	MS_TRACE();
 
 	// Set the UV handle.
-	const int err = uv_tcp_init(DepLibUV::GetLoop(), this->uvHandle);
+	int err = uv_tcp_init(DepLibUV::GetLoop(), this->uvHandle);
```
jmillan, line 138: This is not OK. @ibc, I'll check this tomorrow.

jmillan: It happens in different places, at least in this file.

jmillan:
Closing this PR. It's not correct as it removes some `const` too.

Issue #1007, opened 2023-02-03
[RUST] WebRtcMessage cannot be used in an async context


RemiKalbe:
The `WebRtcMessage` given to the `on_message` closure cannot be used if we want to do anything async.

Imagine the following pseudo code 

```rust
...
consumer.on_message(|m| {
    let m = m.clone();
    sender
        .send_future(async move {
            ...
            handle_message(m).await;
            ...
        })
        .unwrap();
});
...
```

Will result in the following error

```
error[E0521]: borrowed data escapes outside of closure
   --> path/to/file:80:13
    |
72  |           let recv_message_handle = consumer.on_message(|m| {
    |                                                          -
    |                                                          |
    |                                                          `m` is a reference that is only valid in the closure body
    |                                                          has type `&'1 WebRtcMessage<'_>`
...
80  | /             sender
81  | |                 .send_future(async move {
82  | |                    handle_message(m).await;
83  | |
...   |
110 | |                     }
111 | |                 })
    | |__________________^ argument requires that `'2` must outlive `'static`
```
Here I used `send_future`, but I could have used `tokio::spawn`.
I tried all that I could think of, using move, clone, and so on.

PR #1008, opened 2023-02-03
Introduce OwnedWebRtcMessage

RemiKalbe:
Not sure if it's the best way to solve #1007 but it compiles and passes the tests.

### DIFF
--- a/rust/benches/direct_data.rs
+++ b/rust/benches/direct_data.rs
@@ -51,7 +51,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
                     let _ = sender.send(());
                 });
 
-                let _ = direct_data_producer.send(WebRtcMessage::Binary(Cow::from(data)));
+                let _ = direct_data_producer.send(WebRtcMessage::Binary(data.clone()));
 
                 let _ = receiver.recv();
             })
--- a/rust/benches/direct_data.rs
+++ b/rust/benches/direct_data.rs
@@ -51,7 +51,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
                     let _ = sender.send(());
                 });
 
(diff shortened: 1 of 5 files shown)


nazar-pc:
Why don't you clone in event handler instead? In some cases you don't need to clone things, so can avoid allocation. This basically forces allocation for everyone all the time.

RemiKalbe:
Could you point me to where event handler is ? I don't know the codebase well enough 😅

nazar-pc:
In `on_message` in https://github.com/versatica/mediasoup/issues/1007 try `.into_owned()` rather than `.clone()`

RemiKalbe:
```
no method named `into_owned` found for reference `&WebRtcMessage<'_>` in the current scope
method not found in `&WebRtcMessage<'_>`
```
To I need to have some trait in scope ?

nazar-pc:
I mean you'll need to match on enum and then to `.into_owned()` on the contents.

Though I'm not opposed to the latest version of changes in this PR, but can you replace `to_vec` with `into_owned`? It seems more idiomatic to me.

RemiKalbe:
Oh yeah right 😅, I did not think about that!

nazar-pc:
The reason of such behavior here is efficiency: the memory slice you're getting is most likely allocated by C++ worker. If you process it right away, you don't need to re-allocate it somewhere. But if you want to move it to another thread then yes, you need to clone it and get owned value.

RemiKalbe:
Got it ! Thanks for the explanation 👍

ibc:
Should we close #1007 then?

nazar-pc:
Yes, it was essentially a question, there was nothing to fix, but this PR makes it a bit more convenient

PR #1012, opened 2023-02-15
Fix consuming data producer from direct transport by data consumer on non-direct transport

nazar-pc:
Should fix issue reported in https://mediasoup.discourse.group/t/rust-confusion-about-dataconsumers-dataproducers/4956

The fix here is to create default SCTP stream parameters (ordered stream) in case producer was created on direct transport.

@RemiKalbe can you confirm, please?

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 0.11.4
+
+* Fix consuming data producer from direct transport by data consumer on non-direct transport.
+
 # 0.11.3
 
 * Updates from mediasoup TypeScript `3.10.13..=3.11.8`.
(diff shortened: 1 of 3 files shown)


ibc (commented):
Cannot check mediasoup Node code now. Is this PR adding same logic in Rust than the existing one in Node for this scenario?

nazar-pc:
Yes. Node created `{}` (essentially default value) if producer's SCTP stream parameters are missing, now Rust does the same, but it has non-optional `ordered` field, which is why it is set explicitly to `true`, which is the default value.

RemiKalbe:
I no longer have the error so I think it works, thank you. I'll try to test further if I can actually receive a message, but I have other issues in my code I need to fix first 🫡

ibc (approved):

nazar-pc:
0.11.4 is published on crates.io, thanks for reporting!

PR #1013, opened 2023-02-16
Allow simulcast with a single encoding (and N temporal layers)

ibc:
- Use case: VP8 or H264 with 1 single stream (1 spatial layer) and multiple temporal layers.
- Related to https://github.com/versatica/mediasoup-client/pull/248
- See new query params in mediasoup-demo: https://github.com/versatica/mediasoup-demo/#configuration-via-query-parameters

CleanShot-2023-02-16-at-01 26 33

### DIFF
--- a/node/src/ortc.ts
+++ b/node/src/ortc.ts
@@ -1064,7 +1064,7 @@ export function getConsumerRtpParameters(
 		{
 			const { temporalLayers } = parseScalabilityMode(scalabilityMode);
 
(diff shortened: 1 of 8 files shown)


jmillan (approved):

ibc:
@nazar-pc I've done minimal Rust changes since this PR makes mediasoup use `L` instead of `S` to indicate number of spatial layers in `scalabilityMode`. Everything is green.

ibc:
Released in mediasoup node 3.11.9

PR #997, opened 2023-01-30
Add support for Opus NACK 

t-mullen:
Adds support for NACK packets and retransmissions when using Opus.
This makes a significant difference when packet loss is high. Here's some samples with 40% packet loss.
Before: https://voca.ro/1drKi1WrGBbL
After: https://voca.ro/1c5NmOL4D42u

Changes summary:
- Added an `ignoreNack` option to ConsumerOptions - this is the TX equivalent of the `codecOptions.opusNack` option in [my other PR,](https://github.com/versatica/mediasoup-client/pull/242).
- Added the `{ type: 'nack' }` RTP capability to Opus.

Depends on client PR: https://github.com/versatica/mediasoup-client/pull/242

PR #1015, opened 2023-02-22
Add support for RTCP NACK in OPUS

ibc:
Inspired by PR #997 by @t-mullen.

### Details

- This PR enables RTP retransmission for OPUS codec in mediasoup by adding NACK support into all supported configurations of OPUS codec.
- `ConsumerOptions` given to `transport.consume()` get a new `enableRtx` boolean option. If unset, it's `true` for video codecs and `false` for audio codecs (default behavior). If set to `true` and if the consuming device supports RTCP NACK for opus, then mediasoup will honour NACK received from the device and will retransmitted requested OPUS RTP packets.
- Currently just libwebrtc supports NACK for OPUS and it does it without using RTX codec, and we do the same in mediasoup.

### Bonus Tracks

- `RtpStreamSend`:  Use 1000ms as max retransmission delay for audio streams (keep using 2000ms for video streams as before).

### DIFF
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -51,8 +51,9 @@ export type ConsumerOptions =
 	preferredLayers?: ConsumerLayers;
 
 	/**
-	 * Whether this Consumer should enable RTP retransmissions upon receipt of
-	 * RTCP NACK from the remote Consumer. If not set, it's by default true for
+	 * Whether this Consumer should enable RTP retransmissions, storing sent RTP
+	 * and processing the incoming RTCP NACK from the remote Consumer. 
+	 ....
 	 * video codecs and false for audio codecs. If set to true, NACK will be
 	 * enabled if both endpoints (mediasoup and the remote Consumer) support NACK
 	 * for this codec. When it comes to audio codecs, just OPUS supports NACK.
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -57,7 +57,7 @@ export type ConsumerOptions =
 	 * enabled if both endpoints (mediasoup and the remote Consumer) support NACK
 	 * for this codec. When it comes to audio codecs, just OPUS supports NACK.
 	 */
-	enableNack?: boolean;
+	enableRtx?: boolean;
 
 	/**
 	 * Whether this Consumer should ignore DTX packets (only valid for Opus codec).
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -52,11 +52,11 @@ export type ConsumerOptions =
 
 	/**
 	 * Whether this Consumer should enable RTP retransmissions, storing sent RTP
-	 * and processing the incoming RTCP NACK from the remote Consumer. 
-	 ....
-	 * video codecs and false for audio codecs. If set to true, NACK will be
-	 * enabled if both endpoints (mediasoup and the remote Consumer) support NACK
-	 * for this codec. When it comes to audio codecs, just OPUS supports NACK.
+	 * and processing the incoming RTCP NACK from the remote Consumer. If not set
+	 * it's true by default for video codecs and false for audio codecs. If set
+	 * to true, NACK will be enabled if both endpoints (mediasoup and the remote
+	 * Consumer) support NACK for this codec. When it comes to audio codecs, just
+	 * OPUS supports NACK.
 	 */
 	enableRtx?: boolean;
 
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -51,6 +51,15 @@ export type ConsumerOptions =
 	preferredLayers?: ConsumerLayers;
 
 	/**
+	 * Whether this Consumer should enable RTP retransmissions upon receipt of
+	 * RTCP NACK from the remote Consumer. If not set, it's by default true for
+	 * video codecs and false for audio codecs. If set to true, NACK will be
+	 * enabled if both endpoints (mediasoup and the remote Consumer) support NACK
+	 * for this codec. When it comes to audio codecs, just OPUS supports NACK.
+	 */
+	enableNack?: boolean;
+
+	/**
 	 * Whether this Consumer should ignore DTX packets (only valid for Opus codec).
 	 * If set, DTX packets are not forwarded to the remote Consumer.
 	 */
--- a/node/src/ortc.ts
+++ b/node/src/ortc.ts
@@ -1114,7 +1114,8 @@ export function canConsume(
 export function getConsumerRtpParameters(
 	consumableParams: RtpParameters,
 	caps: RtpCapabilities,
-	pipe: boolean
+	pipe: boolean,
+	enableNack: boolean
 ): RtpParameters
 {
 	const consumerParams: RtpParameters =
@@ -1221,6 +1222,17 @@ export function getConsumerRtpParameters(
 					fb.type !== 'transport-cc' &&
 					fb.type !== 'goog-remb'
 				));
+		}
+	}
+
+	// If enableNack is false, remove RTCP NACK support. If so, we remove all
+	// RTCP feedbacks with type 'nack' (including those whose parameter is 'pli').
+	if (!enableNack)
+	{
+		for (const codec of consumerParams.codecs)
+		{
+			codec.rtcpFeedback = codec.rtcpFeedback!
+				.filter((fb) => fb.type !== 'nack');
 		}
 	}
 
--- a/node/src/ortc.ts
+++ b/node/src/ortc.ts
@@ -1109,13 +1109,21 @@ export function canConsume(
  *
  * It reduces encodings to just one and takes into account given RTP capabilities
  * to reduce codecs, codecs' RTCP feedback and header extensions, and also enables
- * or disabled RTX.
+ * or disables RTX.
  */
 export function getConsumerRtpParameters(
-	consumableParams: RtpParameters,
-	caps: RtpCapabilities,
-	pipe: boolean,
-	enableNack: boolean
+	{
+		consumableRtpParameters,
+		remoteRtpCapabilities,
+		pipe,
+		enableRtx
+	}:
+	{
+		consumableRtpParameters: RtpParameters;
+		remoteRtpCapabilities: RtpCapabilities;
+		pipe: boolean;
+		enableRtx: boolean;
+	}
 ): RtpParameters
 {
 	const consumerParams: RtpParameters =
@@ -1123,22 +1131,22 @@ export function getConsumerRtpParameters(
 		codecs           : [],
 		headerExtensions : [],
 		encodings        : [],
-		rtcp             : consumableParams.rtcp
+		rtcp             : consumableRtpParameters.rtcp
 	};
 
-	for (const capCodec of caps.codecs!)
+	for (const capCodec of remoteRtpCapabilities.codecs!)
 	{
 		validateRtpCodecCapability(capCodec);
 	}
 
 	const consumableCodecs =
-		utils.clone(consumableParams.codecs) as RtpCodecParameters[];
+		utils.clone(consumableRtpParameters.codecs) as RtpCodecParameters[];
 
 	let rtxSupported = false;
 
 	for (const codec of consumableCodecs)
 	{
-		const matchedCapCodec = caps.codecs!
+		const matchedCapCodec = remoteRtpCapabilities.codecs!
 			.find((capCodec) => matchCodecs(capCodec, codec, { strict: true }));
 
 		if (!matchedCapCodec)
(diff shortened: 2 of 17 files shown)


ibc:
@nazar-pc I'm doing Rust things (test missing)

rust/src/router/transport.rs
```diff
@@ -522,13 +523,24 @@ pub(super) trait TransportImpl: TransportGeneric {
...
 
+        let computed_enable_nack: bool;
+
+        if let Some(flag) = enable_nack {
+            computed_enable_nack = flag;
+        } else if producer.kind() == MediaKind::Video {
+            computed_enable_nack = true;
+        } else {
+            computed_enable_nack = false;
+        }
```
nazar-pc: This would look more idiomatic in Rust:
```suggestion
        let enable_nack = enable_nack.unwrap_or(
            producer.kind() == MediaKind::Video
        );
```

It means you're extracting provided value if it is there, but if not using default value, which is true for video. And I'd probably inline it in `ortc::get_consumer_rtp_parameters` call since it is only necessary for `else` branch.

rust/src/ortc.rs
```diff
@@ -641,24 +641,25 @@ pub(crate) fn can_consume(
...
+    for cap_codec in &remote_rtp_capabilities.codecs {
         validate_rtp_codec_capability(cap_codec)
             .map_err(ConsumerRtpParametersError::InvalidCapabilities)?;
     }
 
     let mut rtx_supported = false;
 
-    for mut codec in consumable_params.codecs.clone() {
-        if let Some(matched_cap_codec) = caps
+    for mut codec in consumable_rtp_parameters.codecs.clone() {
```
ibc, line 661: @nazar-pc I need that this changes to be in sync with Node changes in ortc.ts in `getConsumerRtpParameters()`:

```ts
for (const codec of consumableCodecs)
	{
		const matchedCapCodec = remoteRtpCapabilities.codecs!
			.find((capCodec) => matchCodecs(capCodec, codec, { strict: true }));

		if (!matchedCapCodec)
		{
			continue;
		}

		if (!enableRtx && isRtxCodec(codec))
		{
			continue;
		}

		codec.rtcpFeedback = matchedCapCodec.rtcpFeedback!
			.filter((fb) => (
				(enableRtx || fb.type !== 'nack' || fb.parameter)
			));

		consumerParams.codecs.push(codec);
	}
```

Any help please? I'm a bit lost here.

node/src/ortc.ts
```diff
@@ -1109,43 +1109,60 @@ export function canConsume(
...
 		if (!matchedCapCodec)
 		{
 			continue;
 		}
 
-		codec.rtcpFeedback = matchedCapCodec.rtcpFeedback;
+		if (!enableRtx && isRtxCodec(codec))
+		{
+			continue;
+		}
```
nazar-pc: Looks like cheaper operation that can go at the beginning of `for` loop. No need to iterate over codecs and match if this doesn't pass.

rust/src/ortc.rs
nazar-pc: `if` can be added to the beginning of the `for` loop the same way as in TypeScript (there is `.is_rtx()` call used in this file).

Then in `*codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();` you can replace that with something like this (may not compile):
```rust
*codec.rtcp_feedback_mut() = matched_cap_codec
    .rtcp_feedback()
    .iter()
    .filter(|fb| do checks here)
    .cloned()
    .collect();
```

Examples of matching of feedback type also present in this file.

rust/src/router/transport.rs
ibc: Thanks. Done.

node/src/ortc.ts
ibc: :+1:

rust/src/ortc.rs
ibc: Done in https://github.com/versatica/mediasoup/pull/1015/commits/09b6c085e7919cd2539cb0715dc639974b26f5ce

ibc:
@nazar-pc I've tried everything here but definitely I don't understand why this fails:

https://github.com/versatica/mediasoup/pull/1015/commits/e3f425a5924a17a0016704653c0c600f355234d9

It complains:

```
$ KEEP_BUILD_ARTIFACTS=1 cargo test
   Compiling mediasoup v0.11.4 (/Users/ibc/src/v3-mediasoup/rust)
error[E0308]: mismatched types
   --> rust/src/ortc.rs:673:42
    |
673 |               *codec.rtcp_feedback_mut() = matched_cap_codec
    |  _____________--------------------------___^
    | |             |
    | |             expected due to the type of this binding
674 | |                 .rtcp_feedback().clone()
675 | |                 .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
    | |_____________________________________________________________________^ expected struct `Vec`, found `()`
    |
    = note: expected struct `Vec<RtcpFeedback>`
            found unit type `()`
note: method `retain` modifies its receiver in-place
   --> rust/src/ortc.rs:675:18
    |
673 |               *codec.rtcp_feedback_mut() = matched_cap_codec
    |  __________________________________________-
674 | |                 .rtcp_feedback().clone()
    | |________________________________________- you probably want to use this value after calling the method...
675 |                   .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
    |                    ^^^^^^ this call modifies its receiver in-place
    = note: ...instead of the `()` output of method `retain`

error[E0308]: mismatched types
   --> rust/src/ortc.rs:673:42
    |
673 |               *codec.rtcp_feedback_mut() = matched_cap_codec
    |  _____________--------------------------___^
    | |             |
    | |             expected due to the type of this binding
674 | |                 .rtcp_feedback().clone()
675 | |                 .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
    | |_____________________________________________________________________^ expected struct `Vec`, found `()`
    |
    = note: expected struct `Vec<rtp_parameters::RtcpFeedback>`
            found unit type `()`
note: method `retain` modifies its receiver in-place
   --> rust/src/ortc.rs:675:18
    |
673 |               *codec.rtcp_feedback_mut() = matched_cap_codec
    |  __________________________________________-
674 | |                 .rtcp_feedback().clone()
    | |________________________________________- you probably want to use this value after calling the method...
675 |                   .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
    |                    ^^^^^^ this call modifies its receiver in-place
    = note: ...instead of the `()` output of method `retain`
```

I **do** know that `retain()` modifies its receiver in place, that's why I'm doing `clone()` first, but it doesn't care it seems...

Regarding first error " expected struct `Vec`, found `()`", absolutely no idea what it means.

## UPDATE

I've got a solution that works, but not sure how good/idiomatic it is:

### Code in v3

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();
```

### Desired code in this branch (it doesn't compile as shown above)

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec
    .rtcp_feedback().clone()
    .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
```

### Working code

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();

codec
    .rtcp_feedback_mut()
    .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
```

It compiles, no warnings, and makes the new test pass. Should I go this way?
NOTE: I'm gonna push it since it's way better than the failing thing. Pushed here https://github.com/versatica/mediasoup/pull/1015/commits/a84ab48d923703f991d22f36fd6b22fda9992518

rust/src/ortc.rs
```diff
@@ -641,29 +641,39 @@ pub(crate) fn can_consume(
...
             .codecs
             .iter()
             .find(|cap_codec| match_codecs(cap_codec.deref().into(), (&codec).into(), true).is_ok())
         {
-            *codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();
+            // *codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();
+
+            *codec.rtcp_feedback_mut() = matched_cap_codec
+                .rtcp_feedback().clone()
+                .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
```
ibc: **TODO:** This is not compiling. Help needed.

ibc:
> @nazar-pc I've tried everything here but definitely I don't understand why this fails:
> 
> [e3f425a](https://github.com/versatica/mediasoup/commit/e3f425a5924a17a0016704653c0c600f355234d9)

I've got a solution that works, but not sure how good/idiomatic it is:

### Code in v3

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();
```

### Desired code in this branch (it doesn't compile as shown above)

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec
            .rtcp_feedback().clone()
            .retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
```

### Working code

```rust
*codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();

codec.rtcp_feedback_mut().retain(|fb| enable_rtx || fb != &RtcpFeedback::Nack);
```

It compiles, no warnings, and makes the new test pass. Should I go this way?
NOTE: I'm gonna push it since it's way better than the failing thing.

rust/src/ortc.rs
ibc: Resolved here but waiting for @nazar-pc's thoughts:
https://github.com/versatica/mediasoup/pull/1015/commits/a84ab48d923703f991d22f36fd6b22fda9992518

ibc:
Ready for review.

rust/src/ortc.rs
Hartigan: @ibc you can use ".iter().filter(condition).collect()" without clone

nazar-pc: Yes, cloning just to move things around later makes less sense than just to filter stuff and collect then into a new collection instead. @Hartigan's solution is better. Fixed commit with corresponding change.

nazar-pc (approved):
Makes sense to me

rust/src/ortc.rs
ibc: Thanks guys

rust/src/ortc.rs
```diff
@@ -641,29 +641,40 @@ pub(crate) fn can_consume(
...
             .iter()
             .find(|cap_codec| match_codecs(cap_codec.deref().into(), (&codec).into(), true).is_ok())
         {
-            *codec.rtcp_feedback_mut() = matched_cap_codec.rtcp_feedback().clone();
+            *codec.rtcp_feedback_mut() = matched_cap_codec
+                .rtcp_feedback()
+                .iter()
+                .filter(|&&fb| enable_rtx || fb != RtcpFeedback::Nack)
+                .copied()
+                .collect();
```
ibc, line 676: - Why is iter() needed?
- Why &&fb?
- What does copies() do?
- And collect()

Hartigan: - iter() - create iterator over &T in collection
- && - just double dereference, because T is &A in filter(), and after iter() - &&A
- copies() - map iterator with copy() call on each element(&T before copy and T after copy)
- collect() - create collection from iterator. In current case: collect::<Vec<_>>()

ibc: Thanks. Just a question: why is copies() needed?

Hartigan: Before copied() we have iterator over &T. If we collect without copy of elements, we receive Vec<&T>. This vector contains references to elements from source collection. After that we cannot assign Vec<&T> to Vec\<T\> variable.

node/src/Consumer.ts
```diff
@@ -50,6 +50,15 @@ export type ConsumerOptions =
 	 */
 	preferredLayers?: ConsumerLayers;
 
+	/**
+	 * Whether this Consumer should enable RTP retransmissions upon receipt of
+	 * RTCP NACK from the remote Consumer. If not set, it's by default true for
```
jmillan: Little clarification, this change would we also applicable to `consumer.rs`.
```suggestion
	 * Whether this Consumer should enable RTP retransmissions, storing sent RTP
	 * and processing the incoming RTCP NACK from the remote Consumer. 
	 ....
```

jmillan (approved):

rust/src/ortc.rs
ibc: Thanks, so copied() receives a reference to a "struct", copies the full struct and returns it.

Hartigan: Yes. For `copied()` type `T` should have trait `Copy`. In C++ terms: type T should have copy-constructor.
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.copied

ibc:
@jmillan I've added this: https://github.com/versatica/mediasoup/pull/1015/commits/97e99400621d61ab6c3792227c433d096bad2762

- `RtpStreamSend`:  Use 1000ms as max retransmission delay for audio streams (keep using 2000ms for video streams as before).

jmillan:
> RtpStreamSend: Use 1000ms as max retransmission delay for audio streams (keep using 2000ms for video streams as before).

OK, but I see value 2000 for both media types in that commit.

I'm not sure of those values, honestly. Whether 1000 and 2000 are OK or  not. IMHO we need to check and adjust them if needed.

ibc:
> OK, but I see value 2000 for both media types in that commit.

Fixed.
 
> I'm not sure of those values, honestly. Whether 1000 and 2000 are OK or not. IMHO we need to check and adjust them if needed.

My idea is to make it configurable via API option. But, for sure, 2000ms is way to much for audio delay.

PR #1019, opened 2023-03-03
Fix jitter calculation

### LINKED ISSUE #1018: RtpStream jitter calculation is sometimes inaccurate
### Your environment

- Operating system: OS X
- Node version: 16.5
- npm version: 8.5.5
- gcc/clang version: 14.0
- mediasoup version: 3.11.8
- mediasoup-client version: 3.6.61

### Issue description

I have noticed when tracking stats on the server side using `producer.getStats()` I sometimes get extremely high jitter values which do not make sense. I started debugging locally and I believe there may be some sort of rounding / casting bug in how `transit` is calculated.

The code in question is in `RtpStreamRecv::CalculateJitter`, specifically:

```c++
auto transit =
	static_cast<int>(DepLibUV::GetTimeMs() - (rtpTimestamp * 1000 / this->params.clockRate));
```

I don't fully understand what this is doing, but when logging out values I noticed when the jitter jumps very high, the transit value also does (this makes sense)... the only questionable part of this value is the last part which I am calling `transitFactor` due to ignorance, but this portion: `(rtpTimestamp * 1000 / this->params.clockRate)`

When I pulled that out into it's own variable and logged it, I started seeing logs like this (sorry for verbosity but was just looking for patterns). It is much easier to trigger if you can simulate some packet loss ~10%, but not always necessary. It just seems to happen sooner, but even taking into account a dropped packet, the change in the transit value does not really make sense.

```
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370086445, systime:1717562214, jitter:2, clockRate:48000, transitFactor:89309.000000, lastTransit:1717472896, currentTransit:1717472896, d:0] +40ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370087405, systime:1717562234, jitter:1, clockRate:48000, transitFactor:89329.000000, lastTransit:1717472896, currentTransit:1717472896, d:0] +20ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370090285, systime:1717562294, jitter:0, clockRate:48000, transitFactor:89389.000000, lastTransit:1717472896, currentTransit:1717472896, d:0] +60ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370092205, systime:1717562334, jitter:0, clockRate:48000, transitFactor:89429.000000, lastTransit:1717472896, currentTransit:1717472896, d:0] +40ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370096045, systime:1717562414, jitter:5592, clockRate:48000, transitFactor:30.000000, lastTransit:1717472896, currentTransit:1717562368, d:89472] +80ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | ** HIGH JITTER DETECTED ** +0ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370097005, systime:1717562434, jitter:5250, clockRate:48000, transitFactor:50.000000, lastTransit:1717562368, currentTransit:1717562496, d:128] +20ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | ** HIGH JITTER DETECTED ** +0ms
  mediasoup:Channel [pid:25108] RTC::RtpStreamRecv::CalculateJitter() | [ssrc:3847190658, timestamp:1370101805, systime:1717562534, jitter:4929, clockRate:48000, transitFactor:150.000000, lastTransit:1717562496, currentTransit:1717562368, d:128] +100ms
  ```

We can see the "transitFactor" somehow jumps from 89429 to 30, completely throwing off the rolling calculation. I also added logging to the one location where transit is reset to 0 to ensure it wasn't caused by that.

Nothing about the values which trigger this appear special, and I'm honestly kinda stumped at what might be going on here but I'm also not really an expert on C++ type casting / conversions. 

I modified the code to this and the problem still occurred:
```c++
double transitFactor = rtpTimestamp * 1000 / this->params.clockRate;
auto transit = static_cast<int>(timeMs - transitFactor);
```

but when I did this it went away:
```c++
double transitFactor = (double)rtpTimestamp * 1000.0 / (double)this->params.clockRate;
auto transit = static_cast<int>(timeMs - transitFactor);
```

I'm sure the extra casts are silly overkill but I wanted to be explicit.

I know this is all very anecdotal and it honestly does not make sense to me that it works fine the vast majority of the time. I would love to be able to PR a fix but I don't feel like I understand the problem or this calculation well enough to offer something appropriate.

ibc:
Based on research and solutions given by @alexciarlillo and @snnz in issue https://github.com/versatica/mediasoup/issues/1018.

### DIFF
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -645,9 +645,11 @@ namespace RTC
 		if (this->params.clockRate == 0u)
 			return;
 
-		auto transit =
-		  static_cast<int>(DepLibUV::GetTimeMs() - (rtpTimestamp * 1000 / this->params.clockRate));
-		int d = transit - this->transit;
+		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
+		double transitFactor =
+		  static_cast<double>(rtpTimestamp) * 1000.0 / static_cast<double>(this->params.clockRate);
+		auto transit = static_cast<int>(DepLibUV::GetTimeMs() - transitFactor);
+		int d        = transit - this->transit;
 
 		// First transit calculation, save and return.
 		if (this->transit == 0)
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -646,10 +646,9 @@ namespace RTC
 			return;
 
 		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
-		double transitFactor =
-		  static_cast<double>(rtpTimestamp) * 1000.0 / static_cast<double>(this->params.clockRate);
-		auto transit = static_cast<int>(DepLibUV::GetTimeMs() - transitFactor);
-		int d        = transit - this->transit;
+		auto transit = static_cast<int>(
+		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
+		int d = transit - this->transit;
 
 		// First transit calculation, save and return.
 		if (this->transit == 0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### NEXT
 
+* Fix jitter calculation ([PR #1019](https://github.com/versatica/mediasoup/pull/1019), credits to @alexciarlillo).
 * Update NPM deps.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### NEXT
 
-* Fix jitter calculation ([PR #1019](https://github.com/versatica/mediasoup/pull/1019), credits to @alexciarlillo).
+* Fix jitter calculation ([PR #1019](https://github.com/versatica/mediasoup/pull/1019), credits to @alexciarlillo and @snnz).
 * Update NPM deps.
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
       "devDependencies": {
         "@types/debug": "^4.1.7",
         "@types/jest": "^29.4.0",
-        "@types/node": "^18.14.4",
+        "@types/node": "^18.14.5",
         "@types/uuid": "^9.0.1",
         "@typescript-eslint/eslint-plugin": "^5.54.0",
         "@typescript-eslint/parser": "^5.54.0",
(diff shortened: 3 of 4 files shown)


nazar-pc (approved):

worker/src/RTC/RtpStreamRecv.cpp
```diff
@@ -645,9 +645,11 @@ namespace RTC
 		if (this->params.clockRate == 0u)
 			return;
 
-		auto transit =
-		  static_cast<int>(DepLibUV::GetTimeMs() - (rtpTimestamp * 1000 / this->params.clockRate));
-		int d = transit - this->transit;
+		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
+		double transitFactor =
+		  static_cast<double>(rtpTimestamp) * 1000.0 / static_cast<double>(this->params.clockRate);
+		auto transit = static_cast<int>(DepLibUV::GetTimeMs() - transitFactor);
```
ibc: @nazar-pc would you mind telling me how this would be done in Rust? This is a good change to take advantage of this eventual hatred that I feel towards C++.

nazar-pc: Rust refuses to compile if types don't match exactly, which is one of the most confusing parts of C++ for me.

For overflows it will panic in debug build and have some deterministic non-panicking behavior in release build (I believe it will just wrap around) for performance reasons, but you can make it panic in release too if you want (or use checked math to handle it explicitly, see https://doc.rust-lang.org/stable/std/primitive.u32.html#method.checked_add for instance), that is just not the default behavior.

There are methods for checked, overflowing, wrapping, saturating math, so if you want to have exact control over what is happening to numbers you have a decent toolbox. We use those heavily in blockchain code where accidental invisible overflow can have catastrophic consequences.

Some overflow cases are detected by compiler and generate warnings, but that is for something trivial like `1u32 >> 32`.

So while it wouldn't necessarily prevent something like this, it would likely make it easier to discover or handle properly. You can just build an app with panicking overflows and fuzz it with inputs, uncovering various edge-cases.

snnz: With the templates and inline operators, one can make his own type-safe arithmetics in C++ if he wishes. C++ is intended, among other things, to adapt easily existing C code (and familiar programming style), hence these implicit type conversions.

jmillan (approved):

ibc:
Release in mediasoup node 3.11.12

PR #1020, opened 2023-03-06
avgDBov conflict

buptlsp:
Background: 
- We use audioLevel to determine which producer's audio will be sent.
- mapDBovsProducer is a map that uses avgDBov as the key.
- avgDBov is an int value from -127 to 0. In fact, we will use threshold, which has a value from threshold to 0.

Facts:
When there are two producers in the router, their avgDBov within the interval may be the same (with probability 1/threshold), causing this producer to be overwritten by other producers. In this case, the sound of this producer will sound intermittent. The more producers in the room, the higher the probability of them colliding.

### DIFF
--- a/worker/src/RTC/AudioLevelObserver.cpp
+++ b/worker/src/RTC/AudioLevelObserver.cpp
@@ -156,7 +156,7 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		absl::btree_map<int8_t, RTC::Producer*> mapDBovsProducer;
+		absl::btree_map<RTC::Producer*, int8_t> mapDBovsProducer; 
 
 		for (auto& kv : this->mapProducerDBovs)
 		{
@@ -169,7 +169,7 @@ namespace RTC
 			auto avgDBov = -1 * static_cast<int8_t>(std::lround(dBovs.totalSum / dBovs.count));
 
 			if (avgDBov >= this->threshold)
-				mapDBovsProducer[avgDBov] = producer;
+				mapDBovsProducer[producer] = avgDBov;
 		}
 
 		// Clear the map.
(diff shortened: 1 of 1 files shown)


nazar-pc (commented):
Lol, wondering why no one noticed this earlier :upside_down_face: 

One issue here though: `mapProducerDBovs` is btree_map, which is supposed to be sorted. I suspect it wasn't properly processed before (it preferred quietest 10 producers present), not it prefers 10 random producers (ordered by ID), not loudest.

nazar-pc:
The solution here is likely to store a vector of producers, not a single one. Also iteration should likely be in reverse order, so we pick up to 10 loudest producers.

ibc:
> The solution here is likely to store a vector of producers, not a single one. Also iteration should likely be in reverse order, so we pick up to 10 loudest producers.

I need to remember what we are supposed to store here. The 10 Producers with highest audio level?

ibc:
> I suspect it wasn't properly processed before (it preferred quietest 10 producers present)

Nope. It's iterated from highest to lowest keys:

```c++
for (; idx < this->maxEntries && rit != mapDBovsProducer.crend(); ++idx, ++rit)
```

NOTE: I'm not saying there is no issue. Of course there is.

nazar-pc:
> Nope. It's iterated from highest to lowest keys:

Indeed, read the wrong line, you're right.

ibc:
We should use `absl::btree_multimap` right?

nazar-pc:
Yeah, I think so

ibc:
Closing in favour of https://github.com/versatica/mediasoup/pull/1021.
Thanks for reporting, @buptlsp

PR #1021, opened 2023-03-06
AudioLevelObserver: use multimap rather than map to avoid conflict

ibc:
- Replaces PR #1020 created by @buptlsp. Issue explained in that PR.

### DIFF
--- a/worker/src/RTC/AudioLevelObserver.cpp
+++ b/worker/src/RTC/AudioLevelObserver.cpp
@@ -156,7 +156,7 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		absl::btree_map<int8_t, RTC::Producer*> mapDBovsProducer;
+		absl::btree_multimap<int8_t, RTC::Producer*> mapDBovsProducer;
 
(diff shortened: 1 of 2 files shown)


worker/src/RTC/AudioLevelObserver.cpp
```diff
@@ -156,20 +156,24 @@ namespace RTC
...
+			{
 				continue;
+			}
 
 			auto avgDBov = -1 * static_cast<int8_t>(std::lround(dBovs.totalSum / dBovs.count));
 
 			if (avgDBov >= this->threshold)
-				mapDBovsProducer[avgDBov] = producer;
+			{
+				mapDBovsProducer.insert({ avgDBov, producer });
```
ibc, line 175: std and btree multimap container doesn't implement `[ ]` operator.

ibc:
I'm merging this. I've tested it and verified that 2 entries with same audio level are properly stored and later retrieved.

jmillan (approved):

ibc:
Released in mediasoup-node 3.11.13

Issue #1027, opened 2023-03-21
SIGABRT after upgrading to 3.11.14 version


vpalmisano:
### Your environment

- Operating system: Linux
- Node version: 
- npm version:
- gcc/clang version:
- mediasoup version: 3.11.14
- mediasoup-client version: 3.6.82

### Issue description

After upgrading to 3.11.14 version (containing https://github.com/versatica/mediasoup/pull/1023), we received a SIGABRT, with this stack trace:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x0000ffffabab7aa0 in __GI_abort () at abort.c:79
#2  0x0000ffffabdad238 in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/aarch64-linux-gnu/libstdc++.so.6
#3  0x0000ffffabdaad4c in ?? () from /usr/lib/aarch64-linux-gnu/libstdc++.so.6
#4  0x0000ffffabdaadb0 in std::terminate() () from /usr/lib/aarch64-linux-gnu/libstdc++.so.6
#5  0x0000ffffabdab0a4 in __cxa_throw () from /usr/lib/aarch64-linux-gnu/libstdc++.so.6
#6  0x0000ffffabdd3950 in std::__throw_out_of_range_fmt(char const*, ...) () from /usr/lib/aarch64-linux-gnu/libstdc++.so.6
#7  0x0000aaaabc66aa3c in RTC::RetransmissionBuffer::Insert(RTC::RtpPacket*, std::shared_ptr<RTC::RtpPacket>&) ()
#8  0x0000aaaabc68aadc in RTC::RtpStreamSend::ReceivePacket(RTC::RtpPacket*, std::shared_ptr<RTC::RtpPacket>&) ()
#9  0x0000aaaabc6afca4 in RTC::SvcConsumer::SendRtpPacket(RTC::RtpPacket*, std::shared_ptr<RTC::RtpPacket>&) ()
#10 0x0000aaaabc676384 in RTC::Router::OnTransportProducerRtpPacketReceived(RTC::Transport*, RTC::Producer*, RTC::RtpPacket*) ()
#11 0x0000aaaabc663e7c in RTC::Producer::ReceiveRtpPacket(RTC::RtpPacket*) ()
#12 0x0000aaaabc6b4b18 in RTC::Transport::ReceiveRtpPacket(RTC::RtpPacket*) ()
[...]

PR #1029, opened 2023-03-22
RtpRetransmissionBuffer: get rid of startSeq private member

ibc:
- `this->startSeq` is just the seq of the first element in `this->buffer`, nothing else.
- It doesn't make sense to have a separate member just for it.
- We can just read `sequenceNumber` in the oldest item in the buffer.
- The recent crash reported in https://github.com/versatica/mediasoup/issues/1027 was caused due to a non properly updated `this->startSeq`. So let's get rid of it completely.

### DIFF
--- a/worker/src/RTC/RtpRetransmissionBuffer.cpp
+++ b/worker/src/RTC/RtpRetransmissionBuffer.cpp
@@ -29,19 +29,19 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		if (this->buffer.empty())
+		const auto* oldestItem = GetOldest();
+
+		if (!oldestItem)
 		{
 			return nullptr;
 		}
-
-		auto* oldestItem = GetOldest();
 
 		if (RTC::SeqManager<uint16_t>::IsSeqLowerThan(seq, oldestItem->sequenceNumber))
 		{
 			return nullptr;
 		}
 
-		auto idx = static_cast<uint16_t>(seq - oldestItem->sequenceNumber);
+		const auto idx = static_cast<uint16_t>(seq - oldestItem->sequenceNumber);
 
 		if (idx > static_cast<uint16_t>(this->buffer.size() - 1))
 		{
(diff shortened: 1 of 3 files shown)


worker/src/RTC/RtpRetransmissionBuffer.cpp
```diff
@@ -392,14 +380,24 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		return this->Get(this->startSeq);
+		if (this->buffer.empty())
+		{
+			return nullptr;
+		}
+
+		return this->buffer.front();
```
ibc, line 388: Oldest seq in the buffer is just the first element by design, period. Let's do it easy.

worker/src/RTC/RtpRetransmissionBuffer.cpp
```diff
@@ -392,14 +380,24 @@ namespace RTC
...
 	{
 		MS_TRACE();
 
-		return this->Get(this->startSeq + this->buffer.size() - 1);
+		if (this->buffer.empty())
+		{
+			return nullptr;
+		}
+
+		return this->buffer.back();
```
ibc, line 400: Oldest seq in the buffer is just the last element by design, period. Let's do it easy.

worker/src/RTC/RtpRetransmissionBuffer.cpp
```diff
@@ -34,12 +34,14 @@ namespace RTC
 			return nullptr;
 		}
 
-		if (RTC::SeqManager<uint16_t>::IsSeqLowerThan(seq, this->startSeq))
+		auto* oldestItem = GetOldest();
```
nazar-pc: I think it would have been a bit cleaner to check whether this points to `nullptr` rather that checking for empty buffer above. `GetOldest()` contains the same check internally already.

nazar-pc (approved):

worker/src/RTC/RtpRetransmissionBuffer.cpp
ibc: Done here: https://github.com/versatica/mediasoup/pull/1029/commits/595e2016addb955746486cd0852556e7b5f34be6

nazar-pc (approved):

jmillan (commented):
:+1:

PR #1031, opened 2023-03-24
Fix bandwidth probation dead state

vpalmisano:
Running some tests with high packet loss (~20%), in most cases the server stops sending any video packet and we never exit from that state, even if the network link improves.

Looking at the ALR detector (that should be activated when there are no RTP packets sent over the transport), it seems that it is falling into a dead state:

1. in `NetworkControlUpdate GoogCcNetworkController::OnProcessInterval` we periodically call `start_time_ms = alr_detector_->GetApplicationLimitedRegionStartTime()` setting the output value to `probe_controller_->SetAlrStartTimeMs(start_time_ms)`
2. the AlrDetector state is set by `AlrDetector::OnBytesSent`, so when the transport stops sending packets, we could end up in this state:
  ```
  } else if (alr_budget_.budget_ratio() < stop_budget_level_ratio_ &&
             alr_started_time_ms_) {
    state_changed = true;
    alr_started_time_ms_.reset();
  }
  ```

3. at this point `GoogCcNetworkController::OnProcessInterval` calls `probe_controller_->SetAlrStartTimeMs(<empty value>)` with this empty value set into the `ProbeController`, and the `ProbeController::Process` never calls `InitiateProbing`, so no more probing is created;
4. the bandwidth estimation is never tested again, so we never resume sending packets.

With this PR we use a `alr_timeout_` value (set to 3s by default). After that timeout, the ALR resumes sending probes.

### DIFF
--- a/worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/alr_detector.cc
+++ b/worker/deps/libwebrtc/libwebrtc/modules/congestion_controller/goog_cc/alr_detector.cc
@@ -113,8 +113,7 @@ absl::optional<int64_t> AlrDetector::GetApplicationLimitedRegionStartTime(
     int64_t at_time_ms) {
   if (!alr_started_time_ms_ && *last_send_time_ms_) {
     int64_t delta_time_ms = at_time_ms - *last_send_time_ms_;
-    // If ALR is stopped and we didn't sent any packets in the last for a while,
-    // force resuming the state.
+    // If ALR is stopped and we haven't sent any packets for a while, force start.
     if (delta_time_ms > 1000) {
       MS_WARN_TAG(bwe, "large delta_time_ms: %ld, forcing alr state change",
         delta_time_ms);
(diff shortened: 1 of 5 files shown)


ibc:
Is this something available in latest libwebrtc version? You are basically changing libwebrtc. Also, we have an ongoing PR https://github.com/versatica/mediasoup/pull/922 upgrading to latest libwebrtc so I don't see a point in doing it in current very old version.

ibc:
@sarumjanuch you may want to see this.

vpalmisano:
> Is this something available in latest libwebrtc version?

Yes, it seems the same of the latest libwebrtc code: https://webrtc.googlesource.com/src/+/refs/heads/main/modules/congestion_controller/goog_cc/alr_detector.cc

sarumjanuch:
I am aware of this. In my branch it is handled. This is from 18.11.2022:

image

ibc:
Ok, let me take a proper look on Monday and merge it in v3. But it will generate conflicts in your branch @sarumjanuch, is it ok?

sarumjanuch:
Don't worry at all. My branch is already deviated a lot, both, from ms and libwebrtc.

jmillan (approved):

PR #1033, opened 2023-03-28
Github CI: Remove deprecated Ubuntu 18.04

jmillan:
The version is deprecated and will be removed by 4/1/23.

### DIFF
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -8,12 +8,6 @@ jobs:
       fail-fast: false
       matrix:
         build:
-          - os: ubuntu-18.04
-            cc: gcc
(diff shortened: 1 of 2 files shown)


nazar-pc (approved):

ibc (approved):
Shouldn't we also add a newer version?

nazar-pc:
22.04 is already there, I think it is fine to wait for 24.04, most server are running LTS versions of Ubuntu anyway. And it is highly likely that it works just fine on newer distros too.

jmillan:
There is no image with newer version. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

ibc:
Let me merge this because CI is obviously gonna work and I want to write another PR with related changes.

PR #1034, opened 2023-03-28
GitHub CI: Cancel in progress workflows

ibc:
- Cancel a currently running workflow from the same PR, branch or tag when a new workflow is triggered.
- Bonus track: Use Node 18 (LTS) in mediasoup-worker.yaml instead of Node 16.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### NEXT
 
-* Github workflow: Remove deprecated Ubuntu version 18.04 ([PR #1033](https://github.com/versatica/mediasoup/pull/1033)).
+* GitHub CI: Remove deprecated Ubuntu version 18.04 ([PR #1033](https://github.com/versatica/mediasoup/pull/1033)).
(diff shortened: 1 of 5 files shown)


nazar-pc (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ### NEXT
 
-* Github workflow: Remove deprecated Ubuntu version 18.04 ([PR #1033](https://github.com/versatica/mediasoup/pull/1033)).
+* GitHub CI: Remove deprecated Ubuntu version 18.04 ([PR #1033](https://github.com/versatica/mediasoup/pull/1033)).
+* GitHub CI: Cancel in progress workflows ([PR #1034](https://github.com/versatica/mediasoup/pull/1034)).
```
nazar-pc, line 6: Not sure why users of mediasoup would want to see and care about this. Probably shouldn't be in changelog, if they care about everything that happened in the repo, they can see the list of merged things directly.

ibc: I definitely agree and will remove these not useful entries from Changelog.

ibc: Done in v3 branch.

PR #1036, opened 2023-03-28
Properly handle ksvc bandwidth allocation

vpalmisano:
When we use VP9 with K-SVC configuration, the spatial layers are independent; this means that the bandwidth allocator, when it tries to increase the allocated layer,  should not take into account the lower spatial levels bitrates (same behaviour of the Simulcast case).

### DIFF
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -426,6 +426,28 @@ namespace RTC
 				requiredBitrate =
 				  this->producerRtpStream->GetLayerBitrate(nowMs, spatialLayer, temporalLayer);
 
+				// When using K-SVC we must subtract the bitrate of the current used layer
+				// if the new layer is the temporal layer 0 of an higher spatial layer.
+				//
+				// clang-format off
+				if (
+					this->encodingContext->IsKSvc() &&
+					requiredBitrate &&
+					temporalLayer == 0 &&
+					this->provisionalTargetSpatialLayer > -1 &&
+					spatialLayer > this->provisionalTargetSpatialLayer
+				)
+				// clang-format on
+				{
+					auto provisionalRequiredBitrate = this->producerRtpStream->GetBitrate(
+					  nowMs, this->provisionalTargetSpatialLayer, this->provisionalTargetTemporalLayer);
+
+					if (requiredBitrate > provisionalRequiredBitrate)
+						requiredBitrate -= provisionalRequiredBitrate;
+					else
+						requiredBitrate = 1u; // Don't set 0 since it would be ignored.
+				}
+
 				MS_DEBUG_DEV(
 				  "testing layers %" PRIi16 ":%" PRIi16 " [virtual bitrate:%" PRIu32
 				  ", required bitrate:%" PRIu32 "]",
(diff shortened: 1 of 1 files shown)


ibc (commented):
Are you sure we don't need similar logic in ApplyBitrate() method?

vpalmisano:
> Are you sure we don't need similar logic in ApplyBitrate() method?

Do you mean taking into account that logic when we pass `maxBitrate` to the congestion controller?

ibc:
Nope, I mean another thing but maybe we changed it long ago and what I said no longer makes sense. Will recheck tomorrow.

ibc:
Checking this.

ibc:
Checking it later today.

ibc (changes requested):
Ok @vpalmisano, I found the missing thing:

- See `SimulcastConsumer::GetDesiredBitrate()`. It returns the highest bitrate of all its streams (or `this->rtpParameters.encodings[0].maxBitrate` if given).
- However `SvcConsumer::GetDesiredBitrate()` returns the total bitrate of the whole SVC or K-SVC stream. If K-SVC we should just return the bitrate of the spatial layer that produces the most bitrate, right?

NOTE: `Consumer::GetDesiredBitrate()` is used by `Transport::ComputeOutgoingDesiredBitrate()` method.

vpalmisano:
> * If K-SVC we should just return the bitrate of the spatial layer that produces more bitrate, right?

Correct, I'm taking a look.

ibc (commented):
Will review tomorrow. Thanks.

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -531,8 +553,30 @@ namespace RTC
...
+		auto nowMs = DepLibUV::GetTimeMs();
+		uint32_t desiredBitrate{ 0u };
+
+		if (this->encodingContext->IsKSvc())
+		{
+			// Let's iterate all spatial layers of the Producer (from highest to lowest) and
+			// obtain their bitrate. Choose the highest one.
+			// NOTE: When the Producer enables a higher stream, initially the bitrate of
+			// it could be less than the bitrate of a lower stream. That's why we
+			// iterate all streams here anyway.
```
ibc:
```suggestion
			// Let's iterate all spatial layers of the Producer (from highest to lowest) and
			// obtain their bitrate. Choose the highest one.
			// NOTE: When the Producer enables a higher spatial layer, initially the bitrate
			// oft could be less than the bitrate of a lower one. That's why we iterate all
			// spatial layers here anyway.
```

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -531,8 +553,30 @@ namespace RTC
 		if (!IsActive())
 			return 0u;
 
-		auto nowMs              = DepLibUV::GetTimeMs();
-		uint32_t desiredBitrate = this->producerRtpStream->GetBitrate(nowMs);
+		auto nowMs = DepLibUV::GetTimeMs();
+		uint32_t desiredBitrate{ 0u };
+
+		if (this->encodingContext->IsKSvc())
```
ibc, line 560:
```suggestion
		// When using K-SVC each spatial layer is independent of the others.
		if (this->encodingContext->IsKSvc())
```

ibc (approved):

ibc:
Releasing in 3.11.19 right now.

Issue #1037, opened 2023-03-29
High packet loss after upgrading to 3.11.16 version


vpalmisano:
## Bug Report
After upgrading to the last release (containing https://github.com/versatica/mediasoup/pull/1032), when running with limited networking I got a lot of warnings like these:

```
mediasoup:WARN:Channel [pid:1032678] RTC::RtpRetransmissionBuffer::Insert() | packet timestamp is higher than timestamp of immediate newer packet in the buffer, discarding it [ssrc:138048629, seq:14606, timestamp:2513086702] +52ms
RTC::RtpRetransmissionBuffer::Dump() | <RtpRetransmissionBuffer>
RTC::RtpRetransmissionBuffer::Dump() |   buffer [size:205, maxSize:5000]
RTC::RtpRetransmissionBuffer::Dump() |   oldest item [seq:14466, timestamp:2512420792]
RTC::RtpRetransmissionBuffer::Dump() |   newest item [seq:14670, timestamp:2512600792]
RTC::RtpRetransmissionBuffer::Dump() |   buffer window: 4294921574ms
RTC::RtpRetransmissionBuffer::Dump() | </RtpRetransmissionBuffer>
```

```
mediasoup:WARN:Channel [pid:1032678] RTC::RtpRetransmissionBuffer::Insert() | packet has less seq but higher timestamp than oldest packet in the buffer, discarding it [ssrc:409260887, seq:18365, timestamp:1026593387] +44ms
RTC::RtpRetransmissionBuffer::Dump() | <RtpRetransmissionBuffer>
RTC::RtpRetransmissionBuffer::Dump() |   buffer [size:165, maxSize:5000]
RTC::RtpRetransmissionBuffer::Dump() |   oldest item [seq:24816, timestamp:1024930187]
RTC::RtpRetransmissionBuffer::Dump() |   newest item [seq:24980, timestamp:1025106407]
RTC::RtpRetransmissionBuffer::Dump() |   buffer window: 1958ms
RTC::RtpRetransmissionBuffer::Dump() | </RtpRetransmissionBuffer>
```

Reading the last log I see that:
- the buffer state is: oldest item [seq:24816, timestamp:1024930187], newest item [seq:24980, timestamp:1025106407]
- new packet with seq:18365, timestamp:1026593387 is dropped
- seq:18365 probably belongs to a newer packet counted at the next cycle, so it should became the newest item in the buffer

jmillan:
[...]

PR #1041, opened 2023-03-30
Reset RTP retransmission buffer upon RTP sequence number reset

ibc:
Related to #1037

Make `RtpStreamRecv` and `RtpStreamSend` classes be notified by their `RtpStream` parent when the stream seq number is reset so they can react on it. In case of `RtpStreamSend` it clears the entire buffer retransmission since stored packets probably won't make any sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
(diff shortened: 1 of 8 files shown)


jmillan (approved):

ibc:
@jmillan I've added a test unit.

PR #1044, opened 2023-03-30
RTC::RtpRetransmissionBuffer: Handle corner case in which received packet has lower seq than newest packet in the buffer but higher timestamp

ibc:
Description says everything. Test added.


ibc:
Merging since it's harmless and tests and fuzzer pass.

jmillan (commented):
:+1:

PR #1045, opened 2023-03-31
SeqManager: fix crash and add fuzzer

### LINKED ISSUE #1042: mediasoup-worker crash: free(): invalid pointer
## Bug Report

It's consistently happening in the demo server which is running very latest 3.11.19 version. No idea how to reproduce.

```
2023-03-30T10:06:18.320Z mediasoup:ERROR:Worker (stderr) free(): invalid pointer
2023-03-30T10:06:18.323Z mediasoup:ERROR:Worker worker process died unexpectedly [pid:3465, code:null, signal:SIGABRT]
```

### Your environment

- Operating system: Linux Debian
- gcc/clang version: 6.3.0
- mediasoup version: 3.11.19

jmillan:
Fixes #1042


worker/src/RTC/SeqManager.cpp
```diff
@@ -113,15 +113,25 @@ namespace RTC
...
+		{
+			// New input is higher than the maximum seen. But less than acceptable units higher.
+			// Keep it as the maximum seen. See Drop().
+			if (IsSeqHigherThan(input, this->maxInput))
+				this->maxInput = input;
+
+			// New output is higher than the maximum seen. But less than acceptable units higher.
+			// Keep it as the maximum seen. See Sync().
+			if (IsSeqHigherThan(output, this->maxOutput))
+				this->maxOutput = output;
```
ibc: Please let's add { } in all if/else conditions. I'm doing it little by little (there is a option for clang-tidy but it no longer works in my system due to rules that do not exist in the clang-tidy version that in theory we should use).

ibc (approved):

worker/src/RTC/SeqManager.cpp
jmillan: Do you know which one? we should remove it then.

worker/include/RTC/SeqManager.hpp
```diff
@@ -49,6 +49,8 @@ namespace RTC
 		void ClearDropped();
 
 	private:
+		// Whether at least a sequence number has been inserted.
+		bool started{ false };
```
ibc, line 53: Can you explain how it works? I usually hate those flags specially in stuff that deals with number series that can start with any random number.

ibc (approved):

worker/include/RTC/SeqManager.hpp
jmillan: The very first input value must become the current `maxInput` and `maxOutput`.

Before this flag:

* maxInput is initialized to `0` because it's a number.
* first call to `Insert(input)` arrives.
* `IsSeqHigherThan(input, this->maxInput)` => `IsSeqHigherThan(input, 0)` may be false, so `this-maxInput` remains the default value `0`, which is wrong.
* same for `maxOutput`

ibc (approved):

ibc:
I'm releasing.

ibc:
Deploying in the demo server as well.

ibc:
3.11.20 published in a few minutes.

PR #1047, opened 2023-04-03
Feat(rust): get worker resource

bekirisgor:
It seems worker.get_resource_usage() method is missing.


nazar-pc:
It is missing because it does not make sense. In Node.js it shows resources of the worker process, in Rust everything happens in the same process, so you can use any Rust tool to get resources of the same process, there is no need to call into worker, it'll not give you per-worker resources anyway.

ibc:
Yup, what @nazar-pc says makes sense. Thanks @bekirisgor but this is not needed.

bekirisgor:
> It is missing because it does not make sense. In Node.js it shows resources of the worker process, in Rust everything happens in the same process, so you can use any Rust tool to get resources of the same process, there is no need to call into worker, it'll not give you per-worker resources anyway.

> Yup, what @nazar-pc says makes sense. Thanks @bekirisgor but this is not needed.

Thats great thank you for clarification and maintaining mediasoup.

PR #1048, opened 2023-04-03
RTC Logger

jmillan:
The idea is having a way to see the lifetime of each RTP packet in a single entry (ie: log line). How it arrived (seq num, timestamp, etc) and how it left mediasoup. If it didn't leave, it shows the reason.

Router, Producer, Consumer, Transport IDs are present in each record for a reason.

This initial implementation is purely internal and just logs to stdout the info. A future work would send this via unix socket using flatbuffers which would allow processing the data at application level.

For convenience and readability I have not surrounded every call to the logger with `MS_RTC_LOGGER_RTP`, but the code that prints out the line. I can do it though if preferred.

Initially I thought `RtcLogger` could be a logger for more than RTP but at the moment I think this serves for RTP only at least for now.  Having said this, I could add the definition of the logger and the implementation in the corresponding `RtpPacket.[hpp|cpp]` files. Thoughts @ibc?

```txt
  mediasoup:Worker (stdout) {timestamp: 601293210, recvTransportId: 'a9d0c20c-3f2a-42fd-acce-cd1b519fcac5', sendTransportId: 'undefined', routerId: 'a86746e5-8b72-42cc-9834-1617c6276586', producerId: 'f82380e9-f8e9-4a15-9aa3-676327376623', consumerId: 'ba16321a-1544-4ebf-ac70-3926633f8621', recvRtpTimestamp: 2583604420, sendRtpTimestamp: 0, recvSeqNumber: 9168, sendSeqNumber: 0, dropped: true, dropReason: 'SpatialLayerMismatch'} +0ms
  mediasoup:Worker (stdout) {timestamp: 601293210, recvTransportId: 'a9d0c20c-3f2a-42fd-acce-cd1b519fcac5', sendTransportId: 'undefined', routerId: 'a86746e5-8b72-42cc-9834-1617c6276586', producerId: 'f82380e9-f8e9-4a15-9aa3-676327376623', consumerId: 'ba16321a-1544-4ebf-ac70-3926633f8621', recvRtpTimestamp: 3197226919, sendRtpTimestamp: 0, recvSeqNumber: 5030, sendSeqNumber: 0, dropped: true, dropReason: 'SpatialLayerMismatch'} +0ms
  mediasoup:Worker (stdout) {timestamp: 601293226, recvTransportId: '8b7b112b-c7ca-432c-b8eb-cb2d95ffdf76', sendTransportId: '11e55319-7c14-4f9c-a188-6fb70c4ee5a6', routerId: 'a86746e5-8b72-42cc-9834-1617c6276586', producerId: '416d422c-d1c0-4988-a6e1-939776a92762', consumerId: '489eed52-208a-4fe5-b369-445f42cd99b9', recvRtpTimestamp: 2356655251, sendRtpTimestamp: 0, recvSeqNumber: 12899, sendSeqNumber: 315, dropped: false, dropReason: 'None'} +15ms
  mediasoup:Worker (stdout) {timestamp: 601293226, recvTransportId: 'a9d0c20c-3f2a-42fd-acce-cd1b519fcac5', sendTransportId: 'bfcc1c8b-5950-4539-898e-bd5df834e672', routerId: 'a86746e5-8b72-42cc-9834-1617c6276586', producerId: 'a8501f8f-f5a9-4c56-9761-ab16a9b1b1d1', consumerId: 'e7dede0b-69a4-4d3a-a8ee-5b850e219254', recvRtpTimestamp: 529869993, sendRtpTimestamp: 0, recvSeqNumber: 12436, sendSeqNumber: 313, dropped: false, dropReason: 'None'} +0ms
```

NOTE: Yes, currently logs are printed to stdout using `std::cout`. I know this is not done anywhere like this. It's intentional. Not written in stone though.

### DIFF
--- a/worker/src/RTC/RtcLogger.cpp
+++ b/worker/src/RTC/RtcLogger.cpp
@@ -52,20 +52,37 @@ namespace RTC
 #ifdef MS_RTC_LOGGER_RTP
 			MS_TRACE();
 
-			std::cout << "{"
-			          << "timestamp: " << this->timestamp << ", recvTransportId: '"
-			          << this->recvTransportId << "'"
-			          << ", sendTransportId: '" << this->sendTransportId << "'"
-			          << ", routerId: '" << this->routerId << "'"
-			          << ", producerId: '" << this->producerId << "'"
-			          << ", consumerId: '" << this->consumerId << "'"
-			          << ", recvRtpTimestamp: " << this->recvRtpTimestamp
-			          << ", sendRtpTimestamp: " << this->sendRtpTimestamp
-			          << ", recvSeqNumber: " << this->recvSeqNumber
-			          << ", sendSeqNumber: " << this->sendSeqNumber
-			          << ", dropped: " << (this->dropped ? "true" : "false") << ", dropReason: '"
-			          << dropReason2String[this->dropReason] << "'"
-			          << "}" << std::endl;
+			std::cout << "{";
+			std::cout << "\"timestamp\": " << this->timestamp;
+
+			if (!this->recvTransportId.empty())
+			{
+				std::cout << ", \"recvTransportId\": \"" << this->recvTransportId << "\"";
+			}
+			if (!this->sendTransportId.empty())
+			{
+				std::cout << ", \"sendTransportId\": \"" << this->sendTransportId << "\"";
+			}
+			if (!this->routerId.empty())
+			{
+				std::cout << ", \"routerId\": \"" << this->routerId << "\"";
+			}
+			if (!this->producerId.empty())
+			{
+				std::cout << ", \"producerId\": \"" << this->producerId << "\"";
+			}
+			if (!this->consumerId.empty())
+			{
+				std::cout << ", \"consumerId\": \"" << this->consumerId << "\"";
+			}
+
+			std::cout << ", \"recvRtpTimestamp\": " << this->recvRtpTimestamp;
+			std::cout << ", \"sendRtpTimestamp\": " << this->sendRtpTimestamp;
+			std::cout << ", \"recvSeqNumber\": " << this->recvSeqNumber;
+			std::cout << ", \"sendSeqNumber\": " << this->sendSeqNumber;
+			std::cout << ", \"dropped\": " << (this->dropped ? "true" : "false");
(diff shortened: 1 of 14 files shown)


worker/src/RTC/RtcLogger.cpp
ibc, line 1: Add MS_TRACE() in every method please.

ibc (commented):
I see `sendRtpTimestamp: 0` in all logs you show. I assume it's a bug.

Other than that, wouldn't it be better to generate an event and send it to Node/Rust layers as usual?

worker/src/RTC/RtcLogger.cpp
jmillan: true! done.

jmillan:
> I see sendRtpTimestamp: 0 in all logs you show. I assume it's a bug.

It's 0 when `dropped` is true. Which means that the send RTP timestamp was not set. I'm not taking too much consideration for the format as I'm explaining in the next point.

> Other than that, wouldn't it be better to generate an event and send it to Node/Rust layers as usual?

I'm targeting this for another phase so we do it in flatbuffers and this right now is only for internal purposes. This is why I'm not taking too much consideration for the outgoing format.

ibc:
> > Other than that, wouldn't it be better to generate an event and send it to Node/Rust layers as usual?
> 
> I'm targeting this for another phase so we do it in flatbuffers and this right now is only for internal purposes. This is why I'm not taking too much consideration for the outgoing format.

I just wonder how much useful it will be in current status, but I won't block due to this.

jmillan:
> I just wonder how much useful it will be in current status, but I won't block due to this.

It is useful. In most cases we are going to tell a person that is experiencing an issue to copy/paste the logs or to copy/paste the content of such notification.

If this is a concern let's then think how we want to provide it to node/rust.

IMHO this trace should be enabled by `Router` because it implies:

- routerId
- recvTransportId
- producerId
- consumerId
- sendTransportId

The corresponding filters can then be made at application layer. Giving the possibility to enable this by other elements would IMHO imply too much logic in the worker.

Also consider this is a tool for internal debugging. Anyway I'm open to suggestions.

jmillan:
It could even be enabled even more globally, for `worker`. However I'm not sure if it makes sense to enable it for a specific Consumer only, for example.

jmillan:
Once GH actions pass this will be merged.

PR #1049, opened 2023-04-04
Fix check division by zero in transport congestion control

ggarber:
Add check for expected packets in case an empty TransportFeedback RTCP packet is sent or received with 0 packet statuses.


ibc (approved):
Good catch.

ibc:
Released in 3.11.21

PR #1050, opened 2023-04-04
Fix lost pending statuses in transport feedback

ggarber:
When sending TransportCcFeedbackPacket it doesn't include all the statuses that were not added to a chunk yet because there was no call to TransportCcFeedbackPacket->Finish() method to add those pending ones before sending the message.

Because of that at a low packet rate (<50 packets / sec) the TransportCCFeedbackPacket is empty most of the time and the bandwidth estimation can not be calculated properly.

### DIFF
--- a/worker/src/RTC/TransportCongestionControlServer.cpp
+++ b/worker/src/RTC/TransportCongestionControlServer.cpp
@@ -221,6 +221,8 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		this->transportCcFeedbackPacket->Finish();
+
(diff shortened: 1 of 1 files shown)


worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -221,6 +221,8 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		this->transportCcFeedbackPacket->Finish();
```
ibc, line 224: This is amazing because indeed we were NEVER calling this method (but just in the C++ test).

ibc:
Released in 3.11.21

PR #1058, opened 2023-04-08
Implement fake timers for unit tests

ggarber:
Implement support for fake timers (timers that run "faster than real time") to make unit tests faster and more predictable.

The approach taken in this PR is to have two implementations of the Timer class.  One based on LibUV and one based on a simple map keeping the list of timers (FakeTimerManager).   Which one is compiled depends on the FAKE_TIMERS macro being defined that is defined for the test target now.

One of the use cases is to be able to run long tests (f.e. bandwidth estimation tests that take seconds) in milliseconds in unit tests.   Existing tests like KeyFrameManagerTests are also faster now.

From a design point of view it also removes the dependency with DevLibUV from many files that should be LibUV agnostic. 

Maybe there is a better way to do this, it is open for discussion and I'm happy to take other approach if it makes more sense for mediasoup philosophy or close the PR if it doesn't make sense.

### DIFF
--- a/worker/deps/libwebrtc/libwebrtc/call/rtp_transport_controller_send.cc
+++ b/worker/deps/libwebrtc/libwebrtc/call/rtp_transport_controller_send.cc
@@ -19,7 +19,7 @@
 #include "system_wrappers/source/field_trial.h"
 #include "modules/congestion_controller/goog_cc/goog_cc_network_control.h"
 
-#include "DepLibUV.hpp"
+#include "handles/Timer.hpp"
 #include "Logger.hpp"
 #include "RTC/RTCP/FeedbackRtpTransport.hpp"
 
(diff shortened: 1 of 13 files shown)


jmillan (commented):
This is very convenient @ggarber :+1:

I'd say `FakeTimerManager`  should live in its own file. Also, can you please add some tests which would serve as documentation for its usage?

ibc:
I'll be off for a few days. But as Jose said, please move to a separate file and rename the file and the class to FakeTimeHandle to be in sync with naming changes done in flatbuffers branch.

I don't really know if moving GetTimeXx() methods from Utils to Timer is a good idea. Perhaps yes.

ggarber:
> I'll be off for a few days. But as Jose said, please move to a separate file and rename the file and the class to FakeTimeHandle to be in sync with naming changes done in flatbuffers branch.
> 
> I don't really know if moving GetTimeXx() methods from Utils to Timer is a good idea. Perhaps yes.

Ok, I'll move it as soon as i have time (doing this one in my free time :)).   Thx for taking a look.

ibc:
We are finally adding this. But since this PR is so old we are doing it from scratch here: https://github.com/versatica/mediasoup/pull/1784

PR #1061, opened 2023-04-14
Only perform RTP inactivity checks on simulcast streams

ibc:
We compute the received stream `score` based on many factors, including whether packets are being redeived or not. If the encoder in client side stops sending packets we (in mediasoup) assume that something is wrong (RTP inactivity timer fires) and set score to 0. This PR changes this behavior and only enables the RTP inactivity check on simulcast streams. Why? Because in simulcast we NEED this in order to detect when the higher streams stop so we can quickly switch consumers to the lower stream.

The thing is: From RTP protocol perspective, it's perfectly fine that a device stops sending RTP and it doesn't mean that its uplink connection is bad. It's just that the encoder decided to stop sending packets. This happens, for instance, in mobile apps using libwebrtc with camera enabled: when switching to another app the encoder stops sending more webcam RTP packets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### NEXT
 
+* `RtpStreamRecv`: Only perform RTP inactivity check on simulcast streams ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
 * `SeqManager`: Properly remove old dropped entries ([PR #1054](https://github.com/versatica/mediasoup/pull/1054)).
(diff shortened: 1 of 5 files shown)


jmillan (approved):

thanhnghic4:
@ibc I have a questions,
My program basiclly, receving data from udp then share to other users.
on server, I'm using "PlainTransport",
on client (not browers, just local machine), I run gstreamer pipeline to stream video from webcam to udp port.
And I'm using score event to know when client stop streaming.
After this PR, it's not working any more.
So how can I replace this.
basically, from server, I need a event to know when client stop run gstream pipeline

ibc:
Please use the forum for questions. We don't use GitHub as a forum.

PR #1062, opened 2023-04-15
Replace static variables with regular class variables, remove unused includes

nazar-pc:
Not 100% sure, but I think this is a correct fix, but tests pass and Rust doesn't seem to fail with random memory-related bugs originating from C++ anymore.

Please, don't use static variables for mutable data going forward. Yes, it is convenient, but it is a bad practice IMHO.

### DIFF
--- a/worker/include/Channel/ChannelRequest.hpp
+++ b/worker/include/Channel/ChannelRequest.hpp
@@ -24,15 +24,12 @@ namespace Channel
 		static absl::flat_hash_map<FBS::Request::Method, const char*> method2String;
 
 	public:
-		static flatbuffers::FlatBufferBuilder bufferBuilder;
-
-	public:
 		ChannelRequest(Channel::ChannelSocket* channel, const FBS::Request::Request* request);
 		~ChannelRequest() = default;
 
-		flatbuffers::FlatBufferBuilder& GetBufferBuilder() const
+		flatbuffers::FlatBufferBuilder& GetBufferBuilder()
 		{
-			return bufferBuilder;
+			return this->bufferBuilder;
 		}
 		void Accept();
 		template<class Body>
@@ -42,7 +39,7 @@ namespace Channel
 
 			this->replied = true;
 
-			auto& builder = ChannelRequest::bufferBuilder;
+			auto& builder = this->bufferBuilder;
 			auto response = FBS::Response::CreateResponse(builder, this->id, true, type, body.Union());
 
 			auto message = FBS::Message::CreateMessage(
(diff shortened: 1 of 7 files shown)


worker/include/Channel/ChannelRequest.hpp
```diff
@@ -23,16 +23,13 @@ namespace Channel
 	public:
 		static absl::flat_hash_map<FBS::Request::Method, const char*> method2String;
 
-	public:
-		static flatbuffers::FlatBufferBuilder bufferBuilder;
```
ibc, line 27: This should have been thread_xxxx (I don't remember the name now) as other static members to have one per thread, right? Note: I'm not suggesting going that way.

worker/include/Channel/ChannelRequest.hpp
```diff
@@ -60,13 +57,14 @@ namespace Channel
 
 	private:
 		void Send(uint8_t* buffer, size_t size);
-		void Send(const flatbuffers::Offset<FBS::Response::Response>& reponse);
+		void SendResponse(const flatbuffers::Offset<FBS::Response::Response>& response);
```
ibc, line 60: Why Send() and SendResponse()? Perhaps we should have Notify and Reply() instead?

worker/include/Channel/ChannelRequest.hpp
nazar-pc: There is `thread_local`, but IMHO if we already have a class and methods on it, better have explicit dependency rather than thread local variables everywhere.

worker/include/Channel/ChannelRequest.hpp
nazar-pc: From what I understand this one is calling `Send` internally, `Send` only operates on bytes, while this operates on strongly typed response. Jose should be able to tell more when he is back.

worker/include/Channel/ChannelRequest.hpp
ibc: Yes, that's what I meant.

worker/include/Channel/ChannelRequest.hpp
jmillan: Correct. Both methods are private. `Send(Response)`  sends exactly that and the other the given buffer. I'll remove `Send(buffer, size)` after this PR is merged since it's a one line method used in a single place.

jmillan:
The PR makes a lot of sense. I should have done it this way since the beginning.

jmillan (approved):

ibc:
Let's merge this. Thanks.

PR #1063, opened 2023-04-17
flatbuffers

jmillan:
WIP to remove JSON in favour of [flatbuffers](https://github.com/google/flatbuffers) for as the worker communication format.

* JSON parsing is CPU intensive.
* JSON is not type safe. 
* flatbuffers does not parse the buffer. Reading the message takes 0ms.
* flatbuffers is type safe since code (C++, Typescript, Rust, etc) is autogenerated out of the schema definitions.

Extras that come along with this changes:
* There is now a single Channel for worker communication. 
  * Previously there were 2 (Channel and PayloadChannel)
* Each message sent Node->Worker requires a single write() call to Channel.
  * Previously 2 calls per message were needed for Channel and 4 for PayloadChannel.
* Each message send Worker->Node previously executed two extra calls to `memcpy` that are now executed now.

[Feature request](https://github.com/google/flatbuffers/issues/7656) for making optional field getters return `undefined` instead of `null` if the value is not set.

- [ ] [Planus](https://github.com/planus-org/planus/issues/180) not supporting same table name in different namespaces.
- [ ] flatbuffers (TS) should throw `TypeError` instead of `Error` when there is a missing required field. [PR](https://github.com/google/flatbuffers/pull/7910)


ibc:
So is this PR mergeable in v3 without conflicts?

nazar-pc:
Technically, yes

jmillan:
This branch only contains FBS related changes and those that where only present in `flatbuffers` branch. So it's definitely more clean than the previous one.

jmillan:
Renamed `flatbuffers-new` to `flatbuffers`. Final draft PR here #1064

PR #1064, opened 2023-04-17
flatbuffers

jmillan:
WIP to remove JSON in favour of [flatbuffers](https://github.com/google/flatbuffers) for as the worker communication format.

* JSON parsing is CPU intensive.
* JSON is not type safe. 
* flatbuffers does not parse the buffer. Reading the message takes 0ms.
* flatbuffers is type safe since code (C++, Typescript, Rust, etc) is autogenerated out of the schema definitions.

Extras that come along with this changes:

* There is now a single Channel for worker communication. 
  * Previously there were 2 (Channel and PayloadChannel)
* Each message sent Node->Worker requires a single write() call to Channel.
  * Previously 2 calls per message were needed for Channel and 4 for PayloadChannel.
* Each message sent Worker->Node previously executed two extra calls to `memcpy` that are not executed now.

### TODO: flatbuffers related
- [ ] Try to scope notifications per instance so we don't rely on a global flatbuffers notification event enum but an enum per class (producer, consumer, transport, etc.):  https://github.com/versatica/mediasoup/pull/1162 (non blocker)
- [ ] Do not use heap on flatbuffers generation (non blocker)
- [ ] Do not use head on flatbuffers processing (non blocker)

### TODO: flatbuffers unrealed. Changes made in flatbuffer branch, not present in v3
- [x] Document new `TransportListenInfo` in the website (PR https://github.com/versatica/mediasoup/pull/1084).
- [x] Document new `rtcpListenInfo` in `PlainTransportOptions` once PR https://github.com/versatica/mediasoup/pull/1099 is merged here.
- [x] Document the new pause/resume feature in `DataProducer` and `DataConsumer` (PR https://github.com/versatica/mediasoup/pull/1104):
  - `pause()`and `resume()` in `DataProducer` and `DataConsumer`.
  - `paused` getter in `DataProducer` and `DataConsumer`.
  - `dataproducerpause` and `dataproducerresume` events in `DataConsumer`.
  - `pause` and `resume` events in `DataProducer.observer` and `DataConsumer.observer`.
  - `paused?: boolean` in `transport.consumeData()` and `transport.produceData()`.
- [x] Document the new subchannels stuff (PR https://github.com/versatica/mediasoup/pull/1152).


ibc:
@jmillan this branch uses NPM `flatbuffers` 23.3.3 and there is 23.5.9 already. CHANGELOG here: https://github.com/google/flatbuffers/blob/master/CHANGELOG.md

I'm merging `v3` into `flatbuffers` and will later upgrade this dep.

ibc:
> @jmillan this branch uses NPM `flatbuffers` 23.3.3 and there is 23.5.9 already. CHANGELOG here: https://github.com/google/flatbuffers/blob/master/CHANGELOG.md
> 
> I'm merging `v3` into `flatbuffers` and will later upgrade this dep.

Done and pushed.

ibc:
Questions. Sometimes I see this:

```c++
if (flatbuffers::IsFieldPresent(options->listenInfo(), FBS::Transport::ListenInfo::VT_ANNOUNCEDIP))
```

and sometimes I see this (which is way nicer):

```c++
if (body->port().has_value())
{
    port = body->port().value();
}
```

I'd love if the latter could be used everywhere, no idea if possible.

ibc:
I may be crazy but `make lint` doesn't detect wrong changes in `PlainTransport.cpp` although it does in other `XxxxTransport.cpp` classes...

ibc:
@jmillan, in the TODO items above:

> flatbuffers (TS) should throw `TypeError` instead of `Error` when there is a missing required field. [PR done](https://github.com/google/flatbuffers/pull/7910).

It's been merged. Not sure if there is a new release of `flatbuffers` for Node, but if so we should change some code, right?

jmillan:
>> flatbuffers (TS) should throw TypeError instead of Error when there is a missing required field. https://github.com/google/flatbuffers/pull/7910.

> It's been merged. Not sure if there is a new release of flatbuffers for Node, but if so we should change some code, right?

Removed explicit `TypeError` exceptions.

jmillan:
@ibc, the `DataConsumer.[pause|consumer]()` implementation is making the test fail in GH actions. 

Calling `pause()` or `resume()` is sending the corresponding method to the worker and only after receiving the response it's setting the local `paused` member.

The test just checks if `dataConsumer.paused` in order to increase a variable in the test `effectivelySentMessageBytes`, but it can happen that the `pause()` method has been sent but not terminated, so the Node instance `paused` member is false.

I believe that can be the problem, and probably it's not an error per-se but something to be fixed in the test. I was able to repro the error just once, locally:

`TEST_FILE=node/src/tests/test-DirectTransport.ts npm run test:node`

Screenshot 2023-07-17 at 13 03 55

ibc:
Specially in flatbuffers branch, requests and data messages are sent over the same channel so it's guaranteed that, if I pause the DataProducer and then send a message, such a message won't reach the DataConsumers. I'm 95% sure I took this into account when I wrote the corresponding tests. So can we agree on these things?

1. Code is good and behaves as expected.
2. So maybe tests are wrongly written.

ibc:
I think the problem could he here:

CleanShot-2023-07-17-at-13 41 50@2x

We are setting the `paused = true` flag in `pause()` method ONCE we get ok from the worker. However in the test we do NOT await for `dataProducer.pause()` to resolve, and in lines below the test checks `dataProducer.paused` to decide things.

ibc:
It happened to me too. I'm working on it.
CleanShot-2023-07-17-at-13 45 15@2x

ibc:
@jmillan does it make sense that `dataProducer.pause()` sends a request to the worker (same for `Dataconsumer` and same for `resume()` method in both? For instance, `dataProducer.send()` sends a **notification** to the worker. We have experience on this. Sending a request means that `dataConsumerpause()` will throw if the `DataConsumer` is closed and forces the user to always add `catch()` etc.

I think it's a good timing to refactor those requests and make them notifications. On it.

ibc:
> @jmillan does it make sense that `dataProducer.pause()` sends a request to the worker (same for `Dataconsumer` and same for `resume()` method in both? For instance, `dataProducer.send()` sends a **notification** to the worker. We have experience on this. Sending a request means that `dataConsumerpause()` will throw if the `DataConsumer` is closed and forces the user to always add `catch()` etc.
> 
> I think it's a good timing to refactor those requests and make them notifications. On it.

It's not a good timing for doing this. Let's do this in v4. Milestone created in GH and first item:

https://github.com/versatica/mediasoup/issues/1121

ibc:
@jmillan I've pushed the fix: https://github.com/versatica/mediasoup/pull/1064/commits/40379354cb60634c4051f5c2a9657b5964211aa5
Interesting thing is that the corresponding Rust test was properly done from the beginning (it awaits for puase() and resume() to resolve).

piranna:
How does `flatbuffers` compare in memory and performance to other binary alternatives like [CBOR-X](https://github.com/kriszyp/cbor-x)? I though it was the faster and most memory efficient...

ibc:
There is no reason to compare flatbuffers (available in JS, C++, Rust) with something that is just for JS (so not useful for us).

piranna:
I'm reading about Flatbuffers and looks really interesting, and seems to be on pair in both memory and performance. I've open an issue at https://github.com/google/flatbuffers/issues/8046 asking for some benchmark metrics comparing both of them.

threema-lenny:
> I'm reading about Flatbuffers and looks really interesting, and seems to be on pair in both memory and performance. I've open an issue at [google/flatbuffers#8046](https://github.com/google/flatbuffers/issues/8046) asking for some benchmark metrics comparing both of them.

It makes no sense to compare the two in this particular use case. CBOR needs serialisation and deserialisation whereas FlatBuffers does not. The winner is therefore already known. Everything else is an implementation detail. They target totally different use cases. Using FlatBuffers for IPC is a very sensible approach. CBOR would introduce similar overhead to JSON.

piranna:
So, maybe CBOR can be a advantage on the bandwidth usage, for example by sending small numbers as varlen instead of full ints, isn't It?

ibc:
Can we please avoid discussing in this PR about different solutions to the same problem? Flatbuffers was chosen after long research and we are not gonna change it. Also, I already said that we need something that works in Node, Rust and C++. Flatbuffers does. CBOR is only for JS.

nazar-pc:
> So, maybe CBOR can be a advantage on the bandwidth usage, for example by sending small numbers as varlen instead of full ints, isn't It?

We don't care about bandwidth usage at all, please don't distract us with irrelevant discussions here. Feel free to start a topic on the forum though.

piranna:
I'm not discussing about a different solution, just wanted to know about why was this one choosen, definitely it seems to be one of the best options, just I didn't know about it. Sorry for the inconvenience.

ibc:
It was chosen because it's available in all languages we need, it has very good performance (specially compared to parsing and serializing JSON in C++ using JsonCpp lib, Jose did benchmarks), and doesn't require any memory allocation (well, just a global buffer) since it writes and reads in place.

ibc:
So regarding the Rust `ntapi` issue:

> > https://github.com/versatica/mediasoup/actions/runs/6121294807/job/16619834171?pr=1152
> > ```
> > error[E0793]: reference to packed field is unaligned
> >     --> C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ntapi-0.3.7\src\ntexapi.rs:2783:52
> >      |
> > 2783 |         *tick_count.QuadPart_mut() = read_volatile(&(*USER_SHARED_DATA).u.TickCountQuad);
> >      |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> >      |
> >      = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
> >      = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
> >      = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use 
> > ```
> 
> So there is a related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1826485
> 
> It seems to be fixed in `ntapi` 0.4.0: https://bugzilla.mozilla.org/show_bug.cgi?id=1826485#c2
> 
> However I have no idea which `ntapi` version we are using (whatever `ntapi` is). Output of `cargo tree` doesn't show `ntapi` at all.
> 
> Then people say here that _"For some reason (I don't know why) Cargo picked an older version of tokio that used (indirectly) an older version of ntapi. Force Cargo to use new tokio by setting its version to "1.28.0" (the newest version), and it should work."_. We have `tokio` 1.17.0 (whatever `tokio` is) according to `Cargo tree`. But of course we don't specify `tokio` in any `Cargo.toml` so...

@nazar-pc can we do this in `mediasoup/rust/Cargo.toml`?:
Screenshot 2023-09-11 at 12 27 04

nazar-pc:
> @nazar-pc can we do this in mediasoup/rust/Cargo.toml?:

Just do `cargo update -p tokio` and it'll be updated in lock file. Or `cargo update` to update all dependencies at once.

jmillan:
Update: after running the current benchmarks in Rust (only available for dataproducer.send()), I see an overall performance increase of 15% comparing to v3.

I've not done any other performance test. It would be interesting to write some test implying more transformations to from FBS.

nazar-pc:
Will probably get way better once we stop allocating right and left in Rust code (pretty much everything is boxed right now, sometimes a few times).

ibc:
There is a bug. I run mediasoup `flatbuffers` in the demo with `DEBUG="*"` and I see this wrong "unexpected" output:

```
  mediasoup:Worker constructor() +0ms
  mediasoup:Worker spawning worker process: /Users/ibc/src/v3-mediasoup/worker/out/Release/mediasoup-worker --logLevel=debug --logTag=info --logTag=ice --logTag=dtls --logTag=rtp --logTag=srtp --logTag=rtcp --logTag=rtx --logTag=bwe --logTag=score --logTag=simulcast --logTag=svc --logTag=sctp --rtcMinPort=40000 --rtcMaxPort=49999 +0ms
  mediasoup:Channel constructor() +0ms
  mediasoup:PayloadChannel constructor() +0ms
worker[pid:77521] unexpected data:


_Dmediasoup-worker::mediasoup_worker_run() | starting mediasoup-worker process [version:3.12.13]
worker[pid:77521] unexpected data:


FDmediasoup-worker::mediasoup_worker_run() | little-endian CPU detected
worker[pid:77521] unexpected data:


IDmediasoup-worker::mediasoup_worker_run() | 64 bits architecture detected
worker[pid:77521] unexpected data:


1DSettings::PrintConfiguration() | <configuration>
worker[pid:77521] unexpected data:


@DSettings::PrintConfiguration() |   logLevel             : debug
worker[pid:77521] unexpected data:


wDSettings::PrintConfiguration() |   logTags              : info,ice,dtls,rtp,srtp,rtcp,rtx,bwe,score,simulcast,svc,sctp
worker[pid:77521] unexpected data:


@DSettings::PrintConfiguration() |   rtcMinPort           : 40000
worker[pid:77521] unexpected data:


@DSettings::PrintConfiguration() |   rtcMaxPort           : 49999
worker[pid:77521] unexpected data:


`DSettings::PrintConfiguration() |   libwebrtcFieldTrials : WebRTC-Bwe-AlrLimitedBackoff/Enabled/
worker[pid:77521] unexpected data:


2DSettings::PrintConfiguration() | </configuration>
worker[pid:77521] unexpected data:


3DDepLibUV::PrintVersion() | libuv version: "1.44.2"
worker[pid:77521] unexpected data:


GDDepOpenSSL::operator()() | openssl version: "OpenSSL 3.0.8 7 Feb 2023"
worker[pid:77521] unexpected data:


<DDepLibSRTP::ClassInit() | libsrtp version: "libsrtp2 2.5.0"

[opening Readline Command Console...]
type help to print available commands
cmd> worker[pid:77521] unexpected data:


"DDepUsrSCTP::ClassInit() | usrsctp
worker[pid:77521] unexpected data:


\DDepLibWebRTC::ClassInit() | libwebrtc field trials: "WebRTC-Bwe-AlrLimitedBackoff/Enabled/"
worker[pid:77521] unexpected data:


tDwebrtc::FieldTrial::InitFieldTrialsFromString() | Setting field trial string: WebRTC-Bwe-AlrLimitedBackoff/Enabled/
worker[pid:77521] unexpected data:


�DRTC::DtlsTransport::CreateSslCtx() | setting SRTP cryptoSuites for DTLS: SRTP_AEAD_AES_256_GCM:SRTP_AEAD_AES_128_GCM:SRTP_AES128_CM_SHA1_80:SRTP_AES128_CM_SHA1_32
worker[pid:77521] unexpected data:


�DRTC::DtlsTransport::GenerateFingerprints() | sha-224 fingerprint: DF:1B:4A:DD:8B:B7:F9:2D:4D:EE:25:06:F0:B5:A1:27:C3:A4:C0:E5:65:17:2B:EB:46:D0:C9:AD
worker[pid:77521] unexpected data:


�DRTC::DtlsTransport::GenerateFingerprints() | sha-256 fingerprint: BA:2D:F8:86:1F:77:28:BD:10:56:93:F8:A0:01:04:05:3A:70:A8:D2:07:BD:75:36:9C:F5:B2:2B:1C:F1:7D:2D
worker[pid:77521] unexpected data:


~DRTC::DtlsTransport::GenerateFingerprints() | sha-1   fingerprint: 7F:48:8A:6F:08:78:CB:19:15:28:0A:79:7C:FE:7C:FC:59:28:D9:3D
worker[pid:77521] unexpected data:


�DRTC::DtlsTransport::GenerateFingerprints() | sha-384 fingerprint: 9F:79:AA:43:DE:42:53:50:3A:A9:EA:E5:02:17:16:87:27:F4:AE:29:58:EE:72:6C:74:15:84:F2:4A:B5:D4:9E:19:37:4B:5B:A0:C2:29:9C:8E:4F:D1:3D:70:9C:23:B3
worker[pid:77521] unexpected data:


DRTC::DtlsTransport::GenerateFingerprints() | sha-512 fingerprint: 1E:09:89:46:B0:36:D9:B9:0A:B5:66:FF:1B:63:C2:DC:92:95:58:62:FB:61:38:A9:26:52:F4:23:48:32:6B:D5:14:55:6B:25:38:2D:45:46:3D:C9:37:31:CE:C9:5D:A7:11:DB:03:49:8D:B6:8D:B4:3A:DF:5D:AD:2A:EC:1C:44
worker[pid:77521] unexpected data:




77521
```

Note that it's not "unexpcted". It's data that the worker must send to Node. So it looks like if it's being properly received by Node but then Node is also printing it to the console (no idea, it could be another thing).

ibc:
**BUG:**

```
webRtcServer.dump()
ERROR: webRtcServer.dump() failed: TypeError: Do not know how to serialize a BigInt
```

### STR:

- Run the demo and start a meeting.
- In the interactive console in the server, run `dwrs`.

I've added this item to the TODO list in this PR description.

ibc:
Question: is there any reason for all `dump()` methods to still return `Promise<any>`? We do have those returned TS types defined since they are FBS based, right?

ibc:
> **BUG:**
> 
> ```
> webRtcServer.dump()
> ERROR: webRtcServer.dump() failed: TypeError: Do not know how to serialize a BigInt
> ```

I've created a separate ticket for this: https://github.com/versatica/mediasoup/issues/1180

jmillan:
> Question: is there any reason for all dump() methods to still return Promise<any>? We do have those returned TS types defined since they are FBS based, right?

They are supposed to be typed already https://github.com/versatica/mediasoup/blob/flatbuffers/node/src/WebRtcTransport.ts#L438

If you find any untyped please type it or add a TODO item on this PR.

ibc:
> > Question: is there any reason for all dump() methods to still return Promise<any>? We do have those returned TS types defined since they are FBS based, right?
> 
> 
> 
> They are supposed to be typed already https://github.com/versatica/mediasoup/blob/flatbuffers/node/src/WebRtcTransport.ts#L438
> 
> 
> 
> If you find any untyped please type it or add a TODO item on this PR.

I meant WebRtcServer.dump() and I've seen others. I'll create a separate issue.

ibc:
I've created a separate issue: 
Must add TS return type to all dump() methods: https://github.com/versatica/mediasoup/issues/1181

jmillan:
> Must add TS return type to all dump() methods: https://github.com/versatica/mediasoup/issues/1181

@ibc, I don't agree with this TODO list here. It could have been perfectly typed before, as most of responses are, generally `dump()` responses have not been typed nor documented on purpose. This is not a blocking task for this branch, but can be something to do anytime. This branch already has too many working hours and I don't won't to add extra weight on it.

jmillan:
> Be ready to serialize BigInt in Node

@ibc, what are we supposed to do here? We are casting every BigInt to Number already. I don't think this is a blocker for this branch.

ibc:
> generally dump() responses have not been typed nor documented on purpose. This is not a blocking task for this branch

Agreed, removed from TODO. It's just a separate issue now.

ibc:
> We are casting every BigInt to Number already. I don't think this is a blocker for this branch.

Are we? I just found a missing case yesterday. I'm not saying that there are more, but are we sure? The error I fixed yesterday just happens when a dump (or whatever response or notification) contains a super big number. It doesn't fail in build time but just in runtime in certain cases. Are we sure we are casting to Number absolutely all fields defined as uint64 or int65 in FBS? If so we are good.

ibc:
I've added a blocker: https://github.com/versatica/mediasoup/issues/1184

jmillan:
@nazar-pc, time has come for the first two TODO items:

 - [ ] Build fbs.rs at build time (planus).
 - [ ] Format fbs schemas at build time (planus).

nazar-pc:
What do you mean by formatting?

jmillan:
> What do you mean by formatting?

Planus CLI has a ' format' option that formats the flatbuffers schemas files. We should have this done the same way we run `cargo fmt` on the rest of the code.

nazar-pc:
Can you run it manually or you want to check in CI, what is the goal?

jmillan:
> Can you run it manually or you want to check in CI, what is the goal?

We can run it manually. No need to check in CI since it just overrides current files properly formatted. I'm removing from the TODO list.

jmillan:
About `Build fbs.rs at build time (planus).` task. It's not a blocker for me. It is a single autogenerated file what Planus outputs from all schemas so I'm happy generating it with `planus-cli` and keeping in the repo as we do now.

nazar-pc:
I will work on that, but probably closer to the week end, very busy

ibc:
> About `Build fbs.rs at build time (planus).` task. It's not a blocker for me. It is a single autogenerated file what Planus outputs from all schemas so I'm happy generating it with `planus-cli` and keeping in the repo as we do now.

Please let's not release without this. Having to rewrite those things in `fbs.rs` after every change in any FBS file is tedious and it would just be impossible for any other current or future contributor.

We need a proper command to generate `fbs.rs` without having to manually edit it after it. BTW should this be a new task in `worker/Makefile` or somewhere else? Can we define a custom `cargo run planus` command or similar? Does such a thing exist in Cargo land?


> "Format fbs schemas at build time (planus)"

What does this mean? Format which fbs schemas? You mean the source FBS files in `worker/fbs/` folder? Why do we want that those files are autoformatted by planus?

jmillan:
> What does this mean? Format which fbs schemas? You mean the source FBS files in worker/fbs/ folder? Why do we want that those files are autoformatted by planus?

Planus is the only guy in town that has a formatter for .fbs files.

jmillan:
Some performance notes:

|| V3 | flatbuffers |
|-|----|-----------|
|Producer creation| 42us | 24us|
|Producer dump  | 30us | 14us|

Basically we can say we can now do twice as much operations as in v3 in the same time.

ibc:
@jmillan this TODO item in the PR description references a merged PR:

> - [ ] Try to scope notifications per instance so we don't rely on a global flatbuffers notification event enum but an enum per class (producer, consumer, transport, etc.): https://github.com/versatica/mediasoup/pull/1158/files#r1328919685 (non blocker)

Should be checked?

jmillan:
> @jmillan this TODO item in the PR description references a merged PR:
> 
> > * [ ]  Try to scope notifications per instance so we don't rely on a global flatbuffers notification event enum but an enum per class (producer, consumer, transport, etc.): https://github.com/versatica/mediasoup/pull/1158/files#r1328919685 (non blocker)
> 
> Should be checked?

The link was wrong, fixed.

ibc:
@jmillan @nazar-pc, the most important missing piece here is:

> - [ ] Build `rust/src/fbs.rs` at build time (planus) (blocker)

So I'd like to discuss about this.

### Problem

Currently, as soon as we change/add/delete any file in `worker/fbs` folder, we must perform these steps manually in `rust` directory:

1. We need to have `planus-cli` installed (`cargo install planus-cli`).
2. `planus rust -o src/fbs.rs ../worker/fbs/*.fbs`
3. Then revert some of the changes at the top of `rust/src/fbs.rs` file to make sure it re-exports things the same way and suppresses clippy warnings.

### Questions

1. Does an external contributor have to install `planus-cli` manually? Or are we gonna add `planus-cli` in `[dev-dependencies]` in `rust/Cargo.toml`?
  - Off-topic: How is the `cargo` command to install dev deps?

2. So assuming we have `planus-cli`, how are we gonna automatize steps 2 and 3 above? Is this gonna be a new task in main `Makefile`? Or can we have a custom `cargo run planus` command (does such a thing exist?) and use `cargo-cli` as a lib?

nazar-pc:
1. Neither, we'll use Planus as a library and compile files programmatically from `build.rs` instead
2. They are installed automatically when you run tests, etc. It is not the same notion as Node.js. For equivalent stuff to Node.js there are `[build-dependencies]`, those are dependencies needed for `build.rs`
3. We will programmatically compile files as part of `build.rs` and emit [`cargo:rerun-if-changed`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) to instruct Cargo to re-run `build.rs` only if any of the files change (should also include C++ files)

ibc:
> we'll use Planus as a library and compile files programmatically from `build.rs` instead

- So we have to run `cargo build` or `cargo test` to update fbs.rs, right?
- So we have to add ugly code in `build.rs` to "revert some of the changes at the top of `rust/src/fbs.rs` file to make sure it re-exports things the same way and suppresses clippy warnings", right?

nazar-pc:
> So we have to run cargo build or cargo test to update fbs.rs, right?

Neither and both. We will no longer ship `fbs.rs`, it will be compiled automatically when you run `cargo test` or use `mediasoup` as a dependency, but as mentioned before, only if it wasn't compiled already or corresponding files have changed.

> So we have to add ugly code in build.rs to "revert some of the changes at the top of rust/src/fbs.rs file to make sure it re-exports things the same way and suppresses clippy warnings", right?

I believe we do not, we can just suppress warnings on the whole module. Basically we will have `#[allow(clippy::whatever)] mod fbs {included_code_here}` or something like that in `lib.rs`, no manual search/replace will be necessary in `build.rs`.

ibc:
I'm writing all missing docs in https://github.com/versatica/mediasoup-website/pull/50

.gitignore
```diff
@@ -7,8 +7,10 @@
 ## Node.
 /node_modules
 /node/lib
-# Flatc generated files.
+# flatc generated files.
 /node/src/fbs
+# nvm version file.
+/.nvmrc
```
nazar-pc: Why was this necessary? You can ignore it locally, I don't see why it is necessary in the repo.

ibc: I cannot ignore locally (meaning in my global .gitignore) since there are projects that **indeed** want this file into the repo. Just wondering what is so wrong with this addition in .gitignore taking into account that there are many other similar files in there such as ".idea" and so on.

nazar-pc: You can put whatever you want into `.git/info/exclude` of each repo, no need to put stuff not related to the project in `.gitignore`.

ibc: Do you mean that such a "exclude" folder is per local repo (so you and me can have different things into it)?

BTW why did you never say anything about this? 😅

image

nazar-pc: > Do you mean that such a "exclude" folder is per local repo (so you and me can have different things into it)?

Exactly, I usually put `.idea` in there when contributing to various projects, I don't expect all of them to use or care about JetBrains's IDEs.

> BTW why did you never say anything about this? 😅

Maybe I should have :man_shrugging:

ibc: Ok I'll write PR later. I'll remove all those items but /.cache and /coverage which are autogenerated when running certain tasks. What about .tgz? Are we generating something.tgz and leaving it in root folder?

nazar-pc: Things that are created by certain tasks actually should be in there, stuff that have nothing to do with the repo itself should not.

piranna: I also add `/*.tgz` to `.gitignore` file, and also on `.npmignore`.

ibc: We don't need .npmignore because we are explicit about npm package files in `package.json` (we fill the "files" field). Also, if `.npmignore` is not preset it reads `.gitignore`.

ibc: No need for a PR. I've done basically what is told above: https://github.com/versatica/mediasoup/pull/1064/commits/96f0f45a3d3478e4f1d815f84b06da9b8b0cff75

ibc:
@jmillan I'm marking this PR as ready.

PR #1065, opened 2023-04-17
[flatbuffers] Add ListenInfo in all transports and send/recv buffer size options

### LINKED ISSUE #554: allow setting receive buffer size
## Feature Request

I see no way to set the UDP receive buffer size (i.e. SO_RCVBUF), which can be a problem when trying to ingest 'bursty' streams into a router. Would it make sense to add it?

ibc:
- Fixes #554

*NOTE:* This branch was created from `flatbuffers` branch and we must wait for `flatbuffers` to be merged in `v3` before merging this PR.

### Details

- Add optional `sendBufferSize` and `recvBufferSize` (in bytes) in `ListenInfo` type.
- Deprecate `ListenIp` type and allow passing `listenInfo(s)` in all `router.createXxxxTransport()`.

### Bonus Tracks

- Add `{ }` in more C++ condition blocks.
- Rename `handlers/XxxxHandler.xxx` to `handlers/XxxxHandle.xxx` becase those are "handles" rather than "handlers" (in fact they wrap a "libuv handle").


node/src/WebRtcTransport.ts
```diff
@@ -31,11 +32,19 @@ import { IceState as FbsIceState } from './fbs/web-rtc-transport/ice-state';
...
-export type WebRtcTransportListenIndividual =
+type WebRtcTransportListenIndividual1 =
+{
+	/**
+	 * Listening info.
+	 */
+	listenInfos: TransportListenInfo[];
+};
+
+type WebRtcTransportListenIndividual2 =
```
jmillan: Couldn't these be called `WebRtcTransportListenInfos` and `WebRtcTransportListenIps` ?

jmillan (approved):

node/src/WebRtcTransport.ts
ibc: Nope because the latter also includes an optional port so such a name would be confusing.

ibc:
@nazar-pc please tell me you are happy with this.

node/src/WebRtcTransport.ts
ibc: Renamed here to more suitable names: https://github.com/versatica/mediasoup/pull/1065/commits/70662256731da0e2c4cd202811e9dab2a4c6ced2

ibc:
I'm closing this PR (since somehow it became hell due to some force push actions in `flatbuffers` branch). And I've done it again in https://github.com/versatica/mediasoup/pull/1084.

This is the 3rd time I repeat the same work so I'm definitely wishing to merge this PR in `flatbuffers` branch ASAP. Nope, I cannot merge it first in `v3` because:
1. It's a feature that can wait for `flatbuffers` to land.
2. However it doesn't mean that I can repeat same work over and over to avoid awesome GIT conflicts.
3. Doing it for `v3` first would lead to unavoidable GIT conflicts (I already tried it, never again).

PR #1068, opened 2023-04-19
SeqManager: performance, invert std::distance range

### LINKED ISSUE #1067: High mediasoup-worker CPU usage with 3.11.22 version
## Bug Report
After upgrading to 3.11.22 version, we are experiencing a very high CPU usage when the number of video consumers increases, probably caused by the new rtx buffer implementation.
Here you have a graph of the mediasoup worker CPU usage when running a test with 1 SD producer and 40 consumers:
- at the left: using 3.11.22 version
- at the right: using a custom version (https://github.com/vpalmisano/mediasoup/commits/3-11-13-fork) obtained picking the commits starting from 3.11.13 to 3.11.22 excluding the rtx buffer refactor commits.

image

jmillan:
Fixes #1067.

There are way more dropped elements in the range `[begin(), input]` than in `[input, end()]` and `std::distance` complexity is linear.

All tests pass and performance has gone to normal.

@vpalmisano, could you please give it a try too for double confirmation? Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### 3.11.24
+
+* `SeqManager`: Fix performance regression ([PR #1068](https://github.com/versatica/mediasoup/pull/1068), thanks to @vpalmisano for properly reporting).
+
+
 ### 3.11.23
 
 * Node: Fix `appData` for `Transport` and `RtpObserver` parent classes  ([PR #1066](https://github.com/versatica/mediasoup/pull/1066)).
--- a/worker/src/RTC/SeqManager.cpp
+++ b/worker/src/RTC/SeqManager.cpp
@@ -59,7 +59,10 @@ namespace RTC
 		if (SeqManager<T, N>::IsSeqHigherThan(input, this->maxInput))
 		{
 			this->maxInput = input;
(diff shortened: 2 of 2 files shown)


ibc (approved):

vpalmisano:
Yep, CPU usage is back to normal now :rocket: 
image

jmillan:
Added an extra performance enhancement: 5a4055c83f034fce36aaaed18a63d3b18096161f

Since we know that input values added to `dropped()` are the highest seen, explicitly indicate `std::set()` to add the item at the end. This way it does not need to calculate its position by performing comparisons.

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### 3.11.24
```
ibc, line 4: Version must be added when releasing. Here in a PR it should be "Next".

ibc (approved):

ibc:
@vpalmisano please tell us your preferred personal site (GH, LinkedIn or whatever).

vpalmisano:
> please tell us your preferred personal site (GH, LinkedIn or whatever).

https://github.com/vpalmisano is ok ;)

ibc:
https://mediasoup.org/about/#acknowledgements

Issue #1069, opened 2023-04-21
Large packet loss when estimated bandwidth is close to the lowest SVC layer


vpalmisano:
## Bug Report
When the estimated bandwidth is close to the lowest SVC layer, the allocated layer fluctuates between 0:0 and -1:-1. Concurrently, we observed large packet loss spikes at receiver side (80-100%) that didn't happen when using VP8/Simulcast in the same conditions. Analyzing the captured RTP packets, there are large seqno jumps that could explain the packet loss:

![nopatch](https://user-images.githubusercontent.com/283319/233715362-dc34a9b3-870c-49a5-a550-06e53fd5943d.png)
[nopatch.pcapng.gz](https://github.com/versatica/mediasoup/files/11298614/nopatch.pcapng.gz)

The issue can be reproduced consuming 1 VP9 stream and calling `transport.setMaxOutgoingBitrate(80000)` (set a number close to the lowest encoded SVC layer).

buptlsp:
I also found a strange seqno issue. sqlno will jump every 20 packets in some scenarios. Currently I don't know what scenario. Can you please tell me if this pr of yours can solve this problem?

![image](https://github.com/versatica/mediasoup/assets/3291512/abb5ce85-7502-41a9-870d-c291856ab09d)
![image](https://github.com/versatica/mediasoup/assets/3291512/70b71a87-c8e8-44ef-8bd4-554d8a9fa81c)

jmillan:
I plan to check this issue in the next days.

jmillan:
It's been very complicate taking the time to check this lately. I can't say when I'm going to be able to check this.

Meson option `ms_rtc_logger_rtp` will print a log line for **each** RTP packet sent or dropped by mediasoup indicating the cause in case of drop, which would be a good way to troubleshot this. In case anyone is willing to dig further.

`cd worker &&  MESON_ARGS="-Dms_rtc_logger_rtp=true" make`

PR #1070, opened 2023-04-21
Fix seqno jumps when bwe is close to the lowest SVC layer

vpalmisano:
It seems solving the issue described in https://github.com/versatica/mediasoup/issues/1069

patched
[patch.pcapng.gz](https://github.com/versatica/mediasoup/files/11298624/patch.pcapng.gz)

### DIFF
--- a/worker/src/RTC/SvcConsumer.cpp
+++ b/worker/src/RTC/SvcConsumer.cpp
@@ -1099,6 +1099,8 @@ namespace RTC
 			MS_DEBUG_TAG(
 			  svc, "target layers changed [spatial:-1, temporal:-1, consumerId:%s]", this->id.c_str());
 
+			this->syncRequired = true;
+
 			EmitLayersChange();
 
 			return;


jmillan:
I need to check it in detail but at a first glance I don't see how the issue is produced or how the fix solves it. Do you have an explanation for it?

jmillan:
It makes sense to sync when no layer can be sent :+1:

jmillan (approved):

vpalmisano:
> Do you have an explanation for it?

Probably the sequence numbers keep increasing even if no layers are sent.

ggarber:
@jmillan Isn't syncRequired marking when we are waiting for a keyframe (or something similar)?  If layers are disabled is not clear for me why it should be syncRequired.

I mean, it looks clear that there is an (important) issue but I don't understand if changing syncRequired is fixing the issue or workarounding it. 

Thx

jmillan:
Yes, let's get to the root cause here.

ibc:
We are drinking in an offsite. Sorry for the delay.

buptlsp:
I also found a strange seqno issue. sqlno will jump per 20 packets in some scenarios. Currently I don't know what scenario. Can you please tell me if this pr of yours can solve this problem?

image
image

buptlsp:
@ibc can you reply me about this problem?  this means in some situation, the packet will lost 5% percent.

ibc:
@buptlsp, this is a PR with a possible fix for `SvcConsumer` (AKA VP9 codec with SVC). Please let's focus on the topic of the PR.

buptlsp:
the result is similar, the seq number will jumps.  I don't know if it's a same issue. if this is only about svc, I will try to find the reason. thanks!

ggarber:
Should the "this->syncRequired = true;" be done when the layers are reenabled (layer: -1 -> 0) instead of when the layers are disabled (layer: 0 -> -1)?

ibc:
We'll jump into this after some very busy days at work.

jmillan:
Fixed https://github.com/versatica/mediasoup/pull/1494

The problem was that we were not correctly dropping the sequence number in the `SequenceManager` when a packet cannot be forwarded, hence we were generating non legitimate sequence number jumps.

ibc:
> Fixed #1494

I'm a bit lost. You say that PR #1494 fixes this issue #1070 but the PR says that it fixes issue #1069

jmillan:
Closed by #1494

jmillan:
> I'm a bit lost. You say that PR https://github.com/versatica/mediasoup/pull/1494 fixes this issue https://github.com/versatica/mediasoup/pull/1070 but the PR says that it fixes issue https://github.com/versatica/mediasoup/issues/1069

The fixed issue is #1069, and this PR (#1070) was supposed to fix it, which is now closed in favour of #1494

PR #1072, opened 2023-04-24
Fix AlrDetector crash issue in debug mode

qqpmsc2006:


jmillan (approved):
It fixes the crash. Thanks! :+1:

ibc (approved):

PR #1076, opened 2023-05-06
fix: nack timer not work as expected

o-u-p:


ibc (commented):
Can you explain rationale of these changes? Was what wrong before them?

o-u-p:
> Can you explain rationale of these changes? Was what wrong before them?

if not check 0 rtt
`if (filter == NackFilter::TIME && (nackInfo.sentAtMs == 0 || nowMs - nackInfo.sentAtMs >= this->rtt))`
will send nack too quickly.
if not stop nack timer, when nack list is empty, it will never restart, if stop the timer, it will get started here
```
if (!this->timer->IsActive())
		MayRunTimer();
```

worker/src/RTC/RtpStreamRecv.cpp
```diff
@@ -611,12 +611,11 @@ namespace RTC
...
-		}
-
-		// Tell it to the NackGenerator.
-		if (this->params.useNack)
-		{
-			this->nackGenerator->UpdateRtt(static_cast<uint32_t>(this->rtt));
+			// Tell it to the NackGenerator.
+			if (this->params.useNack)
+			{
+				this->nackGenerator->UpdateRtt(static_cast<uint32_t>(this->rtt));
```
ibc, line 617: Why are you just updating current RTT value in `NackGenerator` **only** if > 0.0f? This is wrong. It means that if negative then `NackGenerator` will keep using its previous given RTT value.

ibc:
> if not stop nack timer, when nack list is empty, it will never restart, if stop the timer, it will get started here
> 
> ```
> if (!this->timer->IsActive())
> 		MayRunTimer();
> ```

Makes sense.

ibc:
I'm including some valid changes of this PR into another one. Thanks @o-u-p

ibc:
Closing this PR in favour of https://github.com/versatica/mediasoup/pull/1082

ibc:
Released in mediasoup 3.11.26

Issue #1073, opened 2023-04-28
Inappropriate rtp header padding, which can lead to wasted bandwidth


buptlsp:
## Bug Report
I use wireshark to capture the rtp audio packets and found that the rtp header had strange padding, as shown below:
![image](https://user-images.githubusercontent.com/3291512/235050329-afbedc3d-e1bb-4337-ba54-a6e674860b85.png)
It makes 8 bytes of padding per packet, and according to RFC5285, the one-byte extension header should look like this:
![image](https://user-images.githubusercontent.com/3291512/235050532-1f954ff1-f9da-4b5e-8f2d-0dc1b5ae1f57.png)

mediasoup uses two extension headers：mid and aiuodLevel. in the consumer, both of which are normally 1 byte in length, so only 4 bytes are needed. I changed the code of mediasoup and got the following result, and then all of them run well:
![image](https://user-images.githubusercontent.com/3291512/235051175-29dfa91f-92d7-4080-bf57-9fb35d0071f2.png)

According to wireshark's packet capture, the length of an audio packet is around 150 bytes, each packet has 8 bytes of waste, which basically results in having 5.3% of bandwidth wasted.

the code is because this line of  worker/src/RTC/Producer.cpp :
![image](https://user-images.githubusercontent.com/3291512/235051324-5e3be3bb-d29f-4422-a79e-a41057f70cd4.png)

The code uses MidMaxLength, which has a value of 8. But the actual mid is less than 8. When UpdateMid, it can be memcpy directly. this method makes updateMid easy, but the extra padding will cause the bandwidth to increase. Bandwidth is money. This place may need resize extensions, but it may bring more memory and cpu consumption. Maybe I need to judge the length change to handle it, but this would be a bit too difficult for me. I hope you guys can help me with this, thanks a lot.

nazar-pc:
Don't we intentionally change 1 byte RTP header extensions to 4 bytes intentionally, @ibc?

ibc:
> Don't we intentionally change 1 byte RTP header extensions to 4 bytes intentionally, @ibc?

Can you please specify the question a bit more? do you mean if we change 1 byte to 4 bytes extension type when adding an extra extension in case there is no more slots when using 1 byte style?

nazar-pc:
I thought we converted either into 4 byte extension internally to simplify handling, didn't we?

ibc:
[...]

PR #1079, opened 2023-05-12
Rtp header padding

buptlsp:
what I have done:
1. npm run test:worker, the result as follow, all test passed.
image
2. test in online process. the result as follow:
-  the bandwidth has decreased by 5% -10%
-  the memory of machine has decreased by about 10%
-  the CPU consumption of machine seems to have remained unchanged
3. make fuzzer . I haven't used fuzzer before, I couldn't run it successfully. Can you help me try this? Thank you

### DIFF
--- a/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
+++ b/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
@@ -165,8 +165,9 @@ void Fuzzer::RTC::RtpPacket::Fuzz(const uint8_t* data, size_t len)
 	packet->HasExtension(14);
 	packet->GetExtension(14, extenLen);
 	packet->ReadTransportWideCc01(wideSeqNumber);
-	packet->UpdateTransportWideCc01(12345u);
+	//packet->UpdateTransportWideCc01(12345u);
 	//packet->SetExtensionLength(14, 2);
+	packet->setExtionValue(packet->transportWideCc01ExtensionId, 2u, "09");
 
 	packet->SetSsrcAudioLevelExtensionId(11);
 	packet->HasExtension(11);
--- a/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
+++ b/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
@@ -167,7 +167,7 @@ void Fuzzer::RTC::RtpPacket::Fuzz(const uint8_t* data, size_t len)
 	packet->ReadTransportWideCc01(wideSeqNumber);
 	//packet->UpdateTransportWideCc01(12345u);
 	//packet->SetExtensionLength(14, 2);
-	packet->setExtionValue(packet->transportWideCc01ExtensionId, 2u, "09");
+	packet->SetExtensionValue(packet->transportWideCc01ExtensionId, 2u, "09");
 
 	packet->SetSsrcAudioLevelExtensionId(11);
 	packet->HasExtension(11);
--- a/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
+++ b/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
@@ -165,8 +165,6 @@ void Fuzzer::RTC::RtpPacket::Fuzz(const uint8_t* data, size_t len)
 	packet->HasExtension(14);
 	packet->GetExtension(14, extenLen);
 	packet->ReadTransportWideCc01(wideSeqNumber);
-	//packet->UpdateTransportWideCc01(12345u);
-	//packet->SetExtensionLength(14, 2);
 	packet->SetExtensionValue(packet->transportWideCc01ExtensionId, 2u, "09");
 
 	packet->SetSsrcAudioLevelExtensionId(11);
--- a/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
+++ b/worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
@@ -166,7 +166,7 @@ void Fuzzer::RTC::RtpPacket::Fuzz(const uint8_t* data, size_t len)
 	packet->GetExtension(14, extenLen);
 	packet->ReadTransportWideCc01(wideSeqNumber);
 	packet->UpdateTransportWideCc01(12345u);
-	packet->SetExtensionLength(14, 2);
+	//packet->SetExtensionLength(14, 2);
 
 	packet->SetSsrcAudioLevelExtensionId(11);
 	packet->HasExtension(11);
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -605,7 +605,7 @@ namespace RTC
 			  currentLen,
 			  value.c_str());
 
-			// If len == currentLen, just memcpy
+			// If len == currentLen, just memcpy.
 			if (len == currentLen)
 			{
 				std::memcpy(extension->value, value.c_str(), len);
@@ -709,7 +709,7 @@ namespace RTC
 
 			auto* extension = it->second;
 			auto currentLen = extension->len;
-			// If len == currentLen, just memcpy
+			// If len == currentLen, just memcpy.
 			if (len == currentLen)
 			{
 				std::memcpy(extension->value, value.c_str(), len);
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -570,8 +570,8 @@ namespace RTC
 	}
 
 	/**
-	 * The caller is responsible of not setting a length higher than the
-	 * available one (taking into account existing padding bytes).
+	 *  This method only can set exists extension. if len of value changed, this method will
+	 * recaculate and fill padding.
 	 */
 	bool RtpPacket::SetExtensionValue(uint8_t id, uint8_t len, const std::string& value)
 	{
@@ -604,10 +604,15 @@ namespace RTC
 			  len,
 			  currentLen,
 			  value.c_str());
-			
-			if (len != currentLen)
+
+			// If len == currentLen, just memcpy
+			if (len == currentLen)
 			{
-				// If len != currentLen, means we may need to shift extension or payload.
+				std::memcpy(extension->value, value.c_str(), len);
+			}
+			// If len != currentLen, means we may need to shift extension or payload.
+			else
+			{
 				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
 				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
 				uint8_t* ptr               = extensionStart;
@@ -617,74 +622,80 @@ namespace RTC
 				std::memset((uint8_t*)extension, 0, currentLen + 1);
 				while (ptr < extensionEnd || ptr1 < extensionEnd)
 				{
+					// Means move has end. Fill the rest buffer to 0u.
 					if (ptr >= extensionEnd)
 					{
 						*ptr1 = 0u;
 						ptr1++;
 						continue;
 					}
-					const uint8_t tempId = (*ptr & 0xF0) >> 4;
-					const size_t tempLen = static_cast<size_t>(*ptr & 0x0F) + 1;
+					const uint8_t tempExtensionId = (*ptr & 0xF0) >> 4;
+					const size_t tempExtensionLen = static_cast<size_t>(*ptr & 0x0F) + 1;
 
-					// TempId=15 in One-Byte extensions means "stop parsing here".
-					if (tempId == 15u)
+					// tempExtensionId=15 in One-Byte extensions means "stop parsing here".
+					if (tempExtensionId == 15u)
 						break;
-					if (tempId != 0u && tempId != id)
+					// Valid extension id.
+					if (tempExtensionId != 0u)
 					{
 						if (ptr1 < ptr)
 						{
-							std::memmove(ptr1, ptr, tempLen + 1);
-							this->oneByteExtensions[tempId - 1] = reinterpret_cast<OneByteExtension*>(ptr1);
+							std::memmove(ptr1, ptr, tempExtensionLen + 1);
+							// the extension has move from ptr to ptr1, store the One-Byte extension element in an
+							// array.
+							// `-1` because we have 14 elements total 0..13 and `id` is in the range 1..14.
+							this->oneByteExtensions[tempExtensionId - 1] =
+							  reinterpret_cast<OneByteExtension*>(ptr1);
 						}
-						extensionsTotalSize += tempLen + 1;
-						ptr += tempLen + 1;
-						ptr1 += tempLen + 1;
-				
+						extensionsTotalSize += tempExtensionLen + 1;
+						ptr += tempExtensionLen + 1;
+						ptr1 += tempExtensionLen + 1;
+
 						MS_DEBUG_DEV(
-						  "tempId: %" PRIu8 " tempLen:%zd, offset:%ld", tempId, tempLen, ptr - extensionStart);
-					
+						  "tempExtensionId: %" PRIu8 " tempExtensionLen:%zd, offset:%ld",
+						  tempExtensionId,
+						  tempExtensionLen,
+						  ptr - extensionStart);
 					}
+					// tempExtensionId=0 means alignment.
 					else
 					{
 						ptr++;
 					}
 				}
-				
+
 				MS_DEBUG_DEV(
 				  "extensionsTotalSize: %zd headerLength:%zd",
 				  extensionsTotalSize,
 				  GetHeaderExtensionLength());
-				
+
 				auto paddedExtensionsTotalSize =
 				  static_cast<size_t>(Utils::Byte::PadTo4Bytes(static_cast<uint16_t>(extensionsTotalSize)));
 				extensionsTotalSize = paddedExtensionsTotalSize;
 				int16_t shift = static_cast<int16_t>(extensionsTotalSize - GetHeaderExtensionLength());
-				
+
 				MS_DEBUG_DEV("shift:%d paddedExtensionsTotalSize: %zd", shift, paddedExtensionsTotalSize);
-				
+
 				std::memmove(this->payload + shift, this->payload, this->payloadLength + this->payloadPadding);
 				// Clear the shift place, if shift > currentLen, may create some strange extension.
 				if (shift > 0)
 				{
 					std::memset(this->payload, 0, shift);
 				}
-				// Begin to set current extension.
+				// Begin to set current extension to the end of the extension buffer.
 				ptr                             = extensionStart + extensionsTotalSize - (len + 1);
 				this->oneByteExtensions[id - 1] = reinterpret_cast<OneByteExtension*>(ptr);
 				extension                       = this->oneByteExtensions[id - 1];
 				*ptr                            = (id << 4) | ((len - 1) & 0x0F);
 				ptr++;
 				std::memcpy(ptr, value.c_str(), len);
-
 				this->payload += shift;
+
+				// Update packet total size.
 				this->size += shift;
+
+				// Update the header extension length.
 				this->headerExtension->length = htons(extensionsTotalSize / 4);
-			}
-			else
-			{
-				// In One-Byte extensions value length 0 means 1.
-				extension->len = len - 1;
-				std::memcpy(extension->value, value.c_str(), len);
 			}
 
 			return true;
@@ -698,38 +709,50 @@ namespace RTC
 
 			auto* extension = it->second;
 			auto currentLen = extension->len;
-			if (len != currentLen)
+			// If len == currentLen, just memcpy
+			if (len == currentLen)
 			{
-				// If len != currentLen, means we may need shift payload or extension.
+				std::memcpy(extension->value, value.c_str(), len);
+			}
+			// If len != currentLen, means we may need shift payload or extension.
+			else
+			{
 				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
 				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
 				uint8_t* ptr               = extensionStart;
 				size_t extensionsTotalSize = static_cast<size_t>(len + 2);
 				uint8_t* ptr1              = ptr;
-				// Clear current extension valueLen + 2 byteheader.
+				// Clear current extension, valueLen + 2 byteheader. we will rewrite after move all
+				// extension in right place.
 				std::memset((void*)extension, 0, currentLen + 2);
+				// Two-Byte extensions can have length 0.
 				while (ptr + 1 < extensionEnd || ptr1 < extensionEnd)
 				{
+					// Means move has end. Fill the rest buffer to 0u.
 					if (ptr + 1 >= extensionEnd)
 					{
 						*ptr1 = 0u;
 						ptr1++;
 						continue;
 					}
-					const uint8_t tempId  = *ptr;
-					const uint8_t tempLen = *(ptr + 1);
+					const uint8_t tempExtensionId  = *ptr;
+					const uint8_t tempExtensionLen = *(ptr + 1);
 
-					if (tempId != 0u && tempId != id)
+					// Valid extension id.
+					if (tempExtensionId != 0u)
 					{
 						if (ptr1 < ptr)
 						{
-							std::memmove(ptr1, ptr, tempLen + 2);
-							this->mapTwoBytesExtensions[tempId] = reinterpret_cast<TwoBytesExtension*>(ptr1);
+							std::memmove(ptr1, ptr, tempExtensionLen + 2);
+							// The extension has moved from ptr to ptr1. Store the Two-Bytes extension element in the map.
+							this->mapTwoBytesExtensions[tempExtensionId] =
+							  reinterpret_cast<TwoBytesExtension*>(ptr1);
 						}
-						extensionsTotalSize += tempLen + 2;
-						ptr += tempLen + 2;
-						ptr1 += tempLen + 2;
+						extensionsTotalSize += tempExtensionLen + 2;
+						ptr += tempExtensionLen + 2;
+						ptr1 += tempExtensionLen + 2;
 					}
+					// id=0 means alignment.
 					else
 					{
 						ptr++;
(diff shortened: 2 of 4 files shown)


worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
```diff
@@ -165,8 +165,9 @@ void Fuzzer::RTC::RtpPacket::Fuzz(const uint8_t* data, size_t len)
 	packet->HasExtension(14);
 	packet->GetExtension(14, extenLen);
 	packet->ReadTransportWideCc01(wideSeqNumber);
-	packet->UpdateTransportWideCc01(12345u);
-	packet->SetExtensionLength(14, 2);
+	//packet->UpdateTransportWideCc01(12345u);
```
ibc: Why are those tests commented? We cannot leave commented tests. Do they no longer work or what?

ibc (changes requested):
I'll review next week in depth but overall code doesn't respect existing code guidelines. Please run "make lint" and "make format" in worker/ folder. Also leave blank space after logs and use auto or auto* or const auto or const auto* where possible as in the whole code. In addition, comments must start with capital letter and end with a dot.

worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
buptlsp: this two line can be replace by setExtensionValue(). should I just delete the 2 line?

buptlsp:
> I'll review next week in depth but overall code doesn't respect existing code guidelines. Please run "make lint" and "make format" in worker/ folder. Also leave blank space after logs and use auto or auto* or const auto or const auto* where possible as in the whole code. In addition, comments must start with capital letter and end with a dot.

It's my first time to write pr of mediasoup. I will run make lint and make format. and I will check the format problem

worker/fuzzer/src/RTC/FuzzerRtpPacket.cpp
ibc: We should test as much as possible

buptlsp: > We should test as much as possible

Extremely agree. This is my first time writing C++, and I have been writing PHP before. So the more tests, the better

buptlsp:
I have fix the format. if there is any problem, notify me any time.  the header padding is as follow:
image

btw, I found other two problems about the rtp packet when i use wireshark to capture the packet.  I think is not related to this pr because it also occured before my code change. I will fix them next time.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+				extension                       = this->oneByteExtensions[id - 1];
+				*ptr                            = (id << 4) | ((len - 1) & 0x0F);
+				ptr++;
+				std::memcpy(ptr, value.c_str(), len);
+
+				this->payload += shift;
+				this->size += shift;
+				this->headerExtension->length = htons(extensionsTotalSize / 4);
+			}
+			else
```
jmillan: I would put this condition (`if (len == currentLen)`) the first one as it's sorter and helps readability. Also `extension->len` does not need to be reset because it has not changed.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+			  "set extension id: %" PRIu8 ", length:%" PRIu8 ", currentLen:%" PRIu8 " value:%s",
+			  id,
+			  len,
+			  currentLen,
+			  value.c_str());
+			
+			if (len != currentLen)
+			{
+				// If len != currentLen, means we may need to shift extension or payload.
+				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
```
jmillan, line 619:
```suggestion
				uint8_t* extensionsStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
```
jmillan, line 620:
```suggestion
				uint8_t* extensionsEnd      = extensionStart + GetHeaderExtensionLength();
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+				uint8_t* ptr               = extensionStart;
+				size_t extensionsTotalSize = static_cast<size_t>(len + 1);
+				uint8_t* ptr1              = ptr;
```
jmillan:
```suggestion
                         	size_t extensionsTotalSize = static_cast<size_t>(len + 1);
				uint8_t* ptr               = extensionStart;
				uint8_t* ptr1              = extensionStart;
```

jmillan: Can you provide a more descriptive names to `ptr` and `ptr1`? How do they differ?

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+				while (ptr < extensionEnd || ptr1 < extensionEnd)
+				{
+					if (ptr >= extensionEnd)
+					{
+						*ptr1 = 0u;
+						ptr1++;
+						continue;
+					}
+					const uint8_t tempId = (*ptr & 0xF0) >> 4;
+					const size_t tempLen = static_cast<size_t>(*ptr & 0x0F) + 1;
```
jmillan:
```suggestion
					const uint8_t tempExtensionId = (*ptr & 0xF0) >> 4;
					const size_t tempExtensionLen = static_cast<size_t>(*ptr & 0x0F) + 1;
```

worker/src/RTC/RtpPacket.cpp
buptlsp: this line was copied from RtpPacket::ParseExtensions(), do you mean I have to change this line or all of other places?

worker/src/RTC/RtpPacket.cpp
buptlsp: yes, It's a good idea!

worker/src/RTC/RtpPacket.cpp
buptlsp: ptr will go forward always, ptr1 only go forward with data.  ptr，and movePtr?  may you give me some advice?

worker/src/RTC/RtpPacket.cpp
buptlsp: ok

worker/src/RTC/RtpPacket.cpp
jmillan: Ok, we can leave for for consistency, and rename it later if needed.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,94 @@ namespace RTC
...
+						}
+						extensionsTotalSize += tempLen + 1;
+						ptr += tempLen + 1;
+						ptr1 += tempLen + 1;
+				
+						MS_DEBUG_DEV(
+						  "tempId: %" PRIu8 " tempLen:%zd, offset:%ld", tempId, tempLen, ptr - extensionStart);
+					
+					}
+					else
```
jmillan, line 657: This `else` in `ParseExtensions()` is only reachable for `tempId == 0.` Now it's reachable for `tempId == 0` and `tempId == id`

jmillan:
@buptlsp, since you've based the new `SetExtensionValue()` on the existing `ParseExtensions()` method do please reuse the code as much as possible for a better readability.  Reuse the structure, the variable names, the comments.., So we can clearly see the difference between both methods at a first glance.

And please write an explanatory comment at least on top of every condition (most of them are already present from original code). The code needs to be understood now and in the coming years, and having to go through all code lines to understand it makes it time consuming in the short and long term.

buptlsp:
> @buptlsp, since you've based the new `SetExtensionValue()` on the existing `ParseExtensions()` method do please reuse the code as much as possible for a better readability. Reuse the structure, the variable names, the comments.., So we can clearly see the difference between both methods at a first glance.
> 
> And please write an explanatory comment at least on top of every condition (most of them are already present from original code). The code needs to be understood now and in the coming years, and having to go through all code lines to understand it makes it time consuming in the short and long term.

in fact,  I have reuse the code as much as I can. the method is differ to parseExtensions, I copy from two method mainly :  ParseExtensions, setExtensions。 most structure and variable names cames from this two methods. but the method setExtensionValue is more complicated because the padding can be everywhere and the data may override.

I can add some more comments to explain the code.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -604,10 +604,15 @@ namespace RTC
 			  len,
 			  currentLen,
 			  value.c_str());
-			
-			if (len != currentLen)
+
+			// If len == currentLen, just memcpy
```
jmillan: Comments start with Uppercase and end with dot (.).

jmillan (commented):
I'll review this properly next week.

jmillan:
@buptlsp, few comments:

`SetExtensionValue()` does not need to consider padding bits _between_ header values because we generate such buffer and only add padding at the end of the header extensions buffer (see `RtpPacket::SetExtensions`).

The previous point allows for `ptr1` not to be needed in `SetExtensionValue()` which will make the code cleaner. Also would require a single `memmove` operation to be done, just for the header we are setting.

In pseudo code, the new extensions would be rewritten as:

```c++
			// If len != currentLen, means we may need to shift extension or payload.
			else
			{
				uint8_t* extensionStart      = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
				uint8_t* extensionEnd        = extensionStart + GetHeaderExtensionLength();
				uint8_t* ptr                          = extensionStart;
				size_t extensionsTotalSize = 0;

				while (ptr < extensionEnd)
				{
					const uint8_t tempExtensionId = (*ptr & 0xF0) >> 4;
					const size_t tempExtensionLen = static_cast<size_t>(*ptr & 0x0F) + 1;

					// tempExtensionId=15 in One-Byte extensions means "stop parsing here".
					if (tempExtensionId == 15u)
                                          {
						break;
                                          }
                                          // Extension id 0. We already reached the end of the extension header.
                                          // NOTE: Padding only exists at the end of the extension header for outgoing RTP packet.
					if (tempExtensionId == 0u)
                                          {
                                               break:
                                          }
					// Valid extension id.
					else
					{
                                                // This is the extension which length and value is being set. 
						if (tempExtensionId == id)
						{
							std::memmove(ptr, value.c_str(), len + 1);
                                                        ptr += len + 1;
            						extensionsTotalSize += len + 1;
						}
                                                  else
						{
							ptr += tempExtensionLen + 1;
            						extensionsTotalSize += tempExtensionLen + 1;
						}
             
                  				// Store the One-Byte extension element.
						// `-1` because we have 14 elements total 0..13 and `id` is in the range 1..14.
						this->oneByteExtensions[tempExtensionId - 1] =
					             reinterpret_cast<OneByteExtension*>(ptr);
					}
				}
```

This is much cleaner approach IMHO and reduces the number of `memmove()` calls to just the header being set.

buptlsp:
@jmillan Have you read the issue https://github.com/versatica/mediasoup/issues/1073 I mentioned? I don't think your code will solve the problem. the padding indeed existed between headers.

jmillan:
> the padding indeed existed between headers.

It's an accidental padding in MID extension:

* We are defining a RTC::MidMaxLength of 8 bytes.
* We are providing `RtpPacket::SetExtensions()` with a buffer of size 8 and exten len 8 for the MID extension.
* Then we set the mid extension value (if the Consumer RtpParameters do define a `mid` value).
* When setting the mid value, it's usually less than 8 bytes, so the exten len is modified and the remaining bits are left to 0, thus creating padding.

IMHO we should:

* In `Producer::MangleRtpPacket()` create the MID extension with length of 1 byte and assign it a value of 0.
  *  This way if the Consumer does not define `mid` we waste the less possible bandwidth.
  *  This way we don't create artificial padding. 
  * NOTE: We already set default values for other extensions (see the indicated method).
* Later on, `RtpPacket::UpdateMid()` needs to check if the provided value is less than `RTC::MidMaxLength` and otherwise log error.
* Then my proposed approach will work (no padding between extension headers) and the header resize will be more clean and efficient.

Do you agree @ibc?

buptlsp:
I don't think it‘s a good idea. Your method may be cleaner, but there are a lot of prerequisites. My method does not need to have prerequisites and assumptions like yours. One day, we may add other extensions, or we may introduce other padding, and at that time my method will still be valid.

jmillan:
> Your method may be cleaner, but there are a lot of prerequisites

What prerequisites?

* The fact that every outgoing extension header should have a default value? That's indeed something desirable to have the minimum possible overhead for unset headers.
* The fact that we don't create artificial padding between header extensions? That's indeed desirable too, and it's not a prerequisite but an implementation detail, which is positive and BTW is already done if the previous point is satisfied.

ibc:
While @jmillan's proposal makes sense I would also like to know about those "prerequisites" and those potential problems if we add more extensions in the future.

jmillan:
> While @jmillan's proposal makes sense I would also like to know about those "prerequisites" and those potential problems if we add more extensions in the future.

There are no prerequisites, but a single contract:
* `Producer::MangleRtpPacket()` always needs to set a default value for every extension.

We already to that for **every** extension but for `MID`.

My proposal is that we set a default value of 0, with a length of 1 for MID. It will optionally be rewritten with a new value/length. In the future, we need to follow the same policy: always set a default value for new extensions. If that extension is optional, like MID, then use the smallest possible value. (value 0, length 1).

buptlsp:
Well, let's go along with your idea that the mid length is now 1 byte. The mid value can appear anywhere in the header. (It's the first header now, but you can't assume it's forever). Suppose the new mid value's length is 3. What do we need to do when updateMid()?

1. there is not enough memory to put the value. we need to move the header to make sure it has enough memory. all headers after mid should change their place.
Note: If you move the header backwards, you have to be careful to overwrite the payload.
3. we need to recalculate the header length. If there is other padding before, we may not need to increase the length again. 
4. we may need to move the payload backwards due to the increased length.

You've done a bunch of operations like setting the default value and setting it to the minimum, which just reduces the situation where you think the length will only increase and the payload will not move forward. and after all operation you have done, your code is not as clean as your think.

now you have change the mid length to 3. but the Router has lots of consumers, this packet will update other mid after you have change it to length 3. not the new length is 1, what should you do now?

My method is called setExtensionValue, and we may make an extension longer or shorter, and the payload may move forward or backward. Your method is entirely for setExtensions, not a method that can be used anywhere.

jmillan:
@buptlsp 

> Your method is entirely for setExtensions, not a method that can be used anywhere.

No, my change was indeed for `SetExtensionValue()`. It's behaviour is identical to yours with the exception that it just moves memory on the position where the extension being set resides. All extensions lie on the same memory buffer so such move will affect all extensions that come after it.

Then total extension length and padding needs to be recalculated, just as in your implementation.

I can directly implement it when I get the time so we can simply compare it.

buptlsp:
I have already put my code online in my environment. the cpu, memory and bandwidth result I have give in the pr.  So, I am not in a hurry to solve this problem． after all, I have already saved money for my company.

I don't know your pseudo well. maybe when you complete the code ,we can compare and discuss later.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,105 @@ namespace RTC
...
+				std::memcpy(extension->value, value.c_str(), len);
+			}
+			// If len != currentLen, means we may need to shift extension or payload.
+			else
+			{
+				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
+				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
+				uint8_t* ptr               = extensionStart;
+				size_t extensionsTotalSize = static_cast<size_t>(len + 1);
+				uint8_t* ptr1              = ptr;
```
jmillan: Please, rename to `lastExtension`, since it points exactly to the last extension that has been evaluated, while `ptr` keeps moving forward among the 0s.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,105 @@ namespace RTC
...
+				// Clear current extension, we will rewrite after move all extension in right place.
+				std::memset((uint8_t*)extension, 0, currentLen + 1);
+				while (ptr < extensionEnd || ptr1 < extensionEnd)
+				{
+					// Means move has end. Fill the rest buffer to 0u.
+					if (ptr >= extensionEnd)
+					{
+						*ptr1 = 0u;
+						ptr1++;
```
jmillan: can we add here a single `memset` in order to fill with zeros at once?

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +709,84 @@ namespace RTC
...
+				std::memcpy(extension->value, value.c_str(), len);
+			}
+			// If len != currentLen, means we may need shift payload or extension.
+			else
+			{
+				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
+				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
+				uint8_t* ptr               = extensionStart;
+				size_t extensionsTotalSize = static_cast<size_t>(len + 2);
+				uint8_t* ptr1              = ptr;
```
jmillan: same naming changes please.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +709,84 @@ namespace RTC
...
+				// extension in right place.
+				std::memset((void*)extension, 0, currentLen + 2);
+				// Two-Byte extensions can have length 0.
+				while (ptr + 1 < extensionEnd || ptr1 < extensionEnd)
+				{
+					// Means move has end. Fill the rest buffer to 0u.
+					if (ptr + 1 >= extensionEnd)
+					{
+						*ptr1 = 0u;
+						ptr1++;
```
jmillan: single memset here.

jmillan (commented):
After re-reviewing I'm fine with the implementation.

I've added few change requests.

jmillan (commented):
After re-reviewing I'm fine with the implementation.

I've added few change requests.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,105 @@ namespace RTC
...
+				  static_cast<size_t>(Utils::Byte::PadTo4Bytes(static_cast<uint16_t>(extensionsTotalSize)));
+				extensionsTotalSize = paddedExtensionsTotalSize;
+				int16_t shift = static_cast<int16_t>(extensionsTotalSize - GetHeaderExtensionLength());
+
+				MS_DEBUG_DEV("shift:%d paddedExtensionsTotalSize: %zd", shift, paddedExtensionsTotalSize);
 
-			// In One-Byte extensions value length 0 means 1.
-			extension->len = len - 1;
+				std::memmove(this->payload + shift, this->payload, this->payloadLength + this->payloadPadding);
+				// Clear the shift place, if shift > currentLen, may create some strange extension.
```
jmillan:
```suggestion
				// Fill padding with zeroes.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,105 @@ namespace RTC
...
+				auto paddedExtensionsTotalSize =
+				  static_cast<size_t>(Utils::Byte::PadTo4Bytes(static_cast<uint16_t>(extensionsTotalSize)));
+				extensionsTotalSize = paddedExtensionsTotalSize;
+				int16_t shift = static_cast<int16_t>(extensionsTotalSize - GetHeaderExtensionLength());
+
+				MS_DEBUG_DEV("shift:%d paddedExtensionsTotalSize: %zd", shift, paddedExtensionsTotalSize);
 
-			// In One-Byte extensions value length 0 means 1.
-			extension->len = len - 1;
+				std::memmove(this->payload + shift, this->payload, this->payloadLength + this->payloadPadding);
```
jmillan, line 678: Is `payloadPadding` needed here? it's only valid to point the padding for parsed packets, but has no relevance otherwise.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,105 @@ namespace RTC
...
+				uint8_t* ptr               = extensionStart;
+				size_t extensionsTotalSize = static_cast<size_t>(len + 1);
+				uint8_t* ptr1              = ptr;
+				// Clear current extension, we will rewrite after move all extension in right place.
+				std::memset((uint8_t*)extension, 0, currentLen + 1);
+				while (ptr < extensionEnd || ptr1 < extensionEnd)
+				{
+					// Means move has end. Fill the rest buffer to 0u.
+					if (ptr >= extensionEnd)
+					{
```
jmillan: In which case are we entering this condition?

worker/src/RTC/RtpPacket.cpp
```diff
@@ -566,16 +566,14 @@ namespace RTC
...
-
-		SetExtensionLength(this->midExtensionId, midLen);
+		SetExtensionValue(this->midExtensionId, midLen, mid);
 	}
 
 	/**
-	 * The caller is responsible of not setting a length higher than the
-	 * available one (taking into account existing padding bytes).
+	 *  This method only can set exists extension. if len of value changed, this method will
+	 * recaculate and fill padding.
```
jmillan:
```suggestion
	 *  This method set the value for an existing extension. If the length of the current value changed,
	 * it will recalculate it and regenerate the header extension accordingly.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,101 @@ namespace RTC
...
-			if (len < currentLen)
-				std::memset(extension->value + len, 0, currentLen - len);
+			MS_DEBUG_DEV(
+			  "set extension id: %" PRIu8 ", length:%" PRIu8 ", currentLen:%" PRIu8 " value:%s",
+			  id,
+			  len,
+			  currentLen,
+			  value.c_str());
+
+			// If len == currentLen, just memcpy.
```
jmillan:
```suggestion
			// If len == currentLen, just copy the new value.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,101 @@ namespace RTC
...
+					// tempExtensionId=15 in One-Byte extensions means "stop parsing here".
+					if (tempExtensionId == 15u)
+						break;
+					// Valid extension id.
+					if (tempExtensionId != 0u)
+					{
+						if (lastExtensionPtr < ptr)
+						{
+							std::memmove(lastExtensionPtr, ptr, tempExtensionLen + 1);
+							// the extension has move from ptr to lastExtensionPtr, store the One-Byte extension
```
jmillan:
```suggestion
							// The extension has moved from ptr to lastExtensionPtr, store the One-Byte extension
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -600,12 +598,101 @@ namespace RTC
...
+						  ptr - extensionStart);
+					}
+					// tempExtensionId=0 means alignment.
+					else
+					{
+						ptr++;
+					}
+				}
+
+				// Means move has end. Fill the rest buffer to 0u.
```
jmillan:
```suggestion
				// Fill the rest buffer to 0u.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +705,82 @@ namespace RTC
 
 			auto* extension = it->second;
 			auto currentLen = extension->len;
+			// If len == currentLen, just memcpy.
```
jmillan:
```suggestion
			// If len == currentLen, just copy the new value.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +705,82 @@ namespace RTC
...
+			}
+			// If len != currentLen, means we may need shift payload or extension.
+			else
+			{
+				uint8_t* extensionStart    = reinterpret_cast<uint8_t*>(this->headerExtension) + 4;
+				uint8_t* extensionEnd      = extensionStart + GetHeaderExtensionLength();
+				uint8_t* ptr               = extensionStart;
+				size_t extensionsTotalSize = static_cast<size_t>(len + 2);
+				uint8_t* lastExtensionPtr  = ptr;
+				// Clear current extension, valueLen + 2 byteheader. we will rewrite after move all
```
jmillan:
```suggestion
				// Clear current extension, valueLen + 2 byteheader. We will rewrite after moving all
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +705,82 @@ namespace RTC
...
+				// extension in right place.
```
jmillan:
```suggestion
				// extensions to the right place.
```

worker/src/RTC/RtpPacket.cpp
```diff
@@ -618,12 +705,82 @@ namespace RTC
...
+					else
+					{
+						ptr++;
+					}
+				}
 
-			// Fill with 0's if new length is minor.
-			if (len < currentLen)
-				std::memset(extension->value + len, 0, currentLen - len);
+				// Means move has end. Fill the rest buffer to 0u.
```
jmillan:
```suggestion
				// Fill the rest buffer to 0u.
```

jmillan (commented):
Just some cosmetic changes.

The PR is good to go for me. I'd like @ibc to review and approve it before merging.

ibc:
I'm on vacation so give some days before reviewing this. For now just some concerns:

1. Does this PR include proper C++ tests for the new behavior/feature? I'd like to see tests that create a RTP packets, add RTP extensions and asserts that the total RTP packet size matches the expected size.
2. Does this PR include the new API into fuzzer code so it's guaranteed that nothing can crash mediasoup?

buptlsp:
@jmillan  I have already changed the comment.
@ibc  I have add some complex test cases. the result is as follow:
image
  
I don't know how to use fuzzer, can you help to solve this problem please?

ibc:
@buptlsp:

@jmillan and me have had a long call today about this PR. We entirely understand the motivation and the goal, plus the code obviously works as expected. However we have decided to go another way and do it even better, so we are gonna close this PR.

Please check this story in which I've documented what we are gonna do next:

https://github.com/versatica/mediasoup/issues/1107

PR #1082, opened 2023-05-18
worker: Fix NACK timer and avoid negative RTT

ibc:
Based on PR #1076 of @o-u-p

### Details

- Stop NACK timer when NACK list is empty as explained in https://github.com/versatica/mediasoup/pull/1076#issuecomment-1537148706.
- Avoid RTT becoming negative in `RtpStreamSend` and `RtpStreamRecv`. If so, assign it with 0.0f.
- And if 0.0f, use `DefaultRtt` (100 ms) in `NackGenerator`.
- Remove useless `RtpStream::hasRtt` and check `RtpStream::rtt > 0.0f` instead.
- Add some brackets in condition blocks.

### DIFF
--- a/worker/src/RTC/NackGenerator.cpp
+++ b/worker/src/RTC/NackGenerator.cpp
@@ -51,14 +51,18 @@ namespace RTC
 			this->lastSeq = seq;
 
 			if (isKeyFrame)
+			{
 				this->keyFrameList.insert(seq);
+			}
 
 			return false;
 		}
 
 		// Obviously never nacked, so ignore.
 		if (seq == this->lastSeq)
+		{
 			return false;
+		}
 
 		// May be an out of order packet, or already handled retransmitted packet,
 		// or a retransmitted packet.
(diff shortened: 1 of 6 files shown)


worker/src/RTC/NackGenerator.cpp
```diff
@@ -277,7 +297,10 @@ namespace RTC
 				continue;
 			}
 
-			if (filter == NackFilter::TIME && (nackInfo.sentAtMs == 0 || nowMs - nackInfo.sentAtMs >= this->rtt))
+			if (
+			  filter == NackFilter::TIME &&
+			  (nackInfo.sentAtMs == 0 ||
+			   nowMs - nackInfo.sentAtMs >= (this->rtt > 0u ? this->rtt : DefaultRtt)))
```
jmillan, line 303:
```suggestion
			   nowMs - nackInfo.sentAtMs >= (this->rtt > 0.0f ? this->rtt : DefaultRtt)))
```

jmillan (approved):
:+1:

worker/src/RTC/NackGenerator.cpp
ibc: Nope. `this->rtt` in `NackGenerator` is `uint32_t`.

PR #1084, opened 2023-05-18
[flatbuffers] Add ListenInfo in all transports and send/recv buffer size options

### LINKED ISSUE #554: allow setting receive buffer size

ibc:
- Fixes #554

*NOTE:* This branch targets `flatbuffers` branch.

### Details

- Add `ListenInfo` in all transport factories and deprecate `ListenIp`.
- Add optional `sendBufferSize` and `recvBufferSize` (in bytes) in `ListenInfo` type.

### Bonus Tracks

- Add `{ }` in more C++ condition blocks.
- Rename `handlers/XxxxHandler.xxx` to `handlers/XxxxHandle.xxx` becase those are "handles" rather than "handlers" (in fact they wrap a "libuv handle").


jmillan (approved):

ibc:
Merging to avoid another hell.

PR #1087, opened 2023-05-24
Automate and publish prebuilt mediasoup-worker binaries

ibc:
Thanks to @barlock for all his work in PR #1083 (current PR is a fork of that).

### How does it work?

- A new `mediasoup-worker-prebuild.yaml` CI action generates prebuilt binaries of `mediasoup-worker` for different archs.
- Those prebuilt binaries are published into [GitHub Releases](https://github.com/versatica/mediasoup/releases) of mediasoup project.
- When users install mediasoup via NPM, the `postinstall` script will try to fetch the proper prebuilt binary from GH. If not found then it builds it locally as before.
- If `MEDIASOUP_WORKER_BIN` env is given then mediasoup-worker prebuilt is neither fetched nor built (no change here).
- If `MEDIASOUP_LOCAL_DEV` or `MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD` env is given (and `MEDIASOUP_FORCE_WORKER_PREBUILT_DOWNLOAD` is not given) then mediasoup-worker prebuilt is not fetched (so it's locally built).
- If mediasoup package is being installed via `git+ssh` (instead of via npm registry) then mediasoup-worker prebuilt is not fetched (so it's locally built).

### NOTES

As @nazar-pc explained below in comments, this effort doesn't make sense for Rust (since Rust doesn't use `mediasoup-worker` but it compiles and links against `libmediasoup-worker`), so this is just for Node.

Also thanks to @piranna for his previous similar effort in PR #777.

### TODO

- [x] Automate GitHub release creation via `npm run release`.
- [x] Make `npm run release` also read entries in `Changelog.md` for the last version and add them into the GH release.
  - Done here: 045bf8359bec8bc262c5b092d6e79438a94f65bf
- [x] Update `Building.md` (document new scripts, env variables, etc).
- [x] The new `mediasoup-worker-prebuild.yaml` CI action must run **after** the others success.
  - Done here: 00267d56d0610c3d41da3d59e822ac795ff86f68
- [x] Document in the website.
  - Done locally in my computer but not yet published, of course.
- [X] `postinstall` task should **not** try to fetch a prebuilt `mediasoup-worker` when installing a mediasoup via GIT branch. 
  - Imagine we create a 'foo' branch in mediasoup (which lot of changes in C++ side) and want to use it in our media server app.
  - So in the `package.json` of our media server app we add a dependency as follows: `"mediasoup": "github:versatica/mediasoup#foo"`.
  - In that case we do **not** want to fetch a prebuilt `mediasoup-worker`. We **do** want to build it locally.
  - We can skip the prebuilt download by setting the `MEDIASOUP_SKIP_PREBUILT_DOWNLOAD` env. However we should skip it by default if mediasoup is being installed via GIT. Can we know that somehow?
  - Done here: e4c160ecaefbac1e4698c32f06bb32f6ec46b11d
- [X] We should probably create a prebuilt for Apple ARM (M1/M2), this is, `mediasoup-worker-x.x.x-darwin-arm64.tgz`.
  - Unfortunately AFAIU we need a `self-hosted` GitHub CI runner, this is, we need to own a macOS ARM machine which is available and connects to GitHub, see https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners.
  - Done here 5b2ded365b4da5d2637fa387a0e4fe14af9e1a1d.


ibc:
There are various concerns or missing things here:

1. When running "npm install mediasoup" and the script fetches the prebuilt binary (and doesn't build local one) we are NOT using the binary later `¯\_(ツ)_/¯`. This is because `Worker.ts` still expects to spawn the binary at `worker/out/Release/mediasoup-worker`. So I'll modify `Worker.ts` in a way that it tries first to load the binary from `worker/prebuild/Release`. And IMHO we should rename the downloaded binary (once extracted) to `mediasoup-worker` to make things easier. And this must work in Rust and Windows.
2. For local development, make command must delete worker/prebuild/* so later it will use the locally build one. Must also apply to Rust.
3. When running CI tasks, npm ci (which will trigger "postinstall" script as usual) should NOT fetch the prebuilt binary but should always build it locally. I'll add an env for this.

ibc:
I think every is done (all but the automation of the GH release creation via `npm run release` script).

> 1. When running "npm install mediasoup" and the script fetches the prebuilt binary (and doesn't build local one) we are NOT using the binary later `¯\_(ツ)_/¯`. This is because `Worker.ts` still expects to spawn the binary at `worker/out/Release/mediasoup-worker`. So I'll modify `Worker.ts` in a way that it tries first to load the binary from `worker/prebuild/Release`. And IMHO we should rename the downloaded binary (once extracted) to `mediasoup-worker` to make things easier. And this must work in Rust and Windows.

In `Worker.ts` we don't check (and we do not want to check) whether a file or folder exists. That's an async or blocking operation that should not be done within `Worker.ts` in runtime. So I've made `tar.extract()` move the extracted `mediasoup-worker` to `worker/out/Release/mediasoup-worker`, which is the official path of the worker.

> 2. For local development, make command must delete worker/prebuild/* so later it will use the locally build one. Must also apply to Rust.

I've just added env `MEDIASOUP_LOCAL_ENV: "true"` in `mediasoup-worker-prebuild.yaml` GH CI task.

4. When running CI tasks, npm ci (which will trigger "postinstall" script as usual) should NOT fetch the prebuilt binary but should always build it locally. I'll add an env for this.

No need for a new env. We should just use `MEDIASOUP_LOCAL_ENV: true`.

ibc:
I have amazing concerns about this `strip: 3`:

```ts
// Extract mediasoup-worker in the official mediasoup-worker path.
res.body
	.pipe(tar.extract(
		{
			// TODO: This is super controversial and weak. No option to just
			// "ignore all folders no matter how many levels there are?".
			strip : 3,
			newer : false,
			cwd   : WORKER_RELEASE_DIR
		}))
```

ibc:
Any thoughts @barlock?

ibc:
@nazar-pc is it enough if we fetch `mediasoup-worker` binary from GitHub and place it in `worker/out/Release/` for it to also be used by Rust? I don't think so since Rust needs to statically link against `libmediasoup-worker`, right? Then how to do it? Should we also publish `libmediasoup-worker` in GH and make `cargo` try to fetch it?

ibc:
> I have amazing concerns about this `strip: 3`:
> 
> ```ts
> // Extract mediasoup-worker in the official mediasoup-worker path.
> res.body
> 	.pipe(tar.extract(
> 		{
> 			// TODO: This is super controversial and weak. No option to just
> 			// "ignore all folders no matter how many levels there are?".
> 			strip : 3,
> 			newer : false,
> 			cwd   : WORKER_RELEASE_DIR
> 		}))
> ```

This is properly solved here https://github.com/versatica/mediasoup/pull/1087/commits/94a4e36648df4b9ca32602f098536b989902064e

npm-scripts.js
```diff
@@ -1,177 +1,209 @@
...
+const fetch = require('node-fetch');
+const tar = require('tar');
+
+const PKG = JSON.parse(fs.readFileSync('./package.json').toString());
+const IS_FREEBSD = os.platform() === 'freebsd';
+const IS_WINDOWS = os.platform() === 'win32';
+const MAYOR_VERSION = PKG.version.split('.')[0];
+const MAKE = process.env.MAKE || (IS_FREEBSD ? 'gmake' : 'make');
+const WORKER_RELEASE_DIR = 'worker/out/Release';
+const WORKER_RELEASE_BIN = 'mediasoup-worker';
```
barlock: I've had challenges with this on windows, the bin is `mediasoup-worker.exe`. 

In my pr, [I had done some searching](https://github.com/versatica/mediasoup/pull/1083/files#diff-0e4b0cf2dfea5eb3fd7fcefa330bbcbe8ec35f53426f00d14f34a94281ed7a2aR75-R83) in the package state to find all of the files at the top level, which I'm not sure is right either.

```suggestion
const WORKER_RELEASE_BIN = `mediasoup-worker${IS_WINDOWS ? '.exe' : ''}`;
```

I think this will fix that problem in a more obvious way.

npm-scripts.js
```diff
@@ -1,177 +1,209 @@
...
-
 const process = require('process');
 const os = require('os');
 const fs = require('fs');
 const { execSync, spawnSync } = require('child_process');
-const { version } = require('./package.json');
+const fetch = require('node-fetch');
+const tar = require('tar');
+
+const PKG = JSON.parse(fs.readFileSync('./package.json').toString());
```
barlock, line 8: Out of curiosity why not require the `.json` like it was before?

npm-scripts.js
```diff
@@ -1,177 +1,209 @@
...
+		}
+
+		case 'release':
+		{
+			checkRelease();
+			executeCmd(`git commit -am '${PKG.version}'`);
+			executeCmd(`git tag -a ${PKG.version} -m '${PKG.version}'`);
+			executeCmd(`git push origin v${MAYOR_VERSION}`);
+			executeCmd(`git push origin '${PKG.version}'`);
+			executeCmd('npm publish');
```
barlock:
```suggestion
			const { Octokit} = require('@octokit/rest');
			const octokit = new Octokit({ auth: process.env.GH_TOKEN });

			checkRelease();
			executeCmd(`git commit -am '${PKG.version}'`);
			executeCmd(`git tag -a ${PKG.version} -m '${PKG.version}'`);
			executeCmd(`git push origin v${MAYOR_VERSION}`);
			executeCmd(`git push origin '${PKG.version}'`);

			await octokit.repos.createRelease(
				{
					owner: 'integrated-cinematics',
					repo: 'mediasoup',
					name: version,
					tag_name: version,
					draft: false,
				}
			);
			
			executeCmd('npm publish');
```

This should automate the creation of a release for you. `@octokit/rest` being a new dev dependency.

nazar-pc:
Don't worry about Rust here, Rust is compiled either way, so I don't think we need pre-compiled stuff there.
I'm at the offsite this and next week, don't have much time to review.

npm-scripts.js
ibc: Thanks.

npm-scripts.js
ibc: This is part of my discarded attempt to move to ESM. Still I think it's not a bad idea. For instance, if you `require('package.json')` the whole JSON will be present in the bundled JS (in case of apps or libs that expose a bundle).

npm-scripts.js
ibc: Amazing, thanks, will add it.

jmillan (approved):

barlock (approved):
LGTM. 

FWIW regarding the macOS ARM builds (which is my primary workstation) I've been building it locally on my machine and just uploading it to the release page by hand b/c I wanted to save myself from having to build it every time I did an install and releases were infrequent enough.

😅 It's also the reason I added `MEDIASOUP_WORKER_DOWNLOAD_BASE` as an environment config, so I had the option to continue doing that after this change was accepted. 

Last I checked a few months ago, there were no free github runners that supported macOS arm for open source. A universal binary is probably the simplest way to cover that use case until GitHub supports arm runners natively.

ibc:
> FWIW regarding the macOS ARM builds (which is my primary workstation) I've been building it locally on my machine and just uploading it to the release page

Wow, that's a good idea. I will add a uploadMacArmPrebuiltWorker() function called in the "postinstall" task in case current platform and arch is macOS ARM.

BTW do you know how to upload it using the same library we already use (in this PR) to create the release? I can look for it tomorrow but maybe you can do some copy&paste here 😀

piranna:
> Also thanks to @piranna for his previous similar effort in PR #777.

I have been maintaining and updating it https://github.com/Mafalda-SFU/mediasoup, just only the nightly scripts are failling to detect new versions for some reason, and publishing of packages fails due to npm credentials, but besides that I have everything working there too :-)

piranna:
By the way, in my approach, I'm building all binaries on Github Actions, and later packaging them together, being selected the correct one at runtime. This is the recommended way, since distribution later is easier because it's a single download that algo gets cached on `~/.npm` folder. That way it's possible to install fully offline, meanwhile by downloading the binary, although it's not needed to compile anymore, it still requires an internet connection.

ibc:
> > FWIW regarding the macOS ARM builds (which is my primary workstation) I've been building it locally on my machine and just uploading it to the release page
> 
> Wow, that's a good idea. I will add a uploadMacArmPrebuiltWorker() function called in the "postinstall" task in case current platform and arch is macOS ARM.

So done here: 5b2ded365b4da5d2637fa387a0e4fe14af9e1a1d

ibc:
> By the way, in my approach, I'm building all binaries on Github Actions, and later packaging them together, being selected the correct one at runtime. This is the recommended way, since distribution later is easier because it's a single download that algo gets cached on `~/.npm` folder. That way it's possible to install fully offline, meanwhile by downloading the binary, although it's not needed to compile anymore, it still requires an internet connection.

Each `mediasoup-worker-x.x.x-platform-version.tgz` file is about 2.7 MB. We are now generating 4 prebuilts, so 10.8 MB. I don't want to include all those in the NPM package.Things may change in the future, let's see.

piranna:
> Things may change in the future, let's see.

My code is already there, in case we want to combine both implementations some day :-)

ibc:
> My code is already there, in case we want to combine both implementations some day :-)

Sure.

ibc:
> It's also the reason I added `MEDIASOUP_WORKER_DOWNLOAD_BASE` as an environment config

@barlock I've renamed env variables. Will point to the documentation when released.

ibc:
Bad news... https://github.com/versatica/mediasoup/issues/1089

Issue #1089, opened 2023-05-26
Prebuilt mediasoup-worker fails in Debian Bullseye 11: version `GLIBC_2.33' not found


ibc:
## Bug Report

After https://github.com/versatica/mediasoup/pull/1087 is merged and released, `npm i mediasoup` in Debian Bullseye 11 fetches `mediasoup-worker-3.12.0-linux-x64.tgz` from mediasoup GitHub Releases:https://github.com/versatica/mediasoup/releases.

Then it fails in runtime. Easy to reproduce by just running the worker directly:

```
~/node_modules/mediasoup$ ./worker/out/Release/mediasoup-worker

./worker/out/Release/mediasoup-worker: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by ./worker/out/Release/mediasoup-worker)
./worker/out/Release/mediasoup-worker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./worker/out/Release/mediasoup-worker)
./worker/out/Release/mediasoup-worker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./worker/out/Release/mediasoup-worker)
./worker/out/Release/mediasoup-worker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./worker/out/Release/mediasoup-worker)
```

### Your environment

- Operating system: Debian Bullseye 11
- Node version: 18.16.0
- npm version: 9.6.7
- gcc/clang version: gcc (Debian 10.2.1-6) 10.2.1 20210110
- mediasoup version: 3.12.0

ibc:
@barlock help please

ibc:
I've found a probably completely unrelated fix (no idea if it makes sense, haven't tested it because no idea about it): https://github.com/anbox/anbox/pull/1892/files

ibc:
Problem seems to be that my Debian host has different version of `libc` (2.31) installed:

```
/lib/x86_64-linux-gnu$ ls -l libc*

-rwxr-xr-x 1 root root 1.9M Apr 19 23:17 libc-2.31.so*
lrwxrwxrwx 1 root root   12 Apr 19 23:17 libc.so.6 -> libc-2.31.so*
[...]

PR #1091, opened 2023-05-29
Trigger prebuilt on Linux using Ubuntu 20.04

ibc:
Related to https://github.com/versatica/mediasoup/issues/1089


ibc:
So it works. I confirm that using Ubuntu 20.04 (instead of 22.04) to build the Linux prebuilt makes the binary work in Debian 11. Closing this testing PR and pushing directly to v3.

PR #1094, opened 2023-05-29
Fix: enable loading .env variables

jorodi:
Install the "dotenv" package and import it in npm-scripts.mjs to be able to load environment variables found in a .env file: e.g. MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD=true.

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -4,6 +4,7 @@ import fs from 'fs';
 import { execSync, spawnSync } from 'child_process';
 import fetch from 'node-fetch';
 import tar from 'tar';
+import 'dotenv/config';
 
 const PKG = JSON.parse(fs.readFileSync('./package.json').toString());
 const IS_FREEBSD = os.platform() === 'freebsd';
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "license": "ISC",
       "dependencies": {
         "debug": "^4.3.4",
+        "dotenv": "^16.0.3",
         "h264-profile-level-id": "^1.0.1",
         "node-fetch": "^3.3.1",
         "supports-color": "^9.3.1",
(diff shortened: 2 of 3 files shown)


ibc:
Why have you closed this PR?

ibc:
Anyway this PR is not needed. You are supposed to install dotenv and have your .env file in your top Node app that later imports mediasoup, and not in mediasoup itself. mediasoup (and any other dep imported by your Node app) will read those envs from process.env.

jorodi:
> Why have you closed this PR?

Sorry. I accidentally deleted the branch when I was pushing a fix.

> Anyway this PR is not needed. You are supposed to install dotenv and have your .env file in your top Node app that later imports mediasoup, and not in mediasoup itself. mediasoup (and any other dep imported by your Node app) will read those envs from process.env.

I added `MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD=true` to my project's .env.  The prebuilt was still used instead of skipping it.  The fix I was going to push allows reading from a project's .env.  Note, the .env is located at the root of the project's folder, which is standard.

ibc:
> I added MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD=true to my project's .env. The prebuilt was still used instead of skipping it.

Because that happens during installation time (npm install or npm ci) instead of during runtime. 

I understand your concern but we cannot make mediasoup import a library to read a potential .env file in the parent folder. There are many ways to setup an environment variable before running any command (such as "npm install" in your case) but that's up to the project rather than being built in mediasoup.

ibc:
Imagine that your project does not directly have mediasoup as a dep but instead includes another dep that includes mediasoup. Your mediasoup may end located at project/node_modules/anotherDep/node_modules/mediasoup (it depends on many npm related things) and your trick wouldn't work. I'm afraid we cannot accept this PR.

jorodi:
In the case of a project (parent app) with a dependency (A) that has a mediasoup dependency (B), then dependency (A) would set the `MEDIASOUP_` .env vars if it had a need to.  This obviously overrides the parent app's .env setting.  But if the parent app was using dependency A, then I'd assume that is the functionality they were looking for.

This is an edge case (and should be considered), but for the majority of apps directly using mediasoup as a dependency, this would be a convenience.  

However, understood.  Thank you for taking the time.

PR #1098, opened 2023-06-01
fix increase layer when current layer has not receive SR

penguinol:
For transport without bandwidth estimation, such as `PlainRtpTransport`, we will only change layers when receiving first SR of each stream or there is a change in the score.
But in follow situation, we can not switch to the preferred layer.
We may receive SR of preferred layer before we receiving keyframe of current target layer, when `currentSpitalLayer` is still -1.
After we received keyframe of target layer, we received another SR of preferred layer, but it's not the first SR of this stream, so we will not call `MayChangeLayers`

```mermaid
sequenceDiagram
Src ->> mediasoup: CreateProducer S3T3
Dst->> mediasoup: CreateConsumer, perferred(2, 2)
Src ->> mediasoup: non-keyframe of S1
mediasoup->>mediasoup: targetSpatialLayer = 1, tsReferenceSpitalLayer = 1, s1 score = 10
Src ->> mediasoup: non-keyframe of S2
mediasoup->>mediasoup: s2 score = 10
Src ->> mediasoup: SenderReport of S1
mediasoup->>mediasoup: save ntp of S1
Src ->> mediasoup: SenderReport of S2
mediasoup->>mediasoup: fisrt SR of S2, but currentSpitalLayer is -1, does not call `MayChangeLayers` 
Src ->> mediasoup: key-frame of S1
mediasoup->>mediasoup: currentSpatialLayer = 1
Src ->> mediasoup: SenderReport of S2 
mediasoup->>mediasoup: not first SR of S2, does not call `MayChangeLayers`
```


ibc:
I don't really understand this part:
CleanShot 2024-05-19 at 19 36 23@2x

If no keyframe of S1 is received, why does the graph says that `tsReferenceSpatialLayer` is 1? AFAIK it's -1 at this time.

ibc:
@penguinol do you think this PR can also fix issue #492 and replace this old but never addressed PR #500?

ibc:
Ok I think I get it. I'll test and merge. Sorry for the delay.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -366,11 +366,11 @@ namespace RTC
 
 		MS_DEBUG_TAG(simulcast, "first SenderReport [ssrc:%" PRIu32 "]", rtpStream->GetSsrc());
 
-		// If our current selected RTP stream does not yet have SR, do nothing since
+		// If our TsReference RTP stream does not yet have SR, do nothing since
 		// we know we won't be able to switch.
```
```suggestion
		// If our RTP timestamp reference stream does not yet have SR, do nothing
		// since we know we won't be able to switch.
```

PR #1099, opened 2023-06-01
PlainTransportOptions: add optional rtcpListenInfo

### LINKED ISSUE #1096: Plain Transport Custom RTCP Port
## Feature Request

When we create PlainTransport with `rtcpMux` `false`, then the transport uses 2 ports. One for RTP and the other for RTCP. While creating we can pass the port to `router.createPlainTransport()`, however, the RTCP port is selected from the worker's port range. I'm requesting an option to pass down a custom RTCP port to  `router.createPlainTransport()`, so the application layer can manage this port. In our app, we're already managing pipe transport ports and with the addition of plain transport, we're only able to manage RTP ports and not RTCP.

ibc:
Fixes #1096

NOTE: PR targeting `flatbuffers` branch since it depends on the new `TransportListenInfo` type.

### DIFF
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -646,7 +646,7 @@ export class Router<RouterAppData extends AppData = AppData>
 		}
 
 		// If rtcpMux is enabled, ignore rtcpListenInfo.
-		if (rtcpMux && rtcpListenInfo !== undefined)
+		if (rtcpMux)
 		{
 			logger.warn('createPlainTransport() | ignoring given rtcpListenInfo since rtcpMux is enabled');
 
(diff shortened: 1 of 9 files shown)


node/src/Router.ts
```diff
@@ -644,6 +645,14 @@ export class Router<RouterAppData extends AppData = AppData>
 			throw new TypeError('if given, appData must be an object');
 		}
 
+		// If rtcpMux is enabled, ignore rtcpListenInfo.
+		if (rtcpMux && rtcpListenInfo !== undefined)
```
jmillan: We can unset `rtcpListenInfo` unconditionally.

```suggestion
		if (rtcpMux)
```

jmillan (approved):
Little change requested. Other than that :+1:

PR #1108, opened 2023-06-30
fix jitter in rr

penguinol:
According to RFC, the unit of jitter in rr should be timestamp
`interarrival jitter: 32 bits
An estimate of the statistical variance of the RTP data packet interarrival time, measured in timestamp units and expressed as an unsigned integer`

### DIFF
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -236,7 +236,7 @@ namespace RTC
 		RTC::RtpStream::FillJsonStats(jsonObject);
 
 		jsonObject["type"]        = "inbound-rtp";
-		jsonObject["jitter"]      = this->jitter;
+		jsonObject["jitter"]      = this->jitter / this->params.clockRate * 1000;
 		jsonObject["packetCount"] = this->transmissionCounter.GetPacketCount();
 		jsonObject["byteCount"]   = this->transmissionCounter.GetBytes();
 		jsonObject["bitrate"]     = this->transmissionCounter.GetBitrate(nowMs);
@@ -510,7 +510,7 @@ namespace RTC
 
 		// Fill the rest of the report.
 		report->SetLastSeq(static_cast<uint32_t>(this->maxSeq) + this->cycles);
-		report->SetJitter(this->jitter * (this->params.clockRate / 1000));
+		report->SetJitter(this->jitter);
 
 		if (this->lastSrReceived != 0)
 		{
@@ -698,8 +698,8 @@ namespace RTC
 		}
 
 		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
-		auto transit = static_cast<int>(
-		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
+		auto transit =
+		  static_cast<int32_t>(DepLibUV::GetTimeMs() * (this->params.clockRate / 1000) - rtpTimestamp);
 		int d = transit - this->transit;
 
 		// First transit calculation, save and return.
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -510,7 +510,7 @@ namespace RTC
 
 		// Fill the rest of the report.
 		report->SetLastSeq(static_cast<uint32_t>(this->maxSeq) + this->cycles);
-		report->SetJitter(this->jitter);
+		report->SetJitter(this->jitter * (this->params.clockRate / 1000));
 
 		if (this->lastSrReceived != 0)
 		{


ibc:
LGTM

jmillan (approved):

ibc:
Nope. RFC 3550 explains how to calculate the jitter:

> A.8 Estimating the Interarrival Jitter
> 
>    The code fragments below implement the algorithm given in Section
>    6.4.1 for calculating an estimate of the statistical variance of the
>    RTP data interarrival time to be inserted in the interarrival jitter
>    field of reception reports.  The inputs are r->ts, the timestamp from
>    the incoming packet, and arrival, the current time in the same units.
>    Here s points to state for the source; s->transit holds the relative
>    transit time for the previous packet, and s->jitter holds the
>    estimated jitter.  The jitter field of the reception report is
>    measured in timestamp units and expressed as an unsigned integer, but
>    the jitter estimate is kept in a floating point.  As each data packet
>    arrives, the jitter estimate is updated:
> 
>       int transit = arrival - r->ts;
>       int d = transit - s->transit;
>       s->transit = transit;
>       if (d < 0) d = -d;
>       s->jitter += (1./16.) * ((double)d - s->jitter);

Note that `r->ts` is the rtp timestamp **which takes into account the clock rate of the stream**.

And we replicate the very same code in mediasoup in `RtpStreamRecv::CalculateJitter()`:

```c++
		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
		auto transit = static_cast<int>(
		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
		int d = transit - this->transit;

		// First transit calculation, save and return.
		if (this->transit == 0)
		{
			this->transit = transit;

			return;
		}

		this->transit = transit;

		if (d < 0)
		{
			d = -d;
		}

		this->jitter += (1. / 16.) * (static_cast<double>(d) - this->jitter);
```

So we are good.

In addition, if the jitter computation was wrong, the fix should be done in `RtpStreamRecv::CalculateJitter()` method rather than modifying its returned value later just when generating stats.

penguinol:
In standard `int transit = arrival - r->ts;`, but ours is `int transit = arrival/clockrate - r->ts/clockrate;`,
So standard is in `timestamp` unit, but ours is in `ms` unit.

Fix here is because the jitter in status json is more readable for human in ms unit.

penguinol:
@ibc 
For example in webrtc:
https://webrtc.googlesource.com/src/+/refs/heads/main/modules/rtp_rtcp/source/receive_statistics_impl.cc#153
```c++
TimeDelta receive_diff = receive_time - *last_receive_time_;
  RTC_DCHECK_GE(receive_diff, TimeDelta::Zero());
int32_t receive_diff_rtp =
      (receive_diff * packet.payload_type_frequency()).seconds<uint32_t>();
  int32_t time_diff_samples =
      receive_diff_rtp - (packet.Timestamp() - last_received_timestamp_);
  time_diff_samples = std::abs(time_diff_samples);
```
They also convert from ms to timestamp by `int32_t receive_diff_rtp =
      (receive_diff * packet.payload_type_frequency()).seconds<uint32_t>();`

And in ion-sfu:
https://github.com/ionorg/ion-sfu/blob/master/pkg/buffer/buffer.go#L367-L376

```go
	arrival := uint32(arrivalTime / 1e6 * int64(b.clockRate/1e3))
	transit := arrival - p.Timestamp
```
When calcuate transit, they convert arrival time from microsecond to timestamp

jmillan:
I definitely see we are working in the same units already. We work in milliseconds since the begining.

```
/* transit (ms) = now (ms) - rtpTimestamp (ms) */
auto transit = static_cast<int>(
		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
/
* d = transit (ms) - this->transit (ms) */
int d = transit - this->transit;
```

ibc:
In the standard:

```
int transit = arrival - r->ts;
```

* where I assume that `r->ts` is the timestamp of the RTP packet.

In mediasoup:

```c++
auto transit = static_cast<int>(
		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
```

* where `rtpTimestamp` is the timestamp of the RTP packet (we call `CalculateJitter(packet->GetTimestamp())`).

The spec says:

> The inputs are `r->ts`, the timestamp from the incoming packet, and `arrival`, the current time **in the same units**.

However we are not honouring that. In mediasoup:

* `arrival` is `DepLibUV::GetTimeMs()`, in ms.
* `r->ts` is `rtpTimestamp * 1000 / clockRate`, in ms as well.

In summary:

* `transit` in the standard is calculated in timestamp units.
* `transit` in mediasoup is calculated in ms.

ibc:
Let's please assume this (hope we agree):

1. We want that `RtpStreamRecv::CalculateJitter(uint32_t rtpTimestamp)` calculates `this->jitter` in timestamp units rather than ms.

2. Apparently we need to change this:
   ```c++
   auto transit = static_cast<int>(
   		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
   ```
   to this:
   ```c++ 
   auto transit = static_cast<int>(CURRENT_TIME_IN_TIMESTAMP_UNITS - rtpTimestamp);
   ```

Yes or not?

ibc:
What about these changes?

```diff
diff --git a/worker/include/RTC/RtpStreamRecv.hpp b/worker/include/RTC/RtpStreamRecv.hpp
index eec7a51a7..16c0d07e7 100644
--- a/worker/include/RTC/RtpStreamRecv.hpp
+++ b/worker/include/RTC/RtpStreamRecv.hpp
@@ -114,7 +114,7 @@ namespace RTC
 		uint64_t lastSrReceived{ 0u };
 		// Relative transit time for prev packet.
 		int32_t transit{ 0u };
-		// Jitter in ms.
+		// Jitter in RTP timestamp units.
 		uint32_t jitter{ 0u };
 		uint8_t firSeqNumber{ 0u };
 		uint32_t reportedPacketLost{ 0u };
diff --git a/worker/src/RTC/RtpStreamRecv.cpp b/worker/src/RTC/RtpStreamRecv.cpp
index 0fe1de02c..f8bb9100f 100644
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -692,15 +692,14 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		if (this->params.clockRate == 0u)
+		if (GetClockRate() == 0u)
 		{
 			return;
 		}
 
 		// NOTE: Based on https://github.com/versatica/mediasoup/issues/1018.
-		auto transit = static_cast<int>(
-		  DepLibUV::GetTimeMs() - (static_cast<uint64_t>(rtpTimestamp) * 1000 / this->params.clockRate));
-		int d = transit - this->transit;
+		auto transit = static_cast<int>((DepLibUV::GetTimeMs() * GetClockRate() / 1000) - rtpTimestamp);
+		int d        = transit - this->transit;
```

jmillan:
> What about these changes?

Makes sense to me.

ibc:
@penguinol I'm closing this PR and replacing it with https://github.com/versatica/mediasoup/pull/1117. Please comment there. Thanks.

PR #1110, opened 2023-07-06
Replace `clone` in utils with `structuredClone` everywhere

threema-lenny:
`structuredClone` should be much more clean (and potentially efficient) to clone an object recursively. It also prevented some type casting.

Hopefully bumping to Node 17 is acceptable [considering that Node 16 is almost EOL](https://github.com/nodejs/release#release-schedule).

### DIFF
--- a/node/src/Transport.ts
+++ b/node/src/Transport.ts
@@ -1,7 +1,6 @@
 import { v4 as uuidv4 } from 'uuid';
 import { Logger } from './Logger';
 import { EnhancedEventEmitter } from './EnhancedEventEmitter';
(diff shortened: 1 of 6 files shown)


ibc:
We cannot require Node 17 (which by definition is an unstable/experimental Node version) and definitely we want to support Node 16for a while. I'll leave this PR open to not forget about this feature in the future once we can require Node >= 18.

ibc:
This is done in `flatbuffers` branch conditionally, meaning that the `clone()` method in `utils.ts` verifies whether `structuredClone()` exists or not and uses it if so.

PR #1112, opened 2023-07-10
Add New API replaceProducer()  in Transport 

kingpeter2015:
After the browser create a new Consumer, if One `Consumer` want to subscribe another `Producer`,  it's not necessary to call `consumer.close()`, then `transport.consume()` to update. With the New `replaceProducer()` API, one can modify the consumer attribute on the mediasoup side while keeping the consumer on the browser side unchanged.

----------
Motivation:
* When the browser has a lot of video windows that need to poll repeatedly, destroying and recreating consumers will reduce performance. New API can speed up video stream.
* To avoid the `webrtc` [issue](https://github.com/versatica/mediasoup-client/issues/237): Destroying and recreating consumers repeatedly may cause memory leaks.


.gitignore
```diff
@@ -41,3 +41,5 @@
 
 # Packaged module.
 *.tgz
+yarn.lock
+package-lock.json
```
ibc, line 45: Why? Why are you changing our project settings? We do want to have exact NPM dep versions specified.

worker/src/Channel/ChannelRequest.cpp
```diff
@@ -70,7 +70,8 @@ namespace Channel
 		{ "rtpObserver.pause",                           ChannelRequest::MethodId::RTP_OBSERVER_PAUSE                               },
 		{ "rtpObserver.resume",                          ChannelRequest::MethodId::RTP_OBSERVER_RESUME                              },
 		{ "rtpObserver.addProducer",                     ChannelRequest::MethodId::RTP_OBSERVER_ADD_PRODUCER                        },
-		{ "rtpObserver.removeProducer",                  ChannelRequest::MethodId::RTP_OBSERVER_REMOVE_PRODUCER                     }
+		{ "rtpObserver.removeProducer",                  ChannelRequest::MethodId::RTP_OBSERVER_REMOVE_PRODUCER                     },
+		{ "consumer.update", ChannelRequest::MethodId::CONSUMER_UPDATE             }
```
ibc, line 74: Syntax doesn't match the existing lines above.

ibc:
I'm afraid we will do this in mediasoup v4 and not before. This PR lacks too many things, doesn't follow basic code guidelines, doesn't pass lint tasks, doesn't include tests, it misses notifications to Node/Rus layers about new consumer parameters after the producer has been replaced, etc. I'm afraid we cannot accept this PR and honestly don't want to deal with it yet.

bianxg:
When will mediasoup v4  be launched, Thanks.

ibc:
> When will mediasoup v4  be launched, Thanks.

We haven't even started it.

PR #1114, opened 2023-07-10
worker: Add Transport::Destroying() protected method

ibc:
Must be called by `Transport` child classes at the top of their destructors.

Related to https://mediasoup.discourse.group/t/rarely-crashes-on-sendsctpdata-when-enablesctp-is-enabled/5340/

@nazar-pc as told in the forum, this bug only affected Rust although I think that it could perfectly affect mediasoup Node as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### Next
+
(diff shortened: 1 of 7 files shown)


ibc:
@satoren check this PR please.

nazar-pc (approved):

PR #1118, opened 2023-07-13
add test with ASAN

satoren:
Run test with Address Sanitizer

This test currently fails due to stack-buffer-overflow, but we will fix that in another PR.

### DIFF
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -154,6 +154,11 @@ Rewrites mediasoup-worker C++ files using [clang-format](https://clang.llvm.org/
 
 Builds and runs the `mediasoup-worker-test` binary at `worker/out/Release/` (or at `worker/out/Debug/` if the "MEDIASOUP_BUILDTYPE" environment variable is set to "Debug"), which uses [Catch2](https://github.com/catchorg/Catch2) to run test units located at `worker/test/` folder.
 
+
+### 'make test-asan'
+
+Run test with Address Sanitizer
+
 ### `make tidy`
 
 Runs [clang-tidy](http://clang.llvm.org/extra/clang-tidy/) and performs C++ code checks following `worker/.clang-tidy` rules.
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -157,7 +157,7 @@ Builds and runs the `mediasoup-worker-test` binary at `worker/out/Release/` (or 
 
 ### 'make test-asan'
 
-Run test with Address Sanitizer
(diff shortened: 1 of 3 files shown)


worker/Makefile
```diff
@@ -225,6 +225,11 @@ else
 	$(BUILD_DIR)/mediasoup-worker-test --invisibles --use-colour=yes $(MEDIASOUP_TEST_TAGS)
 endif
 
+test-asan: setup
```
ibc, line 229: You must add "test-asan" also in a list with all task names at the top of this file.

doc/Building.md
```diff
@@ -154,6 +154,11 @@ Rewrites mediasoup-worker C++ files using [clang-format](https://clang.llvm.org/
 
 Builds and runs the `mediasoup-worker-test` binary at `worker/out/Release/` (or at `worker/out/Debug/` if the "MEDIASOUP_BUILDTYPE" environment variable is set to "Debug"), which uses [Catch2](https://github.com/catchorg/Catch2) to run test units located at `worker/test/` folder.
 
+
+### 'make test-asan'
+
+Run test with Address Sanitizer
```
ibc: Add dot at the end, please.

worker/meson.build
```diff
@@ -333,7 +325,16 @@ mediasoup_worker_test = executable(
     'test/src/Utils/TestJson.cpp',
     'test/src/Utils/TestString.cpp',
     'test/src/Utils/TestTime.cpp',
+]
```
ibc, line 328: Blank line after this ] please.

worker/meson.build
```diff
@@ -350,6 +351,36 @@ test(
...
+    '-DMS_LOG_STD',
+    '-DMS_TEST',
+    '-fsanitize=address',
+  ],
+  link_args: [
+    '-fsanitize=address',
+  ],
+)
+
+
```
ibc: Only one blank line here instead of two, please.

ibc (commented):
Just cosmetic changes requested.

ibc (approved):
Will merge later today.

jmillan:
Why not enable ASAN always rather than create a new target?

ibc:
> Why not enable ASAN always rather than create a new target?

Because it is probably way slower, it doesn't work on Windows and we still want to be able to test worker in Windows and any other OS not supporting ASAN.

jmillan (approved):

PR #1122, opened 2023-07-19
Fix lack of "layerschange" event when all streams in the producer die

ibc:
The initial purpose of this PR was to JUST enable RTP inactivity check in simulcast streams (this was already done in PR https://github.com/versatica/mediasoup/pull/1061) and **just** if the video has more than 1 single stream (remember that if the producer video is VP8/H264 with a single stream/encoding with N temporal layers, it generates a `SimulcastConsumer`). And this has been done already. However I've found another issue (regression that happened no idea when).

So here the problem:

- Simulcast with > 1 streams.
- So RTP inactivity checks are enabled in all streams.
- All Producer streams die so RTP inactivity happens in all them.
- `SimulcastConsumer::ProducerRtpStreamScore()` is called in all them with `score: 0`.
- Such a method should trigger 'consumerlayerschange' with `spatialLayer: null` and `temporalLayer: null`, but it doesn't (pretty sure it did in the past).
- Why? Because `SimulcastConsumer::ProducerRtpStreamScore()` ends calling `this->MayChangeLayers()` which ends calling `this->listener->OnConsumerNeedBitrateChange()`.
- `Transport::OnConsumerNeedBitrateChange()` ends calling `this->DistributeAvailableOutgoingBitrate()` which calls `consumer->GetBitratePriority()`.
- `SimulcastConsumer::GetBitratePriority()` checks `this->IsActive()` and it returns `false` so `Transport` forgets about this consumer and won't call `IncreaseLayer()` etc, so the consumer won't have a chance to notify null layers.
- `SimulcastConsumer::IsActive()` returns false because it's check the score of all the streams in the producer and, if all 0, it returns `false`.
- So here the problem. Not sure how to resolve it. But I did: https://github.com/versatica/mediasoup/pull/1122/files#diff-24e17490eee048f39100d99aee1f3b0f842a73ebca6128162e3e838443c07b4bR1336

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Next
+
+* `SimulcastConsumer`: Fix lack of "consumerlayerschange" event when all streams in the producer die ([PR #1122](https://github.com/versatica/mediasoup/pull/1122)).
+
 
 ### 3.12.6
 
--- a/worker/include/RTC/Consumer.hpp
+++ b/worker/include/RTC/Consumer.hpp
@@ -14,6 +14,7 @@
 #include "RTC/RtpHeaderExtensionIds.hpp"
 #include "RTC/RtpPacket.hpp"
 #include "RTC/RtpStream.hpp"
+#include "RTC/RtpStreamRecv.hpp"
 #include "RTC/RtpStreamSend.hpp"
 #include "RTC/Shared.hpp"
 #include <absl/container/flat_hash_set.h>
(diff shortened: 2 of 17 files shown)


jmillan:
> SimulcastConsumer::GetBitratePriority() checks this->IsActive() and it returns false so Transport forgets about this consumer and won't call IncreaseLayer() etc, so the consumer won't have a chance to notify null layers.

How can it not be active at this point? `MayChangeLayers()` is only called if the Consumer is active.

ibc:
Because all the streams of the producer now have score 0 so SimulcastConsumer::IsActive() returns false. And because I think we call MayChangeLayers() by just checking if parent Consumer::IsActive() which is true.

jmillan:
I wonder if the following change in `SimulcastConsumer::ProducerRtpStreamScore` would be more clear and explicit:

```diff
diff --git a/worker/src/RTC/SimulcastConsumer.cpp b/worker/src/RTC/SimulcastConsumer.cpp
index 3527f3c0f..a5d4a44aa 100644
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -324,9 +324,14 @@ namespace RTC

                if (RTC::Consumer::IsActive())
                {
+                       // All Producer streams are dead.
+                       if (!IsActive())
+                       {
+                               UpdateTargetLayers(-1, -1);
+                       }
+
                        // Just check target layers if the stream has died or reborned.
                        // clang-format off
-                       if (
+                       else if (
                                !this->externallyManagedBitrate ||
                                (score == 0u || previousScore == 0u)
```

EDIT: diff lines may not match with 'v3' since I was in a different branch.

ibc:
@jmillan pushed here. Tested. It works. https://github.com/versatica/mediasoup/pull/1122/commits/331d42bdcbabfed98c9391fd62505a16a69cd759

PR #1125, opened 2023-07-23
Properly handle errors on `setInterval()`

piranna:
I've found a bug on the tests where if `dataProducer.send(message, ppid);` throws an exception, that's unhandled and the full tests crashes. This PR replaces `setTimeout()` for Node.js promised one, so instead of errors happening in a error handler function that ends up to global scope, now they are properly captured by the scope of the async function.

### DIFF
--- a/node/src/tests/test-DirectTransport.ts
+++ b/node/src/tests/test-DirectTransport.ts
@@ -1,3 +1,5 @@
+import { setInterval } from 'node:timers/promises';
+
 import * as mediasoup from '../';
 
 const { createWorker } = mediasoup;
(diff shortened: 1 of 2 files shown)


ibc:
Thanks. I will check this maybe on next week. The thing is that I already fixed one of those tests but just in the flatbuffers branch (because we added pause()/resume() in DataProducer and DataConsumer and the test uses them). So I must check carefully and before doing it I have a lot of work to be done.

piranna:
> Thanks. I will check this maybe on next week. The thing is that I already fixed one of those tests but just in the flatbuffers branch (because we added pause()/resume() in DataProducer and DataConsumer and the test uses them). So I must check carefully and before doing it I have a lot of work to be done.

I can be able to merge this change with the flatbuffers branch if it's working to see if there's any conflict.

ibc:
I'm closing this PR since changes were done in the meantime and this PR fixes the problem as well (https://github.com/versatica/mediasoup/pull/1228). But it's nice to know about Node promised `setInterval`.

piranna:
> it's nice to know about Node promised `setInterval`

Yes, it's pretty nice, there's several nice and not so much known APIs added lately in the build-in library :-)

PR #1131, opened 2023-07-30
Implement RtpParameters::from_fbs() and RtpParameters::into_fbs()

nazar-pc:
As requested. Note that CI fails because APIs I have introduced are private and not used yet, which is expected.


jmillan (approved):
👍

PR #1132, opened 2023-07-30
Get Error subclasses in tests from mediasoup.types collection

piranna:
Error subclasses are being directly imported from their path in tests, but they are already being exported on `mediasoup.types` object. This change get the error subclasses from it, so there's only a single Mediasoup import and all tests are run only against the public API as acceptance tests. Only remaining local import is `ortc.ts` file, that I have found no way to access it otherwise to using it directly (since they are all of them stateless pure functions, maybe moving the file to its own dependency package?), but that can be reviewed in another PR.

There are some extra minor changes, like end of lines whitespace removal done automatically, or a couple of files where the only remaining usage of of `mediasoup` namespace was to get the `Worker` class, so I've included it in the object destructuring for cleanest.

### DIFF
--- a/node/src/tests/test-Consumer.ts
+++ b/node/src/tests/test-Consumer.ts
@@ -1,7 +1,6 @@
 import * as mediasoup from '../';
-import { UnsupportedError } from '../errors';
 
-const { createWorker } = mediasoup;
+const { createWorker, types: { UnsupportedError } } = mediasoup;
 
 let worker: mediasoup.types.Worker;
 let router: mediasoup.types.Router;
@@ -529,7 +528,7 @@ test('transport.consume() with enableRtx succeeds', async () =>
 	audioConsumer2.close();
 }, 2000);
 
-test('transport.consume() can be created with user provided mid', async () => 
+test('transport.consume() can be created with user provided mid', async () =>
 {
 	const audioConsumer1 = await transport2.consume(
 		{
--- a/node/src/tests/test-Producer.ts
+++ b/node/src/tests/test-Producer.ts
@@ -1,7 +1,6 @@
 import * as mediasoup from '../';
-import { UnsupportedError } from '../errors';
 
-const { createWorker } = mediasoup;
+const { createWorker, types: { UnsupportedError } } = mediasoup;
 
 let worker: mediasoup.types.Worker;
 let router: mediasoup.types.Router;
--- a/node/src/tests/test-Router.ts
+++ b/node/src/tests/test-Router.ts
@@ -1,7 +1,6 @@
 import * as mediasoup from '../';
-import { InvalidStateError } from '../errors';
 
-const { createWorker } = mediasoup;
+const { createWorker, types: { InvalidStateError } } = mediasoup;
 
 let worker: mediasoup.types.Worker;
 
--- a/node/src/tests/test-WebRtcServer.ts
+++ b/node/src/tests/test-WebRtcServer.ts
@@ -1,11 +1,10 @@
 // @ts-ignore
 import * as pickPort from 'pick-port';
 import * as mediasoup from '../';
-import { InvalidStateError } from '../errors';
 
-const { createWorker } = mediasoup;
+const { createWorker, types: { InvalidStateError, Worker } } = mediasoup;
 
-let worker: mediasoup.types.Worker;
+let worker: Worker;
 
 beforeEach(() => worker && !worker.closed && worker.close());
 afterEach(() => worker && !worker.closed && worker.close());
--- a/node/src/tests/test-Worker.ts
+++ b/node/src/tests/test-Worker.ts
@@ -2,11 +2,12 @@ import * as os from 'os';
 import * as process from 'process';
 import * as path from 'path';
 import * as mediasoup from '../';
-import { InvalidStateError } from '../errors';
 
-const { createWorker, observer } = mediasoup;
+const {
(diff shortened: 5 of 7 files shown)


ibc:
I don't want to make this huge syntax change. It's of course correct and legit but it's not the syntax used everywhere else. In fact, in test files we want to import mediasoup files directly from `mediasoup/n/src` folder rather than importing the compiled JS files in `mediasoup/node/lib` folder.

Regarding where things are imported from, those are tests so we can perfectly import "internal" files for testing purposes, even if they were not public.

I agree we have to rethink about the purpose of `types.ts`. Perhaps it should just export pure TypeScript types rather than also JS classes, but we'll reconsider it in next version.

So thanks, nothing is wrong in this PR but I see no reason to make cosmetic changes and adopt a syntax we are not used to and it's not used anywhere else in the code.

(I've fixed the extra space directly in master branch).

piranna:
> In fact, in test files we want to import mediasoup files directly from `mediasoup/n/src` folder rather than importing the compiled JS files in `mediasoup/node/lib` folder.

What's the difference about them in practical terms?


> I agree we have to rethink about the purpose of `types.ts`. Perhaps it should just export pure TypeScript types rather than also JS classes, but we'll reconsider it in next version.

So, maybe fully remove the `mediasoup.types` field, and instead generate a `types.d.ts` file with definitions of the classes signatures, or maybe better, define them as interfaces, having the real classes as private implementation detail? I agree on the idea :-)


> So thanks, nothing is wrong in this PR but I see no reason to make cosmetic changes and adopt a syntax we are not used to and it's not used anywhere else in the code.

To be honest... this PR and the issues and bugs I'm finding in the tests are due to the fact [I'm extracting Mediasoup tests](https://github.com/Mafalda-SFU/Mediasoup-node-tests) so I can use them as acceptance and compatibility tests of the Mediasoup API in Mafalda SFU :-) I said nothing in the past because I'm still improving them, but my idea was to propose later to use them as API compliance tests for both Mediasoup and Mafalda SFU (provide them a Mediasoup-like object, in this case the `mediasoup` library itself, and you are go). I'm totally capable of do the changes on the tests just with a couple of regexp, just only this PR would make code simple and more aesthetic, independently of the extraction process.

ibc:
> > In fact, in test files we want to import mediasoup files directly from `mediasoup/n/src` folder rather than importing the compiled JS files in `mediasoup/node/lib` folder.
> 
> 
> 
> What's the difference about them in practical terms?

Tests run under Jest with TypeScript so we don't need to transpile TS to JS in order to run the tests.

  
> > I agree we have to rethink about the purpose of `types.ts`. Perhaps it should just export pure TypeScript types rather than also JS classes, but we'll reconsider it in next version.
> 
> 
> 
> So, maybe fully remove the `mediasoup.types` field, and instead generate a `types.d.ts` file with definitions of the classes signatures, or maybe better, define them as interfaces, having the real classes as private implementation detail? I agree on the idea :-)

Honestly we do want to export classes (such as Worker, Router, etc) because those are needed when coding an app in TypeScript (to have a map of Workers, etc), but I see little benefit in having class interfaces just to avoid exporting the classes themselves.


> > So thanks, nothing is wrong in this PR but I see no reason to make cosmetic changes and adopt a syntax we are not used to and it's not used anywhere else in the code.
> 
> 
> 
> To be honest... this PR and the issues and bugs I'm finding in the tests are due to the fact [I'm extracting Mediasoup tests](https://github.com/Mafalda-SFU/Mediasoup-node-tests) so I can use them as acceptance and compatibility tests of the Mediasoup API in Mafalda SFU :-) I said nothing in the past because I'm still improving them, but my idea was to propose later to use them as API compliance tests for both Mediasoup and Mafalda SFU (provide them a Mediasoup-like object, in this case the `mediasoup` library itself, and you are go). I'm totally capable of do the changes on the tests just with a couple of regexp, just only this PR would make code simple and more aesthetic, independently of the extraction process.

I'll consider all this in v4 and there will be a ticket to propose ideas/changes, but I don't want to make breaking or huge changes in v3, specially now that we are focused on migration to flatbuffers.

piranna:
> Tests run under Jest with TypeScript so we don't need to transpile TS to JS in order to run the tests.

I know, that's why I was asking, since classes are already being exported from `mediasoup.types` object both in Typescript and transpiled Javascript, only thing that's not being available in transpiled `mediasoup.types` object are obviously the Typescript types, since they are being removed on tranpilation process.


> Honestly we do want to export classes (such as Worker, Router, etc) because those are needed when coding an app in TypeScript (to have a map of Workers, etc), but I see little benefit in having class interfaces just to avoid exporting the classes themselves.

I see it from an egoist perspective for obvious reasons :-) Besides that, I think it makes sense to don't export classes that can't be instantiated, just for the shake of being able to use `instanceof`. For coding in Typescript, interfaces are equally useful, since you can build "a map of Worker-like objects" instead of "a map of Worker instances", for example. Embrace the duck-typing :-)


> I'll consider all this in v4 and there will be a ticket to propose ideas/changes, but I don't want to make breaking or huge changes in v3, specially now that we are focused on migration to flatbuffers.

Thank you for being open to consider this ideas :-) I'm already writting my ideas in a Notion doc to discuss them when the opportunity arises, and I totally agree on NOT doing huge breaking changes, so far most of the ones I have are related to simplify the API, and migration would be minimal. No problem, let's talk about them after Summer :-)

PR #1139, opened 2023-08-09
Fix RTCP SDES

ibc:
### Details

- Initially this PR was intended to remove a non legit warning log. A SDES chunk may perfecly terminate with a single null octet (AKA item type === 0) and, if so, we MUST NOT check the next byte (item length) because it may not exist (if chunk is padded to 4 bytes) and, if it exists, it should be one or more zeroes (up to 3) to pad the chunk to 4 bytes.
- However then I've added more tests and replicated this one I've written in rtp.js library: https://github.com/versatica/rtp.js/blob/master/src/tests/rtcp/SdesPacket.test.ts#L195 and I also refactored the existing SDES test to test more things. And LOT of problems show up.

### Bonus Tracks

- Use Node 18 in `Dockerfile`.

### Issue 2

*UPDATE:* Fixed in this commit: https://github.com/versatica/mediasoup/pull/1139/commits/1d107e0188f3729ca2404a1ec89f63a8d7646520

Parsing of this SDES proper packet fails miserably ("parse packet 2" section in `TestSdes.cpp` in this branch):

```
0x82, 0xca, 0x00, 0x0c, // Type: 202 (SDES), Count: 2, Length: 12
// Chunk 2
0x00, 0x00, 0x04, 0xd2, // SSRC: 1234
0x01, 0x06, 0x71, 0x77, // Item Type: 1 (CNAME), Length: 6, Text: "qwerty"
0x65, 0x72, 0x74, 0x79,
0x06, 0x06, 0x69, 0xc3, // Item Type: 6 (TOOL), Length: 6, Text: "iñaki"
0xb1, 0x61, 0x6b, 0x69,
0x00, 0x00, 0x00, 0x00, // 4 null octets
// Chunk 3
0x00, 0x00, 0x16, 0x2e, // SSRC: 5678
0x05, 0x11, 0x73, 0x6f, // Item Type: 5 (LOC), Length: 17, Text: "somewhere œæ€"
0x6d, 0x65, 0x77, 0x68,
0x65, 0x72, 0x65, 0x20,
0xc5, 0x93, 0xc3, 0xa6,
0xe2, 0x82, 0xac, 0x00  // 1 null octet
```

1. `packet->GetSize()` should return 52 bytes however it returns 28 which is the byte length of the SDES header (4 bytes) plus only the first chunk (24 bytes), but there are 2 chunks!
2. However `packet->GetCount()` returns 2 as expected, and given that the code it runs is `return this->chunks.size();` it clearly means that it's parsing the second chunk with length 0 (wrong, it's 24).

I'm pretty sure that the SDES parsing in rtp.js is way better than here. For instance, by checking `Sdes.cpp` code I don't see any evidence that it's properly processing chunk null octets.
* As per spec, each SDES chunk MUST be padded to 4 bytes with null octets at the end.
* In addition, each chunk MUST end with 1, 2, 3 or 4 mandatory null octets. Why? Because "item type === 0" is the only way to say "this is the end of this chunk", and the remaining 1, 2 or 3 null octets are needed in case of padding.

So in packet above, chunk 1 ends with 4 null octets (which are part of the chunk) because the 2 items it contains are already padded to 4 bytes, so we need to add a "item type 0" null octet to define "end of items in this chunk", so we also need to add another 3 null octets for padding.

**To make it clear:** DO NOT try to parse as many null octets as possible. We must only parse from 1 to 4 null octet and stop once the chunk is padded to 4 bytes. So for example, in the packet above, look at the second/last item item in the first chunk:
```
0x06, 0x06, 0x69, 0xc3, // Item Type: 6 (TOOL), Length: 6, Text: "iñaki"
0xb1, 0x61, 0x6b, 0x69,
0x00, 0x00, 0x00, 0x00, // 4 null octets
```

And then look at the SSRC line of the second chunk:
```
0x00, 0x00, 0x16, 0x2e, // SSRC: 5678
```

**Yes!** it starts with 2 zeroes, but this is because SSRC is 5678 so `0x00, 0x00, 0x16, 0x2e`!

I'm pretty sure that parsing in mediasoup `Sdes.cpp` is not considering this at all.

- Check the SDES parser of **rtp.js** to see how it does it properly: https://github.com/versatica/rtp.js/blob/master/src/rtcp/SdesPacket.ts?ts=2#L111
- And also the constructor of `SdesChunk`: https://github.com/versatica/rtp.js/blob/master/src/rtcp/SdesPacket.ts#L336

### Issue 3

*UPDATE:* Fixed in this commit: https://github.com/versatica/mediasoup/pull/1139/commits/1114f978612d8bc85d17b96e971b763ff5286055

I've added "parse packet 3" https://github.com/versatica/mediasoup/pull/1139/commits/e303b74c6040e9da0f96bcdeba2b3743595eee14 and the test fails since it cannot parse packet 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### Next
+
+* Fix RTCP SDES ([PR #1139](https://github.com/versatica/mediasoup/pull/1139)).
+
+
 ### 3.12.8
 
 * Export `workerBin` absolute path ([PR #1123](https://github.com/versatica/mediasoup/pull/1123)).
--- a/worker/Dockerfile
+++ b/worker/Dockerfile
@@ -12,7 +12,7 @@ RUN \
 	   clang
 
 # Install node 16.
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash -
 RUN apt-get install -y nodejs
 
 # Make CC and CXX point to clang/clang++ installed above.
--- a/worker/include/RTC/RTCP/Sdes.hpp
+++ b/worker/include/RTC/RTCP/Sdes.hpp
@@ -121,6 +121,9 @@ namespace RTC
 					size += item->GetSize();
 				}
 
+				// Add the mandatory null octet.
+				++size;
+
 				// Consider pading to 32 bits (4 bytes) boundary.
 				// http://stackoverflow.com/questions/11642210/computing-padding-required-for-n-byte-alignment
 				return (size + 3) & ~3;
--- a/worker/src/RTC/RTCP/Sdes.cpp
+++ b/worker/src/RTC/RTCP/Sdes.cpp
@@ -61,7 +61,9 @@ namespace RTC
 			auto it = SdesItem::type2String.find(type);
 
 			if (it == SdesItem::type2String.end())
+			{
 				return Unknown;
+			}
 
 			return it->second;
 		}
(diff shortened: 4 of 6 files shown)


jmillan (approved):

ibc:
I have fixed **issue 2** in this commit (but tests are missing until **issue 1** is fixed):

https://github.com/versatica/mediasoup/pull/1139/commits/1d107e0188f3729ca2404a1ec89f63a8d7646520

ibc:
And **issue 1** is not a bug because I am stupid. Properly using vector iterator now. Will push soon.

ibc:
I've fixed Issues 1 and 2, but have found a new issue 3. See description above.

ibc:
Everything is fixed and tested. Running fuzzer before merging.

ibc:
We are good.

PR #1140, opened 2023-08-09
Replace `uuid` package with built-in `crypto.randomUUID()`

piranna:
This PR replace `uuid` package with Node.js built-in `crypto.randomUUID()` function. Also, it fixes a couple of typos where the random UUID would be generated if an empty string would be provided as `id`, now only it's being generated if `undefined` is provided (note, we should validate the provided `id` is an UUID instead of a random string, though).

OTOH, I've found that the `id` field of `DataProducerOptions` is not documented. Not fully sure if it makes sense to provide it all, since bandwidth and CPU are low, or it's just there only to mimic `Producer` API...

### DIFF
--- a/node/src/PipeTransport.ts
+++ b/node/src/PipeTransport.ts
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'crypto';
 import { Logger } from './Logger';
 import * as ortc from './ortc';
 import {
@@ -322,7 +322,7 @@ export class PipeTransport<PipeTransportAppData extends AppData = AppData>
 
 		const reqData =
 		{
-			consumerId             : uuidv4(),
+			consumerId             : randomUUID(),
 			producerId,
 			kind                   : producer.kind,
 			rtpParameters,
--- a/node/src/Transport.ts
+++ b/node/src/Transport.ts
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'crypto';
 import { Logger } from './Logger';
 import { EnhancedEventEmitter } from './EnhancedEventEmitter';
 import * as utils from './utils';
@@ -595,7 +595,7 @@ export class Transport
 			// do not include CNAME, create a random one.
 			else if (!this.#cnameForProducers)
 			{
-				this.#cnameForProducers = uuidv4().substr(0, 8);
+				this.#cnameForProducers = randomUUID().substring(0, 8);
 			}
 
 			// Override Producer's CNAME.
@@ -615,7 +615,7 @@ export class Transport
 
 		const reqData =
 		{
-			producerId : id || uuidv4(),
+			producerId : id ?? randomUUID(),
 			kind,
 			rtpParameters,
 			rtpMapping,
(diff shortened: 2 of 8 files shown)


nazar-pc:
This is not mergable, `crypto.randomUUID` is only supported starting Node.js 19

node/src/PipeTransport.ts
```diff
@@ -1,4 +1,5 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'node:crypto';
```
ibc: I know that importing from `node:xxxx` instead of from `xxxx` is nicer nowadays but we cannot change this just here and I'd appreciate if PRs don't change cosmetic opinionated styles despite they are valid.

ibc: I see that `node:xxxx` is added in other places but I don't want unrelated unneeded style changes in a PR.

node/src/Transport.ts
```diff
@@ -1,4 +1,5 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'node:crypto';
+
```
ibc: No reason for this blank line.

piranna:
> This is not mergable, `crypto.randomUUID` is only supported starting Node.js 19

According to https://nodejs.org/api/crypto.html#cryptorandomuuidoptions, it's available since v14.17.0... :-/

ibc:
As told above, we need to support Node >= 16 so we cannot merge this and won't be able to merge it until we deprecate Node 18 which may not happen in less than a year.

node/src/PipeTransport.ts
piranna: Ok, I'll remove that changes, and add `node:` in a new PR.

node/src/Transport.ts
piranna: I use to separate import types, it's easier to read. I can remove it too if you want.

nazar-pc:
Please use reviews for comments both ways :pray: I have 7 notifications already within minutes here.

node/src/Transport.ts
ibc: No file in mediasoup TypeScript code separates system and local imports with a blank line. Please, let's always honor the project style rather than trying to introduce our preferred one.

ibc:
CI fails because `uuid` is still being imported in `worker/scripts/xxxx` files.

node/src/PipeTransport.ts
ibc: Please don do that PR yet. We are actively working on the flatbuffers branch and we want to avoid huge changes in v3 branch until we merge flatbuffers.

piranna:
I've fixed all the concerns.

`uuid` package in `worker/scripts` is a transitive dependency of `request` package, that's deprecated, and it's a dependency itself of `nugget`, that's a dependency itself of `clang-tools-prebuilt`. Proper fix should be done there, meanwhile I've set `uuid` as `peerDependency` of worker scripts so tests can pass.

Zobbobb:
Cristiano Ronaldo

ibc:
Replacing this PR with https://github.com/versatica/mediasoup/pull/1193 which targets `flatbuffers` branch instead. Thanks.

PR #1141, opened 2023-08-10
Renamed WebRtcTransport `webRtcServerClosed()` to `listenServerClosed()`

piranna:
I've found a typo where the `WebRtcTransport.webRtcServerClosed()` was never called. Instead, the `Transport.listenServerClosed()` method is being directly called by WebRtcServer, so the clean-ups done at `webRtcServerClosed()` are never done. Since `listenServerClose` event is at Transport level with the idea maybe of have other listener servers in the future (raw TCP? RTP?), I've renamed `webRtcServerClosed()` to the more generic one `listenServerClosed()`, calling to the parent class.

### DIFF
--- a/node/src/WebRtcTransport.ts
+++ b/node/src/WebRtcTransport.ts
@@ -374,11 +374,11 @@ export class WebRtcTransport<WebRtcTransportAppData extends AppData = AppData>
 	}
 
 	/**
-	 * Called when closing the associated WebRtcServer.
+	 * Called when closing the associated listenServer (WebRtcServer).
 	 *
 	 * @private
 	 */
-	webRtcServerClosed(): void
+	listenServerClosed(): void
 	{
 		if (this.closed)
 		{


ibc:
And here a new added test that indeed fails in v3 branch and will pass once this PR is merged:

https://github.com/versatica/mediasoup/commit/a34ca02f9ae6856e5f550382a08d6460e6157154

ibc:
@nazar-pc I am not brave enough to know how this is done in Rust, I couldn't find any similar code so the bug probably doesn't happen in Rust.

nazar-pc:
Cleanups are done in `Drop` implementation in Rust, so most likely it works correctly there

piranna:
> And here a new added test that indeed fails in v3 branch and will pass once this PR is merged:
> 
> [a34ca02](https://github.com/versatica/mediasoup/commit/a34ca02f9ae6856e5f550382a08d6460e6157154)

Glad you liked this PR :-)

PR #1143, opened 2023-08-25
fix/make dtls fragment stay within mtu size range

### LINKED ISSUE #1100: DTLS fragment exceeds MTU size if the certificate is large.

Reproduce the bug by:

```bash
```

Generate a large certificate:

```bash
mkdir -p server/certs &&
```

Run server:

```bash
cd server
npm start
```

Run app:

```bash
cd app
npm start
```

image

pnts-se:
closes #1100 

This PR will make DTLS fragment stay within MTU size range. Currently that's not always the case, e.g. on large certificate size.

The solution is to stop calling `BIO_get_mem_data()` manually for pending out going data, and instead rely on a callback function, set with `BIO_set_callback_ex()`.

Packet capture during usage when running a worker build from this branch:
mediasoup_dtls_wireshark-230825

### DIFF
--- a/worker/include/RTC/DtlsTransport.hpp
+++ b/worker/include/RTC/DtlsTransport.hpp
@@ -172,6 +172,7 @@ namespace RTC
 			return this->localRole;
 		}
 		void SendApplicationData(const uint8_t* data, size_t len);
+		void SendDtlsData(void*, int);
 
 	private:
 		bool IsRunning() const
@@ -193,7 +194,6 @@ namespace RTC
 		}
 		void Reset();
 		bool CheckStatus(int returnCode);
-		void SendPendingOutgoingDtlsData();
 		bool SetTimeout();
 		bool ProcessHandshake();
 		bool CheckRemoteFingerprint();
--- a/worker/src/RTC/DtlsTransport.cpp
+++ b/worker/src/RTC/DtlsTransport.cpp
@@ -53,6 +53,19 @@ inline static unsigned int onSslDtlsTimer(SSL* /*ssl*/, unsigned int timerUs)
 		return 2 * timerUs;
 }
 
+long sslBioOutCallback(
+  BIO* bio, int operationType, const char* argp, size_t len, int argi, long argl, int ret, size_t* processed)
+{
+	long resultOfcallback = (operationType == BIO_CB_RETURN) ? ret : 1;
+	if (operationType == BIO_CB_WRITE && argp && len > 0)
+	{
+		MS_DEBUG_DEV("%i bytes of DTLS data ready to sent to the peer", static_cast<int>(len));
+		RTC::DtlsTransport* dtls = (RTC::DtlsTransport*)BIO_get_callback_arg(bio);
+		dtls->SendDtlsData((void*)argp, len);
+	}
+	return resultOfcallback;
+}
+
 namespace RTC
 {
 	/* Static. */
@@ -538,6 +551,8 @@ namespace RTC
 			goto error;
 		}
 
+		BIO_set_callback_ex(this->sslBioToNetwork, sslBioOutCallback);
+		BIO_set_callback_arg(this->sslBioToNetwork, (char*)this);
 		SSL_set_bio(this->ssl, this->sslBioFromNetwork, this->sslBioToNetwork);
 
 		// Set the MTU so that we don't send packets that are too large with no fragmentation.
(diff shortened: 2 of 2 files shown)


ibc:
Amazing. Thanks a lot.

However, probably it will take several days before we can review and merge this PR (super busy days after summer and lot of pending work).

pnts-se:
> However, probably it will take several days before we can review and merge this PR

Yes of course, I'm in no hurry. I will try to find something else mediasoup-related to help out with.

As for this PR: I'm just beginning to learn the code base and I'm not sure if I got it right in terms of encapsulation and data hiding. Might want to refactor that.

ibc:
I'm testing this PR right now, but I'll do in the `flatbuffers` branch to avoid conflicts once that is merged.

ibc:
I see a potential problem:

```c++
	void DtlsTransport::Reset()
	{
		MS_TRACE();

		int ret;

		if (!IsRunning())
			return;

		MS_WARN_TAG(dtls, "resetting DTLS transport");

		// Stop the DTLS timer.
		this->timer->Stop();

		// We need to reset the SSL instance so we need to "shutdown" it, but we
		// don't want to send a Close Alert to the peer, so just don't call
		// SendPendingOutgoingDTLSData().
		SSL_shutdown(this->ssl);

		this->localRole.reset();
		this->state            = DtlsState::NEW;
		this->handshakeDone    = false;
		this->handshakeDoneNow = false;

		// Reset SSL status.
		// NOTE: For this to properly work, SSL_shutdown() must be called before.
		// NOTE: This may fail if not enough DTLS handshake data has been received,
		// but we don't care so just clear the error queue.
		ret = SSL_clear(this->ssl);

		if (ret == 0)
			ERR_clear_error();
	}
```

Note the:

```c++
// We need to reset the SSL instance so we need to "shutdown" it, but we// don't want to send a Close Alert to the peer, so just don't call
// SendPendingOutgoingDTLSData().
SSL_shutdown(this->ssl);
```

However in your PR we no longer call `SendPendingOutgoingDTLSData()` or any other method to send data. Instead it is done automatically by the `sslBioOutCallback` callback. The idea is that we reset mediasoup side DTLS **without** sending DTLS Close Alert to the endpoint and I'm afraid we are loosing this.

ibc:
> I see a potential problem:
> 
> ```c++
> 	void DtlsTransport::Reset()
> 	{
> 		MS_TRACE();
> 
> 		int ret;
> 
> 		if (!IsRunning())
> 			return;
> 
> 		MS_WARN_TAG(dtls, "resetting DTLS transport");
> 
> 		// Stop the DTLS timer.
> 		this->timer->Stop();
> 
> 		// We need to reset the SSL instance so we need to "shutdown" it, but we
> 		// don't want to send a Close Alert to the peer, so just don't call
> 		// SendPendingOutgoingDTLSData().
> 		SSL_shutdown(this->ssl);
> 
> 		this->localRole.reset();
> 		this->state            = DtlsState::NEW;
> 		this->handshakeDone    = false;
> 		this->handshakeDoneNow = false;
> 
> 		// Reset SSL status.
> 		// NOTE: For this to properly work, SSL_shutdown() must be called before.
> 		// NOTE: This may fail if not enough DTLS handshake data has been received,
> 		// but we don't care so just clear the error queue.
> 		ret = SSL_clear(this->ssl);
> 
> 		if (ret == 0)
> 			ERR_clear_error();
> 	}
> ```
> 
> Note the:
> 
> ```c++
> // We need to reset the SSL instance so we need to "shutdown" it, but we// don't want to send a Close Alert to the peer, so just don't call
> // SendPendingOutgoingDTLSData().
> SSL_shutdown(this->ssl);
> ```
> 
> However in your PR we no longer call `SendPendingOutgoingDTLSData()` or any other method to send data. Instead it is done automatically by the `sslBioOutCallback` callback. The idea is that we reset mediasoup side DTLS **without** sending DTLS Close Alert to the endpoint and I'm afraid we are loosing this.

Let's ignore this use case since AFAIR it never worked fine. This is, if `DtlsTransport::Reset()` is called **once** the DTLS session was established, it never works again, and probably it's what the spec says. So whether we send a DTLS Close Alert or not is irrelevant because:
1. In most of the cases we do want to send it (when we really close/reset the `DtlsTransport` due to a fatal failure).
2. The use case of reseting the DTLS session once established is not a real/needed use case.

ibc:
@pnts-se I'm gonna close this PR in favour of this one https://github.com/versatica/mediasoup/pull/1156 which is basically a clone of yours but it targets `flatbuffers` branch instead of `v3` branch (to avoid complex git conflicts).

Thanks a lot!

PR #1146, opened 2023-08-29
Rust examples: Improve rooms registry logic

Jurshsmith:
This change removes the redundant logic when
getting or creating a room.

### DIFF
--- a/rust/examples/videoroom.rs
+++ b/rust/examples/videoroom.rs
@@ -232,7 +232,6 @@ mod rooms_registry {
     use crate::room::{Room, RoomId, WeakRoom};
     use async_lock::Mutex;
     use mediasoup::prelude::*;
-    use std::collections::hash_map::Entry;
     use std::collections::HashMap;
     use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


nazar-pc:
Thanks, I'll have to look at it carefully later

Jurshsmith:
Thanks @nazar-pc 

I am looking into the linting issue.

So that you know, this is just 1 of many PRs I will be creating. My main goal is to decouple the `ParticipantConnection` from the WebSocket connection to serve as a more generic integration template when working with `non-Actix` frameworks like `Warp`, `Axum`, or bare `Tokio`

nazar-pc:
These examples are not designed to be end-user applications even remotely, there is quite a lot more involved in production application. They are just a trivial showcase how to use the library and how to have a few more things around it. Supporting more frameworks is really not the goal here and decoupling will likely require more abstractions that will make code harder to understand, not easier.

Jurshsmith:
> These examples are not designed to be end-user applications even remotely, there is quite a lot more involved in production applications.

Agreed. I plan to keep existing features as is.

> ...and decoupling will likely require more abstractions that will make code harder to understand, not easier.

Not really. Personally, it was initially difficult to understand until I saw the boundary between the participant connection and the WebSocket connection. I think explicitly showing that boundary will simplify it a bit. I don't foresee a lot of abstractions besides a `WebsocketMessageHandler` trait. What do you think?


By the way, it seems the CI failure is environment-related (Something with the Rust version).

nazar-pc:
`v3` branch doesn't have `Cargo.lock` committed (WIP `flatbuffers` branch does), feel free to upgrade `rust-toolchain.toml` to latest stable release and fix all of the clippy lints afterwards if there are any. Ideally as a separate PR.

ibc:
As per above discussion, can we close this PR?

nazar-pc:
I still need to look into this, but as postponing, will try to review this week

Razzwan:
Let me express my opinion. I am trying to add Mediasoup to my Axum project right now and I am having a huge difficulty in understanding the code that uses Actix lib. And although the solution in this particular pull request does not completely solve this problem, however, I am sure that the problem is very relevant

ibc:
> I still need to look into this, but as postponing, will try to review this week

It's already next week

rust/examples/videoroom.rs
```diff
@@ -250,57 +249,23 @@ mod rooms_registry {
...
-                            std::thread::spawn(move || {
-                                futures_lite::future::block_on(async move {
-                                    rooms.lock().await.remove(&room_id);
-                                });
-                            });
-                        }
-                    })
-                    .detach();
-                    Ok(room)
+                    None => self.create_room(worker_manager).await,
```
nazar-pc, line 255: This will deadlock. `self.rooms.lock()` will create `MutexGuard`, which will be help for the whole scope of `if` block. At the the same time `self.create_room()` tries to take the same exact lock, meaning it will deadlock. This is why there was a new thread created before, ensuring deadlock doesn't happen.

nazar-pc (changes requested):
> It's already next week

Certainly 😅

UPD:
> This change removes the redundant logic when getting or creating a room.

Next time I'll appreciate a description of what and why is redundant because it doesn't seem to be redundant to me.

nazar-pc:
> Let me express my opinion. I am trying to add Mediasoup to my Axum project right now and I am having a huge difficulty in understanding the code that uses Actix lib. And although the solution in this particular pull request does not completely solve this problem, however, I am sure that the problem is very relevant

You'll certainly need to familiarize yourself with the way Actix works to understand this example. Actor model might take a bit of getting used to, but it does make a lot of sense for this particular use case. Note that I don't expect you to copy-paste it into a production application with minor changes, you really need to understand what is going on and why before writing something for production.

Echo example is a better place too start that videoroom, it is much simpler, while teaching you the basics of the available APIs. The rest is a business logic that heavily depends on what your application is supposed to do.

BTW you're welcome to write a simpler example without Actix, but I don't think it'll be dramatically simpler. There is certain amount of boilerplate needed either way and Actix was something I actively used back when I wrote these examples and that contained necessary building blocks.

ibc:
Can we close this PR? Let's please not keep PRs open forever just to discuss things on them.

nazar-pc:
I think so. Simplifying examples is a great idea, but this PR seems to introduce deadlocks instead.

Feel free to open a new PR.

PR #1147, opened 2023-08-30
Upgrade rust version in toolchain

Jurshsmith:


nazar-pc:
As mentioned in the other PR, expectation is that you will not only upgrade Rust version, but also fix all new clippy lints that pop up as the result of this

nazar-pc (approved):

PR #1149, opened 2023-09-01
rust: move PipeTransport to flatbuffers

jmillan:
PipeTransport tests are not disabled because they depend on WebRtcTransport, which will be done next.


ibc (approved):

nazar-pc (approved):
Makes sense

PR #1150, opened 2023-09-02
fix: template-id not allowed for destructor in c++20

o-u-p:
when compiled with c++20, it will get error:"template-id not allowed for destructor"


ibc:
Thanks

ibc:
Released in mediasoup 3.12.12.

PR #1153, opened 2023-09-11
fbs: FBS WebRtcTransport

jmillan:
New FBS types:

* CryptoSuite
* SrtpParameters
* FingerprintAlgorithm
* DtlsRole
* IceCandidateType
* IceCandidateTcpType
* IceRole
* IceState

Extra:
- [x] Remove IceState::Close from everywhere.


node/src/tests/test-WebRtcTransport.ts
```diff
@@ -306,6 +311,7 @@ test('webRtcTransport.connect() with wrong arguments rejects with TypeError', as
 		fingerprints :
 		[
 			{
+				// @ts-ignore.
 				algorithm : 'sha-256000',
```
jmillan, line 315: Now that we have types, does it make any sense to test with wrong values that the typing system would not allow to happen?

I know some of the values come from the network, but the client side is also typed, so IMHO we should consider undefined behaviour if wrong values are provided in runtime. Otherwise our code needs to consider such values and not simply rely on the types, which defeats the purpose of the type system IMHO.

worker/include/RTC/DtlsTransport.hpp
```diff
@@ -28,7 +30,6 @@ namespace RTC
 	public:
 		enum class Role
 		{
-			NONE = 0,
```
jmillan, line 31: Removed in favour of `std::optional`

worker/include/RTC/DtlsTransport.hpp
```diff
@@ -37,7 +38,6 @@ namespace RTC
 	public:
 		enum class FingerprintAlgorithm
 		{
-			NONE = 0,
```
jmillan, line 40: Removed in favour of `std::optional`

worker/include/RTC/DtlsTransport.hpp
jmillan: Do not use `NONE` as an internal way to indicate that there is no `Role`. Make use of c++17 `std::optional` functionality.

worker/include/RTC/IceServer.hpp
```diff
@@ -17,9 +18,14 @@ namespace RTC
 			NEW = 1,
 			CONNECTED,
 			COMPLETED,
-			DISCONNECTED
+			DISCONNECTED,
+			CLOSED
```
jmillan: Added `CLOSED` for consistency with Node and Rust. This can probably be removed though.

I'll review this.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -149,6 +149,101 @@ namespace RTC
 			SSL_CTX_free(DtlsTransport::sslCtx);
 	}
 
+	DtlsTransport::Role DtlsTransport::RoleFromFbs(FBS::WebRtcTransport::DtlsRole role)
```
jmillan, line 152: Please @ibc, can you take a look at this file?

I've reviewed multiple times and all tests pass.

node/src/tests/test-WebRtcTransport.ts
ibc: > Now that we have types, does it make any sense to test with wrong values that the typing system would not allow to happen?

Yes, because we cannot assume that every user uses TypeScript or validates network received data no matter it's created by media soup-client or not.

worker/include/RTC/DtlsTransport.hpp
ibc: Good.

worker/include/RTC/IceServer.hpp
ibc: When is CLOSED needed?

worker/src/Channel/ChannelSocket.cpp
```diff
@@ -259,9 +259,9 @@ namespace Channel
 		auto* message = FBS::Message::GetMessage(msg);
 
 		// TMP: For debugging.
-		// auto s = flatbuffers::FlatBufferToString(
-		//   reinterpret_cast<uint8_t*>(msg), FBS::Message::MessageTypeTable());
-		// MS_ERROR("%s", s.c_str());
+		auto s = flatbuffers::FlatBufferToString(
+		  reinterpret_cast<uint8_t*>(msg), FBS::Message::MessageTypeTable());
+		MS_ERROR("%s", s.c_str());
```
ibc: Can we remove this?

worker/src/RTC/DtlsTransport.cpp
ibc: It's a simple 1:1 mapping between C++ enums and FBS enums, what can be wrong? :)

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -913,7 +1011,7 @@ namespace RTC
 			this->timer->Stop();
 
 			// Process the handshake just once (ignore if DTLS renegotiation).
-			if (!wasHandshakeDone && this->remoteFingerprint.algorithm != FingerprintAlgorithm::NONE)
```
ibc, line 916: Is this condition wrapped into some `if (this->remoteFingerprint.algorithm.has_value())`? It's hard to see in the diff.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1031,8 +1129,6 @@ namespace RTC
 		MS_TRACE();
 
 		MS_ASSERT(this->handshakeDone, "handshake not done yet");
-		MS_ASSERT(
```
ibc, line 1034: Here we should assert that `this->remoteFingerprint.algorithm.has_value()` right?

worker/src/RTC/IceCandidate.cpp
```diff
@@ -1,39 +1,66 @@
...
 {
-	/* Instance methods. */
+	/* Class methods. */
 
-	flatbuffers::Offset<FBS::WebRtcTransport::IceCandidate> IceCandidate::FillBuffer(
-	  flatbuffers::FlatBufferBuilder& builder) const
+	IceCandidate::CandidateType IceCandidate::CandidateTypeFromFbs(
+	  FBS::WebRtcTransport::IceCandidateType type)
 	{
-		MS_TRACE();
```
ibc, line 14: Please don't remove `MS_TRACE()` calls.

worker/src/Channel/ChannelSocket.cpp
jmillan: I'll put it under a flag. It's intended for development and it's extremely useful.

worker/include/RTC/IceServer.hpp
jmillan: It is not. But it's present in TS and RUST. I'll create a TODO to remove them in this PR.

node/src/tests/test-WebRtcTransport.ts
jmillan: I think we should at some point. Ie: in Rust we don't do it that way and we rely on the type system.

worker/src/RTC/DtlsTransport.cpp
jmillan: I meant `file` not block :-)

worker/src/RTC/DtlsTransport.cpp
jmillan: Nop, `FingerPrint::algorithm` is not optional. If the handshake was done it must have a valid value.

worker/src/RTC/DtlsTransport.cpp
jmillan: Fingerprint.algorithm is not optional.

worker/src/RTC/IceCandidate.cpp
jmillan: I didn't. `IceCandidate::FillBuffer` has it's `MS_TRACE`. The diff just shows it moved.

rust/tests/integration/webrtc_transport.rs
```diff
@@ -161,6 +160,8 @@ fn create_succeeds() {
                         .try_into()
                         .unwrap(),
                     );
+                    // TODO: I wonder how this was not needed before...
```
jmillan, line 163: Nazar? `enable_sctp` must be true and it was not explicit before. I wonder how was it working before. I don't see what I'm missing.

worker/src/Channel/ChannelSocket.cpp
ibc: Please don't use `MS_ERROR`. Use `MS_DUMP`.

worker/src/RTC/DtlsTransport.cpp
ibc: ok

rust/tests/integration/webrtc_transport.rs
```diff
@@ -194,6 +195,29 @@ fn create_succeeds() {
...
+                assert_eq!(ice_candidates[2].r#type, IceCandidateType::Host);
+                assert_eq!(ice_candidates[2].tcp_type, None);
+                assert_eq!(ice_candidates[2].ip, "127.0.0.1".parse::<IpAddr>().unwrap());
+                assert_eq!(ice_candidates[2].protocol, Protocol::Udp);
+                assert_eq!(ice_candidates[2].r#type, IceCandidateType::Host);
+                assert_eq!(ice_candidates[2].tcp_type, None);
+                assert!(ice_candidates[0].priority > ice_candidates[1].priority);
+                assert!(ice_candidates[1].priority > ice_candidates[2].priority);
+
+                // TODO: How comes UDP and TCP where created before?
```
jmillan: We were creating explicitly UDP sockets, how could we be expecting UDP and TCP? cc @nazar-pc.

node/src/tests/test-WebRtcTransport.ts
ibc: In Rust you cannot pass an invalid enum or type, in JS you can.

worker/src/RTC/DtlsTransport.cpp
ibc: ok, next time click here to add a file comment:

Screenshot 2023-09-11 at 12 19 37

ibc: File looks good.

ibc (approved):

node/src/tests/test-WebRtcTransport.ts
jmillan: > In Rust you cannot pass an invalid enum or type, in JS you can.

>> or validates network received data no matter it's created by media soup-client or not.

Such argument applies to rust too.

ibc: In Rust you **cannot** pass a random string.

jmillan: > In Rust you cannot pass a random string.

I'm talking about runtime. We can't control what's coming from the network.

worker/include/RTC/IceServer.hpp
jmillan: Done.

node/src/tests/test-WebRtcTransport.ts
ibc: I don't care what we receive from the network in a Rust app. Whatever we receive from the network we have to convert into a proper `webrtc_transport_options` Rust type **first** before we pass it to `router.create_webrtc_transport()`.

jmillan: This is not being done. I'm removing the `skip` from the test and properly throw when receiving wrong data.

jmillan:
**EDIT**: After merging `flatbuffers` the wrong line appeared and it's now fixed by a new commit.

worker linting is failing, giving this error:

Screenshot 2023-09-12 at 12 00 28

Such line in this branch does NOT look like the one in the red block! Do you have any idea of what can be happening here?

rust/src/data_structures.rs
```diff
@@ -389,6 +477,93 @@ pub enum DtlsFingerprint {
...
+            },
+        }
+    }
+
+    pub(crate) fn from_fbs(fingerprint: &web_rtc_transport::Fingerprint) -> DtlsFingerprint {
+        match fingerprint.algorithm {
+            web_rtc_transport::FingerprintAlgorithm::Sha1 => {
+                let mut value_result = [0_u8; 20];
+
+                hex_as_bytes(fingerprint.value.as_str(), &mut value_result);
```
nazar-pc, line 519: Why is it sent as hex and not binary from worker in the first place?

node/src/WebRtcTransport.ts
```diff
@@ -744,8 +848,10 @@ function parseIceCandidate(binary: FbsWebRtcTransport.IceCandidate): IceCandidat
 		ip         : binary.ip()!,
 		protocol   : parseProtocol(binary.protocol()),
 		port       : binary.port(),
-		type       : 'host',
-		tcpType    : binary.tcpType() === 'passive' ? 'passive' : undefined
+		type       : iceCandidateTypeFromFbs(binary.type()),
+		tcpType    : binary.tcpType() === null ?
```
ibc, line 852: Here we are relying on `null` value. However we want it to be `undefined` when not present in the FBS object. If in the future this feature is accepted (you wrote a PR AFAIR) this is gonna fail. So wouldn't it be better if ew check `!binary.tcpType()` instead?

rust/src/data_structures.rs
```diff
@@ -389,6 +477,93 @@ pub enum DtlsFingerprint {
...
 }
 
+fn hex_as_bytes(input: &str, output: &mut [u8]) {
+    for (i, v) in output.iter_mut().enumerate() {
+        *v = u8::from_str_radix(&input[i * 3..(i * 3) + 2], 16)
+            .map_err(|error| {
+                panic!("Failed to parse value {input} as series of hex bytes: {error}")
+            })
+            .unwrap();
+    }
```
nazar-pc: Generally try to avoid manual indexing in loops in Rust, use higher level operations to get an iterator and zip it with the other iterator to move in lock step. This also typically results in vectorized machine code that is good for performance too.

Something like this:
```suggestion
    for (i, o) in input.splot(':').zip(&mut output) {
        *o = u8::from_str_radix(i, 16)
            .unwrap_or_else(|error| {
                panic!("Failed to parse value {i} as series of hex bytes: {error}")
            });
    }
```

rust/src/data_structures.rs
```diff
@@ -826,6 +1001,30 @@ pub struct DtlsParameters {
     pub fingerprints: Vec<DtlsFingerprint>,
 }
 
+impl DtlsParameters {
+    pub(crate) fn to_fbs(&self) -> web_rtc_transport::DtlsParameters {
+        web_rtc_transport::DtlsParameters {
+            role: Some(self.role.to_fbs()),
```
nazar-pc: Why is `role` optional in flatbuffers?

rust/src/messages.rs
```diff
@@ -863,6 +885,22 @@ impl RouterCreateWebrtcTransportRequest {
             is_data_channel: true,
         }
     }
+
+    pub(crate) fn to_fbs(&self) -> web_rtc_transport::WebRtcTransportOptions {
+        web_rtc_transport::WebRtcTransportOptions {
+            base: Box::new(transport::Options {
+                direct: false,
+                max_message_size: 0,
```
nazar-pc: Why do we have these fields in "base" options that do not make sense to generic transport and set to fake values?

rust/src/router/webrtc_transport.rs
```diff
@@ -851,10 +973,10 @@ impl WebRtcTransport {
         let response = self
             .inner
             .channel
-            .request(self.id(), TransportRestartIceRequest {})
+            .request_fbs(self.id(), TransportRestartIceRequest {})
             .await?;
 
-        Ok(response.ice_parameters)
+        Ok(response)
```
nazar-pc:
```suggestion
        self
            .inner
            .channel
            .request_fbs(self.id(), TransportRestartIceRequest {})
            .await
```

rust/tests/integration/webrtc_transport.rs
nazar-pc: Not sure, was it set in TypeScript before?

rust/tests/integration/webrtc_transport.rs
nazar-pc: IIRC `ListenInfo` is a relatively new thing, apparerntly something was off there? Check git history for this section of the code, I do not remember off top of my head.

rust/src/router/webrtc_transport.rs
```diff
@@ -201,26 +286,27 @@ pub struct WebRtcTransportStat {
...
-    pub probation_bytes_sent: usize,
+    pub probation_bytes_sent: u64,
     pub probation_send_bitrate: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub available_outgoing_bitrate: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub available_incoming_bitrate: Option<u32>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_incoming_bitrate: Option<u32>,
+    pub max_incoming_bitrate: u32,
```
nazar-pc: Why not an option anymore? I thought it was an option that may not be set, in which case it would be `None`.

nazar-pc (commented):
Only looked at Rust changes

rust/src/data_structures.rs
ibc: AFAIR we allow it to be optional so we decide in worker side.

nazar-pc: Not sure what benefit it has because `None` case now has to be handled by both worker and library (we have `unwrap_or` below). It would be easier for everyone if it wasn't optional in the first place.

rust/tests/integration/webrtc_transport.rs
ibc: What do we mean here? Is the test expecting a TCP socket even if only UDP was requested? Is this a bug in Rust create_webrtc_transport? In Rust webrtc_transport constructor? Or just in the test?

nazar-pc: If test was passing before worker must have been creating TCP socket in the past as well. I don't think Rust side was hallucinating things that were not there. Maybe it was fixed with flatbuffers update? IDK.

ibc: I migrated `ListenIps` to `ListenInfos` in Rust entirely (while in Node we still allow `listenIps` for backwards compatibility) so it's imposible that Rust side is creating TCP sockets if not requested by the app.

ibc: Wow, something must be wrong since the test is indeed just creating UDP sockets:

```rust
let transport1 = router
                .create_webrtc_transport({
                    let mut webrtc_transport_options = WebRtcTransportOptions::new(
                        vec![
                            ListenInfo {
                                protocol: Protocol::Udp,
                                ip: IpAddr::V4(Ipv4Addr::LOCALHOST),
                                announced_ip: Some("9.9.9.1".parse().unwrap()),
                                port: None,
                                send_buffer_size: None,
                                recv_buffer_size: None,
                            },
                            ListenInfo {
                                protocol: Protocol::Udp,
                                ip: IpAddr::V4(Ipv4Addr::UNSPECIFIED),
                                announced_ip: Some("9.9.9.2".parse().unwrap()),
                                port: None,
                                send_buffer_size: None,
                                recv_buffer_size: None,
                            },
                            ListenInfo {
                                protocol: Protocol::Udp,
                                ip: IpAddr::V4(Ipv4Addr::LOCALHOST),
                                announced_ip: None,
                                port: None,
                                send_buffer_size: None,
                                recv_buffer_size: None,
                            },
                        ]
                        .try_into()
                        .unwrap(),
                    );
                    webrtc_transport_options.num_sctp_streams = NumSctpStreams {
                        os: 2048,
                        mis: 2048,
                    };
                    webrtc_transport_options.max_sctp_message_size = 1000000;
                    webrtc_transport_options.app_data = AppData::new(CustomAppData { foo: "bar" });

                    webrtc_transport_options
                })
```

ibc: Sorry, now I don't understand. What is wrong @jmillan? I don't see that the test is expecting TCP sockets. Where do you see that?

nazar-pc: Check commented-out code below

ibc: Oh hell, sorry.

ibc: So the test passes in `v3` and `flatbuffers` branch because:

```
#[test]
#[ignore]
```

¯\_(ツ)_/¯

@jmillan I'm trying to fix it in this branch directly.

ibc: Guys, there is nothing to test. The test was disabled in `v3` and `flatbuffers` branches, and in this branch it is enabled, fixed and it passes. That's all.

@jmillan I'll remove the commented-out code and push after confirming everything is ok.

ibc: Pushed. Resolving this thread.

node/src/WebRtcTransport.ts
jmillan: TS flatbuffers does provide `null` rather than `undefined` for optionals.

> So wouldn't it be better if ew check !binary.tcpType() instead?

Nop, because the corresponding enum value can be 0, and thus such check cannot be done as you would get a false negative.

rust/src/data_structures.rs
jmillan: This is probably because the client needs to use it in hex mode, so everyone in the chain, starting from worker, provides it in this form. I wouldn't be totally against having it in binary form in server side in a future.

rust/src/data_structures.rs
jmillan: applied, thanks.

rust/src/router/webrtc_transport.rs
jmillan: It should be an option and be using `std::optional` in C++ in the first place. But it's initialized to 0 instead. We need to review all the places in the worker where `std::optional` make sense.

I'll make this field optional in this PR.

node/src/WebRtcTransport.ts
ibc: I know that TS flatbuffers generates null instead of undefined, but we consider that unfortunate and you wrote a PR changing that, am I wrong?

ibc: > Nop, because the corresponding enum value can be 0, and thus such check cannot be done as you would get a false negative.

Right. What about checking "typeof xxxx === "number"?

Yes, I insist in being ready for the lib to produce undefined instead of null.

jmillan: > Right. What about checking "typeof xxxx === "number"?

I have nothing agains that. We can do this in a separate PR for all optional values.

I made a [feature request](https://github.com/google/flatbuffers/issues/7656), but I'm not planning into doing the PR myself mid term.

ibc: Ah ok, I thought you wrote a PR. We are good then.

rust/src/messages.rs
jmillan: I'm setting this to optional. Both 'direct'  and 'max_message_size'  are used by the base Transport class and only applied to direct transport.

I'm setting it to optional in this PR.

ibc: AFAIR we inspect and use those fields in Transport.cpp so they need to be in the base options of message.

rust/src/data_structures.rs
ibc: I may be wrong but role None is used to indicate that this->role in DtlsTransport.cpp is not yet set (None is its default initial value). However role is optional in arguments given to webRtcTransport.connect() method. Those are different things.

rust/src/messages.rs
jmillan: Done.

rust/src/router/webrtc_transport.rs
jmillan: done.

nazar-pc:
In order to minimize notifications, please just resolve comments, no need to comment "done" on each one. I get way too many notifications of various kind from mediasoup repo. Thanks for understanding :pray:

rust/src/data_structures.rs
jmillan: IMHO it makes sense that the **DtlsParameters** type has non optional DtlsRole, and having it defaulted to 'auto'.

worker/include/RTC/DtlsTransport.hpp
```diff
@@ -2,12 +2,14 @@
...
 
 #include "common.hpp"
+#include "FBS/webRtcTransport_generated.h"
 #include "RTC/SrtpSession.hpp"
 #include "handles/TimerHandle.hpp"
 #include <openssl/bio.h>
 #include <openssl/ssl.h>
 #include <openssl/x509.h>
 #include <absl/container/flat_hash_map.h>
+#include <optional>
```
ibc: Please don't include this. It's already included in `common.hpp`.

ibc: I'll remove it.

ibc: No, I can't due to (explained in Slack).

PR #1156, opened 2023-09-13
Make DTLS fragment stay within MTU size range

### LINKED ISSUE #1100: DTLS fragment exceeds MTU size if the certificate is large.

ibc:
- Fixes #1100
- All credits to @pnts-se and his PR #1143

### Details

- As explained in https://github.com/versatica/mediasoup/issues/1100. OpenSSL doesn't respect MTU size when sending the server certificate.
- In fact, issue #1100 clearly shows how mediasoup is sending a DTLS Server Hello message of 2008 bytes which exceeds the UDP MTU.
- This PR fixes the issue by using `BIO_set_callback_ex` callback as other projects as done.
- This can be tested by adding this dump into `WebRtcTransport.cpp` in `OnDtlsTransportSendData()` method:
   ```c++
   MS_DUMP(">>> DTLS data sent [len:%zu]", len);
   ```

### DIFF
--- a/worker/src/RTC/DtlsTransport.cpp
+++ b/worker/src/RTC/DtlsTransport.cpp
@@ -53,6 +53,31 @@ inline static unsigned int onSslDtlsTimer(SSL* /*ssl*/, unsigned int timerUs)
 		return 2 * timerUs;
 }
 
+inline static long onSslBioOut(
+  BIO* bio,
+  int operationType,
+  const char* argp,
+  size_t len,
+  int /*argi*/,
+  long /*argl*/,
+  int ret,
+  size_t* /*processed*/)
+{
+	long resultOfcallback = (operationType == BIO_CB_RETURN) ? static_cast<long>(ret) : 1;
+
+	if (operationType == BIO_CB_WRITE && argp && len > 0)
+	{
+		MS_DEBUG_DEV("%zu bytes of DTLS data ready to be sent", len);
+
+		RTC::DtlsTransport* dtlsTransport =
+		  reinterpret_cast<RTC::DtlsTransport*>(BIO_get_callback_arg(bio));
+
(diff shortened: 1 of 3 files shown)


jmillan (approved):

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -929,9 +956,8 @@ namespace RTC
 		// Stop the DTLS timer.
 		this->timer->Stop();
 
-		// We need to reset the SSL instance so we need to "shutdown" it, but we
-		// don't want to send a Close Alert to the peer, so just don't call
-		// SendPendingOutgoingDTLSData().
+		// NOTE: We need to reset the SSL instance so we need to "shutdown" it, but we
+		// don't want to send a Close Alert to the peer. However this is gonna happen.
```
pnts-se, line 960: If this is shown to be a problem, perhaps I can write a failing test verifying the behavior. Then do some work to try to avoid sending the Close Alert and pass the test.

ibc: Sure, could you do it once this PR is merged and try that in flatbuffers branch?

NOTE: the problem I see is that we DO WANT to send DTLS Close Alert in most of the cases, and there is just one case in which we do not want to send it (this one). So no idea this can be done. Even if we set a this->allowOutgoingDtlsCloseAlert = true flag before that scenario and set it to false immediately after ignoring such a DTLS Close Alert, the problem is that this is driven by a OpenSSL callback that probably runs asynchronously.

PR #1158, opened 2023-09-15
Rust: port transport notifications to FBS

jmillan:


worker/fbs/notification.fbs
```diff
@@ -53,20 +53,20 @@ union Body {
   FBS.Producer.SendNotification,
   FBS.DataProducer.SendNotification,
   // Notifications from worker.
-  FBS.Transport.TraceNotification,
+  Transport_TraceNotification: FBS.Transport.TraceNotification,
```
ibc, line 56: Why this?

ibc: I mean: why just here and in a very few cases below?

jmillan: Due to [Same table names in different namespaces not supported](https://github.com/planus-org/planus/issues/180).

The solution is creating an alias which adds the namespace to the type.  I'll make a PR that creates an alias for all notification and request bodies, so we are consistent, which in fact results in shorter and more consistent names, taking `FBS` out of the type name.

ibc (approved):

rust/src/router/direct_transport.rs
```diff
@@ -533,25 +525,34 @@ impl DirectTransport {
...
-                }
-            })
+                    notification::Event::DirecttransportRtcp => {
+                        let Ok(Some(notification::BodyRef::RtcpNotification(_body))) =
+                            notification.body()
+                        else {
+                            panic!("Wrong message from worker: {notification:?}");
+                        };
+
+                        // TODO.
```
nazar-pc: TODO what?

rust/src/router/direct_transport.rs
```diff
@@ -533,25 +525,34 @@ impl DirectTransport {
...
+                    }
+                    _ => unimplemented!(),
```
nazar-pc: Hm, what are the other variants?

rust/src/router/pipe_transport.rs
```diff
@@ -585,25 +575,39 @@ impl PipeTransport {
...
+                            panic!("Wrong message from worker: {notification:?}");
+                        };
+
+                        let trace_event_data =
+                            transport::TraceNotification::try_from(body).unwrap();
+                        handlers
+                            .trace
+                            .call_simple(&TransportTraceEventData::from_fbs(trace_event_data));
+                    }
+                    _ => unimplemented!(),
```
nazar-pc: Same question here. List all variants explicitly and have TODO in each of them to make sure things no longer compile when new variant is introduced. Things like this tend to eat compilation errors and cause frustration down the line.

rust/src/router/plain_transport.rs
```diff
@@ -627,35 +610,69 @@ impl PlainTransport {
...
+                            panic!("Wrong message from worker: {notification:?}");
+                        };
+
+                        let trace_event_data =
+                            transport::TraceNotification::try_from(body).unwrap();
+                        handlers
+                            .trace
+                            .call_simple(&TransportTraceEventData::from_fbs(trace_event_data));
+                    }
+                    _ => unimplemented!(),
```
nazar-pc: Same here

rust/src/router/webrtc_transport.rs
```diff
@@ -706,51 +683,84 @@ impl WebRtcTransport {
...
+                            panic!("Wrong message from worker: {notification:?}");
+                        };
+
+                        let trace_event_data =
+                            transport::TraceNotification::try_from(body).unwrap();
+                        handlers
+                            .trace
+                            .call_simple(&TransportTraceEventData::from_fbs(trace_event_data));
+                    }
+                    _ => unimplemented!(),
```
nazar-pc: And here

rust/src/router/direct_transport.rs
jmillan: The same as it was before :-). TODO: Handle the RTCP notification, since we are in the RTCP notification handler.

rust/src/router/direct_transport.rs
jmillan: Notification event types live in the [FBS notification scope `Notification::Event`](https://github.com/versatica/mediasoup/blob/fbs_transport_events/worker/fbs/notification.fbs#L27). Since all notifications for `DirectTransport` are already handled here, receiving any other notification event for `DirectTransort` would be a bug. Would a `panic!()` fit better here?

Ideally notification/request/reponse types should live in their specific scope, but this is what the schema allows us to do. I would be happy to re-review that once after all Rust part has been ported to FBS.

rust/src/router/pipe_transport.rs
jmillan: Same, listing all variants would mean listing variants for all other classes which are unrelated to`PipeTransport`.

rust/src/router/direct_transport.rs
nazar-pc: Worth spelling it out in full then, not obvious at all

rust/src/router/direct_transport.rs
nazar-pc: Hm, I don't like the design where notifications for all transports are located in the same enum, this is error-prone. Previously each transport had their own notifications. We can still share code with enum like this if contents is identical:
```rust
enum DirectTransportNotification {
    Rtcp(RtcpTransportNotification),
}
```

`RtcpTransportNotification` above can be shared between different transports for example.

Also I don't like `Directtransport` prefix, `DirectTransport` would have been nicer. Why did it change?

ibc: > Also I don't like Directtransport prefix, DirectTransport would have been nicer. Why did it change?

This should be a typo in the PR and indeed must be fixed.

jmillan: > RtcpTransportNotification above can be shared between different transports for example.

I'll try the approach we had before.

>> Also I don't like Directtransport prefix, DirectTransport would have been nicer. Why did it change?

> This should be a typo in the PR and indeed must be fixed.

That's auto-generated code. Nothing we can freely choose.

nazar-pc: > That's auto-generated code. Nothing we can freely choose.

I don't believe this, `DirecttransportRtcp` has `Rtcp`, so it capitalizes some things, just not others.

ibc: Auto-generated code from where? How does the source look like? AFAIK flatbuffers converts underscore syntax to camelCase.

jmillan: > I don't believe this, DirecttransportRtcp has Rtcp, so it capitalizes some things, just not others.

`DIRECTTRANSPORT_RTCP` is renamed to `DirecttransortRtcp`. If the event in the schema was `DIRECT_TRANSPORT_RTCP` then we would have `DirectTransportRtcp`, but then in the worker we would be dealing with `DIRECT_TRANSPORT_RTCP`, and we consider DirectTransport a single word.

Reference to the schema: https://github.com/versatica/mediasoup/blob/67ef2f91991d7be9e8fca90ef246b10bf02bd665/worker/fbs/notification.fbs#L27

ibc: IMHO we should adhere to this syntax:

- Use CamelCase when possible.
- When it's not possible (fbs requires lowcase and underscores) use xxx_xxx syntax.
- When forced to write in capital letters, use _ to separate those sub words that otherwise would be in capital letter in the cases above.

This is:

- DirectTransport, WebRtcTransport
- direct_transport, webrtc_transport (exception here I know)
- DIRECT_TRANSPORT, WEBRTC_TRANSPORT (exception here as well).

ibc: But the problem is that we also want to have names like this:

DIRECTTRANSPORT_XXXX

which causes the above discussion...

Ok, my thoughts:

We should focus on what we can control, and forget about auto generated names.

rust/src/router/direct_transport.rs
```diff
@@ -533,19 +564,15 @@ impl DirectTransport {
         let subscription_handler = {
             let handlers = Arc::clone(&handlers);
 
-            channel.subscribe_to_notifications(id.into(), move |notification| {
-                match serde_json::from_slice::<Notification>(notification) {
+            channel.subscribe_to_fbs_notifications(id.into(), move |notification| {
+                match Notification::from_fbs(notification) {
```
jmillan, line 572: @nazar-pc, seems this is the way to go. Definitely cleaner as each class will just care about it's own notifications.

Once confirmed this is the way to go, I'll implement in the rest of transports.

nazar-pc: This side looks fine, btu outer side seems to still have all events in a flat data structure, which is what I wanted to avoid. Basically when/if we add another notification in flatbuffers and C++ I want Rust to stop compiling and tell you to handle another enum variant. Currently this is not the case due to `_ =>` above.

jmillan: That's something that I couldn't cleanly achieve by the time I defined the schemas. We can go back to this matter before merging flatbuffers into v3. I'm adding it as a TODO in the `flatbuffers` PR.

That would affect the schemas, Node, C++ and Rust. I'll keep going with Rust and once we get that, it will only require to change the methods `Notification::from_fbs`.

Do you have any other concern in this PR?

If that's the only concern then I'll do the same with all the remaining transports and keep going.

nazar-pc: That is my biggest concern here

PR #1159, opened 2023-09-15
FBS: normalize FBS types

jmillan:
Use aliases for Message, Request and Notification bodies.

Needed for Rust[*], plus the resulting type names are much more convenient, which remove the redundant 'FBS_' prefix.

[*] (https://github.com/planus-org/planus/issues/180).

NOTE: This PR is based on #1158 and will be merged once the former is accepted and merged.


ibc (approved):
Wow

PR #1160, opened 2023-09-18
FBS: format schemas

jmillan:
Using planus formatter.


ibc:
How does it work? Is it some command that automatically does this? If so, is there also a linter command for FBS?

jmillan:
> How does it work? Is it some command that automatically does this? If so, is there also a linter command for FBS?

This is the exact command I applied:

```sh
for file in `ls`; do planus format $file | tee ${file}.new && mv ${file}.new $file; done
```
 
Before having flatbuffers in V3 this will be automatic and have it's own command, the same way the generation of Rust FBS code will be part of the build process.

ibc (approved):

nazar-pc (approved):

PR #1161, opened 2023-09-18
FBS: type for RtpHeaderExtensionUri

jmillan:


ibc (approved):

jmillan:
I'll merge this PR and the schema formatting one after https://github.com/versatica/mediasoup/pull/1158 is merged since the formers depend on the latest.

I'll continue then porting `transport.produce()` and `transport.consume()`.

PR #1164, opened 2023-09-26
FBS: Producer

jmillan:
Move Rust Producer related communication to FBS. 

UPDATE:
* [Consumer FBS PR](https://github.com/versatica/mediasoup/pull/1165) sources this one.
* subscribe_to_notifications implementation is missing here. It will be done afterwards, once this PR is merged to avoid conflicts with the ongoing Consumer one.


rust/src/rtp_parameters.rs
```diff
@@ -204,11 +204,27 @@ pub enum MediaKind {
...
+    }
+}
+
 /// Error that caused [`MimeType`] parsing error.
 #[derive(Debug, Error, Eq, PartialEq)]
 pub enum ParseMimeTypeError {
-    /// Invalid input string
-    #[error("Invalid input string")]
+    /// Invalid MIME type input string
+    #[error("Invalid MIME type input string")]
```
jmillan, line 227: This helped me during debugging as I did not knw which input string it was referring to.

rust/src/scalability_modes.rs
```diff
@@ -118,7 +118,7 @@ impl Default for ScalabilityMode {
 #[derive(Debug, Error, Eq, PartialEq)]
 pub enum ParseScalabilityModeError {
     /// Invalid input string
-    #[error("Invalid input string")]
+    #[error("Invalid Scalability Mode input string")]
```
jmillan, line 121: same.

rust/src/router/producer.rs
```diff
@@ -179,28 +240,72 @@ pub struct ProducerStat {
...
+            panic!("Wrong message from worker");
+        };
+
+        Self {
+            timestamp: base.timestamp,
+            ssrc: base.ssrc,
+            rtx_ssrc: base.rtx_ssrc,
+            rid: base.rid.clone(),
+            kind: MediaKind::from_fbs(base.kind),
+            mime_type: base.mime_type.to_string().parse().unwrap(),
```
jmillan, line 281: cc @nazar-pc. I see I'm abusing calling to `unwrap()`. I'll review FBS transform methods and probably need to make them return most of the times a Result.

node/src/Producer.ts
```diff
@@ -603,14 +618,84 @@ export class Producer<ProducerAppData extends AppData = AppData>
...
+	switch (type)
+	{
+		case FbsRtpParameters.Type.SIMPLE:
+			return 'simple';
+		case FbsRtpParameters.Type.SIMULCAST:
+			return 'simulcast';
+		case FbsRtpParameters.Type.SVC:
+			return 'svc';
+		default:
+			throw new TypeError(`invalid ProducerType: ${type}`);
```
ibc: It's not an invalid "ProducerType" but an invalid "FbsRtpParameters.Type".

node/src/Producer.ts
```diff
@@ -603,14 +618,84 @@ export class Producer<ProducerAppData extends AppData = AppData>
...
+		case FbsProducer.TraceEventType.FIR:
+			return 'fir';
+		case FbsProducer.TraceEventType.NACK:
+			return 'nack';
+		case FbsProducer.TraceEventType.PLI:
+			return 'pli';
+		case FbsProducer.TraceEventType.RTP:
+			return 'rtp';
+		default:
+			throw new TypeError(`invalid TraceEventType: ${eventType}`);
```
ibc:
```suggestion
			throw new TypeError(`invalid FbsProducer.TraceEventType: ${eventType}`);
```

ibc (approved):
Just cosmetic things.

node/src/Producer.ts
jmillan: As I commented [here](https://github.com/versatica/mediasoup/pull/1164/files#diff-ce1ef95b294b370ff3bc7484654d563c89a6faf7888bc49b819b2cd8443270d3R174) I believe we should have another type `ProducerType` which does not have the PIPE entry. Since a producer will never be of type PIPE.

node/src/Producer.ts
jmillan: I'm removing this line as the switch is already exhausted.

node/src/Producer.ts
ibc: That's irrelevant here. I just meant that what it's invalid is `FbsRtpParameters.Type` (the `type` passed to the function) rather then `ProducerType` (which is the **result** of that function).

jmillan: Transport.Consume() receives a `producerId` and a `type` arguments, among others.

The `type` is the type of the Producer being consumed unless we want to create a **pipe** Consumer. Hence, IMHO rather than having such arguments we should have `producerId` and an optional `pipe` argument. Take the `type` from the Producer (giving it's id) and use it to choose which type of Consumer we need to create. If `pipe` is enabled then we ignore the Producer type and create a PipeConsumer.

This is not a blocker for this PR but I'll create a TODO in flatbuffers PR.

EDIT: Added in the [TODO](https://github.com/versatica/mediasoup/pull/1064)

ibc: Trying to clarify agsin: I just lilterally meant the log text. It said "invalid ProducerType: ${type}". However it was wrong. It should have been "invalid FbsRtpParameters.Type: ${type}" becuase that's what `type` is.

ibc: Regarding pipe stuff: Could we please not enter into that thing now? For v4 I'd like to have a single `Consumer` class that can dynamically handle any number of streams and layers and where "pipe" is just an option to make the Consumer select a single incoming stream or deliver them all to the remote endpoint.

node/src/Producer.ts
```diff
@@ -498,9 +498,24 @@ export class Producer<ProducerAppData extends AppData = AppData>
...
+
+		for (const eventType of types)
+		{
+			try
+			{
+				fbsEventTypes.push(producerTraceEventTypeToFbs(eventType));
+			}
+			catch (error)
+			{
+				// Ignore invalid event types.
```
nazar-pc: Should really print an error here, we don't expect this to ever happen

node/src/Producer.ts
```diff
@@ -498,9 +498,24 @@ export class Producer<ProducerAppData extends AppData = AppData>
...
+			}
+		}
+
 		/* Build Request. */
 		const requestOffset = new FbsProducer.EnableTraceEventRequestT(
-			types
+			fbsEventTypes
```
nazar-pc, line 518: Do we need `fbs` and `Fbs` prefixes in TypeScript? I thought everything was already converted here.

rust/tests/integration/producer.rs
```diff
@@ -636,6 +630,7 @@ fn produce_already_used_mid_ssrc() {
                     parameters.encodings = vec![RtpEncodingParameters {
                         ssrc: Some(22222222),
                         rtx: Some(RtpEncodingParametersRtx { ssrc: 6667 }),
+                        dtx: Some(false),
```
nazar-pc: Why did this need to change from `None` to `Some(false)`?

node/src/Producer.ts
jmillan: I'm enhancing the FBS schemas on the way an creating new types when needed where there were mere strings before. So they need conversion too.

rust/tests/integration/producer.rs
jmillan: Will put it back to None. Option<boolean> already seems a bit rare to begin with.

PR #1165, opened 2023-09-29
FBS: Consumer

jmillan:
Move Rust Consumer related communication to FBS.


jmillan:
@nazar-pc, could you please take a look at the Rust side of this and [Producer PR](https://github.com/versatica/mediasoup/pull/1164) whenever you have the time? Thanks!

This PR takes the former as the base. I did it this way no to get blocked until the former was reviewed, but I don't want to stack more PRs.

nazar-pc:
Will look, there are 3 PRs in my list, sorry for delays

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -109,24 +109,21 @@ namespace RTC
...
-		auto rtpStream = this->rtpStream->FillBuffer(builder);
+		std::vector<flatbuffers::Offset<FBS::RtpStream::Dump>> rtpStreams;
+		rtpStreams.emplace_back(this->rtpStream->FillBuffer(builder));
 
-		auto svcConsumerDump = FBS::Consumer::CreateSvcConsumerDump(
+		auto dump = FBS::Consumer::CreateConsumerDumpDirect(
 		  builder,
 		  base,
-		  rtpStream,
+		  &rtpStreams,
```
ibc, line 118: I like using `std::addressof(rtpStreams)` instead of C style `&rtpStreams`, but it's not critical.

ibc (approved):

node/src/Consumer.ts
```diff
@@ -789,9 +790,24 @@ export class Consumer<ConsumerAppData extends AppData = AppData>
...
+
+		for (const eventType of types)
+		{
+			try
+			{
+				fbsEventTypes.push(consumerTraceEventTypeToFbs(eventType));
+			}
+			catch (error)
+			{
+				// Ignore invalid event types.
```
nazar-pc: Same as in producer PR, I think error should be printed and not swallowed in such cases, might be hard to debug otherwise.

node/src/Consumer.ts
```diff
@@ -1104,27 +1141,27 @@ function parseSimpleConsumerDump(data: FbsConsumer.SimpleConsumerDump): SimpleCo
...
-		currentSpatialLayer    : data.currentSpatialLayer(),
-		preferredTemporalLayer : data.preferredTemporalLayer(),
-		targetTemporalLayer    : data.targetTemporalLayer(),
-		currentTemporalLayer   : data.currentTemporalLayer()
+		preferredSpatialLayer  : data.preferredSpatialLayer()!,
+		targetSpatialLayer     : data.targetSpatialLayer()!,
+		currentSpatialLayer    : data.currentSpatialLayer()!,
+		preferredTemporalLayer : data.preferredTemporalLayer()!,
+		targetTemporalLayer    : data.targetTemporalLayer()!,
+		currentTemporalLayer   : data.currentTemporalLayer()!
```
nazar-pc, line 1159: The number of `!` looks worrying to me, something feels off if we need it so much in this function

rust/src/data_structures.rs
```diff
@@ -1117,6 +1117,8 @@ pub enum OwnedWebRtcMessage {
     EmptyBinary,
 }
 
+/**
+ * TODO. Implement FBS conversion.
```
nazar-pc, line 1121: When do you plan to do this?

rust/src/router/consumer.rs
```diff
@@ -395,6 +589,55 @@ enum Notification {
...
+
+                let trace_notification_fbs = consumer::TraceNotification::try_from(body).unwrap();
+                let trace_notification = ConsumerTraceEventData::from_fbs(trace_notification_fbs);
+
+                Ok(Notification::Trace(trace_notification))
+            }
+            _ => Err(NotificationParseError::InvalidEvent),
+        }
+    }
+}
```
nazar-pc, line 641: New line between structs/impls, please (here and in other places):
```suggestion
}

```

rust/tests/integration/consumer.rs
```diff
@@ -993,18 +985,14 @@ fn dump_succeeds() {
...
                         codec_payload_type: None,
                         rtx: None,
                         max_bitrate: None,
-                        max_framerate: None,
-                        dtx: None,
+                        dtx: Some(false),
                         scalability_mode: ScalabilityMode::None,
-                        spatial_layers: None,
-                        temporal_layers: None,
-                        ksvc: None
```
nazar-pc: Hm, why were all these removed? I recall Producer and consumer data structures were distinct for a reason.

rust/tests/integration/consumer.rs
```diff
@@ -974,7 +966,7 @@ fn dump_succeeds() {
                 vec![RtpEncodingParameters {
                     codec_payload_type: Some(100),
                     rtx: None,
-                    dtx: None,
+                    dtx: Some(false),
```
nazar-pc: Same question as in producer PR: why was this change necessary?

node/src/Consumer.ts
jmillan: For simulcast and SVC we know these have values coming from the worker, this is the reason of using `!` here. For `SimpleConsumer` they are not being populated and we don't even access them.

rust/src/data_structures.rs
jmillan: I've added a TODO item in the flatbuffers PR. It must be done for sure before merging that branch into v3.

rust/tests/integration/consumer.rs
jmillan: It was not necessary and I'll put it back to `None`. So when the value coming from the server is `false` it will be `None` rather than `Some(false)`. I find the field type `Option<bool>` confusing in itself. Doesn't it make more sense for that such member `dtx` to just be `bool` and be defaulted to `false`?

Anyway I'm putting it back to `None`.

rust/tests/integration/consumer.rs
jmillan: Nop, they are indeed the same. Verified. Those removed ones are represented by `scalability_mode`.

rust/tests/integration/consumer.rs
nazar-pc: `None` basically means it wasn't specified, use the default. We can make it just `bool`, but since it was optional I made it `Option`. While here `false` is the default, for some options the default for `None` is `true`.

PR #1168, opened 2023-10-03
Fbs rtp observers

jmillan:
And missing `router.create_pipe_transport`.

### DIFF
--- a/worker/fbs/router.fbs
+++ b/worker/fbs/router.fbs
@@ -46,7 +46,7 @@ table CreateAudioLevelObserverRequest {
 }
 
 table CreateActiveSpeakerObserverRequest {
-    active_speaker_observer_id: string (required);
+    rtp_observer_id: string (required);
     options: FBS.ActiveSpeakerObserver.ActiveSpeakerObserverOptions (required);
 }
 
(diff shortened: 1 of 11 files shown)


worker/fbs/router.fbs
```diff
@@ -46,7 +46,7 @@ table CreateAudioLevelObserverRequest {
 }
 
 table CreateActiveSpeakerObserverRequest {
-    active_speaker_observer_id: string (required);
+    rtp_observer_id: string (required);
```
jmillan, line 49: It should have been `rtp_observer_id` since the beginning.

worker/src/RTC/PipeTransport.cpp
```diff
@@ -168,7 +168,12 @@ namespace RTC
 			}
 
 			tuple = FBS::Transport::CreateTupleDirect(
-			  builder, localIp.c_str(), this->udpSocket->GetLocalPort(), "", 0, FBS::Transport::Protocol::UDP);
+			  builder,
+			  localIp.c_str(),
+			  this->udpSocket->GetLocalPort(),
+			  nullptr,
```
jmillan, line 174: This is a fix here, remote ip is optional and hence must be `nullptr` rather than empty string here.

ibc (approved):

PR #1169, opened 2023-10-06
FBS: DataProducer, DataConsumer

jmillan:
Plus remove code that is not used anymore. Ie:

- `request()` method since `request_fbs()` is used everywhere.
- `notify()` method since `notify_fbs()` is used everywhere.
- All types that were only related to `request()` and `notify()`

Some new TODOs added to flatbuffers branch:
- Remove all temporal FBS suffixes in types and classes.
- Properly test SCTP and enable subchannels.
- Enable remaining tests.
- Test the demo.

### DIFF
--- a/worker/src/RTC/DataConsumer.cpp
+++ b/worker/src/RTC/DataConsumer.cpp
@@ -26,19 +26,20 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		this->typeString = data->type()->str();
+		switch (data->type())
+		{
+			case FBS::DataProducer::Type::SCTP:
+			{
+				this->type = DataConsumer::Type::SCTP;
 
-		if (this->typeString == "sctp")
-		{
-			this->type = DataConsumer::Type::SCTP;
-		}
-		else if (this->typeString == "direct")
-		{
-			this->type = DataConsumer::Type::DIRECT;
-		}
-		else
-		{
-			MS_THROW_TYPE_ERROR("invalid type");
+				break;
+			}
+			case FBS::DataProducer::Type::DIRECT:
+			{
+				this->type = DataConsumer::Type::DIRECT;
+
+				break;
+			}
 		}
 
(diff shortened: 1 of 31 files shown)


ibc:
> notify() method since request() is used everywhere.

What does it mean? Is it a typo?

ibc:
> Properly test SCTP and enable subchannels.

subchannels were tested already and there is a test for them. In flatbuffers branch.

worker/src/RTC/DataConsumer.cpp
```diff
@@ -66,9 +67,12 @@ namespace RTC
 		// paused is set to false by default.
 		this->paused = data->paused();
 
-		for (const auto subchannel : *data->subchannels())
+		if (flatbuffers::IsFieldPresent(data, FBS::Transport::ConsumeDataRequest::VT_SUBCHANNELS))
```
ibc, line 70: This check wasn't needed before. Why is it now?

jmillan: Because `subchannels` is an optional field, and not providing it (which is legitimate for an optional field) would crash. 

[Here](https://github.com/versatica/mediasoup/blob/flatbuffers/node/src/Transport.ts#L1730) you added an empty  array if none was provided, but not providing it, which is legitimate, would crash.

jmillan:
>> Properly test SCTP and enable subchannels.

> subchannels were tested already and there is a test for them. In flatbuffers branch.

"Properly test SCTP and enable subchannels in Rust"

worker/src/RTC/DataConsumer.cpp
ibc: True

ibc (approved):
Good work

jmillan:
I'm merging this so I can move forward. @nazar-pc if you can briefly review I'll take the comments into consideration.

nazar-pc (commented):
Looks fine on the surface

PR #1174, opened 2023-10-11
Enhance flatc build system

### LINKED ISSUE #1154: [flatbuffers] flatc build system is not reliable enough
- `make flatc` doesn't recompile existing `.fbs` files, so `make flatc-force` is needed every time a change is done in any `.rfb.  file.
- If some FBS `table` is renamed, the `FBS` folder keeps the old one forever (a folder with the old name) and it may produce build errors if it references a type/name that no longer exists, so manual delete or `make flatc-force` is needed.

IMHO this is too weak and error prune.

jmillan:
Fixes #1154

Do only generate them when the flatbuffers schemas change by using a proper Meson target.


worker/Makefile
```diff
@@ -209,17 +207,8 @@ libmediasoup-worker: setup flatc
 	$(MESON) install -C $(BUILD_DIR) --no-rebuild --tags libmediasoup-worker
 
 flatc: setup
-	$(MESON) compile -C $(BUILD_DIR) flatc
-	[ -d $(FLATC_OUT_DIR) ] || \
-		$(FLATC) --cpp --cpp-field-case-style lower --reflect-names --scoped-enums \
-			-o $(FLATC_OUT_DIR) ./fbs/*
-
-flatc-clean:
```
ibc, line 217: No need for this anymore?

worker/meson.build
```diff
@@ -277,6 +277,57 @@ executable(
...
+  'fbs/transport.fbs',
+  'fbs/webRtcServer.fbs',
+  'fbs/webRtcTransport.fbs',
+  'fbs/worker.fbs',
+]
+
+flatbuffers_cpp_out_dir = 'FBS'
+flatbuffers_cpp_install_dir = meson.current_source_dir() + '/include'
+
+flatc = find_program('flatc')
```
ibc: Where this this find_program() look for flatc binary? In the PATH? Because it shouldn't.

ibc (commented):
So a question (because this happens to me in the past):

If I delete a source .fbs file because it's no longer needed (or I rename it), will it be also removed from the compiled fbs dir?

worker/meson.build
jmillan: It's not in the path. flatc is a subproject of mediasoup. It finds it there as the subproject installation is a dependency for the target.

worker/Makefile
jmillan: nop, we don't need a specific clear for flatc anymore. the cleanup is done for the regular 'clean'  target as the generated sources reside in the same build directory.

worker/meson.build
ibc: What I would like to hear is that it just uses the binary provided by the flat subproject. If it uses whatever `flatc` in the `PATH` this means problems.

ibc: So can't we just make it use `flatc` provided by the subproject and fail otherwise if not found?

jmillan:
> If I delete a source .fbs file because it's no longer needed (or I rename it), will it be also removed from the compiled fbs dir?

It seems that Meson does not solve that for us. So if we do that, apart from renaming or deleting the corresponding source file name in the files list in meson.build, we should remove the old generated file from both the build and the install dir.

worker/meson.build
jmillan: > So can't we just make it use flatc provided by the subproject and fail otherwise if not found?

Not as per the docs. 

I've pushed a commit where we force the version given by our flatbuffers subproject.

jmillan:
>> If I delete a source .fbs file because it's no longer needed (or I rename it), will it be also removed from the compiled fbs dir?

> It seems that Meson does not solve that for us. So if we do that, apart from renaming or deleting the corresponding source file name in the files list in meson.build, we should remove the old generated file from both the build and the install dir.

I honestly don't see a solution for this rather than go and remove the files from the build and install folders.

jmillan:
@ibc, feel free to give it a try in case you see any possible enhancement here.

jmillan:
> >> If I delete a source .fbs file because it's no longer needed (or I rename it), will it be also removed from the compiled fbs dir?
> 
> > It seems that Meson does not solve that for us. So if we do that, apart from renaming or deleting the corresponding source file name in the files list in meson.build, we should remove the old generated file from both the build and the install dir.
> 
> I honestly don't see a solution for this rather than go and remove the files from the build and install folders.

Fixed now. A `make clean` will make it.

ibc (changes requested):
flatc-clean must be added in .PHONY in Makefile.

ibc:
> > If I delete a source .fbs file because it's no longer needed (or I rename it), will it be also removed from the compiled fbs dir?
> 
> 
> 
> It seems that Meson does not solve that for us. So if we do that, apart from renaming or deleting the corresponding source file name in the files list in meson.build, we should remove the old generated file from both the build and the install dir. 

What are the build and the install dir? I assume the install dir is worker/include/FBS, is it? And not sure what the build dir is but I expect it not to be part of the repo (so gitignored), am I wrong?

jmillan:
> What are the build and the install dir? I assume the install dir is worker/include/FBS, is it? And not sure what the build dir is but I expect it not to be part of the repo (so gitignored), am I wrong?

Build dir is within the build folder: 
* ./out/Release/build/FBS
* ./out/Debug/build/FBS

Install dir is `include/FBS`.

ibc (approved):
Nice

ibc:
So this doesn't work as expected right? https://github.com/versatica/mediasoup/commit/0147ef27240f72aa0c9902ee53b95adaf55d2779

I mean, there is literally no way in any Unix based OS to know which version a binary in the PATH has other than running it first with --version (in case it supports it) and parsing the random output that it generates. Honestly I didn't like that thing in the commit you have reverted. It just couldn't work.

So, can we just force Meson to run a binary included in a subproject and ignore PATH completely?

jmillan:
> So, can we just force Meson to run a binary included in a subproject and ignore PATH completely?

As said, our build depends on the subproject flatbuffers which comes with the flatc binary. Such subproject **forces** future meson `find_binary('flatc')` to return the just generated binary. So we are 100% sure that we will always use our flatc in our subproject.

jmillan:
Summary of this PR, which I consider ready to be merge:

I'm added a `meson.build` file to `worker/fbs` file:
- It generates the corresponding .h files
- Worker build depends on it

As a bonus track, generated `.h` files where sufficed with `_generated` before and I've removed such suffix.

EDIT: header files are not set in `worker/include/` anymore, hence there is no need to make that specific clean. The headers are imported from the folder where they are built, simply.

ibc (approved):

PR #1177, opened 2023-10-16
CI: Use Node.js version 20

jmillan:


nazar-pc (approved):

ibc (approved):
What about the Node version in the Docker file?

ibc (approved):

jmillan:
For some reason CodeQL is complaining about not finding the C++ code..

As per the logs of the action it builds mediasoup correctly.

ibc:
> For some reason CodeQL is complaining about not finding the C++ code..
> 
> As per the logs of the action it builds mediasoup correctly.

I'd just ignore it. It cannot be related to this PR. We can investigate later and even remove it if whatever it complains about is not easy to fix.

jmillan:
> I'd just ignore it. It cannot be related to this PR. We can investigate later and even remove it if whatever it complains about is not easy to fix.

As per the docs, it seems that the code needs to be compiled by the action (it seems weird to me, but..). I've just disabled the prebuilt download and giving it a try.

EDIT: [Docs]( https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/no-source-code-seen-during-build)

jmillan:
> As per the docs, it seems that the code needs to be compiled by the action (it seems weird to me, but..). I've just disabled the prebuilt download and giving it a try.

That fixed it.

Now trying to fix a issue (present in V3 already) about not being able to remove `msys` on Windows.

ibc:
> As per the docs, it seems that the code needs to be compiled by the action (it seems weird to me, but..). I've just disabled the prebuilt download and giving it a try.
> 
> EDIT: Docs

I've read those docs and I don't get it. What does it mean that you have disabled the prebuilt download and give it a try?

jmillan:
> I've read those docs and I don't get it. What does it mean that you have disabled the prebuilt download and give it a try?

The docs say the code needs to be compiled. I'm not trying to answer the question 'how was it working before'.

Screenshot 2023-10-17 at 10 06 43

ibc:
> The docs say the code needs to be compiled. 

Which code? Is this about prebuilt worker binaries or what? Also, we do obviously compile those...

jmillan:
> Which code? Is this about prebuilt worker binaries or what? Also, we do obviously compile those...

The C++ code.... We were not compiling it within the CodeQL workflow because we were downloading the prebuilt binaries instead. Try removing the change I added in the workflow file to see the logs.

.github/workflows/codeql.yml
```diff
@@ -36,6 +36,9 @@ jobs:
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
+    env:
+      MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD: "true"
```
ibc, line 40: Wow I completely missed this!

PR #1182, opened 2023-10-18
Fix IceServer crash when client uses ICE renomination

### LINKED ISSUE #1176: Worker Crash SIGABRT
## Bug Report

**IMPORTANT:** We primarily use GitHub as an issue tracker. Just open an issue here if you have encountered a bug in mediasoup.

If you have questions or doubts about mediasoup or need support, please use the mediasoup Discourse Group instead:

https://mediasoup.discourse.group

If you got a crash in mediasoup, please try to provide a core dump into the issue report:

https://mediasoup.org/support/#crashes-in-mediasoup-get-a-core-dump

I don't have core dumps yet. 

### Your environment

- Operating system: Linux
- Node version: Node v20 + v21-nightly 
- npm version:
- gcc/clang version:
- mediasoup version: 3.12.12
- mediasoup-client version:

### Issue description
2023-10-15T20:33:22.685Z mediasoup:ERROR:Channel Producer Channel error: Error: write EPIPE
2023-10-15T20:33:22.731Z mediasoup:ERROR:PayloadChannel Producer PayloadChannel error: Error: write EPIPE
2023-10-15T20:33:22.751Z mediasoup:ERROR:Worker worker process died unexpectedly [pid:19, code:null, signal:SIGABRT]

ibc:
Fixes #1176

### Details

- Problem was in this PR 756 that added support for ICE renomination: https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination
- Before that PR, we always assumed that if there is a tuple then we are in 'connected' or 'completed' state, so there is a selected ICE tuple.
- That's no longer true when client uses ICE nomination stuff.

### DIFF
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -122,8 +122,8 @@ namespace RTC
 		std::string password;
 		std::string oldUsernameFragment;
 		std::string oldPassword;
-		uint32_t remoteNomination{ 0u };
 		IceState state{ IceState::NEW };
+		uint32_t remoteNomination{ 0u };
 		std::list<RTC::TransportTuple> tuples;
 		RTC::TransportTuple* selectedTuple{ nullptr };
 	};
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -94,7 +94,7 @@ namespace RTC
 		bool IsValidTuple(const RTC::TransportTuple* tuple) const;
 		void RemoveTuple(RTC::TransportTuple* tuple);
 		/**
-		 * This should be just called in 'connected' or completed' state and the
+		 * This should be just called in 'connected' or 'completed' state and the
 		 * given tuple must be an already valid tuple.
 		 */
 		void MayForceSelectedTuple(const RTC::TransportTuple* tuple);
--- a/worker/include/RTC/IceServer.hpp
+++ b/worker/include/RTC/IceServer.hpp
@@ -93,9 +93,11 @@ namespace RTC
 		}
 		bool IsValidTuple(const RTC::TransportTuple* tuple) const;
 		void RemoveTuple(RTC::TransportTuple* tuple);
-		// This should be just called in 'connected' or completed' state
-		// and the given tuple must be an already valid tuple.
-		void ForceSelectedTuple(const RTC::TransportTuple* tuple);
+		/**
+		 * This should be just called in 'connected' or completed' state and the
+		 * given tuple must be an already valid tuple.
+		 */
+		void MayForceSelectedTuple(const RTC::TransportTuple* tuple);
 
 	private:
 		void HandleTuple(
--- a/worker/src/RTC/WebRtcTransport.cpp
+++ b/worker/src/RTC/WebRtcTransport.cpp
@@ -1058,10 +1058,11 @@ namespace RTC
 		}
 
 		// Trick for clients performing aggressive ICE regardless we are ICE-Lite.
-		// NOTE: Only do it if we already have a selected ICE candidate tuple.
-		if (this->iceServer->GetSelectedTuple())
+		if (
+		  this->iceServer->GetState() == RTC::IceServer::IceState::CONNECTED ||
+		  this->iceServer->GetState() == RTC::IceServer::IceState::COMPLETED)
 		{
-			this->iceServer->ForceSelectedTuple(tuple);
+			this->iceServer->MayForceSelectedTuple(tuple);
 		}
 
 		// Check that DTLS status is 'connecting' or 'connected'.
@@ -1146,10 +1147,11 @@ namespace RTC
 		}
 
 		// Trick for clients performing aggressive ICE regardless we are ICE-Lite.
-		// NOTE: Only do it if we already have a selected ICE candidate tuple.
-		if (this->iceServer->GetSelectedTuple())
+		if (
+		  this->iceServer->GetState() == RTC::IceServer::IceState::CONNECTED ||
+		  this->iceServer->GetState() == RTC::IceServer::IceState::COMPLETED)
 		{
-			this->iceServer->ForceSelectedTuple(tuple);
+			this->iceServer->MayForceSelectedTuple(tuple);
 		}
 
 		// Pass the packet to the parent transport.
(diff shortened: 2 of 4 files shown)


ibc:
@penguinol could you please take a look to this?

penguinol:
Maybe we do not need to remove `MS_ASSERT` in `IceState::NEW` and `IceState::DISCONNECTED`, just reset `this->remoteNomination = 0` when state change to `IceState::DISCONNECTED` is enough.

ibc:
> Maybe we do not need to remove `MS_ASSERT` in `IceState::NEW` and `IceState::DISCONNECTED`, just reset `this->remoteNomination = 0` when state change to `IceState::DISCONNECTED` is enough.

I don't think so. In the original code, when HandleTuple() is called, we first assert that tuple is empty and later we check nomination, but the assertion already happened and there were tuples so it aborted the process.

The thing here is that, before introducing renomination stuff, IceServer logic assumed that if there is any tuple then state was connected or completed and there was a selectedTuple. This is no longer true. With renomination it can happen that the first time we call to HandleTuple() the tuple is added to the list of tuples but selectedTuple remains unset and state remains new (or disconnected).

So I have a question. Scenario:
- State is new with no tuples.
- HandleTuple() is called for first time with a candidate with nomination 0.
- In that case the tuple is added to the tuple list but it doesn't become selectedTuple and state remains NEW.

Is this correct? Or should state change to connected and the tuple become the selectedTuple?

worker/include/RTC/IceServer.hpp
```diff
@@ -93,9 +93,11 @@ namespace RTC
 		}
 		bool IsValidTuple(const RTC::TransportTuple* tuple) const;
 		void RemoveTuple(RTC::TransportTuple* tuple);
-		// This should be just called in 'connected' or completed' state
-		// and the given tuple must be an already valid tuple.
-		void ForceSelectedTuple(const RTC::TransportTuple* tuple);
+		/**
+		 * This should be just called in 'connected' or completed' state and the
```
jmillan:
```suggestion
		 * This should be just called in 'connected' or 'completed' state and the
```

worker/src/RTC/WebRtcTransport.cpp
```diff
@@ -1058,7 +1058,12 @@ namespace RTC
 		}
 
 		// Trick for clients performing aggressive ICE regardless we are ICE-Lite.
-		this->iceServer->ForceSelectedTuple(tuple);
+		if (
+		  this->iceServer->GetState() == RTC::IceServer::IceState::CONNECTED ||
+		  this->iceServer->GetState() == RTC::IceServer::IceState::COMPLETED)
+		{
+			this->iceServer->MayForceSelectedTuple(tuple);
+		}
```
jmillan: You can just call `MayForceSelecteTuple()` as it will return if the state is not correct. There is otherwise a duplication of conditions in the caller and the target method.

worker/src/RTC/WebRtcTransport.cpp
```diff
@@ -1142,7 +1147,12 @@ namespace RTC
...
```
jmillan: same as in previous comment.

jmillan (approved):
Looks good :+1:

penguinol:
> > Maybe we do not need to remove `MS_ASSERT` in `IceState::NEW` and `IceState::DISCONNECTED`, just reset `this->remoteNomination = 0` when state change to `IceState::DISCONNECTED` is enough.
> 
> I don't think so. In the original code, when HandleTuple() is called, we first assert that tuple is empty and later we check nomination, but the assertion already happened and there were tuples so it aborted the process.
> 
> The thing here is that, before introducing renomination stuff, IceServer logic assumed that if there is any tuple then state was connected or completed and there was a selectedTuple. This is no longer true. With renomination it can happen that the first time we call to HandleTuple() the tuple is added to the list of tuples but selectedTuple remains unset and state remains new (or disconnected).
> 
> So I have a question. Scenario:
> 
> * State is new with no tuples.
> * HandleTuple() is called for first time with a candidate with nomination 0.
> * In that case the tuple is added to the tuple list but it doesn't become selectedTuple and state remains NEW.
> 
> Is this correct? Or should state change to connected and the tuple become the selectedTuple?

According to https://webrtc.googlesource.com/src/+/refs/heads/main/p2p/base/connection.cc#1037
Nomination will not be 0

worker/src/RTC/WebRtcTransport.cpp
ibc: done

ibc:
> According to https://webrtc.googlesource.com/src/+/refs/heads/main/p2p/base/connection.cc#1037
> Nomination will not be 0

Yes, but we cannot crash if client sends a STUN binding request with nomination 0.

ibc:
@jmillan commented in https://github.com/versatica/mediasoup/pull/756#pullrequestreview-1687085076 about those conditions in `HandleTuple()`. My response here:

Those conditions are basically like this:

```ts
case IceState::NEW:
{
	if (!hasUseCandidate && !hasNomination)
	{
		// Store the tuple.
		auto* storedTuple = AddTuple(tuple);

		// Mark it as selected tuple.
		SetSelectedTuple(storedTuple);

		// Update state.
		this->state = IceState::CONNECTED;

		// Notify the listener.
		this->listener->OnIceServerConnected(this);
	}
	else
	{
		// Store the tuple.
		auto* storedTuple = AddTuple(tuple);

		if ((hasNomination && nomination > this->remoteNomination) || !hasNomination)
		{

			// Mark it as selected tuple.
			SetSelectedTuple(storedTuple);

			// Update state.
			this->state = IceState::COMPLETED;

			// Update nomination.
			if (hasNomination && nomination > this->remoteNomination)
			{
				this->remoteNomination = nomination;
			}

			// Notify the listener.
			this->listener->OnIceServerCompleted(this);
		}
	}

	break;
}
```

@jmillan said:
> The condition previous to this `else` block assures that here `hasUseCandidate` is `true` and `hasNominamtion` is also `true`, so checking that again here is not needed, and makes it harder to read. I had to go to the previous block to make sure this conditions were repeated.

But that's not true:
- The fist condition is `if (!hasUseCandidate && !hasNomination)`.
- So if we are in the `else` block it means that `hasUseCandidate` **OR** `hasNomination` is true (or both).

jmillan:
> The fist condition is if (!hasUseCandidate && !hasNomination).
> So if we are in the else blockit means thathasUseCandidate**OR**hasNomination` is true (or both).

Totally true :+1:

PR #1183, opened 2023-10-19
Fbs RTP trace info

jmillan:
NOTE: We are not typing Node side for trace info and it was not typed before, which could have been. Let's not block the `flatbuffers` branch because it does not type everything even though it was not typed before neither.

### DIFF
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -3,7 +3,7 @@
 #[cfg(test)]
 mod tests;
 
-use crate::fbs::{common, sctp_association, transport, web_rtc_transport};
+use crate::fbs::{common, rtp_packet, sctp_association, transport, web_rtc_transport};
 use serde::de::{MapAccess, Visitor};
 use serde::ser::SerializeStruct;
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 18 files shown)


rust/src/data_structures.rs
```diff
@@ -1133,9 +1131,9 @@ pub struct RtpPacketTraceInfo {
     /// Whether packet contains a key frame.
     pub is_key_frame: bool,
     /// Packet size.
-    pub size: usize,
+    pub size: u64,
```
ibc, line 1134: Sorry isn't it uint64? Or both are valid? Or am I wrong?

worker/include/RTC/RtpPacket.hpp
```diff
@@ -3,10 +3,12 @@
 
 #include "common.hpp"
 #include "Utils.hpp"
+#include "FBS/rtpPacket.h"
 #include "RTC/Codecs/PayloadDescriptorHandler.hpp"
 #include "RTC/RtcLogger.hpp"
 #include <absl/container/flat_hash_map.h>
 #include <array>
+#include <flatbuffers/flatbuffers.h>
```
ibc: Can we not place this in the middle of std standard includes?

rust/src/data_structures.rs
ibc: Hello? Hola?

ibc: Ok this is not FBS but Rust.

ibc (approved):

PR #1185, opened 2023-10-19
[flatbuffers] Fix missing options in router.createWebRtcTransport()

### LINKED ISSUE #1184: [flatbuffers] enableUdp/Tcp and forceUdp/Tcp in createWebRtcTransport() are completely missing
## Bug Report

- https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/Router.cpp#L281C5-L282C82
- https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/WebRtcServer.cpp#L306

Those things are missing in flatbuffers branch so passing `enableUdp: false` (for example) to `router.createWebRtcTransport()` is completely ignored.

ibc:
Fixes #1184

### Details

- No idea how, but flatbuffers branch misses `enableUdp`, `enableTcp`, and `preferUdp` and `preferTcp` options when creating a `WebRtcTransport` with a `WebRtcServer`. This was implemented in v3.

### DIFF
--- a/rust/src/messages.rs
+++ b/rust/src/messages.rs
@@ -651,6 +651,10 @@ pub(crate) struct RouterCreateWebrtcTransportData {
     #[serde(flatten)]
     listen: RouterCreateWebrtcTransportListen,
     initial_available_outgoing_bitrate: u32,
+    enable_udp: bool,
+    enable_tcp: bool,
+    prefer_udp: bool,
+    prefer_tcp: bool,
     enable_sctp: bool,
     num_sctp_streams: NumSctpStreams,
     max_sctp_message_size: u32,
@@ -679,6 +683,10 @@ impl RouterCreateWebrtcTransportData {
             },
             initial_available_outgoing_bitrate: webrtc_transport_options
                 .initial_available_outgoing_bitrate,
+            enable_udp: webrtc_transport_options.enable_udp,
+            enable_tcp: webrtc_transport_options.enable_tcp,
+            prefer_udp: webrtc_transport_options.prefer_udp,
+            prefer_tcp: webrtc_transport_options.prefer_tcp,
             enable_sctp: webrtc_transport_options.enable_sctp,
             num_sctp_streams: webrtc_transport_options.num_sctp_streams,
             max_sctp_message_size: webrtc_transport_options.max_sctp_message_size,
@@ -700,6 +708,10 @@ impl RouterCreateWebrtcTransportData {
                 is_data_channel: true,
             }),
             listen: self.listen.to_fbs(),
+            enable_udp: self.enable_udp,
+            enable_tcp: self.enable_tcp,
+            prefer_udp: self.prefer_udp,
+            prefer_tcp: self.prefer_tcp,
         }
     }
 }
--- a/rust/src/router/webrtc_transport.rs
+++ b/rust/src/router/webrtc_transport.rs
@@ -117,6 +117,18 @@ pub struct WebRtcTransportOptions {
     /// Initial available outgoing bitrate (in bps).
     /// Default 600000.
     pub initial_available_outgoing_bitrate: u32,
+    /// Enable UDP.
+    /// Default true.
+    pub enable_udp: bool,
+    /// Enable TCP.
+    /// Default true if webrtc_server is given, false otherwise.
+    pub enable_tcp: bool,
+    /// Prefer UDP.
+    /// Default false.
+    pub prefer_udp: bool,
+    /// Prefer TCP.
+    /// Default false.
+    pub prefer_tcp: bool,
     /// Create a SCTP association.
     /// Default false.
     pub enable_sctp: bool,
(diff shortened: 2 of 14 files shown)


jmillan (approved):
> I'm extremely worried that something else may have been lost, specially in WebRtcServer cpp and hpp files.

It was you in [this commit](https://github.com/versatica/mediasoup/commit/148090a3aeb06f1f9513f94997d2ed3cd36fb2a5). One of those in which most of changes are unrelated to the initial intent :-).

ibc:
Can you please recheck that commit link? It doesn't work for me.

jmillan:
> Can you please recheck that commit link? It doesn't work for me.

Edited the link.

ibc:
@nazar-pc @jmillan I desperately need help with this: https://github.com/versatica/mediasoup/pull/1185/commits/1e46e3507a50ce8347fae76c6315e28c19a71b5c

nazar-pc:
We have to/from_fbs methods, have you updated them all?

ibc (commented):
Here I have updated the to_fbs method:

https://github.com/versatica/mediasoup/pull/1185#discussion_r1365895788

rust/src/messages.rs
```diff
@@ -700,6 +708,10 @@ impl RouterCreateWebrtcTransportData {
                 is_data_channel: true,
             }),
             listen: self.listen.to_fbs(),
+            enable_udp: self.enable_udp,
```
ibc, line 711: Here I have updated the to_fbs method.

rust/src/router/webrtc_transport.rs
```diff
@@ -152,6 +168,10 @@ impl WebRtcTransportOptions {
         Self {
             listen: WebRtcTransportListen::Server { webrtc_server },
             initial_available_outgoing_bitrate: 600_000,
+            enable_udp: true,
```
ibc, line 171: If I change this to enable_udp: false then yes, it reaches the worker and the test fails.

nazar-pc: These are different methods. Test calls `::new()`, this is `::new_with_server()`, they are different things.

ibc: There is no `new_with_server` method anywhere in `rust/` :-|

ibc: Ah, I see, sorry.

nazar-pc: Maybe we are calling the wrong thing in tests :man_shrugging:

ibc: No. for some reason I assume that that test was using webrtc_server. It's not.

ibc:
Rust done and test modified to use the new stuff. PR ready.

@nazar-pc can you please press the "Approve" button?

rust/src/router/webrtc_transport/tests.rs
```diff
@@ -124,18 +130,13 @@ fn create_with_webrtc_server_succeeds() {
...
         {
             let ice_candidates = transport.ice_candidates();
-            assert_eq!(ice_candidates.len(), 2);
+            assert_eq!(ice_candidates.len(), 1);
             assert_eq!(ice_candidates[0].ip, IpAddr::V4(Ipv4Addr::LOCALHOST));
-            assert_eq!(ice_candidates[0].protocol, Protocol::Udp);
-            assert_eq!(ice_candidates[0].port, port1);
+            assert_eq!(ice_candidates[0].protocol, Protocol::Tcp);
+            assert_eq!(ice_candidates[0].port, port2);
             assert_eq!(ice_candidates[0].r#type, IceCandidateType::Host);
```
nazar-pc, line 136: We had UDP and TCP before, why removing UDP?

ibc: Because I am testing `enable_udp: false` as I said. Just a few lines above:

```rust
let transport = router
            .create_webrtc_transport({
                let mut webrtc_transport_options = WebRtcTransportOptions::new_with_server(
                    webrtc_server.clone()
                );
                // Let's disable UDP so resulting ICE candidates should only contain TCP.
                webrtc_transport_options.enable_udp = false;

                webrtc_transport_options
            })
            .await
            .expect("Failed to create WebRTC transport");
```

nazar-pc: Why is TCP testing no longer necessary here then?

ibc: - That test creates a `WebRtcServer` that listens in UDP and TCP.
- Then it creates a transport with `webrtc_server` and `enable_udp: false`, meaning that the resulting transport should only use the TCP ICE candidates of the given `webrtc_server`.
- So the test asserts that there is only a ICE candidate (`assert_eq!(ice_candidates.len(), 1);`)  and it's TCP.

nazar-pc: I understand what test is doing with these changes, but I don't understand what was wrong with the old test that you needed to change it.

Should you maybe add another test case instead? Also why no such change in TS, did TS test contain different conditions? (tests are usually symmetrical).

ibc: Fair. I've done same change in Node test: https://github.com/versatica/mediasoup/pull/1185/commits/cf0b0332438f73716ace87cc06ad6eb65db9668a

> Should you maybe add another test case instead?

There is little value in just asserting that a transport created with `webRtcServer` has the same ICE candidates than `webRtcServer`. So IMHO in those tests we can also test those `enable_udp` etc fields.

nazar-pc: I mean if it should do that, why not test it? :slightly_smiling_face:

nazar-pc (approved):

rust/src/router/webrtc_transport/tests.rs
ibc: Because:

- Jose and me have lot of work these days. I'm exhausted working on 3 different mediasoup PRs in parallel and doing things at work.
- The current API sucks. It got way worse after adding `webRtcServer` since then we allow `listenIps` OR `webRtcServer` but not both, and some other options only make sense for one of them, etc etc. I know it and I'd love to refactor it in v4, but we have to keep backwards compatibility.

But all those are excuses and you are right.

Issue #1179, opened 2023-10-18
npm postinstall fails in Windows: rd /s /q worker\out\msys => The system cannot find the file specified


ibc:
## Bug Report

This is a recent error happening in Windows CI: https://github.com/versatica/mediasoup/actions/runs/6536142903/job/17747178461?pr=1178

```
npm-scripts [INFO] [postinstall] executeCmd(): rd /s /q worker\out\msys
The system cannot find the file specified.
npm-scripts [ERROR] [postinstall] executeCmd() failed, exiting: Error: Command failed: rd /s /q worker\out\msys
```

Nothing has changed so this is related to Windows. No idea. We need to make it work reliably in all supported Windows versions.

Somehow the issue doesn't make sense. When installing mediasoup on Windows (and forcing it to compile the worker instead of fetching a prebuilt worker), this is the flow:

1. `MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD=true npm install mediasoup --foreground-scripts` (well, whichever syntax is needed in Windows to set environment variables).
2. The npm 'postinstall' task will run the 'postinstall' section in `npm-scripts.mjs`.
3. It will first fun `buildWorker()`:
  ```ts
  function buildWorker()
  {
	logInfo('buildWorker()');

	if (IS_WINDOWS)
	{
		if (!fs.existsSync('worker/out/msys/bin/make.exe'))
		{
			installMsysMake();
		}

		const msysPath = `${process.cwd()}\\worker\\out\\msys\\bin`;

		if (!process.env.PATH.includes(msysPath))
		{
			process.env.PATH = `${msysPath};${process.env.PATH}`;
		}
	}

	executeCmd(`${MAKE} -C worker`);
  }
  ```
4. Note that if `worker/out/msys/bin/make.exe` doesn't exist then it calls `installMsysMake()` which does this:
  ```ts
  function installMsysMake()
  {
	logInfo('installMsysMake()');

	let res = spawnSync('where', [ 'python3.exe' ]);

	if (res.status !== 0)
	{
		res = spawnSync('where', [ 'python.exe' ]);

		if (res.status !== 0)
		{
			logError('`installMsysMake() | cannot find Python executable');

			exitWithError();
		}
	}

	executeCmd(`${String(res.stdout).trim()} worker\\scripts\\getmake.py`);
  }
  ```
5. So here it ensures that we have `python3.exe` and then runs `worker/scripts/getmake.py` which does this:
  ```py
[...]

PR #1186, opened 2023-10-19
Use given PYTHON env variable (if given) when running worker/scripts/getmake.py

ibc:
Just that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### NEXT
 
 * CI: Use Node.js version 20 ([PR #1177](https://github.com/versatica/mediasoup/pull/1177)).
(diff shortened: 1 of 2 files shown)


ibc:
@nazar-pc I assume we need this also for Rust (although I don't see that Rust code calls `getmake.py` at all).

jmillan (approved):

jmillan (approved):

nazar-pc (commented):
When is this useful?

ibc:
> When is this useful?

When you are desperate asking workmates to test the installation of mediasoup in Windows to debug https://github.com/versatica/mediasoup/issues/1179 and they need to pass a specific python path because the `python` they have in the PATH is a link to somewhere else and somehow in Windows links do not work as nicely as in *nix.

PR #1187, opened 2023-10-20
Fix NPM 'postinstall' task in Windows

### LINKED ISSUE #1179: npm postinstall fails in Windows: rd /s /q worker\out\msys => The system cannot find the file specified

ibc:
Fixes #1179

Provide `getmake.py` with the absolute path in which it must place fetched stuff.

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -445,7 +445,9 @@ function installMsysMake()
 		pythonPath = String(res.stdout).trim();
 	}
 
-	executeCmd(`${pythonPath} worker\\scripts\\getmake.py`);
+	const dir = path.resolve('worker/out/msys');
+
+	executeCmd(`${pythonPath} worker\\scripts\\getmake.py --dir="${dir}"`);
 }
 
 function ensureDir(dir)
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -1,4 +1,5 @@
 use std::env;
+#[cfg(target_os = "windows")]
 use std::path::Path;
 use std::process::Command;
 
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -19,8 +19,10 @@ fn main() {
     // Force forward slashes on Windows too so that is plays well with our dumb `Makefile`.
     let mediasoup_out_dir = format!("{}/out", out_dir.replace('\\', "/"));
 
-    // Store original PATH so we make `make clean-all` use it, otherwise
-    // our modified PATH may include rm.exe in Windows and delete itself.
+    // Store original PATH so we make `make clean-all` use it. This is because, in Windows,
+    // we may need to fetch `make` and we store it in out/msys and then we add out/msys/bin
+    // to the PATH, and that folder may contain rm.exe, so `make clean-all` would use
+    // that rm.exe and delete itself and make the task fail.
     let original_path = env::var("PATH").unwrap();
 
     // Add C++ std lib
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -1,6 +1,4 @@
 use std::env;
-#[cfg(target_os = "windows")]
-use std::path::Path;
 use std::process::Command;
 
 fn main() {
@@ -94,7 +92,7 @@ fn main() {
             };
 
             let worker_abs_path = env::current_dir().unwrap();
-            let dir = Path::new(&worker_abs_path).join("out/msys");
+            let dir = worker_abs_path.join("out/msys");
 
             if !Command::new(python)
                 .arg("scripts\\getmake.py")
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -16,7 +16,7 @@ fn main() {
     };
 
     let out_dir = env::var("OUT_DIR").unwrap();
-    // Force forward slashes on Windows too so that is plays well with our dumb `Makefile`
+    // Force forward slashes on Windows too so that is plays well with our dumb `Makefile`.
     let mediasoup_out_dir = format!("{}/out", out_dir.replace('\\', "/"));
 
     // Add C++ std lib
@@ -91,8 +91,7 @@ fn main() {
                 "python".to_string()
             };
 
-            let worker_abs_path = env::current_dir().unwrap();
-            let dir = worker_abs_path.join("out/msys");
+            let dir = format!("{}/msys", mediasoup_out_dir.replace('\\', "/"));
 
             if !Command::new(python)
                 .arg("scripts\\getmake.py")
@@ -104,6 +103,11 @@ fn main() {
             {
                 panic!("Failed to install MSYS/make")
             }
+
+            env::set_var(
+                "PATH",
+                format!("{}\\bin;{}", dir, env::var("PATH").unwrap()),
+            );
         }
 
         env::set_var(
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -19,6 +19,10 @@ fn main() {
     // Force forward slashes on Windows too so that is plays well with our dumb `Makefile`.
     let mediasoup_out_dir = format!("{}/out", out_dir.replace('\\', "/"));
 
+    // Store original PATH so we make `make clean-all` use it, otherwise
+    // our modified PATH may include rm.exe in Windows and delete itself.
+    let original_path = env::var("PATH").unwrap();
+
     // Add C++ std lib
     #[cfg(target_os = "linux")]
     {
@@ -172,6 +176,7 @@ fn main() {
         // Clean
         if !Command::new("make")
             .arg("clean-all")
+            .env("PATH", original_path)
             .env("MEDIASOUP_OUT_DIR", &mediasoup_out_dir)
             .spawn()
             .expect("Failed to start")
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -79,7 +79,7 @@ fn main() {
     {
         if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
             let python = if env::var("PYTHON").is_ok() {
-                env::var("PYTHON").unwrap().to_string()
+                env::var("PYTHON").unwrap()
             } else if Command::new("where")
                 .arg("python3")
                 .status()
@@ -92,10 +92,10 @@ fn main() {
             };
 
             let worker_abs_path = env::current_dir().unwrap();
-            let dir = format!("{}\\out\\msys", worker_abs_path.display());
+            let dir = Path::new(&worker_abs_path).join("out/msys");
 
             if !Command::new(python)
-                .arg("scripts\\getmake.py --dir {}", dir)
+                .arg("scripts\\getmake.py --dir {}", dir.display())
                 .status()
                 .expect("Failed to start")
                 .success()
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::path::Path;
 use std::process::Command;
 
 fn main() {
@@ -95,7 +96,9 @@ fn main() {
             let dir = Path::new(&worker_abs_path).join("out/msys");
 
             if !Command::new(python)
-                .arg("scripts\\getmake.py --dir {}", dir.display())
+                .arg("scripts\\getmake.py")
+                .arg("--dir")
+                .arg(dir)
                 .status()
                 .expect("Failed to start")
                 .success()
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -96,7 +96,7 @@ fn main() {
             if !Command::new(python)
                 .arg("scripts\\getmake.py")
                 .arg("--dir")
-                .arg(dir)
+                .arg(dir.clone())
                 .status()
                 .expect("Failed to start")
                 .success()
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -78,19 +78,24 @@ fn main() {
     #[cfg(target_os = "windows")]
     {
         if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
-            let python = if Command::new("where")
-                .arg("python3.exe")
+            let python = if env::var("PYTHON").is_ok() {
+                env::var("PYTHON").unwrap().to_string()
+            } else if Command::new("where")
+                .arg("python3")
                 .status()
                 .expect("Failed to start")
                 .success()
             {
-                "python3"
+                "python3".to_string()
             } else {
-                "python"
+                "python".to_string()
             };
 
+            let worker_abs_path = env::current_dir().unwrap();
+            let dir = format!("{}\\out\\msys", worker_abs_path.display());
+
             if !Command::new(python)
-                .arg("scripts\\getmake.py")
+                .arg("scripts\\getmake.py --dir {}", dir)
                 .status()
                 .expect("Failed to start")
                 .success()
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -80,8 +80,8 @@ fn main() {
     #[cfg(target_os = "windows")]
     {
         if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
-            let python = if env::var("PYTHON").is_ok() {
-                env::var("PYTHON").unwrap()
+            let python = if let Ok(python) = env::var("PYTHON") {
+                python
             } else if Command::new("where")
                 .arg("python3")
                 .status()
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -97,7 +97,7 @@ fn main() {
                 "python".to_string()
             };
 
-            let dir = format!("{}/msys", mediasoup_out_dir.replace("\\", "/"));
+            let dir = format!("{}/msys", mediasoup_out_dir.replace('\\', "/"));
 
             if !Command::new(python)
                 .arg("scripts\\getmake.py")
--- a/worker/build.rs
+++ b/worker/build.rs
@@ -97,7 +97,7 @@ fn main() {
                 "python".to_string()
             };
 
-            let dir = format!("{}/msys", mediasoup_out_dir.replace('\\', "/"));
+            let dir = format!("{}/msys", mediasoup_out_dir.replace("\\", "/"));
 
             if !Command::new(python)
                 .arg("scripts\\getmake.py")
--- a/worker/scripts/getmake.py
+++ b/worker/scripts/getmake.py
@@ -1,10 +1,21 @@
-import io, hashlib, tarfile, urllib.request
+import argparse, io, hashlib, tarfile, urllib.request
+
+argParser = argparse.ArgumentParser()
+
+argParser.add_argument(
+  '--dir',
+  type=str,
+  required=True,
+  help='absolute path of the directoy in which fetched content will be placed'
+)
+
+args = argParser.parse_args()
 
 def get(url, digest):
   data = urllib.request.urlopen(url).read()
   assert hashlib.sha256(data).hexdigest() == digest
   tar = tarfile.open(fileobj=io.BytesIO(data))
-  tar.extractall('worker/out/msys')
+  tar.extractall(args.dir)
   tar.close()
 
 get('https://sourceforge.net/projects/mingw/files/MSYS/Base/msys-core/msys-1.0.19-1/msysCORE-1.0.19-1-msys-1.0.19-bin.tar.xz/download', '8c4157d739a460f85563bc4451e9f1bbd42b13c4f63770d43b9f45a781f07858')


ibc:
Let's see if CI passes.

ibc:
@nazar-pc please, tell me where mediasoup Rust installation calls the `getmake.py` script. It obviously does so now it fails since it must pass a mandatory `--dir` arg. However there is no "getmake" in `rust/` directory...

ibc:
> @nazar-pc please, tell me where mediasoup Rust installation calls the `getmake.py` script. It obviously does so now it fails since it must pass a mandatory `--dir` arg. However there is no "getmake" in `rust/` directory...

Sorry, found it. In `worker/build.rs`.

ibc:
@nazar-pc may you help here please? Here in `worker/build.rs`:

```rust
if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
            let python = if Command::new("where")
                .arg("python3.exe")
                .status()
                .expect("Failed to start")
                .success()
            {
                "python3"
            } else {
                "python"
            };

            if !Command::new(python)
                .arg("scripts\\getmake.py")
                .status()
                .expect("Failed to start")
                .success()
            {
                panic!("Failed to install MSYS/make")
            }
        }
```

1. I need to make it check if `PYTHON` env is given so it will use it instead of trying `where`. Is it ok?
  ```rust
            let python = if env::var("PYTHON") {
                env::var("PYTHON")
            } else if Command::new("where")
                .arg("python3.exe")
                .status()
                .expect("Failed to start")
                .success()
            {
                "python3"
            } else {
                "python"
            };
  ```

2. I need to call `getmake.py` with `--dst ABSOLUTE_PATH_TO_worker\\out\\msys`. We are in `worker/build.rs` so current folder is `/xxx/mediasoup/worker` but current os working dir is `/xxx/mediasoup` (not sure about this). How can I get current absolute path to `worker` directory in Rust so I can append `\\out\\msys`? I couldn't find any nice way to go.

BTW this PR is making Node+Windows CI pass :) https://github.com/versatica/mediasoup/actions/runs/6585288341/job/17891419544?pr=1187

nazar-pc:
1. [`env::var`](https://doc.rust-lang.org/stable/std/env/fn.var.html) gives you `Result`, you can use `.unwrap_or_else(|| where_stuff)` to get path or use `match`, depending on what looks better
2. https://doc.rust-lang.org/stable/std/fs/fn.canonicalize.html or https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.canonicalize depending on which type you have, https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.join is used to combine paths. Rust docs are pretty good and have search at the top.

ibc:
> 1. [`env::var`](https://doc.rust-lang.org/stable/std/env/fn.var.html) gives you `Result`, you can use `.unwrap_or_else(|| where_stuff)` to get path or use `match`, depending on what looks better

What do I need that for? I mean, I assume that `env::var("PYTHON")` returns a string or None so the condition below is enough, am I wrong?

```rust
          let python = if env::var("PYTHON") {
              env::var("PYTHON")
          } else if Command::new("where")
```

ibc:
> > 1. [`env::var`](https://doc.rust-lang.org/stable/std/env/fn.var.html) gives you `Result`, you can use `.unwrap_or_else(|| where_stuff)` to get path or use `match`, depending on what looks better
> 
> What do I need that for? I mean, I assume that `env::var("PYTHON")` returns a string or None so the condition below is enough, am I wrong?
> 
> ```rust
>           let python = if env::var("PYTHON") {
>               env::var("PYTHON")
>           } else if Command::new("where")
> ```

No :) Solving it.

nazar-pc:
> returns a string or None

No, it returns `Result`, which is either `Ok` or `Err` (explicit form is `Result::Ok` and `Result::Err`, it is an enum).

jmillan:
> Hell, why is this failing? https://github.com/versatica/mediasoup/actions/runs/6586227674/job/17894163137?pr=1187

It's complaining about no permissions error:

`PermissionError: [Errno 13] Permission denied: 'D:\\a\\mediasoup\\mediasoup\\worker\\out\\msys\\bin\\bashbug'`

ibc:
> > Hell, why is this failing? https://github.com/versatica/mediasoup/actions/runs/6586227674/job/17894163137?pr=1187
> 
> It's complaining about no permissions error:
> 
> `PermissionError: [Errno 13] Permission denied: 'D:\\a\\mediasoup\\mediasoup\\worker\\out\\msys\\bin\\bashbug'`

There is no "bashbug" in the whole mediasoup repo.


```
error: failed to run custom build command for `mediasoup-sys v0.5.4 (D:\a\mediasoup\mediasoup\worker)`

Caused by:
  process didn't exit successfully: `D:\a\mediasoup\mediasoup\target\debug\build\mediasoup-sys-609a6396dde901ad\build-script-build` (exit code: 101)
  --- stdout
  C:\hostedtoolcache\windows\Python\3.9.13\x64\python3.exe

  --- stderr
  Traceback (most recent call last):
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 27, in <module>
      get('https://sourceforge.net/projects/mingw/files/MSYS/Base/bash/bash-3.1.23-1/bash-3.1.23-1-msys-1.0.18-bin.tar.xz/download', '38da5419969ab883058a96322bb0f51434dd4e9f71de09cd4f75b96750944533')
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 18, in get
      tar.extractall(args.dir)
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\tarfile.py", line 2045, in extractall
      self.extract(tarinfo, path, set_attrs=not tarinfo.isdir(),
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\tarfile.py", line 2086, in extract
      self._extract_member(tarinfo, os.path.join(path, tarinfo.name),
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\tarfile.py", line 2159, in _extract_member
      self.makefile(tarinfo, targetpath)
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\tarfile.py", line 2200, in makefile
      with bltn_open(targetpath, "wb") as target:
  PermissionError: [Errno 13] Permission denied: 'D:\\a\\mediasoup\\mediasoup\\worker\\out\\msys\\bin\\bashbug'
  thread 'main' panicked at 'Failed to install MSYS/make', worker\build.rs:107:17
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
Error: The process 'C:\Users\runneradmin\.cargo\bin\cargo.exe' failed with exit code 101
```

worker/build.rs
```diff
@@ -78,19 +80,26 @@ fn main() {
     #[cfg(target_os = "windows")]
     {
         if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
-            let python = if Command::new("where")
-                .arg("python3.exe")
+            let python = if env::var("PYTHON").is_ok() {
+                env::var("PYTHON").unwrap()
```
nazar-pc: It is anti-pattern to check `.is_ok()` just to `.unwrap()` right after this (clippy should complain about this). Use pattern matching instead:
```rust
let python = if let Ok(python) = env::var("PYTHON") {
    python
} else if {
    // ...
};
```

worker/build.rs
```diff
@@ -78,19 +80,26 @@ fn main() {
...
             {
-                "python3"
+                "python3".to_string()
             } else {
-                "python"
+                "python".to_string()
             };
 
+            let worker_abs_path = env::current_dir().unwrap();
+            let dir = Path::new(&worker_abs_path).join("out/msys");
```
nazar-pc: I'm pretty sure this will also work:
```suggestion
            let dir = worker_abs_path.join("out/msys");
```

worker/build.rs
```diff
@@ -78,19 +80,26 @@ fn main() {
...
                 .success()
             {
-                "python3"
+                "python3".to_string()
             } else {
-                "python"
+                "python".to_string()
             };
 
+            let worker_abs_path = env::current_dir().unwrap();
```
nazar-pc: This is incorrect, Rust crates are not supposed to touch anything anywhere in the file system except a location where they are allowed to modify things defined by `OUT_DIR` environment variable and already stored in `out_dir` variable (also see `mediasoup_out_dir` in case that is what you actually need here).

nazar-pc (changes requested):

worker/build.rs
ibc: What is `worker_abs_path`? Is it a simple string? or is it a `Path` instance? If it's a string I prefer to use `path` utils to properly join them given `/` or `\\` syntax to separate folders in Windows and Unix.

nazar-pc: It is faster to check docs than asking me to reply with a link: https://doc.rust-lang.org/stable/std/env/fn.current_dir.html
Also IDE should ideally tell you that :confused:

worker/build.rs
ibc:
```rust
println!("--- mediasoup_out_dir: {}", mediasoup_out_dir);

// => mediasoup_out_dir: /Users/ibc/src/v3-mediasoup/target/debug/build/mediasoup-sys-11923be8cb8ca4cd/out/out
```

That's far from what I want.

nazar-pc: That is actually precisely what you want

ibc: I don't see that.

```rust
if !std::path::Path::new("worker/out/msys/bin/make.exe").exists() {
    let python = if let Ok(python) = env::var("PYTHON") {
        python
    } else if Command::new("where")
        .arg("python3")
        .status()
        .expect("Failed to start")
        .success()
    {
        "python3".to_string()
    } else {
        "python".to_string()
    };

    let worker_abs_path = env::current_dir().unwrap();
    let dir = worker_abs_path.join("out/msys");

    if !Command::new(python)
        .arg("scripts\\getmake.py")
        .arg("--dir")
        .arg(dir)
        .status()
        .expect("Failed to start")
        .success()
    {
        panic!("Failed to install MSYS/make")
    }
}

env::set_var(
    "PATH",
    format!(
        "{}\\worker\\out\\msys\\bin;{}",
        env::current_dir()
            .unwrap()
            .into_os_string()
            .into_string()
            .unwrap(),
        env::var("PATH").unwrap()
    ),
);
```

- Here we first check `mediasoup/worker/out/msys/bin/make.exe`.
- If it doesn't exist then we run `getmake.py` and we want to store things in `dir` (let's see below what `dir` should be).
- Then we do this:
  ```rust
  env::set_var(
    "PATH",
    format!(
        "{}\\worker\\out\\msys\\bin;{}",
        env::current_dir()
            .unwrap()
            .into_os_string()
            .into_string()
            .unwrap(),
        env::var("PATH").unwrap()
    ),
  );
  ```
- There we are **clearly** adding current dir + \\worker\\out\\msys\\bin to the `PATH`.
- So it's very clear that `dir` given to `getmake.py` must point to current dir + \\worker\\out\\msys as well.

ibc: So everything was wrong before or there is something I miss.

In addition, it looks suspicious to me that `mediasoup_out_dir` in my macOS points to "/Users/ibc/src/v3-mediasoup/target/debug/build/mediasoup-sys-11923be8cb8ca4cd/out/out" (note the double "out" at the end).

nazar-pc: > Here we first check mediasoup/worker/out/msys/bin/make.exe.

This is fine because it might actually exist there in local development due to being installed by Node.js mediasoup

> If it doesn't exist then we run getmake.py and we want to store things in dir (let's see below what dir should be).

I think this is actually a problem, whoever introduced it should not have downloaded files where they did.

> There we are clearly adding current dir + \worker\out\msys\bin to the PATH.

Current directory is fine, but we need to add the correct location within `OUT_DIR` where we download stuff just like we download PIP, compile static library and do other things in `OUT_DIR`.

> So it's very clear that dir given to getmake.py must point to current dir + \worker\out\msys as well.

It is not clear at all, both old and new code is incorrect as far as I'm concerned. In fact, Rust will likely refuse to publish package at all if it modifies anything outside of `OUT_DIR`, crates are supposed to be self-contained.

nazar-pc: > In addition, it looks suspicious to me that mediasoup_out_dir in my macOS points to "/Users/ibc/src/v3-mediasoup/target/debug/build/mediasoup-sys-11923be8cb8ca4cd/out/out" (note the double "out" at the end).

Looks like that, yes. One `out` is from Cargo, second is "ours".

ibc: ok, so I'll **also** add the missing path to `PATH`.

But you know the funniest thing? **I'm trying to fix an issue that NO longer happens in v3 branch**: https://github.com/versatica/mediasoup/pull/1189 (it's basically v3 branch and CI are perfectly passing).

nazar-pc: It might, but probably because CI worked before too due to GitHub Actions containing correct `make`, but when people install mediasoup on their machines, suddenly things break because `make` in `PATH` is nmake or whatever the thing is that is on Windows by default.

ibc: > Looks like that, yes. One out is from Cargo, second is "ours".

But how can it be wrong? I mean, `mediasoup_out_dir` is being used by build.rs to give env `MEDIASOUP_OUT_DIR` to `make libmediasoup-worker`. How is possible that it was wrong all this time? I mean, can I just make `mediasoup_out_dir` be `out_dir`:

```diff
- let mediasoup_out_dir = format!("{}/out", out_dir.replace('\\', "/"));
+ let mediasoup_out_dir = out_dir.replace('\\', "/");
```

and things will work?

nazar-pc: That part isn't wrong, I meant `get_make` and where it places downloaded files wasn't correct.

ibc: This is gonna be fine: https://github.com/versatica/mediasoup/pull/1187/commits/cca0dabc2e00f12f94eb767c853a9612b6843cde

ibc: Rust is amazing. I declare a `let dir` and it later complains because I have "moved" it so I cannot use it in two places and must clone it in one of them.

```
error[E0382]: borrow of moved value: `dir`
   --> worker\build.rs:109:39
    |
94  |             let dir = format!("{}/msys", mediasoup_out_dir.replace('\\', "/"));
    |                 --- move occurs because `dir` has type `std::string::String`, which does not implement the `Copy` trait
...
99  |                 .arg(dir)
    |                      --- value moved here
...
109 |                 format!("{}\\bin;{}", dir, env::var("PATH").unwrap()),
    |                                       ^^^ value borrowed here after move
    |
    = note: this error originates in the macro `$crate::__export::format_args` which comes from the expansion of the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider cloning the value if the performance cost is acceptable
    |
99  |                 .arg(dir.clone())
    |                         ++++++++
```

ibc:
Now this:

https://github.com/versatica/mediasoup/actions/runs/6587295108/job/17897281193?pr=1187

```
[...]

Deleting D:\a\mediasoup\mediasoup\worker\subprojects\packagecache\openssl-3.0.8.tar.gz
  Deleting D:\a\mediasoup\mediasoup\worker\subprojects\packagecache\openssl_3.0.8-1_patch.zip
  Deleting D:\a\mediasoup\mediasoup\worker\subprojects\openssl-3.0.8
  Progress: 8 / 8 ()

rm -f -rf D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out

  --- stderr

  rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/make.exe': Permission denied
  rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-1.0.dll': Permission denied
  rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-iconv-2.dll': Permission denied
  rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-intl-8.dll': Permission denied
  rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/rm.exe': Permission denied
  rm: cannot chdir from `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys' to ..: No such file or directory
  "make": *** [clean-all] Error 1
```

I give up.

nazar-pc:
> I give up.

Looks like you're trying to remove `rm`, same `rm.exe` you are using to remove itself :upside_down_face: 

Try https://doc.rust-lang.org/stable/std/fs/fn.remove_dir_all.html instead

ibc:
> Now this:
> 
> https://github.com/versatica/mediasoup/actions/runs/6587295108/job/17897281193?pr=1187
> 
> ```
> [...]
> 
> Deleting D:\a\mediasoup\mediasoup\worker\subprojects\packagecache\openssl-3.0.8.tar.gz
>   Deleting D:\a\mediasoup\mediasoup\worker\subprojects\packagecache\openssl_3.0.8-1_patch.zip
>   Deleting D:\a\mediasoup\mediasoup\worker\subprojects\openssl-3.0.8
>   Progress: 8 / 8 ()
> 
> rm -f -rf D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out
> 
>   --- stderr
> 
>   rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/make.exe': Permission denied
>   rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-1.0.dll': Permission denied
>   rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-iconv-2.dll': Permission denied
>   rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/msys-intl-8.dll': Permission denied
>   rm: cannot remove `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys/bin/rm.exe': Permission denied
>   rm: cannot chdir from `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys' to ..: No such file or directory
>   "make": *** [clean-all] Error 1
> ```

So `make clean-all` is failing. It does this:

```Makefile
clean-all: clean-subprojects
$(RM) -rf $(MEDIASOUP_OUT_DIR)
```

- It's clear that `clean-subprojects` does work.
- `MEDIASOUP_OUT_DIR' is supposed to be Rust `mediasoup_out_dir` in `build.rs` because it's passed as env to `make`.
- So it tries to delete it: `rm -f -rf D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out`
- How is it possible that it later fails because "rm: cannot chdir from `D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out/msys' to ..: No such file or directory"?
- Why such a syntax with `/` instead of `\\`... oh well, wait. In `build.rs`:
  ```rust
  // Force forward slashes on Windows too so that is plays well with our dumb `Makefile`.
  let mediasoup_out_dir = format!("{}/out", out_dir.replace('\\', "/"));
  ```
- So this is NOT good right? Because we are passing it to our `Makefile`. But anyway the **whole** paths in `Makefile` use`/` Unix syntax instead of `\\` so...
- And anyway, why does it complain about such a "D:/a/mediasoup/mediasoup/target/debug/build/mediasoup-sys-0443c0f553b7fb4e/out/out" path? `make clean-all` task doesn't do that at all!!!

ibc:
> > I give up.
> 
> Looks like you're trying to remove `rm`, same `rm.exe` you are using to remove itself 🙃
> 
> Try https://doc.rust-lang.org/stable/std/fs/fn.remove_dir_all.html instead

WHAT??? Do you mean that the `rm` command used by Rust+Windows is the one being downloaded with `getmake.py`?????

nazar-pc:
> WHAT??? Do you mean that the rm command used by Rust+Windows is the one being downloaded with getmake.py?????

I mean it doesn't download `make`, it downloads the whole thing and once it is in `PATH`... Is there an API to extract a single file instead?

ibc:
> > > I give up.
> > 
> > 
> > Looks like you're trying to remove `rm`, same `rm.exe` you are using to remove itself 🙃
> > Try https://doc.rust-lang.org/stable/std/fs/fn.remove_dir_all.html instead
> 
> WHAT??? Do you mean that the `rm` command used by Rust+Windows is the one being downloaded with `getmake.py`?????

Hell, there is `rm.exe` in `msys/bin/`... so now what?? and how is possible that this didn't happen before? So this is because, indeed, `msys` was NOT being stored under `mediasoup_out_dir`, right??

ibc:
> > > > I give up.
> > > 
> > > 
> > > Looks like you're trying to remove `rm`, same `rm.exe` you are using to remove itself 🙃
> > > Try https://doc.rust-lang.org/stable/std/fs/fn.remove_dir_all.html instead
> > 
> > 
> > WHAT??? Do you mean that the `rm` command used by Rust+Windows is the one being downloaded with `getmake.py`?????
> 
> Hell, there is `rm.exe` in `msys/bin/`... so now what?? and how is possible that this didn't happen before? So this is because, indeed, `msys` was NOT being stored under `mediasoup_out_dir`, right?? So we had a bug that prevented another bug!

ibc:
> > WHAT??? Do you mean that the rm command used by Rust+Windows is the one being downloaded with getmake.py?????
> 
> I mean it doesn't download `make`, it downloads the whole thing and once it is in `PATH`... Is there an API to extract a single file instead?

Yes, I get it. So I have no idea about a proper solution for this other than placing `msys` somewhere else and leaving it there (do not delete it at the end).

nazar-pc:
Maybe using `C:\Windows\system32\rm.exe` or wherever the original is on Windows.

ibc:
> Maybe using `C:\Windows\system32\rm.exe` or wherever the original is on Windows.

So I have to tell Makefile to use `C:\Windows\system32\rm.exe` or whatever only in Windows. And it could be that the user doesn't have C: but D: or whatever...

nazar-pc:
I have a better option: call `make` by its path on Windows instead of modifying `PATH`, that would definitely solve it.

ibc:
> I have a better option: call `make` by its path on Windows instead of modifying `PATH`, that would definitely solve it.

yes

ibc:
@nazar-pc CI is passing!!!!

npm-scripts.mjs
```diff
@@ -445,7 +445,9 @@ function installMsysMake()
 		pythonPath = String(res.stdout).trim();
 	}
 
-	executeCmd(`${pythonPath} worker\\scripts\\getmake.py`);
+	const dir = path.resolve('worker/out/msys');
+
+	executeCmd(`${pythonPath} worker\\scripts\\getmake.py --dir="${dir}"`);
```
jmillan, line 450: cosmetic, this path could aslo be created with `path.resolve()` and avoid the double back slashes.

worker/build.rs
```diff
@@ -78,25 +82,36 @@ fn main() {
...
                 .success()
             {
-                "python3"
+                "python3".to_string()
             } else {
-                "python"
+                "python".to_string()
             };
 
+            let dir = format!("{}/msys", mediasoup_out_dir.replace('\\', "/"));
```
jmillan, line 100:
```suggestion
            let dir = format!("{}/msys", mediasoup_out_dir.replace("\\", "/"));
```

worker/build.rs
```diff
@@ -78,25 +82,36 @@ fn main() {
...
                 .status()
                 .expect("Failed to start")
                 .success()
             {
                 panic!("Failed to install MSYS/make")
             }
+
+            env::set_var(
+                "PATH",
+                format!("{}\\bin;{}", dir, env::var("PATH").unwrap()),
```
jmillan, line 115: Is it OK to use back slash here?

npm-scripts.mjs
ibc: Yes, but we don't follow an specific approach for that in the whole file. I agree we need consistency but I don't want to decide such a consistency within this PR.

worker/build.rs
ibc: Yes, it was already the same some lines below where `env::set_var()` is also called.

worker/build.rs
ibc: Thanks @jmillan !

CleanShot-2023-10-20-at-16 44 04@2x

Reverting.

jmillan (approved):

PR #1189, opened 2023-10-20
FOO DELETE ME

ibc:
Let's see now, Rust + Windows!

This PR is just v3. Let's see if CI also fails in Windows+Rust or what.


nazar-pc:
Why do you even open a PR? CI runs just fine on branches. All of these attempts generate a huge number of notifications. Every comment, every commit, just too many of them.

ibc:
> Why do you even open a PR? CI runs just fine on branches.

Oh that's true...

Issue #1188, opened 2023-10-20
TS: expect() statements inside event handlers are not taking any effect 


jmillan:
Because mediasoup instances use `EnhancedEventEmitter.safeEmit()` which absorbs any exception occurred in the event handler.

This makes that `expect(1).toBe(2);` in a event handler does not fail and makes the test pass.

Example: https://github.com/versatica/mediasoup/blob/v3/node/src/tests/test-DirectTransport.ts#L154

ibc:
We should **only** use `safeEmit()` in internal listeners and never use it in public events. The app is free to add try/catch into its listeners.

NOTE: Same must be done in mediasoup-client and mediasoup-aiortc-client.

NOTE: Let's do this directly in flatbuffers.

jmillan:
> NOTE: Let's do this directly in flatbuffers.

I have just discovered this issue that has always been there. Can we wait for `flatbuffers` to be moved to `v3` and do it then? 

It will be impossible for anyone to look at the PR otherwise. 

I don't find this being so urgent. We can do it as soon as `flatbuffers` is merged.

ibc:
Ok, I was under the impression that you need this done to verify tests in flatbuffers. If not, let's wait.

ibc:
Hell, why is this failing? https://github.com/versatica/mediasoup/actions/runs/6586227674/job/17894163137?pr=1187

```
error: failed to run custom build command for `mediasoup-sys v0.5.4 (D:\a\mediasoup\mediasoup\worker)`

Caused by:
  process didn't exit successfully: `D:\a\mediasoup\mediasoup\target\debug\build\mediasoup-sys-609a6396dde901ad\build-script-build` (exit code: 101)
  --- stdout
  C:\hostedtoolcache\windows\Python\3.9.13\x64\python3.exe

  --- stderr
  Traceback (most recent call last):
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 27, in <module>
      get('https://sourceforge.net/projects/mingw/files/MSYS/Base/bash/bash-3.1.23-1/bash-3.1.23-1-msys-1.0.18-bin.tar.xz/download', '38da5419969ab883058a96322bb0f51434dd4e9f71de09cd4f75b96750944533')
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 18, in get
      tar.extractall(args.dir)
[...]

PR #1190, opened 2023-10-20
C++: Fix DataProducer::Send invalid memory access

jmillan:
'subchannels' are optional and hence they need to be checked before accessing them.

Remove test expectation that assumed we were receiving all messages in incremental mode. TS tests failure was being masked by [1], I don't know why Rust wasn't failing but definitely consumers are not receiving everything sent by the producer since in the tests both are paused at some time.

[1]: https://github.com/versatica/mediasoup/issues/1188

### DIFF
--- a/worker/src/RTC/DataProducer.cpp
+++ b/worker/src/RTC/DataProducer.cpp
@@ -224,11 +224,14 @@ namespace RTC
 
 				std::vector<uint16_t> subchannels;
 
-				subchannels.reserve(body->subchannels()->size());
-
-				for (const auto subchannel : *body->subchannels())
+				if (flatbuffers::IsFieldPresent(body, FBS::DataProducer::SendNotification::VT_SUBCHANNELS))
 				{
-					subchannels.emplace_back(subchannel);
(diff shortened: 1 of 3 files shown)


nazar-pc (approved):

worker/src/RTC/DataProducer.cpp
```diff
@@ -224,11 +224,14 @@ namespace RTC
 
 				std::vector<uint16_t> subchannels;
 
-				subchannels.reserve(body->subchannels()->size());
```
ibc, line 227: Isn't `subchannels` mandatory in FBS?

ibc (approved):

worker/src/RTC/DataProducer.cpp
jmillan: No: https://github.com/versatica/mediasoup/blob/flatbuffers/worker/fbs/dataProducer.fbs#L44

It happened the same with `DataConsumer.send()`, TS creates a default empty array but the type is Optional and hence legit not to add it.

PR #1192, opened 2023-10-24
Rust: add bench for producer creation/dump

jmillan:


jmillan:
@nazar-pc, I've added some types to the prelude as I'm doing things directly using Rust types. Is there anything against that or a better way to do it?

BTW, most part of the bench code is copied from the producer test.

ibc (commented):
Just in Rust? There are built-in benchmark utils in Node.

rust/benches/producer.rs
```diff
@@ -0,0 +1,168 @@
...
+        });
+
+        group.bench_function("create", |b| {
+            b.iter(|| {
+                let _ = futures_lite::future::block_on(async {
+                    transport_1
+                        .produce(audio_producer_options())
+                        .await
+                        .expect("Failed to produce audio")
+                });
```
nazar-pc, line 145: You are not only testing creation here, you also trigger destruction by dropping the value here. Destruction is async though, so I guess the bench is fine for comparison reasons.

nazar-pc (approved):
There is a better way to run async benches, but this will do for now.

nazar-pc:
> @nazar-pc, I've added some types to the prelude as I'm doing things directly using Rust types. Is there anything against that or a better way to do it?

Prelude should contain most popular types. Some more exotic ones have to be imported separarely and explicitly. What is "most popular" is, of course, subjective. In Rust `Vec` doesn't need to be imported explicitly, but `VecDeque` does for example.

jmillan:
> Just in Rust? There are built-in benchmark utils in Node.

Any pointer to those?

jmillan:
> > Just in Rust? There are built-in benchmark utils in Node.
> 
> Any pointer to those?

I see https://nodejs.org/api/perf_hooks.html.

I'll add the same ones we have for Rust.

jmillan:
> Just in Rust? There are built-in benchmark utils in Node.

@ibc, how do you propose to do this?

We have no benches in TS so far. Should it be done within the jest tests? Should we have a `benches` directory and build our own simple benching suite? 

I've just added a bench in Rust which has a built in bench mechanism run by `cargo bench`, but we don't have such thing in TS. If we want to have this in Node we should think about it.

Is this really a blocker for this PR?

ibc:
> > Just in Rust? There are built-in benchmark utils in Node.
> 
> @ibc, how do you propose to do this?

Honestly I just asked about Node :)

> We have no benches in TS so far. Should it be done within the jest tests? Should we have a `benches` directory and build our own simple benching suite?
> 
> I've just added a bench in Rust which has a built in bench mechanism run by `cargo bench`, but we don't have such thing in TS. If we want to have this in Node we should think about it.
> 
> Is this really a blocker for this PR?

No. We should add such a Node benchmark stuff separately after flatbuffers is merged.

ibc (approved):
`cargo clippy` failing.

jmillan:
> Honestly I just asked about Node :)

That was not how I read it :-)

ibc:
> > Honestly I just asked about Node :)
> 
> That was not how I read it :-)

Because I didn't realize that, indeed, we don't have a proper benchmark mechanism/place for mediasoup Node.

PR #1193, opened 2023-10-25
[flatbuffers] Improve Node imports

ibc:
- Add `node:` prefix to all built-in Node imports (replaces PR #1142).
- Replace `uuid` package with built-in `crypto.randomUUID()`(replaces PR #1140).
- Cosmetic changes in TS files:
  - Avoid unnecessary `const { createWorker } = mediasoup` in almost all tests.
  - Move all FBS related imports to the bottom.
  - Some formating fixes that bypassed our ESLint rules.

### DIFF
--- a/node/src/utils.ts
+++ b/node/src/utils.ts
@@ -1,4 +1,4 @@
-import { randomInt } from 'crypto';
+import { randomUUID, randomInt } from 'node:crypto';
 import { ProducerType } from './Producer';
 import { Type as FbsRtpParametersType } from './fbs/rtp-parameters';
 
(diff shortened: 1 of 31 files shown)


jmillan (approved):

node/src/utils.ts
```diff
@@ -26,6 +26,14 @@ export function clone<T>(value: T): T
 	}
 }
 
+/**
+ * Generates a random UUID v4.
+ */
+export function generateUUIDv4(): string
+{
+	return randomUUID();
+}
```
piranna, line 35: Why is that? Why not using directly the `node:crypto` export?

piranna (changes requested):

node/src/utils.ts
ibc: To make it clear that it's a UUID version 4 and to be consistent with other utils such as `generateRandomNumber()` which also uses `crypto` API inside. It's not bad to have wrappers that allow us to change the internal implementation at any time if needed.

piranna: Not my taste of cake, but ok 👍🏻

piranna (approved):

PR #1194, opened 2023-10-25
Rust: FBS do not use heap to process responses

jmillan:
Avoid heap allocations on response processing. Provides 10% of performance improvement[*].

This is a test bed for rewriting the rest of `from_fbs` to `from_fbs_ref`.

NOTE: Once all `from_fbs` are moved to `from_fbs_ref`, rename the methods back to `from_fbs`

[*]: benched producer::dump()

NOTE: I'd like to merge this once approved. Let's have a TODO for making this improvement everywhere once flatbuffers branch is merged. This is totally a non blocker.

NOTE: I'm going to do some checks in Node too. But that's not a blocker for this PR.

### DIFF
--- a/rust/src/ortc.rs
+++ b/rust/src/ortc.rs
@@ -101,8 +101,9 @@ impl RtpMapping {
                         ssrc: mapping?.ssrc()?,
                         scalability_mode: mapping?
                             .scalability_mode()?
-                            .unwrap_or(String::from("S1T1").as_str())
-                            .parse()?,
+                            .map(|maybe_scalability_mode| maybe_scalability_mode.parse())
+                            .transpose()?
+                            .unwrap_or_default(),
                         mapped_ssrc: mapping?.mapped_ssrc()?,
                     })
                 })
--- a/rust/src/ortc.rs
+++ b/rust/src/ortc.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
+use std::error::Error;
 use std::mem;
 use std::num::{NonZeroU32, NonZeroU8};
 use std::ops::Deref;
@@ -77,32 +78,36 @@ impl RtpMapping {
         }
     }
 
-    pub(crate) fn from_fbs(mapping: rtp_parameters::RtpMapping) -> Self {
-        Self {
+    pub(crate) fn from_fbs_ref(
+        mapping: rtp_parameters::RtpMappingRef<'_>,
+    ) -> Result<Self, Box<dyn Error>> {
+        Ok(Self {
             codecs: mapping
-                .codecs
+                .codecs()?
                 .iter()
-                .map(|mapping| RtpMappingCodec {
-                    payload_type: mapping.payload_type,
-                    mapped_payload_type: mapping.mapped_payload_type,
+                .map(|mapping| {
+                    Ok(RtpMappingCodec {
+                        payload_type: mapping?.payload_type()?,
+                        mapped_payload_type: mapping?.mapped_payload_type()?,
+                    })
                 })
-                .collect(),
+                .collect::<Result<Vec<_>, Box<dyn Error>>>()?,
             encodings: mapping
-                .encodings
+                .encodings()?
                 .iter()
-                .map(|mapping| RtpMappingEncoding {
-                    rid: mapping.rid.clone().map(|rid| rid.to_string()),
-                    ssrc: mapping.ssrc,
-                    scalability_mode: mapping
-                        .scalability_mode
-                        .clone()
-                        .unwrap_or(String::from("S1T1"))
-                        .parse()
-                        .unwrap(),
-                    mapped_ssrc: mapping.mapped_ssrc,
+                .map(|mapping| {
+                    Ok(RtpMappingEncoding {
+                        rid: mapping?.rid()?.map(|rid| rid.to_string()),
+                        ssrc: mapping?.ssrc()?,
+                        scalability_mode: mapping?
+                            .scalability_mode()?
+                            .unwrap_or(String::from("S1T1").as_str())
+                            .parse()?,
+                        mapped_ssrc: mapping?.mapped_ssrc()?,
+                    })
                 })
-                .collect(),
-        }
+                .collect::<Result<Vec<_>, Box<dyn Error>>>()?,
+        })
     }
 }
 
--- a/rust/src/router/producer.rs
+++ b/rust/src/router/producer.rs
@@ -110,9 +110,13 @@ impl RtpStreamRecv {
         Ok(Self {
             params: RtpStreamParams::from_fbs_ref(dump.params()?)?,
             score: dump.score()?,
-            rtx_stream: dump.rtx_stream()?.map(|stream| RtxStream {
-                params: RtxStreamParams::from_fbs_ref(stream.params().unwrap()).unwrap(),
-            }),
+            rtx_stream: if let Some(rtx_stream) = dump.rtx_stream()? {
+                Some(RtxStream {
+                    params: RtxStreamParams::from_fbs_ref(rtx_stream.params()?)?,
+                })
+            } else {
+                None
+            },
         })
     }
 }
--- a/rust/src/router/producer.rs
+++ b/rust/src/router/producer.rs
@@ -106,14 +106,14 @@ pub struct RtpStreamRecv {
 }
 
 impl RtpStreamRecv {
-    pub(crate) fn from_fbs(dump: &rtp_stream::Dump) -> Self {
-        Self {
-            params: RtpStreamParams::from_fbs(&dump.params),
-            score: dump.score,
-            rtx_stream: dump.rtx_stream.clone().map(|stream| RtxStream {
-                params: RtxStreamParams::from_fbs(&stream.params),
+    pub(crate) fn from_fbs_ref(dump: rtp_stream::DumpRef<'_>) -> Result<Self, Box<dyn Error>> {
+        Ok(Self {
+            params: RtpStreamParams::from_fbs_ref(dump.params()?)?,
+            score: dump.score()?,
+            rtx_stream: dump.rtx_stream()?.map(|stream| RtxStream {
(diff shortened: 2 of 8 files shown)


ibc (approved):
Wouldn't it make more sense to not merge this PR in flatbuffers branch but wait and have a separate and complete PR for v3 once flatbuffers lands in v3?

rust/src/router/producer.rs
```diff
@@ -106,14 +106,14 @@ pub struct RtpStreamRecv {
...
-            params: RtpStreamParams::from_fbs(&dump.params),
-            score: dump.score,
-            rtx_stream: dump.rtx_stream.clone().map(|stream| RtxStream {
-                params: RtxStreamParams::from_fbs(&stream.params),
+    pub(crate) fn from_fbs_ref(dump: rtp_stream::DumpRef<'_>) -> Result<Self, Box<dyn Error>> {
+        Ok(Self {
+            params: RtpStreamParams::from_fbs_ref(dump.params()?)?,
+            score: dump.score()?,
+            rtx_stream: dump.rtx_stream()?.map(|stream| RtxStream {
+                params: RtxStreamParams::from_fbs_ref(stream.params().unwrap()).unwrap(),
```
jmillan: @nazar-pc, is there any way to avoid calling those `unwrap` and use the `?` here instead? I couldn't find it as the `map()` closure does not return a `Result` or `Option`.

rust/src/ortc.rs
```diff
@@ -77,32 +78,36 @@ impl RtpMapping {
...
-                        .parse()
-                        .unwrap(),
-                    mapped_ssrc: mapping.mapped_ssrc,
+                .map(|mapping| {
+                    Ok(RtpMappingEncoding {
+                        rid: mapping?.rid()?.map(|rid| rid.to_string()),
+                        ssrc: mapping?.ssrc()?,
+                        scalability_mode: mapping?
+                            .scalability_mode()?
+                            .unwrap_or(String::from("S1T1").as_str())
```
nazar-pc: I'd say you should parse it first and then use `unwrap_or_default()` rather than using a string here (I know it is not the change of this PR).

rust/src/router/producer.rs
ibc: Unless I miss something, `map()` in any language constructs a new array so it's impossible that it returns a reference (a reference to what? it's returning something that it has internally created).

rust/src/worker/channel.rs
```diff
@@ -279,12 +279,8 @@ impl Channel {
...
                                 match response.body().expect("failed accessing response body") {
                                     // Response has body.
-                                    Some(body_ref) => {
-                                        let _ = sender.send(Ok(Some(
-                                            body_ref
-                                                .try_into()
-                                                .expect("failed to retrieve response body"),
-                                        )));
+                                    Some(_) => {
+                                        let _ = sender.send(Ok(Some(Vec::from(&message[4..]))));
```
nazar-pc: I do not like that we are manually slicing `message` here. Is there a way around it? If we already have `Body`, why not sending it through? I suspect you can get `BodyRef` out of is and in happy case will not have to do extra allocation (see `DataProducerDumpRequest::Response` and some other.

jmillan:
> Wouldn't it make more sense to not merge this PR in flatbuffers branch but wait and have a separate and complete PR for v3 once flatbuffers lands in v3?

It's a proof of concept that I wanted to try just to see if there was any perf improvement. It turns out that it comes with 10% of improvement so we can say we want it that way.  This little PR took me almost a day to write, even though the porting the rest of `from_fbs` methods will theoretically be faster,  it's tedious.

I would like this to be merged, which prepares the land for porting the remaining methods, which could be done in a single PR or in various, since as I say, it takes some time doing this manual stuff and right now I'm reaching my human limit.

rust/src/router/producer.rs
nazar-pc: Why not? There is [`and_then`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.and_then) and [`transpose`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.transpose) depending on what you want to do here.

ibc:
> I would like this to be merged, which prepares the land for porting the remaining methods, which could be done in a single PR or in various, since as I say, it takes some time doing this manual stuff and right now I'm reaching my human limit.

Fair.

ibc:
Wow, this error in CI is amazing. I've restarted it assuming it is someone's issue.

https://github.com/versatica/mediasoup/actions/runs/6641579661/job/18044437060?pr=1194

```
npm-scripts [INFO] [worker:build] buildWorker()
npm-scripts [INFO] [worker:build] installMsysMake()
npm-scripts [INFO] [worker:build] executeCmd(): C:\hostedtoolcache\windows\Python\3.9.[13](https://github.com/versatica/mediasoup/actions/runs/6641579661/job/18044437060?pr=1194#step:8:14)\x64\python3.exe worker\scripts\getmake.py --dir="D:\a\mediasoup\mediasoup\worker\out\msys"
Traceback (most recent call last):
  File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 28, in <module>
    get('https://sourceforge.net/projects/mingw/files/MSYS/Base/make/make-3.81-3/make-3.81-3-msys-1.0.13-bin.tar.lzma/download', '847f0cbbf07135801c8e67bf692d29b1821e816ad828753c997fa869a9b89988')
  File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 16, in get
    assert hashlib.sha256(data).hexdigest() == digest
AssertionError
npm-scripts [ERROR] [worker:build] executeCmd() failed, exiting: Error: Command failed: C:\hostedtoolcache\windows\Python\3.9.13\x64\python3.exe worker\scripts\getmake.py --dir="D:\a\mediasoup\mediasoup\worker\out\msys"
```

It's also happening in Windows + Rust:

```
I--- stderr
  Traceback (most recent call last):
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 26, in <module>
      get('https://sourceforge.net/projects/mingw/files/MSYS/Base/termcap/termcap-0.20050421_1-2/libtermcap-0.20050421_1-2-msys-1.0.13-dll-0.tar.lzma/download', '62b58fe0880f0972fcc84a819265989b02439c1c5185870227bd25f870f7adb6')
    File "D:\a\mediasoup\mediasoup\worker\scripts\getmake.py", line 16, in get
      assert hashlib.sha256(data).hexdigest() == digest
  AssertionError
  thread 'main' panicked at 'Failed to install MSYS/make', worker\build.rs:110:17
```

rust/src/worker/channel.rs
jmillan: We don't have a `Body` (we don't want a `Body` as the whole thing here is accessing the data out of `Ref`s to avoid allocation) but a `BodyRef`. With the slicing I'm just removing the size (first 4 bytes) of the buffer that is no longer needed, then passing the binary to the thread waiting for the response. The receiving side will then re-interpret the buffer as a `BodyRef` and from then read everything it needs from the buffer directly.

rust/src/router/producer.rs
jmillan: This is not the same `map`. It's not related to arrays. https://doc.rust-lang.org/std/option/enum.Option.html#method.map

rust/src/worker/channel.rs
nazar-pc: The thing is `4` here is a magic number. Extremely easy to change something elsewhere in the app and break it. Is there a `body_ref.as_slice()` or something? If not `deserialize_message` above or some extra function should give you the split bytes instead. It is very fragile and accidental that you remove 4 bytes at the beginning of something deep in the library.

rust/src/ortc.rs
jmillan: `scalability_mode()` is of type `Result<Option<String>>`. I think it would not apply in this case, right?

nazar-pc: Why not? `.map(|maybe_scalability_mode| maybe_scalability_mode.parse()).transpose()?.unwrap_or_default()`

You map `Option<String>` into `Option<Result<ScalabilityMode>, Error>`, then turn it into `Result<Option<ScalabilityMode>, Error>`, bubble up result and default to default value if it is still `None`.

rust/src/worker/channel.rs
jmillan: There is no such thing like `body_ref.as_slice()` as far as I can see. I've asked in discord tough.

We need that 32bit sized prefix for Node only. Once we don't use UnixSocket to communicate Node and Worker we can totally get rid of this.

We could maybe use a constant rather than a harcoded `4` in the meantime though.

nazar-pc: `deserialize_message` already removes 4 bytes. Extract removal of those bytes into a separate function and use it before calling `deserialize_message`, such that everything down from it will not have to know prefix exists.

ibc:
Guys, I hate Python. This is consistently happening despite I restart failing jobs over and over:

https://github.com/versatica/mediasoup/actions/runs/6643130351/job/18049505764?pr=1194
```
urllib.error.URLError: <urlopen error [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond>
npm-scripts [ERROR] [worker:build] executeCmd() failed, exiting: Error: Command failed: C:\hostedtoolcache\windows\Python\3.9.13\x64\python3.exe worker\scripts\getmake.py --dir="D:\a\mediasoup\mediasoup\worker\out\msys"
```

rust/src/router/producer.rs
jmillan: I can't get rid of those.., will retry tomorrow or ask for help if I can't.

jmillan:
> Guys, I hate Python. This is consistently happening despite I restart failing jobs over and over:

I'm not saying this is 100% related but I've found issues happening only in certain branches that don't happen once merged. Windows CI issues are getting very annoying.

rust/src/router/producer.rs
jmillan: Done

jmillan:
Concerns noted here have been handled plus ConsumerDump construction has also been reimplemented not to make use of heap. I'll merge once CI passes.

PR #1197, opened 2023-10-26
[flatbuffers] Fix CI in node 20

ibc:
Same as https://github.com/versatica/mediasoup-client/pull/283

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -279,7 +279,7 @@ function replaceVersion()
 	const files = fs.readdirSync('node/lib',
 		{
 			withFileTypes : true,
-			recursive     : false
+			recursive     : true
 		});
 
 	for (const file of files)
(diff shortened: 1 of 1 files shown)


ibc:
Next in a new PR I'm gonna remove `mediasoup.version` getter. It's a complete hack (in involves `fs.writeFileSync()` in JS transpiled files in `node/lib` folder). I know this is a breaking change since `mediasoup.version` is a public getter but I think we can live with it. @nazar-pc @jmillan thoughts?

ibc:
So @nazar-pc suggest making `mediasoup.version` getter run `fs.readfileSync(path.join(__dirname__, '..', 'package.json')` and parse it and return the value of `version` field. It's an overkill IMHO.

nazar-pc:
It is not a getter, it is a constant, so you'll need to do that in `(() => {})()` so it is called immediately.

This is a public API you've committed to, breaking it isn't nice. I'm not sure why people would use that field, but maybe a bit less bad alternative than removing it would be to fix it at soem value like current release and mark as deprecated for people to not rely on anymore.

ibc:
I'll check keep it as it is. The issue is fixed in this PR so I won't complicate my life more than it is already. In v4 this will be removed.

PR #1198, opened 2023-10-27
[flatbuffers] Remove '__MEDIASOUP_VERSION__' replacement in JS transpiled files

ibc:
- Instead of writing into all transpiled JS files in `node/lib/`, read `package.json` just once in launch time and read "version" from it.
- I'm not happy with this, not sure if better than before. But clearly we are gonna remove this useless `mediasoup.version` public getter in v4.


nazar-pc (approved):
I think it is much better because we no longer need to post-process files

ibc:
Ok let's go.

jmillan (commented):
:+1:

PR #1199, opened 2023-10-27
FBS: Data[Consumer|Producer]  Send() simplifyfication

jmillan:
For some reason we were not sending everything as binary in this branch.

### DIFF
--- a/rust/src/messages.rs
+++ b/rust/src/messages.rs
@@ -2878,12 +2878,10 @@ impl Notification for DataProducerSendNotification {
     fn into_bytes(self, handler_id: Self::HandlerId) -> Vec<u8> {
         let mut builder = Builder::new();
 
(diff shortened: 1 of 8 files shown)


ibc (approved):
Good catch

rust/src/messages.rs
```diff
@@ -365,7 +365,7 @@ impl Request for WebRtcServerDumpRequest {
                 .tuple_hashes
                 .into_iter()
                 .map(|tuple_hash| WebRtcServerTupleHash {
-                    tuple_hash: tuple_hash.local_ice_username_fragment,
+                    tuple_hash: tuple_hash.tuple_hash,
```
jmillan, line 368: `fbs.rs` has been updated and this previous change in the schema has arisen.

PR #1202, opened 2023-10-29
Compile flatbuffers for Rust in build.rs

nazar-pc:
`flatbuffers` file is now auto-generated during compilation of `mediasoup-sys`, imports were changed accordingly, also `fbs` was added to `worker/Cargo.toml` since otherwise it'll fail to publish.


worker/build.rs
```diff
@@ -16,6 +16,35 @@ fn main() {
...
+                maybe_entry
+                    .map(|entry| {
+                        let path = entry.path();
+                        if path.extension() == Some("fbs".as_ref()) {
+                            Some(path)
+                        } else {
+                            None
+                        }
+                    })
+                    .transpose()
```
ibc, line 34: What does transport() do?

rust/src/fbs.rs
ibc, line 1: Assuming this fbs.es file is being generated on compile time, is it gitignored?

worker/build.rs
```diff
@@ -16,6 +16,35 @@ fn main() {
...
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .expect("Failed to collect flatbuffers files"),
+    )
+    .expect("Failed to translate flatbuffers files");
+
+    fs::write(
+        format!("{out_dir}/fbs.rs"),
+        planus_codegen::generate_rust(&flatbuffers_declarations)
```
ibc, line 43: That's all? No need to tricks in generated fbs.ts file to make clippy happy and so on?

worker/build.rs
nazar-pc: https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.transpose

worker/build.rs
nazar-pc: I was also a bit surprised given the fact that we do literal `include!()`, but apparently Rust tracks where source code comes from and from path it is clear that file is auto-generated and it ignores all warnings from there :man_shrugging:

I'm not sure why that happens and lazy to search, but it works out for well for us here :slightly_smiling_face:

rust/src/fbs.rs
nazar-pc: Nope, it doesn't exist in source tree at all, we write it into `OUT_DIR` that is specifically designed for that purpose just like the rest of stuff, see `build.rs`.

Read notes on https://doc.rust-lang.org/cargo/reference/environment-variables.html that contain `OUT_DIR` for details, there are a few of them there.

ibc (approved):

jmillan (commented):
:+1:

PR #1205, opened 2023-11-02
FBS: Node, Fix RtpParameters

jmillan:
Allways pass encodings, header extensions and rtcp.

All those are `required` by the schemas, so let's pass the default values if not defined by user.

NOTE: Even if they are not mandatory in the user API, sending an empty vector or default values, in the case of `rtcp`, to the worker is convenient.

### DIFF
--- a/node/src/tests/test-Producer.ts
+++ b/node/src/tests/test-Producer.ts
@@ -235,6 +235,69 @@ test('transport2.produce() succeeds', async () =>
 			});
 }, 2000);
 
+test.only('transport1.produce() without header extensions and rtcp succeeds', async () =>
+{
+	const onObserverNewProducer = jest.fn();
+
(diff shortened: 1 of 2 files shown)


node/src/tests/test-Producer.ts
```diff
@@ -235,6 +235,69 @@ test('transport2.produce() succeeds', async () =>
 			});
 }, 2000);
 
+test.only('transport1.produce() without header extensions and rtcp succeeds', async () =>
```
ibc, line 238: No similar test for Rust?

ibc (approved):

node/src/tests/test-Producer.ts
jmillan: No need because in Rust all those three types are not optional but mandatory and all have default values already.

PR #1207, opened 2023-11-03
Improve npm-scripts

ibc:
- Do not call `cleanWorkerArtifacts()` in `flatcNode()`. No reason for that plus it is bypassing the `MEDIASOUP_LOCAL_DEV` env purpose.
- Use more readable object arguments.


jmillan (approved):
💯

PR #1216, opened 2023-11-08
CI: remove unmaintained actions-rs/cargo actions

### LINKED ISSUE #1215: actions-rs/cargo are outdated and unmaintained
We run `actions-rs/cargo@v1` actions in `mediasoup-rust.yaml`. As warned in https://github.com/versatica/mediasoup/actions/runs/6796993961:

> ci (ubuntu-22.04)The following actions uses node12 which is deprecated and will be forced to run on node16: actions-rs/cargo@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default
> ci (ubuntu-20.04)The following actions uses node12 which is deprecated and will be forced to run on node16: actions-rs/cargo@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Problem is that [actions-rs/cargo](https://github.com/actions-rs/cargo) CI actions are not maintained... "This repository has been archived by the owner on Oct 13, 2023. It is now read-only.".

So what now?

ibc:
Fixes #1215

Also cosmetic changes in other CI files.

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -23,6 +23,9 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    env:
+      KEEP_BUILD_ARTIFACTS: '1'
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/mediasoup-rust.yaml
```diff
@@ -23,6 +23,9 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    env:
+      KEEP_BUILD_ARTIFACTS: '1'
```
nazar-pc, line 27: Why was this necessary?

nazar-pc (approved):

.github/workflows/mediasoup-rust.yaml
ibc: It was there already: https://github.com/versatica/mediasoup/pull/1216/files#diff-03b3d6e48484149f7e40c987fc6b29683f6a22bb39bd9095e210ae3b1cef70b3L53

Yes, just in `cargo test` step, but IMHO it makes sense that we set that env from the beginning because we don't want to fetch and rebuild everything if already built in a previous step.

nazar-pc: I see, missed that the first time :see_no_evil:

PR #1217, opened 2023-11-10
Node: extract version from package.json using require()

camilleoudot:
This allows JS bundlers such as webpack to embed the version string in the bundle. Otherwise a bundled JS app would look for mediasoup's package.json in "../.." on the filesystem (and it's very likely not there).

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -1,5 +1,3 @@
-import * as fs from 'node:fs';
-import * as path from 'node:path';
 import { Logger } from './Logger';
 import { EnhancedEventEmitter } from './EnhancedEventEmitter';
 import { workerBin, Worker, WorkerSettings } from './Worker';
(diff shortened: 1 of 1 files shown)


nazar-pc:
Why would you use bundlers for server-side JS?

camilleoudot:
In my case it produces a smaller and single artifact, and loads faster, in a context where this really matters.

nazar-pc (approved):
This makes sense to me

ibc:
Bufff, I'm about to just remove that `version` getter despite it would break API compatibility in v3...

Thanks, merging and releasing.

piranna:
> Bufff, I'm about to just remove that `version` getter despite it would break API compatibility in v3...

Maybe time to start with v4 roadmap.

ibc:
mediasoup node 3.13.1 released.

PR #1218, opened 2023-11-14
liburing

jmillan:
Enable liburing usage for linux (kernel versions >= 6).

Use liburing to relay incoming RTP to Consumers. This greately reduces the number of system calls as just a single system call is needed to relay the incoming RTP packet to "every" Consumer, opposite to libuv which requires a system call to relay the RTP packet to "each" Consumer.

For efficiency, memory buffers are preallocated holding the data payload being sent.

- Building mediasoup in a Linux machine with kernel >= 6 will enable the feature.
- Running mediasoup in a Linux machine with kernel >= 6 will make use of the feature, if it was also enabled in build time.

NOTE: The numbers are really positive.  Not only the CPU usage is reduced but it's also more stable.


nazar-pc:
Okay, I'm glad you've tested it with positive outcome!
From my experience it is not always faster.

ibc:
> but it's also more stable

What does this mean exactly?

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* Enable liburing usage for linux (kernel versions >= 6).
```
ibc: Add PR link now that it exists, please.

jmillan:
> Okay, I'm glad you've tested it with positive outcome!
> From my experience it is not always faster.

In my tests you'll see the difference when larg-ish Routers, but not with small/normal ones. Testing with >200 consumers you'll see a noticeable difference.

nazar-pc:
> In my tests you'll see the difference when larg-ish Routers, but not with small/normal ones. Testing with >200 consumers you'll see a noticeable difference.

That makes a lot of sense and is to be expected. Squeezing more performance out of the same CPU core is always nice.

jmillan:
>> The numbers are really positive. Not only the CPU usage is reduced but it's also more stable.

> What does this mean exactly?

The CPU usage is more stable, it shows less spikes.

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	if (error < 0)
+	{
+		MS_THROW_ERROR("io_uring_register_eventfd() failed: %s", std::strerror(-error));
+	}
+
+	// Watch the event file descriptor.
+	this->uvHandle = new uv_poll_t;
+
+	error = uv_poll_init(DepLibUV::GetLoop(), uvHandle, efd);
```
ibc:
```suggestion
	error = uv_poll_init(DepLibUV::GetLoop(), this->uvHandle, this->efd);
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	if (error != 0)
+	{
+		delete uvHandle;
```
ibc:
```suggestion
		delete this->uvHandle;
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+		MS_THROW_ERROR("uv_poll_init() failed: %s", uv_strerror(error));
+	}
+
+	uvHandle->data = this;
```
ibc:
```suggestion
	this->uvHandle->data = this;
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	error = uv_poll_start(uvHandle, UV_READABLE, static_cast<uv_poll_cb>(onFdEvent));
```
ibc:
```suggestion
	error = uv_poll_start(this->uvHandle, UV_READABLE, static_cast<uv_poll_cb>(onFdEvent));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	if (error != 0)
+	{
+		MS_THROW_ERROR("uv_poll_start() failed: %s", uv_strerror(error));
+	}
+
+	// Initialize available UserData entries.
+	for (size_t i = 0; i < DepLibUring::QueueDepth; ++i)
```
ibc:
```suggestion
	for (size_t i{ 0 }; i < DepLibUring::QueueDepth; ++i)
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+	{
+		MS_THROW_ERROR("io_uring_queue_init() failed: %s", std::strerror(-error));
+	}
+
+	/* Create an eventfd instance */
+	this->efd = eventfd(0, 0);
+
+	if (this->efd < 0)
+	{
+		MS_THROW_ERROR("eventfd() failed: %s", std::strerror(-efd));
```
ibc:
```suggestion
		MS_THROW_ERROR("eventfd() failed: %s", std::strerror(-this->efd));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+	MS_TRACE();
+
+	// Unset active flag.
+	this->active = false;
+
+	auto error = io_uring_submit(&this->ring);
+
+	if (error >= 0)
+	{
+		MS_DEBUG_DEV("%i submission queue entries submitted", ret);
```
ibc: What is this `ret`? It's not declared here at all.

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -232,6 +233,26 @@ void TcpConnectionHandle::Write(
...
 	}
 
+#ifdef MS_LIBURING_ENABLED
+	{
+		if (!DepLibUring::liburing->IsActive())
+		{
+			goto write_libuv;
+		}
+
+		auto prepared = DepLibUring::liburing->PrepareWrite(this->fd, data1, len1, data2, len2, cb);
```
ibc: Could you add some long comment here explaining why this could fail?

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -157,6 +158,26 @@ void UdpSocketHandle::Send(
...
 	}
 
+#ifdef MS_LIBURING_ENABLED
+	{
+		if (!DepLibUring::liburing->IsActive())
+		{
+			goto send_libuv;
+		}
+
+		auto prepared = DepLibUring::liburing->PrepareSend(this->fd, data, len, addr, cb);
```
ibc: Same here.

worker/src/DepLibUring.cpp
jmillan: Fixed, should be `error` instead.

worker/include/DepLibUring.hpp
```diff
@@ -0,0 +1,80 @@
...
+	{
+		this->active = true;
+	}
+	bool IsActive() const
+	{
+		return this->active;
+	}
+	io_uring* GetRing()
+	{
+		return &this->ring;
```
ibc: Let's use C++:

```suggestion
		return std::addressof(this->ring);
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+	MS_TRACE();
+
+	delete DepLibUring::liburing;
+}
+
+DepLibUring::DepLibUring()
+{
+	MS_TRACE();
+
+	auto error = io_uring_queue_init(DepLibUring::QueueDepth, &this->ring, 0);
```
ibc:
```suggestion
	auto error = io_uring_queue_init(DepLibUring::QueueDepth, std::addressof(this->ring), 0);
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	/* Create an eventfd instance */
+	this->efd = eventfd(0, 0);
+
+	if (this->efd < 0)
+	{
+		MS_THROW_ERROR("eventfd() failed: %s", std::strerror(-this->efd));
+	}
+
+	error = io_uring_register_eventfd(&this->ring, this->efd);
```
ibc:
```suggestion
	error = io_uring_register_eventfd(std::addressof(this->ring), this->efd);
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+	// Stop polling the event file descriptor.
+	uv_poll_stop(this->uvHandle);
+
+	delete this->uvHandle;
+
+	// Close the event file descriptor.
+	close(this->efd);
+
+	// Close the ring.
+	io_uring_queue_exit(&this->ring);
```
ibc:
```suggestion
	io_uring_queue_exit(std::addressof(this->ring));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+	auto* userData = this->GetUserData();
+
+	if (!userData)
+	{
+		MS_WARN_DEV("no user data entry available");
+
+		return false;
+	}
+
+	auto* sqe = io_uring_get_sqe(&this->ring);
```
ibc:
```suggestion
	auto* sqe = io_uring_get_sqe(std::addressof(this->ring));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
```
ibc:
```suggestion
	auto* sqe = io_uring_get_sqe(std::addressof(this->ring));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+}
+
+void DepLibUring::Submit()
+{
+	MS_TRACE();
+
+	// Unset active flag.
+	this->active = false;
+
+	auto error = io_uring_submit(&this->ring);
```
ibc:
```suggestion
	auto error = io_uring_submit(std::addressof(this->ring));
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,265 @@
...
+		MS_WARN_DEV("no user data entry available");
+
+		return nullptr;
+	}
+
+	auto idx = this->availableUserDataEntries.front();
+
+	this->availableUserDataEntries.pop();
+
+	auto* userData = &this->userDataBuffer[idx];
```
ibc:
```suggestion
	auto* userData = std::addressof(this->userDataBuffer[idx]);
```

worker/src/handles/TcpConnectionHandle.cpp
jmillan: io_uring has a fixed capacity which is registered upon initialization. If the rings are in use completely at the moment of requesting, then we just fallback to libuv for this packet.

This should not happen as we are creating 1024  * 4 entries, meaning we should be able to send up to 1024 * 4 RTP packets in one shot. Usually a single Router has less than so many Consumers so this should not be a problem.

worker/include/DepLibUring.hpp
jmillan: > Let's use C++:


hahaha

jmillan: We need to rewrite mediasoup to C++!

worker/src/DepLibUring.cpp
```diff
@@ -96,7 +94,7 @@ void DepLibUring::ClassInit()
 	// Enable liburing for kernel versions greather than or equal to 6.
 	if (kernelMayorLong >= 6)
```
ibc, line 95: I'd appreciate a static `IsSupported()` method in `LibUring` class instead.

ibc: No.

ibc: Please add a "info" tag log saying whether liburing is enabled or not.

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+	auto count = io_uring_peek_batch_cqe(liburing->GetRing(), cqes, DepLibUring::QueueDepth);
+
+	// libuv uses level triggering, so we need to read from the socket to reset
+	// the counter in order to avoid libuv calling this callback indefinitely.
+	eventfd_t v;
+	int err = eventfd_read(liburing->GetEventFd(), std::addressof(v));
+	if (err < 0)
+	{
+		MS_ERROR("eventfd_read() failed: %s", std::strerror(-err));
```
ibc: If this error happens then what? There is no return; here. Code below weeks running.

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,411 @@
...
+	};
+
+	for (unsigned int i = 0; i < count; ++i)
```
ibc: `unsigned int i{ 0 }`

worker/src/RTC/Router.cpp
```diff
@@ -660,6 +661,11 @@ namespace RTC
 			// Clone only happens if needed.
 			std::shared_ptr<RTC::RtpPacket> sharedPacket;
 
+#ifdef MS_LIBURING_ENABLED
+			// Activate liburing usage.
+			DepLibUring::SetActive();
+#endif
```
ibc: This is very confusing. Is it enabled because a macro/define is given? Or is it supported? And what does active mean?

ibc (commented):
Please clarify in the code (add comments) about what enabled, supported and active is. AFAIU:

- If there is a macro/defined set by Meson, it should be about "supported" rather than "enabled".
- And then "active" means whether it is being used right now or not.
- I would completely avoid and get rid of "enabled" term.

worker/src/DepLibUring.cpp
jmillan: If this error happens we should abort. This call should never fail.

ibc: Good. Then change it to `MS_ABORT()`? ¯\_(ツ)_/¯

worker/meson.build
```diff
@@ -257,6 +257,28 @@ if host_machine.system() == 'windows'
...
+      liburing_proj.get_variable('uring'),
+    ]
+    link_whole += [
+      liburing_proj.get_variable('liburing')
+    ]
+    common_sources += [
+      'src/DepLibUring.cpp',
+    ]
+    cpp_args += [
+      '-DMS_LIBURING_ENABLED',
```
ibc: This should be `MS_LIBURING_SUPPORTED`.

worker/src/RTC/Router.cpp
```diff
@@ -2,6 +2,7 @@
 // #define MS_LOG_DEV_LEVEL 3
 
 #include "RTC/Router.hpp"
+#include "DepLibUring.hpp"
```
ibc, line 6: I'm a bit worried about having to wrap every call to `DepLibUring` into a `#ifdef MS_LIBURING_SUPPORTED` block. IMHO `DepLibUring.hpp` should ALWAYS expose the same class and should only include `liburing.h` if the define is given. And it should expose a `IsSupported()` static method that returns true/false depending on that macro.

worker/src/DepLibUring.cpp
jmillan: I'll modify it.

jmillan: > Good. Then change it to MS_ABORT()? ¯_(ツ)_/¯

yes yes..., of course.

worker/src/RTC/Router.cpp
jmillan: So you are asking to change:

```
#ifdef MS_LIBURING_SUPPORTED
```

by:

```
if (DepLibUring::IsSupported())
````

is that it?

Comment (ibc) [reply worker/src/RTC/Router.cpp:6]: Yes. So only `DepLibUring::IsSupported()` method checks the `MS_LIBURING_SUPPORTED` given macro. Wdyt? Personally I don't like having all those `#if macro ... #end`. It's not readable. Or it's not real code.

File: worker/src/RTC/Router.cpp (hunk 22 above)
Comment (jmillan) [reply worker/src/RTC/Router.cpp]: `MS_LIBURING_ENABLED` means literaly => mediasoup-worker binary has liburing built in. This is, the capability is within the binary.

What would happen if a linux kernel >= 6 built the binary but a linux kernel < 6 is running it? Then even if the capability is within the binary, it cannot  be used in runtime. So `DepLibUring::SetActive()` will do nothing, indeed the thread local `LibUring` instance will be `nullptr`.

What would happen if a linux kernel >= 6 built the binary and a linux kernel >= 6 is running it? Then `DepLibUring::SetActive()` will effectively call `LibUring::SetActive`. We only send RTP packets while `LibUring` is active, and that's while the `Router` is iterating the `Consumers`, since we only want to use liburing for sending such packets for now.

Comment (jmillan) [reply worker/src/RTC/Router.cpp]: @ibc, let me know if this is clear.

Comment (ibc) [reply worker/src/RTC/Router.cpp]: So we use "enabled" to say that it's included in the binary and "supported" to say that it's effectively available in runtime? That's confusing, even more when we ALSO have "active", meaning that it's "enabled" in the binary, "supported" in runtime and in use right now.

Comment (ibc) [reply worker/src/RTC/Router.cpp]: I would rename MS_LIBURING_ENABLED to MS_LIBURING_SUPPORTED and then have a static DepLibUring::IsSupported() that returns true if that macro is given and kernel version is good. Anyway, with my other proposal the macro is gonna be used just within DepLibUring hpp file (which now should be always included) and the rest of the code would just check that static method.

File: worker/src/RTC/Router.cpp (hunk 24 above)
Comment (jmillan) [reply worker/src/RTC/Router.cpp:6]: The thing with the macro is that MacOS, Windows and binaries built with a linux kernel version <6 will not even have that symbol in the binary, and hence there is no code to run. Such code will not exist, simply. lf we put the macro behind a function everyone will call it even if liburing is not even present in the binary.

We are not going to burn the CPU for calling an empty method **every time** we send **anything** to the network.., but we can avoid making the call when the feature does not even exist in the binary by using a macro.

There are 3 places in the code making use of the macro: Router.cpp, UdpSocketHandler.cpp and TcpConnectionHanlder.cpp. Maybe we'll end up using liburing for every single thing we send over the network which will remove the usage in Router.cpp. It's two places where this macro is used.

EDIT: Also lib.cpp and Worker.cpp do use the macro to load the class.

File: worker/src/RTC/Router.cpp (hunk 22 above)
Comment (jmillan) [reply worker/src/RTC/Router.cpp]: I'll rename `MS_LIBURING_ENABLED` by `MS_LIBURING_SUPPORTED` .

File: worker/src/RTC/Router.cpp (hunk 24 above)
Comment (ibc) [reply worker/src/RTC/Router.cpp:6]: Here you are describing your original approach in this PR. Yes, I understand it. That's why I proposed another approach 😀. With my approach we must only check the macro in 3 places in DepLibUring.hpp (to only include liburing.h if supported, within the IsSupported() static method, and to only add that liburing struct/object member in the class). DepLibUring.hpp should always be included despite is Windows or Linux and DepLibUring.cpp should only be build if the macro is given (as it's already done in the PR).

Why don't you like this?

Comment (jmillan) [reply worker/src/RTC/Router.cpp:6]: > Why don't you like this?

I don't find it bad using a macro not to get code compiled in architectures that is not going to be used. And as said in the previous comment, I don't find a reason to call a method in architectures and older linux machines that we know it will have no effect.

Any library out there has macros when it deals with multiple architectures and we are a library. This feature only works for linux.

Having said this, if given my rationale you insist on putting the macro inside a function that everyone will call then I will do it.

Comment (ibc) [reply worker/src/RTC/Router.cpp:6]: > I don't find a reason to call a method in architectures and older linux machines that we know it will have no effect.

With my proposal the only method that would be called in NON supported platforms will be `DepLibUring::IsSupported()`.

> Having said this, if given my rationale you insist on putting the macro inside a function that everyone will call then I will do it.

No, I don't want to insist or mandate this change. I'm just exposing my concern. Let me check again how the code looks as it is.

File: worker/include/DepLibUring.hpp [hunk 25]
```diff
@@ -0,0 +1,98 @@
+#ifndef MS_DEP_LIBURING_HPP
+#define MS_DEP_LIBURING_HPP
+
+#ifdef MS_LIBURING_SUPPORTED
```
ibc, line 4: I think it makes more sense this way: If not supported don't do absolutely anything.

```suggestion
#ifdef MS_LIBURING_SUPPORTED
#ifndef MS_DEP_LIBURING_HPP
#define MS_DEP_LIBURING_HPP

```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,411 @@
+#define MS_CLASS "DepLibUring"
+
+// #define MS_LOG_DEV_LEVEL 3
```
ibc: For consistency:

```suggestion
#define MS_CLASS "DepLibUring"
// #define MS_LOG_DEV_LEVEL 3
```

worker/src/RTC/Router.cpp
ibc, line 6: So if we check the macro we must do it everywhere. We do it in Worker.cpp so must also be done here:

```suggestion
#ifdef MS_LIBURING_SUPPORTED
#include "DepLibUring.hpp"
#endif
```

worker/include/DepLibUring.hpp
```diff
@@ -0,0 +1,98 @@
...
+		bool active{ false };
+		// Pre-allocated UserData entries.
+		UserData userDataBuffer[QueueDepth]{};
+		// Indexes of available UserData entries.
+		std::queue<size_t> availableUserDataEntries;
+	};
+};
+
+#endif
+#endif
```
ibc: So let's not check the macro here. Instead, the rest of the code should NOT include `DepLibUring.hpp` if the macro is not given. Consistency with your own proposal:
- We only include `DepLibUring.hpp` if the macro is given.
- We only call methods in `DepLibUring` if the macro is given.
- Everywhere.
- We don't add any macro check in `DepLibUring.hpp` or `DepLibUring.cpp` since those files should ONLY be included and built if the macro is given.

```suggestion
#ifndef MS_DEP_LIBURING_HPP
#define MS_DEP_LIBURING_HPP

#include "DepLibUV.hpp"
#include <functional>
#include <liburing.h>
#include <queue>

class DepLibUring
{
public:
	using onSendCallback = const std::function<void(bool sent)>;

	/* Struct for the user data field of SQE and CQE. */
	struct UserData
	{
		uint8_t store[1500]{};
		onSendCallback* cb{ nullptr };
		size_t idx{ 0 };
	};

	/* Number of submission queue entries (SQE). */
	static constexpr size_t QueueDepth{ 1024 * 4 };

	static bool IsSupported();
	static void ClassInit();
	static void ClassDestroy();
	static void StartPollingCQEs();
	static void StopPollingCQEs();
	static bool PrepareSend(
	  int sockfd, const void* data, size_t len, const struct sockaddr* addr, onSendCallback* cb);
	static bool PrepareWrite(
	  int sockfd, const void* data1, size_t len1, const void* data2, size_t len2, onSendCallback* cb);
	static void Submit();
	static void SetActive();
	static bool IsActive();

	class LibUring;

	thread_local static LibUring* liburing;

public:
	class LibUring
	{
	public:
		LibUring();
		~LibUring();
		void StartPollingCQEs();
		void StopPollingCQEs();
		bool PrepareSend(
		  int sockfd, const void* data, size_t len, const struct sockaddr* addr, onSendCallback* cb);
		bool PrepareWrite(
		  int sockfd, const void* data1, size_t len1, const void* data2, size_t len2, onSendCallback* cb);
		void Submit();
		void SetActive()
		{
			this->active = true;
		}
		bool IsActive() const
		{
			return this->active;
		}
		io_uring* GetRing()
		{
			return std::addressof(this->ring);
		}
		int GetEventFd() const
		{
			return this->efd;
		}
		void ReleaseUserDataEntry(size_t idx)
		{
			this->availableUserDataEntries.push(idx);
		}

	private:
		UserData* GetUserData();

	private:
		// io_uring instance.
		io_uring ring;
		// Event file descriptor to watch for completions.
		int efd;
		// libuv handle used to poll uring completions.
		uv_poll_t* uvHandle{ nullptr };
		// Whether we are currently sending RTP over io_uring.
		bool active{ false };
		// Pre-allocated UserData entries.
		UserData userDataBuffer[QueueDepth]{};
		// Indexes of available UserData entries.
		std::queue<size_t> availableUserDataEntries;
	};
};

#endif
```

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "handles/TcpConnectionHandle.hpp"
 #include "DepLibUV.hpp"
+#include "DepLibUring.hpp"
```
ibc, line 7:
```suggestion
#ifdef MS_LIBURING_SUPPORTED
#include "DepLibUring.hpp"
#endif
```

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -2,6 +2,7 @@
 // #define MS_LOG_DEV_LEVEL 3
 
 #include "handles/UdpSocketHandle.hpp"
+#include "DepLibUring.hpp"
```
ibc, line 6:
```suggestion
#ifdef MS_LIBURING_SUPPORTED
#include "DepLibUring.hpp"
#endif
```

ibc:
Sorry for so many comments, I thought I was creating a batch but it was not true.

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,417 @@
...
+inline static void onCloseFd(uv_handle_t* handle)
+{
+	delete reinterpret_cast<uv_poll_t*>(handle);
+}
+
+inline static void onFdEvent(uv_poll_t* handle, int status, int events)
+{
+	auto* liburing = static_cast<DepLibUring::LibUring*>(handle->data);
+
+	auto count = io_uring_peek_batch_cqe(liburing->GetRing(), cqes, DepLibUring::QueueDepth);
```
ibc:
```suggestion
	auto* liburing = static_cast<DepLibUring::LibUring*>(handle->data);
	auto count     = io_uring_peek_batch_cqe(liburing->GetRing(), cqes, DepLibUring::QueueDepth);
```

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -232,6 +235,29 @@ void TcpConnectionHandle::Write(
...
+		auto prepared = DepLibUring::PrepareWrite(this->fd, data1, len1, data2, len2, cb);
+
+		if (!prepared)
+		{
+			// Cannot write via liburing, fallback to libuv.
+			goto write_libuv;
+		}
+
+		return;
+	}
```
ibc: We don't need those brackets.

```suggestion
	if (!DepLibUring::IsActive())
	{
		goto write_libuv;
	}

	// Prepare the data to be sent.
	// NOTE: If all SQEs are currently in use or no UserData entry is available we'll
	// fall back to libuv.
	auto prepared = DepLibUring::PrepareWrite(this->fd, data1, len1, data2, len2, cb);

	if (!prepared)
	{
		// Cannot write via liburing, fallback to libuv.
		goto write_libuv;
	}

	return;
```

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -157,6 +160,29 @@ void UdpSocketHandle::Send(
...
+		}
+
+		// Prepare the data to be sent.
+		// NOTE: If all SQEs are currently in use or no UserData entry is available we'll
+		// fall back to libuv.
+		auto prepared = DepLibUring::PrepareSend(this->fd, data, len, addr, cb);
+
+		if (!prepared)
+		{
+			// Cannot send via liburing, fallback to libuv.
```
ibc: Some MS_DEBUG_DEV saying it?

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -232,6 +235,29 @@ void TcpConnectionHandle::Write(
...
+		}
+
+		// Prepare the data to be sent.
+		// NOTE: If all SQEs are currently in use or no UserData entry is available we'll
+		// fall back to libuv.
+		auto prepared = DepLibUring::PrepareWrite(this->fd, data1, len1, data2, len2, cb);
+
+		if (!prepared)
+		{
+			// Cannot write via liburing, fallback to libuv.
```
ibc: Some MS_DEBUG_DEV saying it?

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -322,6 +348,13 @@ bool UdpSocketHandle::SetLocalAddress()
 	Utils::IP::GetAddressInfo(
 	  reinterpret_cast<const struct sockaddr*>(&this->localAddr), family, this->localIp, this->localPort);
 
+	err = uv_fileno(reinterpret_cast<uv_handle_t*>(this->uvHandle), std::addressof(this->fd));
+
+	if (err != 0)
+	{
+		MS_ERROR("uv_fileno() failed: %s", uv_strerror(err));
```
ibc: Why this? When can this fail? And what would happen if this fail? How is this related to liburing? Also, in p TcpSocketHandle.cpp` here you are also calling `return false;` but here that0s missing.

worker/src/handles/TcpConnectionHandle.cpp
jmillan: The brackets are there for a reason. Try to remove them and it won't compile:

Screenshot 2023-11-20 at 16 58 35

ibc: Hell, ok.

worker/src/handles/UdpSocketHandle.cpp
jmillan: > Why this? When can this fail? 

The method I'm calling to retrieve the file descriptor can fail, so we must handle it. I should not fail but it can.

> And what would happen if this fail?

We should abort, I've added a `return false` right now.

EDIT: The calling method will abort if this method returns `false`.

> How is this related to liburing? 

libuv does know the file descriptor of its handles, but in liburing we need to tell it which file descriptor we want to send data on.

> TcpSocketHandle.cpphere you are also callingreturn false;` but here that0s missing.

Fixed.

ibc: > > And what would happen if this fail?
> 
> We should abort, I've added a `return false` right now.

But what are the implications of returning false? Or should we replace with MS_ABORT()?


> > How is this related to liburing?
> 
> libuv does know the file descriptor of it's handles, but in liburing we need to tell it which file descriptor we want to send data on.

I don't get it. I mean, I do, but why is this now needed? Literally in `UdpSocketHandle` the PR does this:

```c++
	Utils::IP::GetAddressInfo(
	  reinterpret_cast<const struct sockaddr*>(&this->localAddr), family, this->localIp, this->localPort);


	// This is new in this PR:
	err = uv_fileno(reinterpret_cast<uv_handle_t*>(this->uvHandle), std::addressof(this->fd));

	if (err != 0)
	{
		MS_ERROR("uv_fileno() failed: %s", uv_strerror(err));
```

Why isn't that protected by `#ifdef MS_LIBURING_SUPPORTED` macro?

ibc: And why is that code in `SetLocalAddress()` method? That method is only supposed to obtain local IP and port in string/integer format, nothing else.

jmillan: > But what are the implications of returning false? Or should we replace with MS_ABORT()?

If `this.SetLocalAddress()` returns false, then the calling code will already call MS_ABORT.

>  but why is this now needed? 

We need the file descriptor of our local sockets in order to tell to liburing to send data over those file descriptors.

> Why isn't that protected by #ifdef MS_LIBURING_SUPPORTED macro?

I can put the macro sure, and also to the definition of `this->fd`. Do you want it?

> And why is that code in SetLocalAddress() method? That method is only supposed to obtain local IP and port in string/integer format, nothing else.

Let me put out of there.

jmillan: Handled,

- Added the definition and usage of `this->fd` inside the macro.
- Moved the retrieval of `this->fd` out of those methods.

ibc (commented):
Love it now. Is it ready? Are you gonna add those entries we discussed into the dump data? Or in another PR?

jmillan:
> Love it now. Is it ready? Are you gonna add those entries we discussed into the dump data? Or in another PR?

Thanks, I'll do few changes in this PR and add those dump data too. I'll notify you once done.

jmillan:
@ibc, the initial version of liburing is ready to merge. Can you please take a look at today's 4 commits?

* Send Nack-ed RTP using liburing
* Send RTCP compound packets using liburing
* Rename `IsSupported()` method for `IsRuntimeSupported()`
* `LibUring::Dump()`

worker/include/DepLibUring.hpp
```diff
@@ -0,0 +1,102 @@
...
+		// Pre-allocated UserData entries.
+		UserData userDataBuffer[QueueDepth]{};
+		// Indexes of available UserData entries.
+		std::queue<size_t> availableUserDataEntries;
+		// Submission queue entry process count.
+		uint64_t sqeProcessCount{ 0 };
+		// Submission queue entry miss count.
+		uint64_t sqeMissCount{ 0 };
+		// User data miss count.
+		uint64_t userDataMissCount{ 0 };
```
ibc: Consistency.

```suggestion
		uint64_t sqeProcessCount{ 0u };
		// Submission queue entry miss count.
		uint64_t sqeMissCount{ 0u };
		// User data miss count.
		uint64_t userDataMissCount{ 0u };
```

worker/src/DepLibUring.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+void DepLibUring::LibUring::Dump() const
+{
+	MS_TRACE();
+
+	MS_DUMP("<LibUring>");
+	MS_DUMP("  sqeProcessCount        : %" PRIu64, this->sqeProcessCount);
+	MS_DUMP("  sqeMissCount           : %" PRIu64, this->sqeMissCount);
+	MS_DUMP("  userDataMissCount      : %" PRIu64, this->sqeMissCount);
+	MS_DUMP("</LibUring>");
```
ibc:
```suggestion
	MS_DUMP("<DepLibUring>");
	MS_DUMP("  sqeProcessCount   : %" PRIu64, this->sqeProcessCount);
	MS_DUMP("  sqeMissCount      : %" PRIu64, this->sqeMissCount);
	MS_DUMP("  userDataMissCount : %" PRIu64, this->sqeMissCount);
	MS_DUMP("</DepLibUring>");
```

But why do we need this? This is not useful. I expected this info in the worker dump (meaning in the FBS dump response).

jmillan: It's just a mock method. The important thing is that it holds the values.

OK, I'll add this to the worker dump data.

jmillan: It's `LibUring` class where we are doing the `Dump()`.

jmillan: I'm removing this anyway.

ibc (approved):

ashinrot:
> This greately reduces the number of system calls as just a single system call is needed to relay the incoming RTP packet to "every" Consumer

Hello, I'm working on udp syscall tunning on my local server, and trying lots of linux feature. IOUring seems to be a good solution, but how about `sendmmsg` syscall, it seems to be able to reduce the syscall with less code?

jmillan:
> but how about sendmmsg syscall, it seems to be able to reduce the syscall with less code?

You don't send multiple RTP packets in a raw to the same socket in mediasoup, but the "same" RTP packet from a Producer, to multiple sockets.

`sendmmsg` is for sending multiple packets to the same socket.

This PR is about using `liburing` when it's more convenient than `libuv`.

PR #1221, opened 2023-11-15
Modernize clang-format

ibc:
### Details

- Remove `worker/scripts/gulpfile.mjs`.
- Get rid of super old `gulp-clang-format` NPM dependency.
- To clarify about versions:
  - Version of native `clang-format` binary included in the `clang-format` NPM dep required by `gulp-clang-format`: **14.0.0**
  - Verwion of native `clang-format` binary included in the latest `clang-format` NPM package: **15.0.0**
  - Latest version of native `clang-format` binary: **18.0.0**
- In this PR we still depend on `clang-format` NPM package so we got native `clang-format` binary 15.0.0, which comes with `InsertBraces` rule :)
- In a future task, we could try to also get rid of the `clang-format` NPM package and use native `clang-format` binary directly, however it must be different for each OS/arch, etc. So not in this PR.


jmillan (approved):
Awesome 👍

PR #1227, opened 2023-11-17
FBS: LayersChangeNotification body must be optional

jmillan:
It's none when the spatial layer is -1 internally.

This was crashing the worker when trying to create an empty notification, since the schema was requiring the body.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* FBS: Fix, LayersChangeNotification body must be optional ([1227](https://github.com/versatica/mediasoup/pull/1227))
+
+
 ### 3.13.1
 
 * Node: Extract version from `package.json` using `require()` ([PR #1217](https://github.com/versatica/mediasoup/pull/1217) by @arcinston).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### NEXT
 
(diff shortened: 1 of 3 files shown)


nazar-pc (commented):
Layers notification without layers makes little sense to me :thinking:

jmillan:
Yes, same here. But layers are represented by positive numbers. The absence of a positive layer means no layer.

We could make the body mandatory and both fields `spatialLayer` and `temporalLayer` optional though.

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* FBS: Fix, LayersChangeNotification body must be optional ([1227](https://github.com/versatica/mediasoup/pull/1227))
```
```suggestion
* FBS: `LayersChangeNotification` body must be optional ([PR #1227](https://github.com/versatica/mediasoup/pull/1227)).
```

ibc (changes requested):

ibc:
> Yes, same here. But layers are represented by positive numbers. The absence of a positive layer means no layer.
> 
> We could make the body mandatory and both fields `spatialLayer` and `temporalLayer` optional though.

we emit `null` to the user anyway so I don't see the point in notifying with undefined/null layers to Node/Rust to make them emit null.

CHANGELOG.md
ibc: I'm gonna commit this and merge.

ibc (approved):

PR #1228, opened 2023-11-17
Node: Add 'listenererror' in all entities

### LINKED ISSUE #1188: TS: expect() statements inside event handlers are not taking any effect

ibc:
- 'listenererror' event is emitted when an event listener throws (only if it's a synchronous error, otherwise an unhandled rejection will happen anyway).
- Usage:
  ```ts
  consumer.on('listenererror', (eventName, error) => { ... });
  ```
- Fixes issue #1188.

### Bonus Tracks:

- Remove `test.only` in `test-Producer.ts` and fix some tests in that file.

### DIFF
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -173,6 +173,7 @@ export type ConsumerEvents =
 	layerschange: [ConsumerLayers?];
 	trace: [ConsumerTraceEventData];
 	rtp: [Buffer];
+	listenererror: [string, Error];
 	// Private events.
 	'@close': [];
 	'@producerclose': [];
--- a/node/src/DataConsumer.ts
+++ b/node/src/DataConsumer.ts
@@ -85,6 +85,7 @@ export type DataConsumerEvents =
 	message: [Buffer, number];
 	sctpsendbufferfull: [];
 	bufferedamountlow: [number];
+	listenererror: [string, Error];
 	// Private events.
 	'@close': [];
 	'@dataproducerclose': [];
--- a/node/src/DataProducer.ts
+++ b/node/src/DataProducer.ts
@@ -61,6 +61,7 @@ export type DataProducerType = 'sctp' | 'direct';
 export type DataProducerEvents =
 {
 	transportclose: [];
+	listenererror: [string, Error];
 	// Private events.
 	'@close': [];
 };
--- a/node/src/EnhancedEventEmitter.ts
+++ b/node/src/EnhancedEventEmitter.ts
@@ -23,8 +23,6 @@ export class EnhancedEventEmitter<E extends Events = Events> extends EventEmitte
 	 */
 	safeEmit<K extends keyof E & string>(eventName: K, ...args: E[K]): boolean
 	{
-		const numListeners = super.listenerCount(eventName);
-
 		try
 		{
 			return super.emit(eventName, ...args);
@@ -33,9 +31,19 @@ export class EnhancedEventEmitter<E extends Events = Events> extends EventEmitte
 		{
 			logger.error(
 				'safeEmit() | event listener threw an error [eventName:%s]:%o',
-				eventName, error);
+				eventName, error
+			);
 
-			return Boolean(numListeners);
+			try
+			{
+				super.emit('listenererror', eventName, error);
+			}
+			catch (error2)
+			{
+				// Ignore it.
+			}
+
+			return Boolean(super.listenerCount(eventName));
 		}
 	}
 
--- a/node/src/Producer.ts
+++ b/node/src/Producer.ts
@@ -134,6 +134,7 @@ export type ProducerEvents =
 	score: [ProducerScore[]];
 	videoorientationchange: [ProducerVideoOrientation];
 	trace: [ProducerTraceEventData];
+	listenererror: [string, Error];
 	// Private events.
 	'@close': [];
 };
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -172,6 +172,7 @@ type PipeTransportPair =
 export type RouterEvents =
 {
 	workerclose: [];
+	listenererror: [string, Error];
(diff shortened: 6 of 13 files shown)


ibc:
To be clear: in this PR, if anything throws within any dataConsumer or dataProducer event listener in `test-DirectTransport.ts`, the test should fail. But it doesn't (see CI) :(

jmillan (approved):
👍

piranna:
So, if an event listener throws an error, a new `listenerror` is emitted? Why not use `error` event? It's special one because if it's not handled, it's thrown as an unhandled exception, so it doesn't get unnoticed.

ibc:
> So, if an event listener throws an error, a new `listenerror` is emitted? Why not use `error` event? It's special one because if it's not handled, it's thrown as an unhandled exception, so it doesn't get unnoticed.

So you literally mean emitting "error" instead of "listenererror" within safeEmit() right?

It makes sense. Only problem is that we loose the "eventName" argument but we may emit a custom Error class with that field exposed by a public getter.

piranna:
> So you literally mean emitting "error" instead of "listenererror" within safeEmit() right?

What I mean is, from the [Node.js docs](https://nodejs.org/api/events.html#error-events):

> When an error occurs within an EventEmitter instance, the typical action is for an 'error' event to be emitted. These are treated as special cases within Node.js.
>
> If an EventEmitter does not have at least one listener registered for the 'error' event, and an 'error' event is emitted, the error is thrown, a stack trace is printed, and the Node.js process exits.

ibc:
That's the classic controversial documentation that doesn't honor reality:

### Emitting "error" event without an "error" event listener should exit the process (yes, it does):

```js
import { EventEmitter } from 'node:events';
import * as http from 'node:http';

const server = http.createServer((req, res) =>
{
	res.writeHead(200);
	res.end("bye");
});

server.listen(5005, '0.0.0.0', () => console.log('server listening'));

const emitter = new EventEmitter();

emitter.on('foo', () =>
{
	console.log(
		'"foo" event, emitting unhandled "error" event that should terminate the process (but it does not)'
	);

	emitter.emit('error', new TypeError('LALALA'));
});

setTimeout(() =>
{
	console.log('timer expired, emitting "foo"');

	emitter.emit("foo");

	console.log('"foo" emitted');
}, 2000);
```

Output:
```
server listening
timer expired, emitting "foo"
"foo" event, emitting unhandled "error" event that should terminate the process (but it does not)
node:events:492
      throw er; // Unhandled 'error' event
      ^

TypeError: LALALA
    at EventEmitter.<anonymous> (file:///private/tmp/kk/bar2.mjs:25:24)
    at EventEmitter.emit (node:events:514:28)
    at Timeout._onTimeout (file:///private/tmp/kk/bar2.mjs:32:10)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
Emitted 'error' event at:
    at EventEmitter.<anonymous> (file:///private/tmp/kk/bar2.mjs:25:10)
    at EventEmitter.emit (node:events:514:28)
    [... lines matching original stack trace ...]
    at process.processTimers (node:internal/timers:514:7)
```

### Throwing an error within an event listener should generate "error" event (but it doesn't):

```js
import { EventEmitter } from 'node:events';
import * as http from 'node:http';

const server = http.createServer((req, res) =>
{
	res.writeHead(200);
	res.end("bye");
});

server.listen(5005, '0.0.0.0', () => console.log('server listening'));

const emitter = new EventEmitter();

emitter.on('error', (error) =>
{
	console.log('"error" event: %s', String(error));
});

emitter.on('foo', () =>
{
	console.log(
		'"foo" event, throwing an error that should produce an "error" event (but it does not)'
	);

	throw new Error('LALALA');
});

setTimeout(() =>
{
	console.log('timer expired, emitting "foo"');

	emitter.emit("foo");

	console.log('"foo" emitted');
}, 2000);
```

Output:
```
server listening
timer expired, emitting "foo"
"foo" event, throwing an error that should produce an "error" event (but it does not)
file:///private/tmp/kk/bar2.mjs:25
	throw new Error('LALALA');
	^

Error: LALALA
    at EventEmitter.<anonymous> (file:///private/tmp/kk/bar2.mjs:25:8)
    at EventEmitter.emit (node:events:514:28)
    at Timeout._onTimeout (file:///private/tmp/kk/bar2.mjs:32:10)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
```

ibc:
If we make our `safeEmit()` method emit "error" when the user given listener throws, then we would terminate the entire process unless the user creates an "error" listener (in every Transport, Producer, etc). That would be terrible from backwards compatibility. We cannot do that at this point.

piranna:
> Throwing an error within an event listener should generate "error" event (but it doesn't):

I don't think it should emit an `error` event, they must be explicitly emitted, throwing it as an unhandled error is the correct way here.


> If we make our `safeEmit()` method emit "error" when the user given listener throws, then we would terminate the entire process unless the user creates an "error" listener (in every Transport, Producer, etc). That would be terrible from backwards compatibility. We cannot do that at this point.

Agree, fair enough. We can consider it for the 4.0 version, specially regarding all places where `logger.error()` is being call when in fact errors should be thrown instead, maybe as Promises for API homogeneity.

ibc:
> Agree, fair enough. We can consider it for the 4.0 version, specially regarding all places where `logger.error()` is being call when in fact errors should be thrown instead, maybe as Promises for API homogeneity.

Yes.

PR #1230, opened 2023-11-20
Fix issue when `pause`/`resume` events are not emitted

douglaseel:
Currently if we `pause()`/`resume()` a producer/consumer/dataProducer/dataConsumer very quickly like the code above (without awaiting for promise to be resolved), the `resume` event will not be triggered, but it should.

```
producer.observer.on('pause', () => { ... });
producer.observer.on('resume', () => { ... });

let pausePromise = producer.pause();
let resumePromise = producer.resume();
```

### DIFF
--- a/node/src/Consumer.ts
+++ b/node/src/Consumer.ts
@@ -612,14 +612,14 @@ export class Consumer<ConsumerAppData extends AppData = AppData>
 	{
 		logger.debug('pause()');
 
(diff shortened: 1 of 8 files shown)


jmillan (approved):
:+1:

ibc:
I've pulled this branch but I cannot push to it (no permission) so I've cloned it here: https://github.com/versatica/mediasoup/pull/1231

PR #1231, opened 2023-11-20
Fix issue when pause/resume events are not emitted

ibc:
*NOTE:* Forked from https://github.com/versatica/mediasoup/pull/1230 by @douglaseel (since his branch doesn't let me edit it).

----

### Details

- Currently if we `pause()`/`resume()` a producer/consumer/dataProducer/dataConsumer very quickly like the code below (without awaiting for promise to be resolved), the 'resume' event will not be triggered, but it should.
  ```ts
  producer.observer.on('pause', () => { ... });
  producer.observer.on('resume', () => { ... });
  
  let pausePromise = producer.pause();
  let resumePromise = producer.resume();
  ```
- Fix logic of 'pause' and 'resume' events in `Consumer` and `DataConsumer` JS and Rust classes.

### DIFF
--- a/rust/src/router/consumer.rs
+++ b/rust/src/router/consumer.rs
@@ -815,24 +815,23 @@ impl Consumer {
                         }
                         Notification::ProducerPause => {
                             let mut producer_paused = producer_paused.lock();
-                            let was_paused = *paused.lock() || *producer_paused;
+                            let paused = *paused.lock();
                             *producer_paused = true;
 
                             handlers.producer_pause.call_simple();
 
-                            if !was_paused {
+                            if !paused {
                                 handlers.pause.call_simple();
                             }
                         }
                         Notification::ProducerResume => {
                             let mut producer_paused = producer_paused.lock();
                             let paused = *paused.lock();
-                            let was_paused = paused || *producer_paused;
                             *producer_paused = false;
 
                             handlers.producer_resume.call_simple();
 
-                            if was_paused && !paused {
+                            if !paused {
                                 handlers.resume.call_simple();
                             }
                         }
--- a/rust/src/router/data_consumer.rs
+++ b/rust/src/router/data_consumer.rs
@@ -509,24 +509,23 @@ impl DataConsumer {
                         }
                         Notification::DataProducerPause => {
                             let mut data_producer_paused = data_producer_paused.lock();
-                            let was_paused = *paused.lock() || *data_producer_paused;
+                            let paused = *paused.lock();
                             *data_producer_paused = true;
 
                             handlers.data_producer_pause.call_simple();
 
-                            if !was_paused {
+                            if !paused {
                                 handlers.pause.call_simple();
                             }
                         }
                         Notification::DataProducerResume => {
                             let mut data_producer_paused = data_producer_paused.lock();
                             let paused = *paused.lock();
-                            let was_paused = paused || *data_producer_paused;
                             *data_producer_paused = false;
 
                             handlers.data_producer_resume.call_simple();
 
-                            if was_paused && !paused {
+                            if !paused {
                                 handlers.resume.call_simple();
                             }
                         }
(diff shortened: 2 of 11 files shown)


jmillan (approved):

rust/src/router/data_consumer.rs
```diff
@@ -509,24 +509,23 @@ impl DataConsumer {
...
                         Notification::DataProducerPause => {
                             let mut data_producer_paused = data_producer_paused.lock();
-                            let was_paused = *paused.lock() || *data_producer_paused;
+                            let paused = *paused.lock();
                             *data_producer_paused = true;
 
                             handlers.data_producer_pause.call_simple();
 
-                            if !was_paused {
+                            if !paused {
```
jmillan, line 517: No need to check `*data_producer_paused`, anymore?

rust/src/router/consumer.rs
```diff
@@ -815,24 +815,23 @@ impl Consumer {
                         }
                         Notification::ProducerPause => {
                             let mut producer_paused = producer_paused.lock();
-                            let was_paused = *paused.lock() || *producer_paused;
```
jmillan, line 818: No need to check *producer_paused, anymore?

rust/src/router/data_consumer.rs
ibc: If worker sent "data producer pause" notification it means that data producer was not paused. No need to check that since it cannot be changed any other way.

rust/src/router/consumer.rs
ibc: Same.

nazar-pc (commented):
Was there actually an issue in Rust version that it needed changes though? Having tests that prove that would be nice.

ibc:
> Was there actually an issue in Rust version that it needed changes though? Having tests that prove that would be nice.

Changes in Rust are cosmetic to be in sync with cosmetic changes done in Node layer. Real changes done in Node are NOT needed in Rust since Rust code was good already. In other words: real changes made for Node layer in this PR make the code in Node be equivalent to the one in Rust. And of course there were already tests.

nazar-pc:
> Changes in Rust are cosmetic

They are not. Not only variable named changed to less clear, `if` conditions are also changed.

ibc:
> > Changes in Rust are cosmetic
> 
> They are not. Not only variable named changed to less clear, `if` conditions are also changed.

Which exact variable names have changed?

CleanShot-2023-11-20-at-12 20 59@2x

`paused` is just the local paused state.

nazar-pc:
You literally have `*data_producer_paused` removed in the diff on screenshot. Why was it necessary?

ibc:
> You literally have `*data_producer_paused` removed in the diff on screenshot. Why was it necessary?

Explained here: https://github.com/versatica/mediasoup/pull/1231#pullrequestreview-1739535510

Worker only notifies Node/Rust about producer paused or resumed if it was NOT already paused or resumed, so once in Node or Rust we don't need to check previous paused state of the producer.

ibc:
> > You literally have `*data_producer_paused` removed in the diff on screenshot. Why was it necessary?
> 
> Explained here: [#1231 (review)](https://github.com/versatica/mediasoup/pull/1231#pullrequestreview-1739535510)
> 
> Worker only notifies Node/Rust about producer paused or resumed if it was NOT already paused or resumed, so once in Node or Rust we don't need to check previous paused state of the producer.

BTW this is tested in Rust code. If `handlers.resume.call_simple()` or `handlers.pause.call_simple()` is commented then some tests don't complete.

ibc:
@nazar-pc can I merge?

ibc:
> BTW this is tested in Rust code. If `handlers.resume.call_simple()` or `handlers.pause.call_simple()` is commented then some tests don't complete.

Even better. Rust tests were already checking `producer.pause()` / `resume()` and the `on_producer_pause` and `on_producer_resume` events, so we are safe.

CleanShot-2023-11-20-at-13 55 03@2x

Issue #1233, opened 2023-11-21
Refactor RTCP packets using Serializable class


ibc:
In summary: redo RTCP packets following same code used for SCTP packets/chunks.

### Must manage unsupported packet / chunk / block types

For example, a XR packet can contain many different types of Extender Report Blocks. However, if one of those are not implemented in mediasoup then the parsed packet will ignore every block after it.

In (rtp.js](https://github.com/versatica/rtp.js) lib this is solved by having a `GenericExtendedReport` class.

The following XR packet is parsed with zero blocks in mediasoup since the Extended Report LRLE block type is not implemented:

```
0xa0, 0xcf, 0x00, 0x0E, // Padding, Type: 207 (XR), Length: 14
0x5d, 0x93, 0x15, 0x34, // Sender SSRC: 0x5d931534
// Extended Report LRLE
0x01, 0x09, 0x00, 0x04, // BT: 1 (LRLE), T: 9, Block Length: 4
0x03, 0x93, 0x2d, 0xb4, // SSRC of source: 0x03932db4
0x00, 0x11, 0x00, 0x22, // Begin Seq: 0x11, End Seq: 0x22
0b00101010, 0b10101010, // Run Lengh Chunk (zeros)
0b01101010, 0b10101010, // Run Lengh Chunk (ones)
0b11101010, 0b10101010, // Bit Vector Chunk
0b00000000, 0b00000000, // Terminating Null Chunk
// Extended Report DLRR
0x05, 0x00, 0x00, 0x06, // BT: 5 (DLRR), Block Length: 6
0x11, 0x12, 0x13, 0x14, // SSRC 1
0x00, 0x11, 0x00, 0x11, // LRR 1
0x11, 0x00, 0x11, 0x00, // DLRR 1
0x21, 0x22, 0x23, 0x24, // SSRC 2
0x00, 0x22, 0x00, 0x22, // LRR 2
0x22, 0x00, 0x22, 0x00, // DLRR 2
0x00, 0x00, 0x00, 0x04 // Padding (4 bytes)
```

### Must properly handle RTCP padding

Having the above XR packet we can know (from the very beginning) that padding is 4 bytes. We know the RTCP length of the whole packet (14) and we know that the padding bit is set, so we can read the value of the last byte to know how many padding bytes are included and exclude them later when parsing blocks, items, subblocks, etc. In the above XR packet, some custom temporary logs show the problem:

```
[...]

PR #1234, opened 2023-11-21
Fix RTCP DLRR (Delay Since Last Receiver Report) block parsing

### LINKED ISSUE #1211: Issue in sending dlrr block
In [rfc3611](https://www.rfc-editor.org/rfc/rfc3611#section-4.5) ，the dlrr block consists of one or more 3 word sub-
   blocks: one sub-block per Receiver Report.Report. but in mediasoup they are in different blocks.

The client may only resolve 1 block.
rfc3611:
image

mediasoup:
image

ibc:
Fixes #1211

### Details

- Bug fixed but I'm not happy. A complete RTCP refactor is needed. Exposed here: https://github.com/versatica/mediasoup/issues/1233
- Bonus track: In `Makefile` add "flatc" dep to various tasks.
- Bonus track: Get rid of the tedious "variable 'send_delta_ms' set but not used" warning in `remote_bitrate_estimator_abs_send_time.cc`.

### DIFF
--- a/worker/src/RTC/RTCP/XrDelaySinceLastRr.cpp
+++ b/worker/src/RTC/RTCP/XrDelaySinceLastRr.cpp
@@ -75,14 +75,12 @@ namespace RTC
 				{
 					report->AddSsrcInfo(ssrcInfo);
 					offset += ssrcInfo->GetSize();
+					reportLength -= ssrcInfo->GetSize();
 				}
 				else
 				{
 					return report.release();
 				}
-
-				offset += ssrcInfo->GetSize();
-				reportLength -= DelaySinceLastRr::SsrcInfo::BodySize;
 			}
 
 			return report.release();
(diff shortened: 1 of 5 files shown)


worker/src/RTC/RTCP/XrDelaySinceLastRr.cpp
```diff
@@ -75,14 +75,12 @@ namespace RTC
...
 					report->AddSsrcInfo(ssrcInfo);
 					offset += ssrcInfo->GetSize();
+					reportLength -= ssrcInfo->GetSize();
 				}
 				else
 				{
 					return report.release();
 				}
-
-				offset += ssrcInfo->GetSize();
```
ibc, line 84: This was the terrible bug: this was duplicated (see line 77 above) so offset was decremented twice.

jmillan (approved):

PR #1235, opened 2023-11-22
Fix calculate sdes size

ybybwdwd:
`SdesPacket::GetCount()` start with 1 when SdesPacket has chunk, so we need `MaxChunksPerPacket + 1` here.

### DIFF
--- a/worker/include/RTC/RTCP/Sdes.hpp
+++ b/worker/include/RTC/RTCP/Sdes.hpp
@@ -214,7 +214,7 @@ namespace RTC
 				// A serialized packet can contain a maximum of 31 chunks.
 				// If number of chunks exceeds 31 then the required number of packets
 				// will be serialized which will take the size calculated below.
-				size_t size = Packet::CommonHeaderSize * ((this->GetCount() / MaxChunksPerPacket) + 1);
+				size_t size = Packet::CommonHeaderSize * ((this->GetCount() / (MaxChunksPerPacket + 1)) + 1);
 
 				for (auto* chunk : this->chunks)
 				{


jmillan:
@ybybwdwd, the change must come with the corresponding tests.

ibc:
So `MaxChunksPerPacket` is 31.

Original code:
```c++
size_t size = 4 * ((this->GetCount() / 31) + 1);
```

New code:
```c++
size_t size = 4 * ((this->GetCount() / 32) + 1);
```

Let's try some maths with the original code and the new code in this PR:

- SDEP packet with 30 chunks:
  - Original code: `size = 4 * (Math.trunc(30 / 31) + 1) = 4`
  - New code: `size = 4 * (Math.trunc(30 / 32) + 1) = 4`

- SDEP packet with 31 chunks (this is, max number of chunks):
  - Original code: `size = 4 * (Math.trunc(31 / 31) + 1) = 8`
  - New code: `size = 4 * (Math.trunc(31 / 32) + 1) = 4` (ooops)

- SDEP packet with 32 chunks:
  - Original code: `size = 4 * (Math.trunc(32 / 31) + 1) = 8`
  - New code: `size = 4 * (Math.trunc(32 / 32) + 1) = 8`

If the SDES packet has max allowed number of chunks then the initial `size` should be 4 (the same as if there are less chunks than max number).

*So yes, this PR is correct.*

> the change must come with the corresponding tests.

Tests are passing, meaning that there was no test for this anyway.

jmillan:
> Tests are passing, meaning that there was no test for this anyway.

I'm literally asking for a test that asserts the fix. I know there is currently no test for this, otherwise it would fail in this new branch.

ibc:
I'm adding a test for this.

ibc:
@ybybwdwd @jmillan I'm replacing this PR by https://github.com/versatica/mediasoup/pull/1236 which is the same but includes tests. Thanks for the fix, @ybybwdwd.

PR #1237, opened 2023-11-22
RTCP `CompoundPacket`: Use a single DLRR block to hold all ssrc infos

### LINKED ISSUE #1211: Issue in sending dlrr block

ibc:
Fixes #1211

### Details

- Previously, our `CompoundPacket` has a XR packet with many DLRR blocks, each one with a single ssrc info sub-block.
- Now our `CompoundPacket` has a XR pacjet with a single DLRR block with many ssrc info sub-blocks.
- Both approaches are spec compliant **BUT** libwebrtc thinks it makes no sense to have multiple blocks of the same type in an XR packet: https://codereview.webrtc.org/2378113002
- Thanks @ybybwdwd for reporting the issue and for the references.

### TODO

* [x] Same should be done for other blocks. This must be verified. **Done** in https://github.com/versatica/mediasoup/issues/1238.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### NEXT
 
 * Fix RTCP SDES packet size calculation ([PR #1236](https://github.com/versatica/mediasoup/pull/1236) based on PR [PR #1234](https://github.com/versatica/mediasoup/pull/1234) by @ybybwdwd).
+* RTCP Compound Packet: Use a single DLRR report to hold all ssrc info sub-blocks ([PR #1237](https://github.com/versatica/mediasoup/pull/1237)).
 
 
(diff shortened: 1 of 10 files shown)


ibc:
I've created a separate issue to handle the rest of extended reports that may need some changes:

https://github.com/versatica/mediasoup/issues/1238

jmillan (approved):

PR #1245, opened 2023-11-27
Improvements to the build system

ibc:
### Details

- Don't specify Python Invoke version.
- `build.rs`: Idiomaticvs improvements.
- `npm-scripts.mjs`: Use spawn instead of exec to run commands, so we can pass command and each argument separately.
- Honor alphabetical order in "files" in `package.json` and remove `Makefile` from it and `worker/Cargo.toml`.

### Does this PR fix installation under paths with a double quote in a directly name?

No. I'm afraid there is no easy way to do that. For many reasons:

1. Python Invoke doesn't specify an API to pass command name and separate argument strings when running external commands. Honestly I don't even know if Python itself supports it. Ticket reported in Invoke project: https://github.com/pyinvoke/invoke/issues/977
2. Node does offer an API for that: [`child_process.spawn(command[, args][, options])`](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options). However it doesn't escape or encode properly individual arguments. For example:
  ```js
  const path = '/tmp/123 foo "bar';

  cont result = child_process.spawnSync('ls', [ '-ls', path ]);

  result.stderr.toString();
  // => 'ls: /tmp/123 foo "bar: No such file or directory\n'
  ```

At least installation in a path with whitespaces keep working (it was already since previos PR). I just give up here and honestly nobody should expect thst any library work properly when installing it under a too fancy path. BTW this is not the purpose of this PR anyway.


worker/Cargo.toml
```diff
@@ -21,7 +21,6 @@ include = [
     "/test/src",
     "/build.rs",
     "/Cargo.toml",
-    "/Makefile",
```
nazar-pc, line 24: Please add `tasks.py` here

nazar-pc (commented):
I like less shell scripts, just one nit in `build.rs`

worker/Cargo.toml
ibc: Done.

nazar-pc (approved):

ibc:
I'm reverting changes in npm-scripts and use `execSync()` again. Why? Because mediasoup-node CI task fails on Windows due to "tsc ENOENT": https://github.com/versatica/mediasoup/actions/runs/7008109231/job/19063676923?pr=1245

Revert commit here: https://github.com/versatica/mediasoup/pull/1245/commits/e39731d2b2da3e2d90faa5d142f7983a8dd7cb96

When running any npm script, npm adds `node_modules/.bin` to the `PATH`. Otherwise calling `tsc` here wouldn't work at all. But for whatever reason it's failing when using `spawn()` despite the created process inherits `process.env` as documented. I don't want to waste time investigating this. Using `exec()` is way more comfortable than using `spawn()` and using the latter has proven to not fix ANYTHING.

PR #1249, opened 2023-12-01
LibUring: Enable liburing usage for SCTP data delivery

jmillan:
There are two SCTP delivery scenarios that would benefit of liburing:

1- When Router iterates DataConsumers to deliver a message from a DataProducer.
2- When we call `usrsctp_handle_timers()` periodically, which will synchronously call the send/recv callbacks for the SctpAssociations.


ibc:
> When Router iterates DataConsumers to deliver a message from a DataProducer.

This doesn't make sense, do you mean the opposite?

worker/src/DepUsrSCTP.cpp
```diff
@@ -249,7 +252,21 @@ void DepUsrSCTP::Checker::OnTimer(TimerHandle* /*timer*/)
...
+	// sent over the network then we will send those messages within a single
+	// system call.
+	DepLibUring::SetActive();
+#endif
+
 	usrsctp_handle_timers(elapsedMs);
 
+#ifdef MS_LIBURING_SUPPORTED
+	// Submit all prepared submission entries.
+	DepLibUring::Submit();
```
ibc, line 268: Why here? Submit should be called when messages are about to be dents here we just called to handle timers.

jmillan: As indicated in the comment few lines above, the call to `usrsctp_handle_timers()` synchronously call the send/recv static callbacks that we provided to `usrsctp_socket` in `SctpAssociation` for all the send/recv enqueued tasks in the usrsctp sockets, so it's very probable that multiple messages will be sent over the network between the new `DepLibUring::SetActive()` and `DepLibUring::Submit()` calls. Hence, it's important that when those messages are about to be sent `DepLibUring` is active so they are prepared in the ring or order to be later sent at once upon calling `DepLibUring::Submit()`

jmillan:
>> When Router iterates DataConsumers to deliver a message from a DataProducer.

> This doesn't make sense, do you mean the opposite?

The same as for RTP Producer and Consumers. When Router iterates the Consumers to deliver the RTP packet that we received into a Producer, we activate the ring, which make that those messages will be prepared to use liburing. Once all Consumers have been iterated then `DepLibUring::Submit()` is called, meaning, all previously prepared messages will now be delivered at once.

ibc (approved):
Everything was good in the description. I should not review after wines.

ibc:
Merge?
BTW Changelog changes missing.

jmillan:
Adding changelog..

PR #1251, opened 2023-12-04
Node: jest, add modulePathIgnorePatters option

jmillan:
Indicate Jest not to look in worker/ nor rust/ for node modules.

Since flatbuffers is a npm module and also a subproject in worker folder, jest was complaining that there were two different modules for flatbuffers

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
       "node/src/fbs",
       "node/src/tests"
     ],
-    "cacheDirectory": ".cache/jest"
+    "cacheDirectory": ".cache/jest",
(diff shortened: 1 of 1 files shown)


ibc (approved):

nazar-pc (commented):
Is there include option? I'd prefer include option instead of exclude such that random things don't get included accidentally in the future.

jmillan:
> Is there include option? I'd prefer include option instead of exclude such that random things don't get included accidentally in the future.

Unfortunately there is not https://jestjs.io/docs/configuration

ibc:
Let's merge?

Issue #1252, opened 2023-12-04
npm install fails with compilation error


lelikg:
# Operating system:
Debian Buster
- Node version:
18.14
- npm version:
9.3.1
- mediasoup version:
latest

### Issue description
When running npm install, fails with:
```
#12 177.3 npm ERR! [778/1349] Compiling C++ object mediasoup-worker.p/src_RTC_Transport.cpp.o
#12 177.3 npm ERR! FAILED: mediasoup-worker.p/src_RTC_Transport.cpp.o 
[...]

PR #1253, opened 2023-12-05
Disable liburing by default

jmillan:
While investigating issue #1252, disable liburing by default.

In order to enable it, use MESON_ARGS as usual. ie:

`MESON_ARGS=-Dms_enable_liburing=true npm install`


ibc (approved):

ibc:
Changelog and release needed then.

PR #1258, opened 2023-12-11
liburing: avoid extra memcpy on RTP

jmillan:
Use an already allocated send buffer to directly encrypt the RTP rather than using the static EncryptBuffer that needs to be memcpy-ed before sending.

If there are no send buffers available, the EncryptBuffer will be used.

This saves 1% CPU. (Tested in 1 Producer -> 40 audio/video Consumers Router).

### DIFF
--- a/worker/src/DepLibUring.cpp
+++ b/worker/src/DepLibUring.cpp
@@ -335,7 +335,7 @@ uint8_t* DepLibUring::LibUring::GetSendBuffer()
 
 	if (this->availableUserDataEntries.empty())
 	{
-		MS_WARN_DEV("no user data entry available");
+		MS_DEBUG_DEV("no user data entry available");
 
 		return nullptr;
 	}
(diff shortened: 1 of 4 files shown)


worker/src/DepLibUring.cpp
```diff
@@ -319,8 +329,24 @@ void DepLibUring::LibUring::StopPollingCQEs()
 	uv_close(reinterpret_cast<uv_handle_t*>(this->uvHandle), static_cast<uv_close_cb>(onCloseFd));
 }
 
+uint8_t* DepLibUring::LibUring::GetSendBuffer()
+{
+	MS_TRACE();
+
+	if (this->availableUserDataEntries.empty())
+	{
+		MS_WARN_DEV("no user data entry available");
```
ibc: Should this really be a warning (despite it's a dev log). I think it's a perfectly legitimate case so a debug should be fine.

ibc (approved):
Minor comment, approved

ibc:
Conflicts

PR #1260, opened 2023-12-12
libsrtp: use our own fork

jmillan:
It uses a list hash rather than a single list to store stream contexts.

By distributing the stream contexts in multiple lists (32) we enhance the list traversal time for the stream we are looking, by 32 times in the best case.

I made a PR in the project but it didn't succeed https://github.com/cisco/libsrtp/pull/659

They'll work on a different solution. In the meantime we can use this approach. It saves 2% CPU for a Router with 1 AV Producer and 40 A/V Consumers. The gains are bigger as the number of Consumers increase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### NEXT
 
 * liburing: avoid extra memcpy on RTP ([PR #1258](https://github.com/versatica/mediasoup/pull/1258)).
+* libsrtp: use our own fork ([PR #1260](https://github.com/versatica/mediasoup/pull/1260)).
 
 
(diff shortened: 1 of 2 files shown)


nazar-pc (commented):
Is list traversal that expensive, what kind of difference are we talking about here?

jmillan:
> Is list traversal that expensive, what kind of difference are we talking about here?

Data added in the PR description. As the number of streams per context increase, the difference is bigger. Here an extreme case where the list traversal takes more than encrypting:

Screenshot 2023-11-17 at 18 00 49 (1)

ibc (approved):

PR #1261, opened 2023-12-12
Make subchannels be ordered

ibc:
### Details

- Use an ordered set in `DataConsumer` in worker so `subchannels` will be always shown in numerical order (no need to sort them later in Node/Rust layer).
- Also make `dataConsumer.subchannels` be **immediately** (and optimistically) updated before `dataChannel.setSubchannels()` completes.

### DIFF
--- a/node/src/DataConsumer.ts
+++ b/node/src/DataConsumer.ts
@@ -565,6 +565,18 @@ export class DataConsumer<DataConsumerAppData extends AppData = AppData>
 	{
 		logger.debug('setSubchannels()');
 
+		// Optimistic subchannels update. We do this to make the |subchannels|
+		// getter be immediately updated with given values. Later it will be
+		// updated again with the response from the worker.
+		const subchannelsSet = new Set<number>();
+
+		for (const subchannel of subchannels)
+		{
+			subchannelsSet.add(Math.abs(subchannel % 65536));
+		}
+
+		this.#subchannels = Array.from(subchannelsSet).sort();
+
 		/* Build Request. */
 		const requestOffset = new FbsDataConsumer.SetSubchannelsRequestT(
 			subchannels
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### NEXT
 
 * liburing: avoid extra memcpy on RTP ([PR #1258](https://github.com/versatica/mediasoup/pull/1258)).
+* Make subchannels be ordered ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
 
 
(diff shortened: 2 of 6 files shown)


jmillan (approved):

node/src/DataConsumer.ts
```diff
@@ -565,6 +565,18 @@ export class DataConsumer<DataConsumerAppData extends AppData = AppData>
...
+		// getter be immediately updated with given values. Later it will be
+		// updated again with the response from the worker.
+		const subchannelsSet = new Set<number>();
+
+		for (const subchannel of subchannels)
+		{
+			subchannelsSet.add(Math.abs(subchannel % 65536));
+		}
+
+		this.#subchannels = Array.from(subchannelsSet).sort();
```
jmillan, line 578: Why don't we simply wait for the response to update `#subchannels`?

ibc: Please read the PR description.

jmillan: > Also make dataConsumer.subchannels be immediately (and optimistically) updated before dataChannel.setSubchannels() completes.

I read, but I wonder *why* do we need this.

ibc: * Some part of the app does: `dataConsumer.setSubchannels([ 1, 2, 3 ])`. It takes some times to complete.
* In the meantime `dataConsumer.subchannels` remains `[ ]` (empty).
* Another part of the app wants to add subchannel 4: `dataConsumer.setSubchannels([ ...dataConsumer.subchannels, 4 ])`.
* Result? subchannels are `[ 4 ]` instead of the intended `[ 1, 2, 3, 4 ]`.

jmillan: I see, so the only concern is that some `temporal` subchannel does not end up being a real one. I don't remember and there may be a reason for not doing it, I'm just asking: wouldn't `addSubchannels()`, `removeSubchannels()` make sense? Or is it too much clutter creating those new methods?

ibc: > I see, so the only concern is that some temporal subchannel does not end up being a real one.

That's why it's called optimistic update, but if you look at the code it's guaranteed that it behaves fine.

> I'm just asking: wouldn't addSubchannels(), removeSubchannels() make sense? 

Too much clutter IMHO.

ibc:
I'm closing this PR since, as @jmillan said, adding `addSubchannel()` and `emoveSubchannel()` makes more sense. Also, in other async methods (such as `pause()`) we don't update `this.paused` until we get the response fro the worker, so same must be done in `setSubchannels()`.

PR #1262, opened 2023-12-12
Fix Rust DataConsumer

ibc:
- `set_subchannels()` method must exist not only in `DirectDataConsumer` but also in `RegularDataConsumer`.
- Fix wrong usage of `DataConsumer` in tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 * liburing: avoid extra memcpy on RTP ([PR #1258](https://github.com/versatica/mediasoup/pull/1258)).
 * libsrtp: use our own fork ([PR #1260](https://github.com/versatica/mediasoup/pull/1260)).
-* Fix Rust `DataConsumer` in tests ([PR #1262](https://github.com/versatica/mediasoup/pull/1262)).
+* Fix Rust `DataConsumer` ([PR #1262](https://github.com/versatica/mediasoup/pull/1262)).
 
 
 ### 3.13.10
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * liburing: avoid extra memcpy on RTP ([PR #1258](https://github.com/versatica/mediasoup/pull/1258)).
 * libsrtp: use our own fork ([PR #1260](https://github.com/versatica/mediasoup/pull/1260)).
+* Fix Rust `DataConsumer` in tests ([PR #1262](https://github.com/versatica/mediasoup/pull/1262)).
 
 
 ### 3.13.10
--- a/node/src/tests/test-DataConsumer.ts
+++ b/node/src/tests/test-DataConsumer.ts
@@ -71,7 +71,7 @@ test('transport.consumeData() succeeds', async () =>
 	expect(dataConsumer1.label).toBe('foo');
 	expect(dataConsumer1.protocol).toBe('bar');
 	expect(dataConsumer1.paused).toBe(false);
-	expect(dataConsumer1.subchannels.sort((a, b) => a - b)).toEqual([ 0, 1, 2, 100, 65535 ]);
+	expect(dataConsumer1.subchannels).toEqual([ 0, 1, 2, 100, 65535 ]);
 	expect(dataConsumer1.appData).toEqual({ baz: 'LOL' });
 
 	const dump = await router.dump();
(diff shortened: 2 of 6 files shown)


ibc:
### Issue (NOTE: FIXED)

cargo fails with this:

```
cargo test
   Compiling mediasoup-sys v0.7.0 (/Users/ibc/src/v3-mediasoup/worker)
   Compiling mediasoup v0.13.0 (/Users/ibc/src/v3-mediasoup/rust)
error[E0599]: no method named `set_subchannels` found for reference `&DirectDataConsumer` in the current scope
   --> rust/tests/integration/direct_transport.rs:518:14
    |
517 | /         direct_data_consumer_2
518 | |             .set_subchannels(subchannels)
    | |             -^^^^^^^^^^^^^^^ method not found in `&DirectDataConsumer`
    | |_____________|
    |

For more information about this error, try `rustc --explain E0599`.
error: could not compile `mediasoup` (test "integration") due to previous error
```


Comments by @nazar-pc:

> There is probably `Deref` implementation, that dereferences to underlying data structure that provides generic methods.

> The same way as `Vec<u8>` dereferences to `&[u8]`, so you can call slice methods on `Vec<u8>` directly.

> https://doc.rust-lang.org/std/ops/trait.Deref.html
> It is pretty neat actually, but I admit might be a bit confusing at first. It also goes exactly one level, so it is fairly straightforward to work with.

> There are a few things. `DirectDataConsumer impl` is in a very different place than struct definition, I don't like that, they should be close to each other. Second, You have `DataConsumer`, it is an `enum`. You can match and get inner variant or you can turn inner variant back into enum with `From` implementation.

> It is literally what it sounds like https://github.com/versatica/mediasoup/blob/042b7f5ba1ce41cdf0236070b027a6ee7e1f9a9b/rust/src/router/data_consumer.rs#L436-L442. `Enum` is just a fancy struct, it can have methods just like any other struct can in Rust.

jmillan:
`data_consumer_2` there should be used as it is. 

```diff
diff --git a/rust/tests/integration/direct_transport.rs b/rust/tests/integration/direct_transport.rs
index 380f3e1b3..6053de020 100644
--- a/rust/tests/integration/direct_transport.rs
+++ b/rust/tests/integration/direct_transport.rs
@@ -430,7 +430,7 @@ fn send_with_subchannels_succeeds() {
             }
         };

-        let direct_data_consumer_2 = match &data_consumer_2 {
+        let _ = match &data_consumer_2 {
             DataConsumer::Direct(direct_data_consumer) => direct_data_consumer,
             _ => {
                 panic!("Expected direct data consumer")
@@ -514,7 +514,7 @@ fn send_with_subchannels_succeeds() {
         let mut subchannels = data_consumer_2.subchannels();
         subchannels.push(1);

-        direct_data_consumer_2
+        data_consumer_2
             .set_subchannels(subchannels)
             .await
             .expect("Failed to set subchannels");
```

All methods are implemented for `DataConsumer` regardless it's a `RegularDataConsumer` or `DirectDataConsumer`. There is no need to duplicate those.

ibc:
```
cargo test
   Compiling mediasoup v0.13.0 (/Users/ibc/src/v3-mediasoup/rust)
warning: unused variable: `router`
   --> rust/tests/integration/data_consumer.rs:323:23
    |
323 |         let (_worker, router, transport1, data_producer) = init().await;
    |                       ^^^^^^ help: if this is intentional, prefix it with an underscore: `_router`
    |
    = note: `#[warn(unused_variables)]` on by default

error: literal out of range for `u16`
   --> rust/tests/integration/data_consumer.rs:338:47
    |
338 |             .set_subchannels([ 999, 999, 998, 65536 ].to_vec())
    |                                               ^^^^^
    |
    = note: the literal `65536` does not fit into the type `u16` whose range is `0..=65535`
    = note: `#[deny(overflowing_literals)]` on by default
```

ibc:
> ```
> cargo test
>    Compiling mediasoup v0.13.0 (/Users/ibc/src/v3-mediasoup/rust)
> warning: unused variable: `router`
>    --> rust/tests/integration/data_consumer.rs:323:23
>     |
> 323 |         let (_worker, router, transport1, data_producer) = init().await;
>     |                       ^^^^^^ help: if this is intentional, prefix it with an underscore: `_router`
>     |
>     = note: `#[warn(unused_variables)]` on by default
> 
> error: literal out of range for `u16`
>    --> rust/tests/integration/data_consumer.rs:338:47
>     |
> 338 |             .set_subchannels([ 999, 999, 998, 65536 ].to_vec())
>     |                                               ^^^^^
>     |
>     = note: the literal `65536` does not fit into the type `u16` whose range is `0..=65535`
>     = note: `#[deny(overflowing_literals)]` on by default
> ```

I'm on it.

jmillan (approved):

PR #1266, opened 2023-12-15
Enables multiple transports to bind to the same port

mmasaki:
This change specifies UV_UDP_REUSEADDR  and SO_REUSEPORT for the socket in Transport::UDP.
In Linux, by combining the multicast address and SO_REUSEPORT, multiple transports across workers can bind to the same port.
When distributing a stream to multiple workers, this eliminates the need to send the stream via unicast to each worker.
This should be convenient for broadcasting.

Example:

```typescript
const same_port = 1234;
const router1 = await worker1.createRouter(...);
const router2 = await worker2.createRouter(...);
transport1 = await router1.createPlainTransport({ listenIp: "224.0.0.1", comedia: true, port: same_port });
transport1 = await router2.createPlainTransport({ listenIp: "224.0.0.1", comedia: true, port: same_port }); // This does not cause an error.
```

### DIFF
--- a/node/src/PlainTransport.ts
+++ b/node/src/PlainTransport.ts
@@ -109,7 +109,8 @@ export type PlainTransportOptions<PlainTransportAppData extends AppData = AppDat
 	srtpCryptoSuite?: SrtpCryptoSuite;
 
 	/**
-	 * Enable Multicast. Default false.
+	 * Enable multicast for UDP in case listening IP is a multicast valid IP.
+	 * Default false.
 	 */
 	enableMulticast?: boolean;
 
@@ -139,7 +140,7 @@ export type PlainTransportObserverEvents = TransportObserverEvents &
 {
 	tuple: [TransportTuple];
 	rtcptuple: [TransportTuple];
-	sctpstatechange: [SctpState];	
+	sctpstatechange: [SctpState];
 };
 
 type PlainTransportConstructorOptions<PlainTransportAppData> =
--- a/node/src/PlainTransport.ts
+++ b/node/src/PlainTransport.ts
@@ -109,6 +109,11 @@ export type PlainTransportOptions<PlainTransportAppData extends AppData = AppDat
 	srtpCryptoSuite?: SrtpCryptoSuite;
 
 	/**
+	 * Enable Multicast. Default false.
+	 */
+	enableMulticast?: boolean;
+
+	/**
 	 * Custom application data.
 	 */
 	appData?: PlainTransportAppData;
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -681,6 +681,7 @@ export class Router<RouterAppData extends AppData = AppData>
 			sctpSendBufferSize = 262144,
 			enableSrtp = false,
 			srtpCryptoSuite = 'AES_CM_128_HMAC_SHA1_80',
+			enableMulticast = false,
 			appData
 		}: PlainTransportOptions<PlainTransportAppData>
 	): Promise<PlainTransport<PlainTransportAppData>>
@@ -726,6 +727,11 @@ export class Router<RouterAppData extends AppData = AppData>
 			};
 		}
 
+		if (enableMulticast)
+		{
+			logger.debug('createPlainTransport() | enableMulticast');
+		}
+
 		const transportId = generateUUIDv4();
 
 		/* Build Request. */
@@ -765,7 +771,8 @@ export class Router<RouterAppData extends AppData = AppData>
 			rtcpMux,
 			comedia,
 			enableSrtp,
-			cryptoSuiteToFbs(srtpCryptoSuite)
+			cryptoSuiteToFbs(srtpCryptoSuite),
+			enableMulticast
 		);
 
 		const requestOffset = new FbsRouter.CreatePlainTransportRequestT(
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -727,11 +727,6 @@ export class Router<RouterAppData extends AppData = AppData>
 			};
 		}
 
-		if (enableMulticast)
-		{
-			logger.debug('createPlainTransport() | enableMulticast');
-		}
-
 		const transportId = generateUUIDv4();
 
 		/* Build Request. */
--- a/node/src/tests/test-PlainTransport.ts
+++ b/node/src/tests/test-PlainTransport.ts
@@ -316,17 +316,39 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
 		.toThrow(Error);
 }, 2000);
 
-test('router.createPlainTransport() with enableMulticast succeeds', async () =>
+if (process.platform === 'linux')
 {
-	// Use default cryptoSuite: 'AES_CM_128_HMAC_SHA1_80'.
-	const transport1 = await router.createPlainTransport(
-		{
-			listenIp   : '127.0.0.1',
-			enableMulticast: true
-		});
+	test('router.createPlainTransport() with enableMulticast succeeds', async () =>
+	{
+		let transport1: mediasoup.types.PlainTransport | undefined = undefined;
+		let transport2: mediasoup.types.PlainTransport | undefined = undefined;
 
-	expect(typeof transport1.id).toBe('string');
-}, 2000);
+		await expect(async () =>
+		{
+			transport1 = await router.createPlainTransport(
+				{
+					listenIp        : '224.0.0.1',
+					enableMulticast : true
+				});
+
+			transport2 = await router.createPlainTransport(
+				{
+					listenIp        : '224.0.0.1',
+					enableMulticast : true
+				});
+		}).not.toThrow();
+
+		if (transport1)
+		{
+			(transport1 as mediasoup.types.PlainTransport).close();
+		}
+
+		if (transport2)
+		{
+			(transport2 as mediasoup.types.PlainTransport).close();
+		}
+	}, 2000);
+}
 
 test('plainTransport.getStats() succeeds', async () =>
 {
--- a/node/src/tests/test-PlainTransport.ts
+++ b/node/src/tests/test-PlainTransport.ts
@@ -316,6 +316,18 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
 		.toThrow(Error);
 }, 2000);
 
+test('router.createPlainTransport() with enableMulticast succeeds', async () =>
+{
+	// Use default cryptoSuite: 'AES_CM_128_HMAC_SHA1_80'.
+	const transport1 = await router.createPlainTransport(
+		{
+			listenIp   : '127.0.0.1',
+			enableMulticast: true
+		});
+
+	expect(typeof transport1.id).toBe('string');
+}, 2000);
+
 test('plainTransport.getStats() succeeds', async () =>
 {
 	const data = await transport.getStats();
--- a/rust/src/router/plain_transport.rs
+++ b/rust/src/router/plain_transport.rs
@@ -73,6 +73,9 @@ pub struct PlainTransportOptions {
     /// The SRTP crypto suite to be used if enableSrtp is set.
     /// Default 'AesCm128HmacSha180'.
     pub srtp_crypto_suite: SrtpCryptoSuite,
+    /// Enable Multicast.
+    /// Default false.
+    pub enable_multicast: bool,
     /// Custom application data.
     pub app_data: AppData,
 }
--- a/rust/src/router/plain_transport.rs
+++ b/rust/src/router/plain_transport.rs
@@ -73,7 +73,7 @@ pub struct PlainTransportOptions {
     /// The SRTP crypto suite to be used if enableSrtp is set.
     /// Default 'AesCm128HmacSha180'.
     pub srtp_crypto_suite: SrtpCryptoSuite,
-    /// Enable Multicast.
+    /// Enable multicast for UDP in case listening IP is a multicast valid IP.
     /// Default false.
     pub enable_multicast: bool,
     /// Custom application data.
@@ -95,6 +95,7 @@ impl PlainTransportOptions {
             sctp_send_buffer_size: 262_144,
             enable_srtp: false,
             srtp_crypto_suite: SrtpCryptoSuite::default(),
+            enable_multicast: false,
             app_data: AppData::default(),
         }
     }
--- a/rust/tests/integration/plain_transport.rs
+++ b/rust/tests/integration/plain_transport.rs
@@ -265,6 +265,51 @@ fn create_succeeds() {
                 assert_eq!(transport_dump.sctp_state, transport2.sctp_state());
             }
         }
+
+        #[cfg(target_os = "linux")]
+        {
+            let transport1 = router
+                .create_plain_transport({
+                    let mut plain_transport_options = PlainTransportOptions::new(ListenInfo {
+                        protocol: Protocol::Udp,
+                        ip: "224.0.0.1".parse::<IpAddr>().unwrap(),
+                        announced_ip: None,
+                        port: Some(1234),
+                        send_buffer_size: None,
+                        recv_buffer_size: None,
+                    });
+                    plain_transport_options.enable_multicast = true;
+
+                    plain_transport_options
+                })
+                .await
+                .expect("Failed to create Plain transport");
+
+            let transport2 = router
+                .create_plain_transport({
+                    let mut plain_transport_options = PlainTransportOptions::new(ListenInfo {
+                        protocol: Protocol::Udp,
+                        ip: "224.0.0.1".parse::<IpAddr>().unwrap(),
+                        announced_ip: None,
+                        port: Some(1234),
+                        send_buffer_size: None,
+                        recv_buffer_size: None,
+                    });
+                    plain_transport_options.enable_multicast = true;
+
+                    plain_transport_options
+                })
+                .await
+                .expect("Failed to create Plain transport");
+
+            let router_dump = router.dump().await.expect("Failed to dump router");
+            assert_eq!(router_dump.transport_ids, {
+                let mut set = HashedSet::default();
+                set.insert(transport1.id());
+                set.insert(transport2.id());
+                set
+            });
+        }
     });
 }
 
--- a/worker/include/RTC/PlainTransport.hpp
+++ b/worker/include/RTC/PlainTransport.hpp
@@ -85,6 +85,7 @@ namespace RTC
 		size_t srtpMasterLength{ 0 };
 		std::string srtpKeyBase64;
 		bool connectCalled{ false }; // Whether connect() was succesfully called.
+		bool multicast{ false };
 	};
 } // namespace RTC
 
--- a/worker/include/RTC/PlainTransport.hpp
+++ b/worker/include/RTC/PlainTransport.hpp
@@ -85,7 +85,6 @@ namespace RTC
 		size_t srtpMasterLength{ 0 };
 		std::string srtpKeyBase64;
 		bool connectCalled{ false }; // Whether connect() was succesfully called.
-		bool multicast{ false };
 	};
 } // namespace RTC
 
--- a/worker/include/Utils.hpp
+++ b/worker/include/Utils.hpp
@@ -86,6 +86,8 @@ namespace Utils
 		}
 
 		static void NormalizeIp(std::string& ip);
+
+		static bool IsMulticast(const struct sockaddr_storage* addr, const int& family);
 	};
 
 	class File
--- a/worker/include/Utils.hpp
+++ b/worker/include/Utils.hpp
@@ -87,7 +87,7 @@ namespace Utils
 
 		static void NormalizeIp(std::string& ip);
 
-		static bool IsMulticast(const struct sockaddr_storage* addr, const int& family);
+		static bool IsMulticast(const struct sockaddr_storage* addr, const int family);
 	};
 
 	class File
--- a/worker/src/RTC/PlainTransport.cpp
+++ b/worker/src/RTC/PlainTransport.cpp
@@ -155,18 +155,17 @@ namespace RTC
 			this->srtpKeyBase64 = Utils::String::Base64Encode(this->srtpKey);
 		}
 
-		this->multicast = options->enableMulticast();
-
 		try
 		{
 			// This may throw.
 			if (this->listenInfo.port != 0)
 			{
-				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->listenInfo.port, this->multicast);
+				this->udpSocket = new RTC::UdpSocket(
+				  this, this->listenInfo.ip, this->listenInfo.port, options->enableMulticast());
 			}
 			else
 			{
-				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->multicast);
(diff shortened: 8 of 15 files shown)


nazar-pc (commented):
Multiple workers can bind to the same port, but only one of them will get the message, so broadcast use case is not possible. Also the rest of the mediasoup code needs to be aware of this and distinguish which worker message belongs to.

mmasaki:
@nazar-pc In Linux, multiple processes can receive the same packets simultaneously when binding UDP sockets to a multicast address with SO_REUSEPORT. This is a special behavior.

nazar-pc:
I'm quite confident you're misunderstanding it. In case of both TCP and UDP it allows distribution of connections/datagrams across listeners, not broadcasting. Distribution means one datagram will be processed by one worker, another by another, all managed by the kernel.

A great example for UDP is DSN server. It would be odd to get multiple DNS responses to a single request, wouldn't it?

mmasaki:
@nazar-pc The normal behavior is as you would expect. However, the behavior is special only when binding to a multicast address with SO_REUSEPORT. In this case, multiple processes can receive the same packets.

nazar-pc:
Okay, I read about it more, and it only has the effect you expect on multicast addresses, not on any address (which this PR doesn't check), which makes this a very niche use case. And even then I'm not sure if some feedback will be necessary, which will not work properly in this case, but might be functional for plain transport specifically.

ibc:
> it only has the effect you expect on multicast addresses, not on any address (which this PR doesn't check), which makes this a very niche use case.

Exactly. This PR is unconditionally applying "reuse port" to the every UDP open port. This means that, if by accident the app binds into a non multicast address in same UDP port in 2 workers, the second transport won't fail but the behavior later will be completely erratic with UDP packets being randomly distributed to one worker or another. This is not an acceptable change since it prevents a legitimate user error from being noticed.

ibc:
I don't know anything about multicast so I need to understand with an easy and detailed use case how this feature is supposed to behave. In which case we want to use a multicast address? What would happen if two clients send RTP packets to different transports in different workers in same multicast IP and port? Will each UDP packet reach both workers? If so each packet maybe ignored in one of those workers due to "unknown ssrc" or similar. And it would be worse in WebRtcTransports where clients will first send STUN and DTLS packets that require a response from server side.

mmasaki:
I've updated the patch. It turns out that SO_REUSEPORT is not necessary for this purpose, so I've removed it. The patch now only adds UV_UDP_REUSEADDR in the case of a multicast address. With this change, multiple workers will be able to receive a stream on the same port. I believe this is useful for broadcast purposes and is not a niche use case.

ibc:
@mmasaki I made some specific questions in my comment above: https://github.com/versatica/mediasoup/pull/1266#issuecomment-1857758624

I promise that we cannot merge something that we don't understand, so please answer those questions.

mmasaki:
> In which case we want to use a multicast address?

This is useful for distributing the same video/audio to multiple workers for live streaming purposes.

>What would happen if two clients send RTP packets to different transports in different workers in same multicast IP and port?

This feature is intended to be used with `comedia: true`. If two clients send packets simultaneously, as before, the client whose packets arrive first will be choosed.

> Will each UDP packet reach both workers? 

Yes. That's why it is useful for broadcast purposes.
This change will allow us to easily distribute video/audio to multiple workers using tools like ffmpeg or gstreamer.

Example:

```shell
same_port=1234
# we can distribute video/audio to workers with one simple command
ffmpeg \
	-re \
	-v info \
	-stream_loop -1 \
	-i ${MEDIA_FILE} \
	-map 0:a:0 \
	-acodec libopus -ab 128k -ac 2 -ar 48000 \
	-map 0:v:0 \
	-pix_fmt yuv420p -c:v libvpx -b:v 1000k -deadline realtime -cpu-used 4 \
	-f tee \
	"[select=a:f=rtp:ssrc=${AUDIO_SSRC}:payload_type=${AUDIO_PT}]rtp://${audioTransportIp}:${same_port}|[select=v:f=rtp:ssrc=${VIDEO_SSRC}:payload_type=${VIDEO_PT}]rtp://${videoTransportIp}:${same_port}"
```

> If so each packet maybe ignored in one of those workers due to "unknown ssrc" or similar. And it would be worse in WebRtcTransports where clients will first send STUN and DTLS packets that require a response from server side.

This feature is intended to be used with plainTransport.

ibc:
Problems of this PR are:

1. It's intended only for plain transport, but nothing in the code verifies it. A user may accidentally use same multicast IP for WebRTC transports and things would work erratically. So the way to do this should be by adding a specific new `enableMulticast` Boolean option in `PlainTransportOptions` in Node and Rust.
2. This PR is checking whether given address is multicast in PortManager.cpp. It should be a new `Utils::IP::IsMulticast(addr)` utility in Utils.hpp instead.
3. PortManager.bind() methods should have a new `enableMulticast` argument and it should only be set to true when called from PlainTransport.cpp constructor and only if PlainTransportOptions has `enableMulticast`, this is Linux and the given IP is verified to be multicast.

I'll leave this PR on hold and make those required changes when possible, unless you want to make them.

nazar-pc:
I agree this is a potentially useful features, but it does need to be developed a bit further to be mergeable.

mmasaki:
I have tried to resolve the issues. Could you please review it?

node/src/PlainTransport.ts
```diff
@@ -108,6 +108,11 @@ export type PlainTransportOptions<PlainTransportAppData extends AppData = AppDat
 	 */
 	srtpCryptoSuite?: SrtpCryptoSuite;
 
+	/**
+	 * Enable Multicast. Default false.
+	 */
```
ibc:
```suggestion
	/**
	 * Enable multicast for UDP in case listening IP is a multicast valid IP. Default false.
	 */
```

node/src/Router.ts
```diff
@@ -726,6 +727,11 @@ export class Router<RouterAppData extends AppData = AppData>
 			};
 		}
 
+		if (enableMulticast)
```
ibc: We don't need this log. Please remove.

node/src/tests/test-PlainTransport.ts
```diff
@@ -316,6 +316,18 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
 		.toThrow(Error);
 }, 2000);
 
+test('router.createPlainTransport() with enableMulticast succeeds', async () =>
+{
+	// Use default cryptoSuite: 'AES_CM_128_HMAC_SHA1_80'.
+	const transport1 = await router.createPlainTransport(
+		{
+			listenIp   : '127.0.0.1',
+			enableMulticast: true
```
ibc: Please run `npm run lint`. It will show errors here.

node/src/tests/test-PlainTransport.ts
```diff
@@ -316,6 +316,18 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
...
+		});
+
+	expect(typeof transport1.id).toBe('string');
```
ibc: Do not do this useless assertion here. Instead use `expect(await xxxxxx).not.toThrow();`.

Also, you are creating a `transport1` here but you are not closing it once the test ends, so it remains alive until all tests end. Please close the transport1 here.

rust/src/router/plain_transport.rs
```diff
@@ -73,6 +73,9 @@ pub struct PlainTransportOptions {
     /// The SRTP crypto suite to be used if enableSrtp is set.
     /// Default 'AesCm128HmacSha180'.
     pub srtp_crypto_suite: SrtpCryptoSuite,
+    /// Enable Multicast.
+    /// Default false.
+    pub enable_multicast: bool,
```
ibc:
```suggestion
    /// Enable multicast for UDP in case listening IP is a multicast valid IP.
    /// Default false.
    pub enable_multicast: bool,
```

worker/include/Utils.hpp
```diff
@@ -86,6 +86,8 @@ namespace Utils
 		}
 
 		static void NormalizeIp(std::string& ip);
+
+		static bool IsMulticast(const struct sockaddr_storage* addr, const int& family);
```
ibc: No need to pass `family` by reference. The method should NOT be able to change its original value.
```suggestion
		static bool IsMulticast(const struct sockaddr_storage* addr, const int family);
```

worker/src/Utils/IP.cpp
```diff
@@ -155,4 +155,39 @@ namespace Utils
 			}
 		}
 	}
+
+	bool IP::IsMulticast(const struct sockaddr_storage* addr, const int& family)
```
ibc:
```suggestion
	bool IP::IsMulticast(const struct sockaddr_storage* addr, const int family)
```

worker/include/RTC/PlainTransport.hpp
```diff
@@ -85,6 +85,7 @@ namespace RTC
 		size_t srtpMasterLength{ 0 };
 		std::string srtpKeyBase64;
 		bool connectCalled{ false }; // Whether connect() was succesfully called.
+		bool multicast{ false };
```
ibc: This `multicast` class member is not needed at all. Please remove.

worker/src/RTC/PlainTransport.cpp
```diff
@@ -155,16 +155,18 @@ namespace RTC
 			this->srtpKeyBase64 = Utils::String::Base64Encode(this->srtpKey);
 		}
 
+		this->multicast = options->enableMulticast();
```
ibc: This member is not needed at all. Please remove.

worker/src/RTC/PlainTransport.cpp
```diff
@@ -155,16 +155,18 @@ namespace RTC
...
+
 		try
 		{
 			// This may throw.
 			if (this->listenInfo.port != 0)
 			{
-				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->listenInfo.port);
+				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->listenInfo.port, this->multicast);
```
ibc:
```suggestion
				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->listenInfo.port, options->enableMulticast());
```

worker/src/RTC/PlainTransport.cpp
```diff
@@ -155,16 +155,18 @@ namespace RTC
...
 			}
 			else
 			{
-				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip);
+				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, this->multicast);
```
ibc:
```suggestion
				this->udpSocket = new RTC::UdpSocket(this, this->listenInfo.ip, options->enableMulticast());
```

worker/src/RTC/PlainTransport.cpp
```diff
@@ -185,11 +187,11 @@ namespace RTC
 				if (this->rtcpListenInfo.port != 0)
 				{
 					this->rtcpUdpSocket =
-					  new RTC::UdpSocket(this, this->rtcpListenInfo.ip, this->rtcpListenInfo.port);
+					  new RTC::UdpSocket(this, this->rtcpListenInfo.ip, this->rtcpListenInfo.port, this->multicast);
```
ibc:
```suggestion
					  new RTC::UdpSocket(this, this->rtcpListenInfo.ip, this->rtcpListenInfo.port, options->enableMulticast());
```

worker/src/RTC/PlainTransport.cpp
```diff
@@ -185,11 +187,11 @@ namespace RTC
...
 				}
 				else
 				{
-					this->rtcpUdpSocket = new RTC::UdpSocket(this, this->rtcpListenInfo.ip);
+					this->rtcpUdpSocket = new RTC::UdpSocket(this, this->rtcpListenInfo.ip, this->multicast);
```
ibc:
```suggestion
					this->rtcpUdpSocket = new RTC::UdpSocket(this, this->rtcpListenInfo.ip, options->enableMulticast());
```

worker/src/RTC/PortManager.cpp
```diff
@@ -217,6 +217,14 @@ namespace RTC
 			{
 				case Transport::UDP:
 				{
+#if defined(__linux__)
+					if (enableMulticast && Utils::IP::IsMulticast(&bindAddr, family))
+					{
+						flags |= UV_UDP_REUSEADDR;
+						MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");
```
ibc: Logs must be placed first and blank line required.

```suggestion
					{
						MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");

						flags |= UV_UDP_REUSEADDR;
```

worker/src/RTC/PortManager.cpp
```diff
@@ -217,6 +217,14 @@ namespace RTC
 			{
 				case Transport::UDP:
 				{
+#if defined(__linux__)
+					if (enableMulticast && Utils::IP::IsMulticast(&bindAddr, family))
```
ibc: Let's use modern C++:

```suggestion
					if (enableMulticast && Utils::IP::IsMulticast(std::addressof(bindAddr), family))
```

worker/src/RTC/PortManager.cpp
```diff
@@ -472,6 +480,14 @@ namespace RTC
 		{
 			case Transport::UDP:
 			{
+#if defined(__linux__)
+				if (enableMulticast && Utils::IP::IsMulticast(&bindAddr, family))
+				{
+					flags |= UV_UDP_REUSEADDR;
+					MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");
+				}
```
ibc:
```suggestion
				if (enableMulticast && Utils::IP::IsMulticast(std::addressof(bindAddr), family))
				{
					MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");
					
					flags |= UV_UDP_REUSEADDR;
				}
```

worker/src/Utils/IP.cpp
```diff
@@ -155,4 +155,39 @@ namespace Utils
...
+					return true;
+				}
+				break;
+			}
+
+			default:
+			{
+				MS_THROW_TYPE_ERROR("invalid family");
+			}
+		}
```
ibc:
```suggestion
		switch (family)
		{
			case AF_INET:
			{
				uint32_t s_addr = ntohl((reinterpret_cast<const struct sockaddr_in*>(addr))->sin_addr.s_addr);

				return (s_addr >= 0xe0000000 && s_addr <= 0xefffffff);
			}

			case AF_INET6:
			{
				uint8_t _s6_addr = (reinterpret_cast<const struct sockaddr_in6*>(addr))->sin6_addr.s6_addr[0];
				
				return (_s6_addr == 0xFF);
			}

			default:
			{
				MS_THROW_TYPE_ERROR("invalid family");
			}
		}
```

worker/src/Utils/IP.cpp
```diff
@@ -155,4 +155,39 @@ namespace Utils
...
+
+		return false;
```
ibc:
```suggestion
```

node/src/tests/test-PlainTransport.ts
```diff
@@ -316,6 +316,18 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
 		.toThrow(Error);
 }, 2000);
 
+test('router.createPlainTransport() with enableMulticast succeeds', async () =>
```
ibc: This test must also be done in Rust.

ibc (changes requested):

mmasaki:
@ibc Thank you for reviewing. I have applied your suggestions.

node/src/tests/test-PlainTransport.ts
```diff
@@ -316,6 +316,16 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
...
+{
+	// Use default cryptoSuite: 'AES_CM_128_HMAC_SHA1_80'.
+	await expect(router.createPlainTransport(
```
ibc: This transport is not being closed within the test. It must be closed. So this must be done another way. Pseudo code:

let transport: PlainTransport;

expect(transport = await router.createPlainTransport(....)).not.toThrow());

transport.close();

rust/tests/integration/plain_transport.rs
```diff
@@ -265,6 +265,32 @@ fn create_succeeds() {
...
+                })
+                .await
+                .expect("Failed to create Plain transport");
+
+            let router_dump = router.dump().await.expect("Failed to dump router");
+            assert_eq!(router_dump.transport_ids, {
+                let mut set = HashedSet::default();
+                set.insert(transport.id());
+                set
+            });
```
ibc, line 311: I think we need the same here: locally created transport must be closed here. Perhaps not true in Rust because it would be closed once its scope ends. Is it right @nazar-pc?

worker/src/RTC/PortManager.cpp
```diff
@@ -217,6 +217,15 @@ namespace RTC
...
 				case Transport::UDP:
 				{
+#if defined(__linux__)
+					if (enableMulticast && Utils::IP::IsMulticast(std::addressof(bindAddr), family))
+					{
+						MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");
+
+						flags |= UV_UDP_REUSEADDR;
+					}
+#endif
```
ibc, line 227: Only one blank line please.

worker/src/RTC/PortManager.cpp
```diff
@@ -472,6 +481,15 @@ namespace RTC
...
 			case Transport::UDP:
 			{
+#if defined(__linux__)
+				if (enableMulticast && Utils::IP::IsMulticast(std::addressof(bindAddr), family))
+				{
+					MS_DEBUG_DEV("enabling UV_UDP_REUSEADDR");
+					
+					flags |= UV_UDP_REUSEADDR;
+				}
+#endif
```
ibc, line 490: Only one blank line please.

rust/tests/integration/plain_transport.rs
ibc: @nazar-pc could you review this part? Definitely I don't understand this "set" thing here.

ibc:
This looks good. Just a bit of patience please since they are complex days and probably will merge this on next week or beginning of 2024.

rust/tests/integration/plain_transport.rs
nazar-pc: This is probably fine for rust, it'll close on drop automatically

ibc: I don't know what that exactly means.

nazar-pc: When `PlainTransport` is dropped (goes out of scope), it's `.close()` will be triggered automatically it is not closed yet:
https://github.com/versatica/mediasoup/blob/8365c62fde60337d8bffd56ee414ef2eab45b676/rust/src/router/plain_transport.rs#L418-L424

ibc: Yes I know that. So any change needed? Can you approve otherwise?

nazar-pc: I don't think this test is actually testing it properly.

Port is not specified, two transports with the same port are not created to test that it still succeeds. I get it that on the backend some flag is set, but there is no observable behavior difference here that is being actually checked.

mmasaki: I fixed the tests to use two transports with the same port.

ibc: Same test should be done in Node, and BTW this test is gonna succeed in Linux and fail in other platforms (by design), so this test must be different for Linux and macOS/Windows.

mmasaki: I did the same in node.
Now these tests will only be executed on Linux.

ibc:
@mmasaki could you please enable permissions for us to push to your branch?

node/src/tests/test-PlainTransport.ts
```diff
@@ -316,6 +316,40 @@ test('router.createPlainTransport() with non bindable IP rejects with Error', as
...
+					listenIp        : '224.0.0.1',
+					enableMulticast : true
+				});
+
+			transport2 = await router.createPlainTransport(
+				{
+					listenIp        : '224.0.0.1',
+					enableMulticast : true
+				});
+		}).not.toThrow();
```
ibc, line 339: What is this test asserting? transport1 and transport2 are using random listening UDP ports. This test passes in macOS in v3 branch without any multicast setup.

ibc: This test should be as follows:

```ts
test('router.createPlainTransport() with enableMulticast succeeds', async () =>
{
	let transport1: mediasoup.types.PlainTransport | undefined;
	let transport2: mediasoup.types.PlainTransport | undefined;

	await expect(async () =>
	{
		const multicastIp = '224.0.0.1';
		const port = await pickPort({ ip: multicastIp, reserveTimeout: 0 });

		transport1 = await router.createPlainTransport(
			{
				listenInfo      :
				{
					protocol : 'udp',
					ip       : multicastIp,
					port     : port
				},
				enableMulticast : true
			});

		transport2 = await router.createPlainTransport(
			{
				listenInfo      :
				{
					protocol : 'udp',
					ip       : multicastIp,
					port     : port
				},
				enableMulticast : true
			});
	}).not.toThrow();

	transport1?.close();
	transport2?.close();
}, 2000);
```

And surprise, by adding the `UV_UDP_REUSEADDR`, this test **also** passes on macOS (because it's BSD which also supports multicast). So why is this PR hardcoding `if Linux` everywhere to enable multicast?

This should be done in `worker/meson.build` by defining a macro, the same way we pass `-DMS_LIBURING_SUPPORTED`.

jmillan: Indeed, we are playing `libuv` here which is multi-platform and does, as shown in the code, use this flag for Windows too. https://github.com/libuv/libuv/blob/master/src/win/udp.c#L559

Meaning this feature needs not to live under an activation flag.

ibc:
Actually I'm reconsidering the way this PR is done. I don't think we need any special `enableMulticast` option just for `PlainTransport`. Instead we can add `flags` to `TransportListenInfo` struct and let the user pass socket flags. I'll write a PR.

jmillan:
@mmasaki, how does this work for RTCP?. In multicast, will the RTCP generated by the worker properly reach the sender?

ibc:
BTW from libuv docs https://docs.libuv.org/en/v1.x/udp.html#c.uv_udp_flags:

```c
    /*
    * Indicates if SO_REUSEADDR will be set when binding the handle in
    * uv_udp_bind.
    * This sets the SO_REUSEPORT socket flag on the BSDs and OS X. On other
    * Unix platforms, it sets the SO_REUSEADDR flag. What that means is that
    * multiple threads or processes can bind to the same address without error
    * (provided they all set the flag) but only the last one to bind will receive
    * any traffic, in effect "stealing" the port from the previous listener.
    */
    UV_UDP_REUSEADDR = 4,
```

ibc:
I'm working on this PR: https://github.com/versatica/mediasoup/pull/1291

ibc:
I'm closing this PR in favour of already merged PR https://github.com/versatica/mediasoup/pull/1291. Thanks.

PR #1268, opened 2023-12-15
JS: Require curly braces in switch cases

ibc:
### Details

- Require `{` `}` in each `case` block.
- There is no real ESLint rule for it so we must use this:
  ```js
  'no-restricted-syntax'      : [ 2,
    {
      'selector' : 'SwitchCase > *.consequent[type!="BlockStatement"]'
      'message'  : 'Switch cases without blocks are disallowed'
    }
  ]
  ```

### DIFF
--- a/node/.eslintrc.js
+++ b/node/.eslintrc.js
@@ -48,7 +48,7 @@ const eslintConfig =
 		// Unfortunatelly `curly` does not apply to blocks in `switch` cases so
 		// this is needed.
 		// NOTE: We disable this rule since it's producing false positives:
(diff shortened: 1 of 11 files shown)


jmillan:
@ibc, this rule is working fine:  `'selector' : 'SwitchCase > *.consequent[type!="BlockStatement"]',`

NOTE: It does not give those false positives.

ibc:
> 'selector' : 'SwitchCase > *.consequent[type!="BlockStatement"]'

AMAZING, applied.

PR #1274, opened 2023-12-20
worker: Do not use references for async callbacks

jmillan:
If the callback is not executed in the same uv_loop iteration, the access to such values upon executing the callback is undefined.

### DIFF
--- a/worker/include/RTC/SenderBandwidthEstimator.hpp
+++ b/worker/include/RTC/SenderBandwidthEstimator.hpp
@@ -89,7 +89,7 @@ namespace RTC
 	public:
 		void TransportConnected();
 		void TransportDisconnected();
-		void RtpPacketSent(SentInfo& sentInfo);
+		void RtpPacketSent(const SentInfo& sentInfo);
 		void ReceiveRtcpTransportFeedback(const RTC::RTCP::FeedbackRtpTransportPacket* feedback);
 		void EstimateAvailableBitrate(CummulativeResult& cummulativeResult);
 		void UpdateRtt(float rtt);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 
+### NEXT
+
+* worker: Do not use references for async callbacks ([PR #1274](https://github.com/versatica/mediasoup/pull/1274)).
+
+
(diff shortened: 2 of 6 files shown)


jmillan:
TODO: CHANGELOG

worker/include/RTC/SenderBandwidthEstimator.hpp
```diff
@@ -89,7 +89,7 @@ namespace RTC
 	public:
 		void TransportConnected();
 		void TransportDisconnected();
-		void RtpPacketSent(SentInfo& sentInfo);
+		void RtpPacketSent(const SentInfo& sentInfo);
```
ibc, line 92: Wow unrelated changes hehe

ibc:
What is this fixing?

ibc (approved):

worker/include/RTC/SenderBandwidthEstimator.hpp
jmillan: Nop :-)

Screenshot 2023-12-20 at 15 30 07

jmillan:
> What is this fixing?

If the packet is not sent in the current loop iteration, ie: if `uv_udp_try_send()` cannot send the packet immediately and we need to use `uv_udp_send()` which will send the packet when possible (not in this iteration) and notify us when that is done,  we will run the send callback in the future. In such future, the variables passed by reference to the callback will NOT exist, because they were in the **stack**, back in such previous iteration when `uv_udp_send()` was called.

Accessing references to stack variables that do not exist anymore has undefined behaviour, so we may be probably calling the transport congestion `PacketSent()` methods with garbage.

Also, this reminded me that we were having a crash with `jemalloc` at work, and this may well be related.

worker/include/RTC/SenderBandwidthEstimator.hpp
jmillan: This is was a needed change as we are not creating the `onSendCallback()` out of references.

ibc:
I see. Good. Approved, like my txuleton. Look:

image

jmillan:
> I see. Good. Approved, like my txuleton

0xdeadbeef

ibc:
Plus good dessert and 3 wines and patxaran

PR #1276, opened 2023-12-23
Try GitHub CI actions v4

ibc:
Note that `actions/setup-node@v4` now uses Node 20 instead of Node 16: https://github.com/actions/setup-node/releases/tag/v4.0.0


ibc:
This is ok.

PR #1278, opened 2023-12-26
Update: SctpAssociation::DataProducerClosed

grootgordon:
update draft-ietf-rtcweb-data-channel-13 to standard rfc8831(WebRTC Data Channels)


ibc:
Thanks

PR #1279, opened 2023-12-28
Fix build on musl based systems (such as Alpine Linux)

### LINKED ISSUE #1223: Build fails on Docker node:20-alpine
## Bug Report

I am trying to build mediasoup on docker with an alpine node image, but this fails.

Very simple to reproduce, just init a node project and add mediasoup, like here: https://github.com/copiltembel/mediasoup-docker-alpine-not-building

### Your environment

- Operating system: Ubuntu 20.04
- Node version: v20.9.0
- npm version: 10.1.0
- gcc/clang version: gcc version 12.2.1 20220924 (Alpine 12.2.1_git20220924-r10)
- mediasoup version: 3.13.1
- mediasoup-client version: n/a

### Issue description

Running `docker build .` with a simple mediasoup node project on docker alpine leads to the build failing with this error:

`../../../subprojects/flatbuffers-23.3.3/include/flatbuffers/util.h:228:39: error: 'strtoull_l' was not declared in this scope; did you mean 'strtoull'?`
(See full stack error in the above repo.)

Seems to be related to the flatbuffers update? Googling the error took me here: https://github.com/google/flatbuffers/issues/7265

ibc:
Fixes #1223


worker/meson.build
```diff
@@ -44,6 +44,18 @@ if get_option('ms_rtc_logger_rtp')
...
+
+# This is a workaround to define FLATBUFFERS_LOCALE_INDEPENDENT=0 outside
+# flatbuffers project, which is needed in case the current arch doesn't support
+# strtoll_l or strtoull_l (such as musl in Alpine Linux). Problem is that
+# flatbuffers build system not only defines it for its own usage but also relies
+# on it in its include/flatbuffers/util.h file, and if such a file is included
+# by mediasoup source files (and it is included) build fails with "error:
+# 'strtoull_l' was not declared in this scope".
+# See issue: https://github.com/versatica/mediasoup/issues/1223
+add_project_arguments('-DFLATBUFFERS_LOCALE_INDEPENDENT=@0@'.format(cpp.has_function('strtoull_l')), language: 'cpp')
```
ibc, line 57: NOTE: this is copy paste from the meson.build in flatbuffers.wrap file.

worker/meson.build
```diff
@@ -305,48 +323,48 @@ executable(
 )
 
 test_sources = [
-    'test/src/tests.cpp',
```
ibc, line 308: Indentation fixes here and blank lines here and there.

worker/meson.build
nazar-pc: Is this something we should report upstream maybe? Having flatbuffers depend on locale is odd.

nazar-pc (approved):
Makes perfect sense, IMHO it should be the default in the flatbuffers itself

ibc:
@nazar-pc I created a (bad) PR in flatbuffers with lot of rationale included: https://github.com/google/flatbuffers/pull/8198

Summary:

- The issue is real in flatbuffers.
- It was reported in https://github.com/google/flatbuffers/pull/6773 long ago. The maintainer closed it "because no time so let's hide well known and documented and real bugs".
- A more complex/proper solution is exposed in same issue but nobody did it.

ibc:
I also tried to use `add_global_arguments(xxx)` in `flatbuffers/meson.build` with the hope that it would define the macro for the parent project and other subprojects, but it's not permitted by meson because "it's unreliable". So this workaround is the best I could figure out.

PR #1280, opened 2023-12-28
Try to make mediasoup-worker-prebuild CI task only run in releases

### LINKED ISSUE #1243: Avoid that mediasoup-worker-prebuild CI task runs for every PR
I'm tired of seeing **so many** `mediasoup-worker-prebuild` CI actions running (and failing) despite the task is supposed to only run after a release. All those tasks fail due to "GitHub Releases requires a tag". Problem is that, before failing, they completely build mediasoup, which means lot of useless and wasted time:

CleanShot-2023-11-25-at-12 04 34@2x

In theory, the `mediasoup-worker-prebuild` task should only run after releases, but obviously it's not working as expected (or perhaps the configuration of the task is wrong, no idea):

`mediasoup-worker-prebuild.yaml`

```yaml
name: mediasoup-worker-prebuild

on:
  push:
    # Only trigger for production releases.
    tags:
      - "*.*.*"

  # Only trigger when other workflows succeeded.
  workflow_run:
    workflows: [mediasoup-worker, mediasoup-node, mediasoup-rust]
    types:
      - completed

jobs:
  # [etc]  
```

ibc:
Fixes #1243

While it's true that this PR is not running `mediasoup-worker-prebuild`, I think it's not gonna work. I assume that given the documentation of [softprops/action-gh-release](https://github.com/softprops/action-gh-release). 

Some info: 
- https://github.com/softprops/action-gh-release
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

Bonus tracks:
- [Use " instead if ' in CI yaml files and use fail-fast. true everywhere](https://github.com/versatica/mediasoup/pull/1280/commits/b29c494bded82bb19a518d24810274d3aca9247d)


.github/workflows/mediasoup-worker-prebuild.yaml
```diff
@@ -44,29 +46,36 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        if: startsWith(github.ref, "refs/tags/")
```
nazar-pc: Why not skipping the whole workflow instead? Skipping individual steps looks odd

.github/workflows/mediasoup-worker-prebuild.yaml
```diff
@@ -2,6 +2,8 @@ name: mediasoup-worker-prebuild
 
 on:
   push:
+    branches:
+      - v3
```
nazar-pc: Why did you add `v3` branch if you only want to run this workflow on releases? I believe these conditions are additive.

.github/workflows/mediasoup-worker-prebuild.yaml
ibc: > Why not skipping the whole workflow instead?

How to do that? 😀
That's exactly what I've been trying to do.

.github/workflows/mediasoup-worker-prebuild.yaml
ibc: I did assuming they are not additive...

.github/workflows/mediasoup-worker-prebuild.yaml
nazar-pc: There is `if` at job level as well: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

But if job is only triggered by tag, this check doesn't have any effect in the first place.

.github/workflows/mediasoup-worker-prebuild.yaml
nazar-pc: Maybe you're right, the documentation isn've very clear about this: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

nazar-pc (commented):
Reading https://github.com/versatica/mediasoup/issues/1243, I suspect you're pushing tag first and then you're creating release for that tag later. The error is confusing, but on GitHub you can only attach assets to releases and pre-releases, so if release it missing, upload will fail too. You need to change the workflow to be triggered by release, not tag. And probably exclude `rust-*` because it is not needed for those releases.

ibc:
> You need to change the workflow to be triggered by release, not tag. And probably exclude rust-*

So you mean that my first commit was way better?

https://github.com/versatica/mediasoup/commit/61a645895c3f2b845f4f8a411a9987811ca58d0a

nazar-pc:
> So you mean that my first commit was way better?

I do think it should have been sufficient, also `branches: [v3]` was not necessary there.

BTW you can test how it performs by forking into personal repo and trying some tags/releases, from my experience that is sometimes the only way to find out how it works in practice.

ibc:
@nazar-pc it looks that `branches` and `tags` are indeed acumulative, so I'll remove `branches` completely:

CleanShot-2023-12-29-at-10 22 43@2x

And yes, I¡ll do this in a fork.

ibc:
I'm closing this PR and trying things in a personal mediasoup fork. Cosmetic changes done here are now in a separate PR https://github.com/versatica/mediasoup/pull/1281.

ibc:
@nazar-pc let's follow the discussion and experiments in the issue report: https://github.com/versatica/mediasoup/issues/1243#issuecomment-1871895475

PR #1281, opened 2023-12-29
Improvements in CI files

ibc:
This PR comes from PR #1280 which will be closed until intended changs are tested in a separate mediasoup fork. So just cosmetic CI changes here.


.github/workflows/mediasoup-node.yaml
```diff
@@ -12,6 +12,7 @@ jobs:
   ci:
     strategy:
       matrix:
+        fail-fast: true
```
nazar-pc: This should have been at the level higher here, but also it defaults to `true`, so I don't think it is really a necessary change: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

ibc: Which could explain why all those CI jobs are not even being runned XDDD. Thanks, fixing.

ibc: OMG it's not being fixed... why this??

CleanShot-2023-12-29-at-10 33 15@2x

ibc: ok typo introduced:

CleanShot-2023-12-29-at-10 34 31@2x

ibc:
Ok, merging. Easy one.

Issue #1282, opened 2023-12-29
io-uring not enabled in Linux mediasoup-worker prebuilt since it uses kernel 5.15


ibc:
In `mediasoup-worker-prebuild` CI job we are using an old Ubuntu version 20.04 to create the Linux `mediasoup-worker` prebuilt binary (for reasons as documented inline). Problem is that Ubuntu 20.04 uses kernel 5.15 (<6) so we are not enabling `io-uring` on these binaries :(

```meson
# For Linux let's use an old version of Ubuntu (20.04) that builds the
# mediasoup-worker binary using an old version of GLib, so it will work
# on Linux hosts running more modern GLib versions.
# See https://github.com/versatica/mediasoup/issues/1089.
- os: ubuntu-20.04
  cc: gcc
  cxx: g++
- os: macos-12
  cc: clang
  cxx: clang++
- os: windows-2022
  cc: cl
  cxx: cl
```

We may need a more specific way to identify a distribution and maybe build once for 20.04 and once for 22.04.
An alternative would be to only build for one and let the other compile from source every time.

ibc:
We are using this file name as distribution identifier:

```ts
const WORKER_PREBUILD_TAR = `mediasoup-worker-${PKG.version}-${os.platform()}-${os.arch()}.tgz`;
```

Proposal:

- When in Linux (and only in Linux) we may also add the kernel major version (5, 6, etc) into that string.

nazar-pc:
We probably don't want "Linux" either, there are many distros with different versions of libraries. IIRC we are testing that the binary runs, but it doesn't guarantee it will not fail in runtime.

ibc:
I don't understand, we need to know if it linux. Yes, it's indeed a limitation but we re not ready to create neither super universal binaries nor binaries for archs that we cannot run in CI.

nazar-pc:
Well, theoretically you can run VMs in CI, so strictly speaking it is possible to run any OS, whether it is worth it or not is a separate question of course.

ibc:
> Well, theoretically you can run VMs in CI, so strictly speaking it is possible to run any OS, whether it is worth it or not is a separate question of course.

Yes, but that's a separate story and super hard to accomplish.

ibc:
I'm printing this in different archs:

```ts
console.log('platform-arch-release:', `${os.platform()}-${os.arch()}-${os.release()}`);
```

**Debian GNU/Linux 12:**
[...]

PR #1283, opened 2023-12-29
mediasoup-worker linux prebuilt with io-uring support (part 1)

ibc:
This PR exposes kernel mayor number in the mediasoup-worker binary name to make it possible to have prebuilts for Linux with kernel 5 and 6 (the latter supports `io-uring`).

**NOTE 1:** This is related to issue #1282 but doesn't complete it.

**TODO:** In `mediasoup-worker-prebuild.yaml` we must include an Ubuntu version with kernel >= 6. Problem is that Ubuntu 22.04 still uses kernel 5.15 and there are no other more modern Ubuntu versions in GitHub CI.

**NOTE 2:** When that is done (in part 2 of this task), we must still keep Ubuntu 20.04 since, as documented in the job file, it uses an older version of GLib and otherwise we'd run into this issue again: https://github.com/versatica/mediasoup/issues/1089.

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -16,7 +16,7 @@ const WORKER_RELEASE_DIR = 'worker/out/Release';
 const WORKER_RELEASE_BIN = IS_WINDOWS ? 'mediasoup-worker.exe' : 'mediasoup-worker';
 const WORKER_RELEASE_BIN_PATH = `${WORKER_RELEASE_DIR}/${WORKER_RELEASE_BIN}`;
 const WORKER_PREBUILD_DIR = 'worker/prebuild';
-const WORKER_PREBUILD_TAR = `mediasoup-worker-${PKG.version}-${os.platform()}-${os.arch()}.tgz`;
+const WORKER_PREBUILD_TAR = getWorkerPrebuildTarName();
 const WORKER_PREBUILD_TAR_PATH = `${WORKER_PREBUILD_DIR}/${WORKER_PREBUILD_TAR}`;
 const GH_OWNER = 'versatica';
 const GH_REPO = 'mediasoup';
@@ -298,6 +298,22 @@ function getPython()
 	}
 
 	return python;
+}
+
+function getWorkerPrebuildTarName()
+{
+	let name = `mediasoup-worker-${PKG.version}-${os.platform()}-${os.arch()}`;
+
+	// In Linux we want to know about kernel version since kernel >= 6 supports
+	// io-uring.
+	if (os.platform() === 'linux')
+	{
+		const kernelMajorVersion = Number(os.release().split('.')[0]);
+
+		name += `-kernel${kernelMajorVersion}`;
+	}
+
+	return `${name}.tgz`;
 }
 
 function installInvoke()


nazar-pc (commented):
I'm not opposed to merging this, but major version doesn't really mean anything in Linux. And there is a pretty wild range of features available between various revisions of each major version.

I think this will have to become more sophisticated over time to make sure proper versions are accounted for.

I'm also wondering how hard it would be to compile a worker that can make decision about whether to enable io_uring in runtime.

Then it would be possible to install kernel headers and such from newer version even on older kernel and compile it with io_uring support on 20.04, while not using it in runtime if kernel is too old.

ibc:
> I'm not opposed to merging this, but major version doesn't really mean anything in Linux. And there is a pretty wild range of features available between various revisions of each major version.
> 
> I think this will have to become more sophisticated over time to make sure proper versions are accounted for.
> 
> I'm also wondering how hard it would be to compile a worker that can make decision about whether to enable io_uring in runtime.
> 
> Then it would be possible to install kernel headers and such from newer version even on older kernel and compile it with io_uring support on 20.04, while not using it in runtime if kernel is too old.

* `meson.build` only compiles `liburing` subproject if current host has kernel >6. Why? Because `liburing` depends on `io-uring` which is only available in Linux kernel >= 6.
* Then in `DepLibUring.IsRuntimeSupported()` (so in runtime) we verify whether current running host has kernel >= 6, otherwise we do not use `liburing` subproject.

In summary:
1. `liburing` can only be built if the host in which the worker is built has kernel > 6.
2. Later that mediasoup-worker prebuilt could work in a kernel 5. In that case `liburing` is disabled in runtime.
3. Bullet 1 is a requirement for bullet 2.

nazar-pc:
> liburing can only be built if the host in which the worker is built has kernel > 6.

I do not necessarily think that is the case. Headers from newer kernel are probably required, but the kernel running as such. It can be installed on a system running older kernel and with some tweaks to default library search paths compiler should be able to see it.

ibc:
> > liburing can only be built if the host in which the worker is built has kernel > 6.
> 
> I do not necessarily think that is the case. Headers from newer kernel are probably required, but the kernel running as such. It can be installed on a system running older kernel and with some tweaks to default library search paths compiler should be able to see it.

Not sure why such a effort is needed. We only need a kernel >= 6 in CI which is not available yet.

nazar-pc (approved):
I'm fine with this, just exploring possibilities

ibc:
NOTE: as a side effect of this PR part 1, those installing mediasoup node in Debian or Ubuntu or RedHat etc with kernel >= 6 will not have prebuilt worker available so will build worker from scratch, and hence with io-uring support, which IMHO is good. What do you think @jmillan?

jmillan:
I think this should not be needed. Once GH action runners run kernel 6 liburing will be built automatically. Until that happens (unless we manage to compile it with proper headers from an older linux) then liburing will be disabled in the prebuilt binaries.

How will this PR help in any way?

ibc:
> I think this should not be needed. Once GH action runners run kernel 6 liburing will be built automatically. Until that happens (unless we manage to compile it with proper headers from an older linux) then liburing will be disabled in the prebuilt binaries.
> 
> How will this PR help in any way?

1. We need a worker prebuilt made with Ubuntu 20.04 due to the GLib issue (see the link in the PR description), so it can work in not super recent Linux hosts, let's say those running kernel 5 (hopefully).
2. This PR forces new Linux hosts (with kernel 6) to compile worker from scratch since there won't be (yet) prebuilt for them, so the worker will support io-uring.

jmillan:
> > I think this should not be needed. Once GH action runners run kernel 6 liburing will be built automatically. Until that happens (unless we manage to compile it with proper headers from an older linux) then liburing will be disabled in the prebuilt binaries.
> > 
> > How will this PR help in any way?
> 
> 1. We need a worker prebuilt made with Ubuntu 20.04 due to the GLib issue (see the link in the PR description), so it can work in not super recent Linux hosts, let's say those running kernel 5 (hopefully).
> 2. This PR forces new Linux hosts (with kernel 6) to compile worker from scratch since there won't be (yet) prebuilt for them, so the worker will support io-uring.

Gotcha, rather than renaming the binary with the kernel version I think a more elegant way is that the binary itself can tell it, via command argument. Ie: `mediasoup-worker -v [--version]` or `--info`. The output of that command being, among others, whether io_uring is built-in.

ibc:
> Gotcha, rather than renaming the binary with the kernel version I think a more elegant way is that the binary itself can tell it, via command argument. Ie: mediasoup-worker -v [--version] or --info. The output of that command being, among others, whether io_uring is built-in.

For what? How would that help? And  how is that related to the prebuilt fetching?

jmillan:
Anyway I'm OK with this change.

jmillan (approved):

ibc:
Imagine we had Ubuntu 24.04 with kernel 6 in GitHub actions so we can produce io-using capable worker. But we still need to produce a prebuilt worker for Linux with older GLib (due to the referenced issue). So we need TWO prebuilt binaries so two different names. This is not about having "npm install mediasoup" downloading a "universal" binary and then checking some command on it to see if it was valid or not.

nazar-pc:
I agree, we do want to be able to make decision before downloading

PR #1285, opened 2023-12-30
Avoid modification of user input data

### LINKED ISSUE #1277: Validation functions at `ortc.ts` file are modifying input data
Validation functions at https://github.com/versatica/mediasoup/blob/43edb3acd467045fc557ad7ba1fd87c11072e24b/node/src/ortc.ts are modifying the input data, in particular the `createRouter()` function `options.mediaCodecs` field. It's ok to normalice it internally, but input data should not be modified as a side effect. That validation functions has a comment about they modify the input data, but that docstrings are not getting into the Mediasoup documentation, so solutions are

1. add a note about that in the docs
2. export the normalization functions
3. (preferably) create a copy of the input data with `utils.clone()`, so original provided data gets untouched

ibc:
Fixes #1277

### Details

**Node**

- Ensure that user data given to public API functions is not internally modified (so clone it before using it).
- Use proper `utils.clone<T>(xxx)` syntax everywhere.
- Remove no longer needed functions in `ortc.ts` (unused since Flatbuffers migration).
- Do not export unneeded functions in `ortc.ts` (keep them private).

### DIFF
--- a/node/src/Transport.ts
+++ b/node/src/Transport.ts
@@ -714,7 +714,7 @@ export class Transport
 			throw new TypeError('if given, appData must be an object');
 		}
 
-		// Clone given RTP parameters to not modify inout data.
+		// Clone given RTP parameters to not modify input data.
 		const clonedRtpParameters = utils.clone<RtpParameters>(rtpParameters);
 
 		// This may throw.
@@ -861,7 +861,7 @@ export class Transport
 			throw new TypeError('if given, mid must be non empty string');
 		}
 
-		// Clone given RTP capabilities to not modify inout data.
+		// Clone given RTP capabilities to not modify input data.
 		const clonedRtpCapabilities = utils.clone<RtpCapabilities>(rtpCapabilities);
 
 		// This may throw.
@@ -1006,7 +1006,7 @@ export class Transport
 
 		let type: DataProducerType;
 
-		// Clone given SCTP stream parameters to not modify inout data.
+		// Clone given SCTP stream parameters to not modify input data.
 		let clonedSctpStreamParameters
 			= utils.clone<SctpStreamParameters | undefined>(sctpStreamParameters);
 
--- a/node/src/Transport.ts
+++ b/node/src/Transport.ts
@@ -714,17 +714,20 @@ export class Transport
 			throw new TypeError('if given, appData must be an object');
 		}
 
+		// Clone given RTP parameters to not modify inout data.
+		const clonedRtpParameters = utils.clone<RtpParameters>(rtpParameters);
+
 		// This may throw.
-		ortc.validateRtpParameters(rtpParameters);
+		ortc.validateRtpParameters(clonedRtpParameters);
 
 		// If missing or empty encodings, add one.
 		if (
-			!rtpParameters.encodings ||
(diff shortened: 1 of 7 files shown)


node/src/Transport.ts
```diff
@@ -714,17 +714,20 @@ export class Transport
 			throw new TypeError('if given, appData must be an object');
 		}
 
+		// Clone given RTP parameters to not modify inout data.
```
jmillan: 'inout'? This is repeated along the PR

jmillan (approved):

node/src/Transport.ts
ibc: Because I just wrote it once (wrongly) and pasted everywhere ☺️

Will fix

nazar-pc (changes requested):
Please revert all Rust changes, they are not idiomatic and not necessary.

Rust APIs should clearly specify what they need and not more. In this example functions already receive owned value of `rtp_parameters` that will be dropped at the end of the function because it will need an owned value and will be modifying it. All you're doing by cloning it is once more unnecessary allocation. If caller does have a spare value to give away, no extra allocation will be necessary with old API, if there isn't, the caller will have to do explicit clone. But getting value as an argument that you know will always be cloned is anti-pattern in Rust because it introduces unnecessary allocations that caller can't prevent.

Also it is not pretty to add `cloned_` prefix or similar suffix to a variable in Rust. You can redefine variable name even with a different type if you need to in contrast to TypeScript.

ibc:
> If caller does have a spare value to give away, no extra allocation will be necessary with old API, if there isn't, the caller will have to do explicit clone. But getting value as an argument that you know will always be cloned is anti-pattern in Rust because it introduces unnecessary allocations that caller can't prevent.

Example:

1. User defines `media_codecs` in a variable.
2. User passes `media_codecs` to `router1 = worker.create_router(media_codecs)`.
3. `media_codecs` is internally modified by mediasoup. User doesn't know it.
4. User passes same `media_codecs` variable to `router2 = worker.create_router(media_codecs)`. At that time he is passing a modified `media_codecs`.
5. Do we really want that?

Basically you are invalidating the whole purpose of this PR by arguing that the user should clone the data by himself if he doesn't want it to be potentially modified by mediasoup. I just don't get it.

nazar-pc:
> 4. User passes same media_codecs variable to router2 = worker.create_router(media_codecs). At that time he is passing a modified media_codecs.

Not possible, ownership of `media_codecs` will be transferred during the first call. You'll have to clone it during the first call or else it will not compile. This is not TypeScript and not C++.

ibc:
> ownership of `media_codecs` will be transferred during the first call

Ok, that solves it. 

May I know how such a ownership transfer is declared in the code?

nazar-pc:
> May I know how such a ownership transfer is declared in the code?

If you specify a variable - you transfer ownership, if you want to give a reference to the data/borrow it, you use `&foo` or `&mut foo` explicitly.

ibc:
Rust changes reverted.

nazar-pc (approved):
Rust changes look fine to me :joy:

ibc:
> Rust changes look fine to me 😂

Thanks, I had to investigate and learn a lot to write them properly.

PR #1287, opened 2024-01-01
Update Catch2 to version 3.4.0

### LINKED ISSUE #987: Upgrade Catch2 to version 3
We must upgrade [Catch2](https://github.com/catchorg/Catch2) from 2.13.7 to latest 3.3.0 (o newer). I did it in libmediasoupclient project in this ongoing PR https://github.com/versatica/libmediasoupclient/pull/151. However, in there I didn't use the proper new approach 2 (which is building Catch2 as a proper static library). Instead I used approach 1. See:

https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md#how-to-migrate-projects-from-v2-to-v3

Now, when it comes to do this in mediasoup I'm really lost:

- In `worker/subprojects/` there is a **custom?** `catch2.wrap` as follows:
  ```
  [wrap-file]
  directory = Catch2-2.13.7
  source_url = https://github.com/catchorg/Catch2/archive/v2.13.7.zip
  source_filename = Catch2-2.13.7.zip
  source_hash = 3f3ccd90ad3a8fbb1beeb15e6db440ccdcbebe378dfd125d07a1f9a587a927e9
  patch_filename = catch2_2.13.7-1_patch.zip
  patch_url = https://wrapdb.mesonbuild.com/v2/catch2_2.13.7-1/get_patch
  patch_hash = 2f7369645d747e5bd866317ac1dd4c3d04dc97d3aad4fc6b864bdf75d3b57158
  
  [provide]
  catch2 = catch2_dep
  ```
- Why can't we just use the `catch2.wrap` (version 2.x.x) available in official Meson wrap registry inhttps://mesonbuild.com/Wrapdb-projects.html?
- The official Meson wrap registry provides with a modern version of Catch2 3.2.1, which looks as follows:
  ```
  [wrap-file]
  directory = Catch2-3.2.0
  source_url = https://github.com/catchorg/Catch2/archive/v3.2.0.tar.gz
  source_filename = Catch2-3.2.0.tar.gz
  source_hash = feee04647e28ac3cbeff46cb42abc8ee2d8d5f646d36e3fb3ba274b8c69a58ea
  wrapdb_version = 3.2.0-1
  
  [provide]
  catch2 = catch2_dep
  catch2-with-main = catch2_with_main_dep
  ```
- It doesn't require any patch. Can we use it as it is?
- First steps for using Catch2 3.x.x as a static library is _"Change your CMakeLists.txt to link against Catch2WithMain target if you use Catch2's default main. (If you do not, keep linking against the Catch2 target.)"_. No idea how to do that.

ibc:
Fixes #987

### Details

- We no longer use the old Catch2 single-header distribution model but the new more standard library distribution model, which comes with much better compile-time performance. Details in https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md.

### DIFF
--- a/worker/tasks.py
+++ b/worker/tasks.py
@@ -393,32 +393,25 @@ def test(ctx):
             shell=SHELL
         );
 
+    mediasoup_worker_test = 'mediasoup-worker-test.exe' if os.name == 'nt' else 'mediasoup-worker-test';
     mediasoup_test_tags = os.getenv('MEDIASOUP_TEST_TAGS') or '';
 
-    # On Windows lcov doesn't work (at least not yet) and we need to add .exe to
-    # the binary path.
-    if os.name == 'nt':
-        with ctx.cd(WORKER_DIR):
-            ctx.run(
-                f'"{BUILD_DIR}/mediasoup-worker-test.exe" --invisibles --use-colour=yes {mediasoup_test_tags}',
-                echo=True,
-                pty=PTY_SUPPORTED,
-                shell=SHELL
-            );
-    else:
+    # On Windows lcov doesn't work (at least not yet).
+    if os.name != 'nt':
         ctx.run(
             f'"{LCOV}" --directory "{WORKER_DIR}" --zerocounters',
             echo=True,
             pty=PTY_SUPPORTED,
             shell=SHELL
         );
-        with ctx.cd(WORKER_DIR):
-            ctx.run(
-                f'"{BUILD_DIR}/mediasoup-worker-test" --invisibles --use-colour=yes {mediasoup_test_tags}',
-                echo=True,
-                pty=PTY_SUPPORTED,
-                shell=SHELL
-            );
+
+    with ctx.cd(WORKER_DIR):
+        ctx.run(
+            f'"{BUILD_DIR}/{mediasoup_worker_test}" --invisibles --colour-mode=ansi {mediasoup_test_tags}',
+            echo=True,
+            pty=PTY_SUPPORTED,
+            shell=SHELL
+        );
 
 
 @task(pre=[setup, flatc])
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -70,18 +70,13 @@ jobs:
       # We need to install npm deps of worker/scripts/package.json.
       - name: npm ci --prefix worker/scripts
         run: npm ci --prefix worker/scripts --foreground-scripts
-        # TODO: Maybe fix this one day.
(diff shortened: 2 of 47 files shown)


worker/tasks.py
```diff
@@ -393,32 +393,25 @@ def test(ctx):
             shell=SHELL
         );
 
+    mediasoup_worker_test = 'mediasoup-worker-test.exe' if os.name == 'nt' else 'mediasoup-worker-test';
```
nazar-pc, line 396: Looks like this refactoring made code harder to read. What is the benefit of extracting the name into a variable if the rest of the executed command is platform-specific anyway?

nazar-pc (approved):
Makes sense to me, but I don't think refactoring made things better

worker/tasks.py
ibc: Because we may want to change command line arguments given to the test binary and I prefer to not have all them duplicated. There is no platform-specific stuff here other than the binary name.

nazar-pc: What I'm saying is that both command and arguments are different between Windows and macOS/Windows. So in terms of code complexity you have one `if` on this line and one more `if` below, which leads to 4 potential branches, while only 2 of them are valid.

You could just inline `mediasoup-worker-text[.exe]` into the commands below like it was before and it would make code easier to read.

This variable doesn't bring any benefit, I think the old version was better.

ibc: Then I don't understand what you mean. Old version had duplicate code:

CleanShot-2024-01-01-at-14 04 44@2x

New version doesn't have duplicate code:

CleanShot-2024-01-01-at-14 05 45@2x

nazar-pc: That is exactly what I mean: old code had duplicate and having a variable might potentially make sense, but now even commands are different.

Why is it better to write `{mediasoup_worker_test}` instead of `mediasou-worker-test`?

ibc: > but now even commands are different.

What? where? 😀

There is a `lcov` command that only runs on Linux/macOS and it was there since day 0. Other than that we have a single call to `mediasoup-worker-test[.exe]` with **same** arguments.

> Why is it better to write `{mediasoup_worker_test}` instead of `mediasou-worker-test`?

Because in Windows it must be `mediasou-worker-test.exe`.

nazar-pc: Okay, now I see. It is not just refactoring, the code is also different.

Now makes sense, next time try to create two commits: one with refactoring, second that introduces new things. That way it is easier to review.

PR #1288, opened 2024-01-01
Remove lcov dep from the repo

ibc:
- `worker/deps/lcov` is removed form the repo and git ignored.
-  It can still be used locally by calling `worker/scripts/get-dep.sh lcov` and then `worker/scripts/lcov.sh`.


nazar-pc (commented):
Hm... do you think we can download it when needed instead of including complete sources in the repo?

ibc:
> Hm... do you think we can download it when needed instead of including complete sources in the repo?

Plus CI is failing due to some missing Perl dependency needed by this moden version of lcov (which BTW is master branch):

https://github.com/versatica/mediasoup/actions/runs/7377722464/job/20072014022?pr=1288

So we are only using lcov to delete all `.da` files in /Users/ibc/src/v3-mediasoup/worker and subdirectories. But funny thing is that we are NOT generating those coverage files when running the tests. Requirements of [lcov](https://github.com/linux-test-project/lcov):

> Requirements: compile the program in question using GCC with the options
'-fprofile-arcs -ftest-coverage' or '--coverage'. During linking, make sure
to specify '-lgcov' or '--coverage'.

So instead I will just remove lcov completely.

ibc:
Done. `worker/deps/lcov` is removed form the repo and git ignored. It can be used locally by calling `worker/scripts/get-dep.sh lcov` and then `worker/scripts/lcov.sh`.

nazar-pc (approved):
Makes sense to me

PR #1289, opened 2024-01-01
Rust: expose DataChannel string message as binary

### LINKED ISSUE #1269: Worker panic in Rust in WebRtcMessage::new
### Issue description

I'm seeing occasional panics in the call to [WebRtcMessage::new](https://github.com/versatica/mediasoup/blob/d7bee8a34cf0880de6508fc57be1c049375e88fc/rust/src/data_structures.rs#L1074) because of that unwrap. I never call that function myself, so I don't know where the invalid byte(s) would be coming from

[Here's](https://gist.github.com/PaulOlteanu/e87c49771df1c792caab8a97863d412c) a log from the panic.

ibc:
Fixes #1269

### Details

- Do not convert bytes to (UTF-8) string but let the user app do it based on `ppid` (`WebRtcMessage` type).

### DIFF
--- a/rust/tests/integration/direct_transport.rs
+++ b/rust/tests/integration/direct_transport.rs
@@ -202,19 +202,19 @@ fn send_succeeds() {
 
             move |message| {
                 let id: usize = match message {
-                    WebRtcMessage::String(string) => {
-                        recv_message_bytes.fetch_add(string.as_bytes().len(), Ordering::SeqCst);
-                        string.parse().unwrap()
+                    WebRtcMessage::String(binary) => {
+                        recv_message_bytes.fetch_add(binary.len(), Ordering::SeqCst);
+                        String::from_utf8(binary.to_vec()).unwrap().parse().unwrap()
                     }
                     WebRtcMessage::Binary(binary) => {
                         recv_message_bytes.fetch_add(binary.len(), Ordering::SeqCst);
                         String::from_utf8(binary.to_vec()).unwrap().parse().unwrap()
                     }
                     WebRtcMessage::EmptyString => {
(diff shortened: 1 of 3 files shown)


ibc:
But then, the `data_producer.send()` and `data_consumer.send()` methods require an additional `ppid` as argument (same as in Node).

ibc:
> But then, the `data_producer.send()` and `data_consumer.send()` methods require an additional `ppid` as argument (same as in Node).

Nope, not needed.

rust/tests/integration/direct_transport.rs
```diff
@@ -444,13 +444,17 @@ fn send_with_subchannels_succeeds() {
 
         // Must be received by dataConsumer1 and dataConsumer2.
         direct_data_producer
-            .send(WebRtcMessage::String(both.to_string()), None, None)
+            .send(
+                WebRtcMessage::String(std::borrow::Cow::Borrowed(both.as_bytes())),
```
nazar-pc: What about importing `std::borrow::Cow` since it is used so many times here?

nazar-pc (approved):
Makes sense to me and this will definitely be a breaking change for `mediasoup` crate.

rust/tests/integration/direct_transport.rs
ibc: It was already imported, my fault. I've simplified to `Cow::Borrowed` everywhere.

nazar-pc (approved):

ibc:
> Makes sense to me and this will definitely be a breaking change for `mediasoup` crate.

https://github.com/versatica/mediasoup/pull/1290

PR #1291, opened 2024-01-02
ListenInfo: Add transport socket flags

ibc:
### Details

- Add a new `flags` option in `TransportListenInfo` to specify socket flags.
- Supported flags are:
  ```ts
  /**
   * Disable dual-stack support so only IPv6 is used (only if ip is IPv6).
   */
    ipv6Only?: boolean;
  /**
   * Make different transports bind to the same ip and port (only for UDP).
   * Useful for multicast scenarios with plain transport. Use with caution.
   */
   udpReusePort?: boolean;
  ```
- Those flags correspond to the following ones exposed by libuv:
  - https://docs.libuv.org/en/v1.x/udp.html#c.uv_udp_flags
  - https://docs.libuv.org/en/v1.x/tcp.html#c.uv_tcp_bind
- Note that `flags.ipv6Only` is `false` by default, which is a behavior change.
- This PR replaces PR #1266 by offering the same functionality in a generic and extensible way, and without making mediasoup responsible of knowing in detail the supported flags/features in each arch/OS.

### TODO

- Document in website.


ibc:
Rust hates me:

```
error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> rust/src/data_structures.rs:52:17
   |
52 | #[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
   |                 ^^^^
...
67 |     pub flags: Option<Vec<SocketFlag>>,
   |     ---------------------------------- this field does not implement `std::marker::Copy`
   |
note: the `std::marker::Copy` impl for `std::option::Option<std::vec::Vec<data_structures::SocketFlag>>` requires that `std::vec::Vec<data_structures::SocketFlag>: std::marker::Copy`
  --> rust/src/data_structures.rs:67:16
   |
67 |     pub flags: Option<Vec<SocketFlag>>,
   |                ^^^^^^^^^^^^^^^^^^^^^^^
   = note: this error originates in the derive macro `Copy` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0204`.
```

nazar-pc:
> Rust hates me:

It helps you write bug-free code. In this case `Copy` can no longer be derived because `Vec` is heap-allocated data structure rather than a bag of bytes. Remove `Copy` from derive if you have to use a vector there.

worker/fbs/transport.fbs
```diff
@@ -13,11 +13,17 @@ enum Protocol: uint8 {
...
+    UDP_REUSEPORT = 2
+}
+
 table ListenInfo {
-    protocol: FBS.Transport.Protocol = UDP;
+    protocol: Protocol = UDP;
     ip: string (required);
     announced_ip: string;
     port: uint16 = 0;
+    flags: uint8 = 0;
```
jmillan: IMO for readability, rather than this being an already formed integer with the contained flags, it could be an array of `SocketFlags`:

```txt
flags: [SocketFlag];
```

The corresponding integer would then only be formed in C++.

ibc: Let's not make it less performant, please. We don't want to iterate all values in worker side and check each bit against supported libuv flags. This is a perfect and optimized usage of bit checking.

nazar-pc: Can't you compress it into bitfield on the worker and leave it well typed in flatbuffers?

jmillan: Or even:

```txt
table SocketFlags {
  ipv6_only: bool = false;
  udp_reuse_port: bool = false;
}
```

Then `ListenInfo` would have:

```txt
  flags: SocketFlags;
```

I'm just commenting this because it looked strange to me forming a flag integer from TS rather than encoding the corresponding meaningful values over flatbuffers.

ibc:
> It helps you write bug-free code. In this case `Copy` can no longer be derived because `Vec` is heap-allocated data structure rather than a bag of bytes. Remove `Copy` from derive if you have to use a vector there.

And what are the consequences of removing Copy from derive in that struct?

nazar-pc:
> And what are the consequences of removing Copy from derive in that struct?

You'll have to call `.clone()` on it rather than value being implicitly copied (like it happens with `u8` for example). `Clone` typically means there is some side effect of the cloning, in case of vector the biggest side effect is allocation of memory.

worker/fbs/transport.fbs
jmillan: > Let's not make it less performant, please

No single iteration with previous proposal. Not in TS/Rust, not in Worker, plus removed the unneeded vector usage for a well known type.

ibc: Will do

ibc:
@jmillan I've pushed this: https://github.com/versatica/mediasoup/pull/1291/commits/a1405e23abee5dcfd13199d87c767fe5c331f4b6

jmillan (approved):

rust/tests/integration/plain_transport.rs
```diff
@@ -403,6 +412,105 @@ fn create_non_bindable_ip() {
...
+        let _ = router
+            .create_plain_transport({
+                PlainTransportOptions::new(ListenInfo {
+                    protocol: Protocol::Udp,
+                    ip: multicast_ip,
+                    announced_ip: None,
+                    port: Some(port),
+                    flags: Some(SocketFlags {
+                        ipv6_only: false,
+                        udp_reuse_port: true,
```
ibc, line 443: So the problem is that this test passes even if you set `udp_reuse_port: false` in transport 1 or transport 2 (below). In Node it behaves correctly (it fails if any of the transports has `udp_reuse_port: false`.

I've made some checks in `PlainTransport.cpp` and my conclusion is that, as far as `flags` is not None, the FBS object received by the worker has `udpReusePort` with value `true`!! I've confirmed it by adding this code in `PlainTransport` constructor:

```c++
if (this->listenInfo.flags.udpReusePort) {
	MS_DUMP("Oh, true?? ok I am converting it to false!!!");
	this->listenInfo.flags.udpReusePort = false;
}
```

In Rust, such a log ONLY shows up when a `PlainTransport` is created (by a test) with `flags` NOT set to `None`, and in that case, even if you pass ` flags: Some(SocketFlags { ipv6_only: false, udp_reuse_port: false }`, the `MS_DUMP()` is printed, meaning that received `udpReusePort` is `true`!!!

HELP.

rust/tests/integration/plain_transport.rs
```diff
@@ -403,6 +412,105 @@ fn create_non_bindable_ip() {
...
+                    }),
+                    send_buffer_size: None,
+                    recv_buffer_size: None,
+                })
+            })
+            .await
+            .expect("Failed to create first Plain transport");
+
+        // Transport 2.
+        assert!(matches!(
```
ibc, line 502: And this test fails because, as explained above, when `flags` is passed (other than `flags: None`) it doesn't matter whether given `udp_reuse_port` is true or false, because the worker always receives `true`. So the second transport creation does NOT fail and hence the test (which expects it to fail) fails.

rust/tests/integration/plain_transport.rs
ibc: Wait. I may be completely wrong. I finally can print those logs. In `PlainTransport` constructor I've added:

```c++
MS_ERROR_STD("options->listenInfo()->flags()->udpReusePort(): %u", options->listenInfo()->flags()->udpReusePort());
```

And I run tests as follows:

```
DEBUG="mediasoup:ERROR:Worker" cargo test
```

and yes, I only see two lines with value true (1):

```
options->listenInfo()->flags()->udpReusePort(): 1
```

which matches the number of PlainTransports created by the tests with `udp_reuse_port: true`, so we are good here.

So then what? How is it possible that this test DOESNT FAIL if we change `udp_reuse_port` to `false`?? It perfectly fails in Node as expected. And Rust is NOT doing some magic by adding `SO_REUSEADDR` to sockets created in worker using libuv API, right? Because that's not possible, right?

rust/tests/integration/plain_transport.rs
ibc: NOTE: As explained above this is not true. Flag value is properly passed to C++. Problem is that the test doesn't fail (it DOES fail in Node as expected).

ibc: Fixed here: https://github.com/versatica/mediasoup/pull/1291/commits/7b935b61136793c0f1644ad3fcef27b65234d9c3

rust/tests/integration/plain_transport.rs
ibc: Fixed here: https://github.com/versatica/mediasoup/pull/1291/commits/7b935b61136793c0f1644ad3fcef27b65234d9c3

jmillan (approved):

threema-lenny:
I stumbled upon this. It would have been nice to mention the switch from setting `UV_UDP_IPV6ONLY` by default to make it optional and user-supplied in the changelog as it was effectively a breaking change in behaviour.

ibc:
> I stumbled upon this. It would have been nice to mention the switch from setting `UV_UDP_IPV6ONLY` by default to make it optional and user-supplied in the changelog as it was effectively a breaking change in behaviour.

True. I've updated the PR description and added a note in the CHANGELOG file: https://github.com/versatica/mediasoup/commit/2854761ff1dc984d1d3f31cbf216319c085b2e12

PR #1294, opened 2024-01-03
ListenInfo: Ignore given socket flags if not suitable for given IP family or transport

ibc:
For example, by design `libuv` will throw if a IPv4 socket is created with flag `UV_UDP_IPV6ONLY` or `UV_TCP_IPV6ONLY`.


ibc:
Merging, this is an easy one.

jmillan:
👍

PR #1296, opened 2024-01-04
Make Node tests independent (part 2)

ibc:
### Details

- Just done in `test-Router.ts` as POC.
- Avoid declaring global variables in test files. Put them into `ctx: TestContext` instead.
- Things in `ctx` should be const/frozen or reset in each `test()` execution.
- So use a new `utils.deepFreeze()` function which freezes the object and would throw upon any modification attempt.

### DIFF
--- a/node/src/utils.ts
+++ b/node/src/utils.ts
@@ -199,7 +199,7 @@ export function parseStringStringArrayVector(
 }
 
 /**
- * Make make an object or array recursively immutable.
+ * Make an object or array recursively immutable.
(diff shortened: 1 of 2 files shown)


node/src/utils.ts
```diff
@@ -197,3 +197,26 @@ export function parseStringStringArrayVector(
 
 	return array;
 }
+
+/**
+ * Make make an object or array recursively immutable.
```
jmillan:
```suggestion
 * Make an object or array recursively immutable.
```

jmillan (approved):

PR #1300, opened 2024-01-05
Add worker prebuild binary for Linux kernel 6

### LINKED ISSUE #1284: Need kernel 6 capable Linux CI host to produce worker prebuilt with io-uring support
Related to PR https://github.com/versatica/mediasoup/pull/1283 "mediasoup-worker linux prebuilt with io-uring support (part 1)".

### Details

In `mediasoup-worker-prebuild` CI job we are using an old Ubuntu version 20.04 to create the Linux `mediasoup-worker` prebuilt binary (for reasons as documented inline). Problem is that Ubuntu 20.04 uses kernel 5.15 (<6) so we are not enabling `io-uring` on these binaries :(

```meson
# For Linux let's use an old version of Ubuntu (20.04) that builds the
# mediasoup-worker binary using an old version of GLib, so it will work
# on Linux hosts running more modern GLib versions.
# See https://github.com/versatica/mediasoup/issues/1089.
- os: ubuntu-20.04
  cc: gcc
  cxx: g++
- os: macos-12
  cc: clang
  cxx: clang++
- os: windows-2022
  cc: cl
  cxx: cl
```

Problem is that Ubuntu 22.05 also comes with Kernel 5.15, so it's not a valid option. Once GH actions include Ubuntu 24.05 (I assume with kernel 6) we must include it in the job.

**NOTE:** When that is done we must still keep Ubuntu 20.04 since, as documented in the job file, it uses an old version of GLib which is needed in systems with GLib equal or newer than that (the opposite is not true: a software built with a modern version of GLib cannot run in systems with an older GLib version). If we don't do this we'd run into this issue again: https://github.com/versatica/mediasoup/issues/1089.

ibc:
Fixes #1284

### Details

- Ubuntu 22.04.3 LTS has kernel 6.2.
- CI job: No need to manually install `invoke` since `npm run worker:build` internally installs it if needed.

### DIFF
--- a/.github/workflows/mediasoup-worker-prebuild.yaml
+++ b/.github/workflows/mediasoup-worker-prebuild.yaml
@@ -10,11 +10,16 @@ jobs:
     strategy:
       matrix:
         build:
-          # For Linux let's use an old version of Ubuntu (20.04) that builds the
+          # Worker prebuild for Linux with kernel version 5 Ubuntu (20.04).
+          # Let's use an old version of Ubuntu (20.04) that builds the
           # mediasoup-worker binary using an old version of GLib, so it will work
           # on Linux hosts running more modern GLib versions.
           # See https://github.com/versatica/mediasoup/issues/1089.
           - os: ubuntu-20.04
+            cc: gcc
+            cxx: g++
+          # Worker prebuild for Linux with kernel version 6 Ubuntu (22.04).
+          - os: ubuntu-22.04
             cc: gcc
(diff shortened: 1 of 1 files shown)


nazar-pc (approved):

ibc:
There we go!

CleanShot-2024-01-05-at-14 02 22@2x

CleanShot-2024-01-05-at-14 04 24@2x

jmillan:
👍

PR #1307, opened 2024-01-08
Node: Add new worker.on('subprocessclose') to fix Node tests

### LINKED ISSUE #1306: Issues in Node tests
Here some issues in Node tests:

- [ ] CI failing due to aborted `PRODUCER_PAUSE` channel request: https://github.com/versatica/mediasoup/actions/runs/7448459449/job/20262892577?pr=1305
- [ ] `npm run test:node -- --testPathPattern test-ActiveSpeakerObserver` fails with:
  ```
  Jest has detected the following 2 open handles potentially keeping Jest from exiting:

  ●  PROCESSWRAP

      339 | 		);
      340 |
    > 341 | 		this.#child = spawn(
          | 		                   ^
      342 | 			// command
      343 | 			spawnBin,
      344 | 			// args

      at new Worker (node/src/Worker.ts:341:22)
      at Object.<anonymous> (node/src/index.ts:64:17)
      at node/src/index.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/index.ts:4:12)
      at Object.createWorker (node/src/index.ts:38:12)
      at node/src/test/test-ActiveSpeakerObserver.ts:26:31
      at node/src/test/test-ActiveSpeakerObserver.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/test/test-ActiveSpeakerObserver.ts:4:12)
      at Object.<anonymous> (node/src/test/test-ActiveSpeakerObserver.ts:25:23)


  ●  PIPEWRAP

      339 | 		);
      340 |
    > 341 | 		this.#child = spawn(
          | 		                   ^
      342 | 			// command
      343 | 			spawnBin,
      344 | 			// args

          at Array.reduce (<anonymous>)
      at new Worker (node/src/Worker.ts:341:22)
      at Object.<anonymous> (node/src/index.ts:64:17)
      at node/src/index.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/index.ts:4:12)
      at Object.createWorker (node/src/index.ts:38:12)
      at node/src/test/test-ActiveSpeakerObserver.ts:26:31
      at node/src/test/test-ActiveSpeakerObserver.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/test/test-ActiveSpeakerObserver.ts:4:12)
      at Object.<anonymous> (node/src/test/test-ActiveSpeakerObserver.ts:25:23)
  ```
- [ ] `npm run test:node -- --testPathPattern "test-PipeTransport.ts"` fails with:
  ```
  Jest has detected the following 2 open handles potentially keeping Jest from exiting:

  ●  PROCESSWRAP

      339 | 		);
      340 |
    > 341 | 		this.#child = spawn(
          | 		                   ^
      342 | 			// command
      343 | 			spawnBin,
      344 | 			// args

      at new Worker (node/src/Worker.ts:341:22)
      at Object.<anonymous> (node/src/index.ts:64:17)
      at node/src/index.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/index.ts:4:12)
      at Object.createWorker (node/src/index.ts:38:12)
      at node/src/test/test-PipeTransport.ts:171:32
      at node/src/test/test-PipeTransport.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/test/test-PipeTransport.ts:4:12)
      at Object.<anonymous> (node/src/test/test-PipeTransport.ts:170:23)


  ●  PIPEWRAP

      339 | 		);
      340 |
    > 341 | 		this.#child = spawn(
          | 		                   ^
      342 | 			// command
      343 | 			spawnBin,
      344 | 			// args

          at Array.reduce (<anonymous>)
      at new Worker (node/src/Worker.ts:341:22)
      at Object.<anonymous> (node/src/index.ts:64:17)
      at node/src/index.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/index.ts:4:12)
      at Object.createWorker (node/src/index.ts:38:12)
      at node/src/test/test-PipeTransport.ts:171:32
      at node/src/test/test-PipeTransport.ts:8:71
      at Object.<anonymous>.__awaiter (node/src/test/test-PipeTransport.ts:4:12)
      at Object.<anonymous> (node/src/test/test-PipeTransport.ts:170:23)
  ```

Probably many similar ones if we run each test file separately.

NOTE: Yes, I added `--detectOpenHandles` to `jest` command, but it should work fine.

ibc:
Fixes #1306

### Details

- Problem only happened when running a test file separately (it worked by accident when running all tests together as usual).
- Problem was that the `ChildProcess` object in `Worker` (the `this.#child` member) remains open for a bit until it emits 'close' event, which is emitted after 'exit' event or after 'error' event.
- But we didn't await for that event at all since out `worker.close()` method is sync and we do not care of what is happening in the subprocess after calling it. And this is good. However it's bad for tests if we use `--detectOpenHandles` in Jest command.
- Here we have added a new 'subprocessclose' event in `Worker` which is emitted when the worker subprocess has closed. It may be (and it will be) emitted AFTER calling `worker.close()` but not immediately after it.
- This event is basically useful for scenarios like testing where we want to verify that nothing remains running/open when a test is finished.
- Also added a `worker.subprocessClosed` getter.
- Removed an ugly `setTimeout` in `Channel::close()` method. If the app wants to get the very latest notifications/logs from worker when closing it, use the above event to wait for final closure.

### Bonus Tracks

- Be consistent in command line arguments and always use `--option-foo bar` instead of `--option-foo=bar`.
- Make `tasks.py` and `Makefile` more resistant to paths with spaces.

### DIFF
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -575,7 +575,6 @@ export class Worker<
 		this.#closed = true;
 
 		// Kill the worker process.
-		this.#child.unref();
 		this.#child.kill('SIGTERM');
 
 		// Close the Channel instance.
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -200,6 +200,7 @@ export type WorkerDump = {
 
 export type WorkerEvents = {
 	died: [Error];
+	subprocessclose: [];
 	listenererror: [string, Error];
 	// Private events.
 	'@success': [];
@@ -244,7 +245,7 @@ export class Worker<
 	WorkerAppData extends AppData = AppData,
 > extends EnhancedEventEmitter<WorkerEvents> {
 	// mediasoup-worker child process.
-	#child?: ChildProcess;
+	#child: ChildProcess;
 
 	// Worker process PID.
 	readonly #pid: number;
@@ -257,6 +258,9 @@ export class Worker<
 
 	// Died dlag.
 	#died = false;
+
+	// Worker subprocess closed flag.
+	#subprocessClosed = false;
 
 	// Custom app data.
 	#appData: WorkerAppData;
@@ -389,7 +393,10 @@ export class Worker<
 		});
 
 		this.#child.on('exit', (code, signal) => {
-			this.#child = undefined;
+			// If killed by ourselves, do nothing.
+			if (this.#child.killed) {
+				return;
+			}
 
 			if (!spawnDone) {
 				spawnDone = true;
@@ -431,7 +438,10 @@ export class Worker<
 		});
 
 		this.#child.on('error', error => {
-			this.#child = undefined;
+			// If killed by ourselves, do nothing.
+			if (this.#child.killed) {
+				return;
+			}
 
 			if (!spawnDone) {
 				spawnDone = true;
@@ -453,6 +463,19 @@ export class Worker<
 
 				this.workerDied(error);
 			}
+		});
+
+		this.#child.on('close', (code, signal) => {
+			logger.debug(
+				'worker process closed [pid:%s, code:%s, signal:%s]',
+				this.#pid,
+				code,
+				signal,
+			);
+
+			this.#subprocessClosed = true;
+
+			this.safeEmit('subprocessclose');
 		});
 
 		// Be ready for 3rd party worker libraries logging to stdout.
@@ -493,6 +516,13 @@ export class Worker<
 	 */
 	get died(): boolean {
 		return this.#died;
+	}
+
+	/**
+	 * Whether the Worker subprocess is closed.
+	 */
+	get subprocessClosed(): boolean {
+		return this.#subprocessClosed;
 	}
 
 	/**
@@ -545,15 +575,8 @@ export class Worker<
 		this.#closed = true;
 
 		// Kill the worker process.
-		if (this.#child) {
-			// Remove event listeners but leave a fake 'error' hander to avoid
-			// propagation.
-			this.#child.removeAllListeners('exit');
-			this.#child.removeAllListeners('error');
-			this.#child.on('error', () => {});
-			this.#child.kill('SIGTERM');
-			this.#child = undefined;
-		}
+		this.#child.unref();
+		this.#child.kill('SIGTERM');
 
 		// Close the Channel instance.
 		this.#channel.close();
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -467,7 +467,7 @@ export class Worker<
 
 		this.#child.on('close', (code, signal) => {
 			logger.debug(
-				'worker process closed [pid:%s, code:%s, signal:%s]',
+				'worker subprocess closed [pid:%s, code:%s, signal:%s]',
 				this.#pid,
 				code,
 				signal,
--- a/node/src/test/test-ActiveSpeakerObserver.ts
+++ b/node/src/test/test-ActiveSpeakerObserver.ts
@@ -27,8 +27,14 @@ beforeEach(async () => {
 	ctx.router = await ctx.worker.createRouter({ mediaCodecs: ctx.mediaCodecs });
 });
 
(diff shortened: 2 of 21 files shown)


node/src/test/test-DirectTransport.ts
```diff
@@ -138,7 +144,7 @@ test('dataProducer.send() succeeds', async () => {
 			);
 		});
 
-		sendNextMessage();
+		await sendNextMessage();
```
ibc, line 147: I have some hope that this will fix this CI random failure: https://github.com/versatica/mediasoup/actions/runs/7448459449/job/20262892577?pr=1305

node/src/test/test-ActiveSpeakerObserver.ts
```diff
@@ -27,8 +27,14 @@ beforeEach(async () => {
 	ctx.router = await ctx.worker.createRouter({ mediaCodecs: ctx.mediaCodecs });
 });
 
-afterEach(() => {
+afterEach(async () => {
 	ctx.worker?.close();
+
+	if (ctx.worker?.subprocessClosed === false) {
```
ibc, line 33: We check `=== false` because if the worker failed to be created in `beforeEach` hook, then `ctx.worker` won't exist here so `ctx.worker?.subprocessClosed` will return `undefined`. In that case we do NOT want to await for any event.

node/src/Worker.ts
```diff
@@ -545,15 +575,8 @@ export class Worker<
...
-		if (this.#child) {
-			// Remove event listeners but leave a fake 'error' hander to avoid
-			// propagation.
-			this.#child.removeAllListeners('exit');
-			this.#child.removeAllListeners('error');
-			this.#child.on('error', () => {});
-			this.#child.kill('SIGTERM');
-			this.#child = undefined;
-		}
+		this.#child.unref();
```
ibc: I've randomly added this unref() call 😌

ibc: And I've removed now since it is useless since we DO HAVE a IPC:

https://nodejs.org/api/child_process.html#subprocessunref

node/src/test/test-ActiveSpeakerObserver.ts
jmillan, line 33: Maybe early return? 

```suggestion
	if (ctx.worker?.subprocessClosed) {
	    return;
	}
```

jmillan (approved):
Just added a cosmetic/style comment.

node/src/test/test-ActiveSpeakerObserver.ts
nazar-pc: There is a comment above why `=== false` is necessary here

jmillan: > There is a comment above why === false is necessary here

I'm suggesting to early return if subprocess is closed so the following block (awaiting for it) is not indented. AFAIK we extensively use this pattern.

ibc: This is not appropriate here IMHO. This is not a method in which if some condition is false we don't want to do anything. See test-DirectTransport (or maybe test-PipeTransport) in which we create 2 workers so this early return fashion cannot be done because we have to do the same for both workers independently. This is not like checking the this.closed flag in a close() method. If I apply your suggestion and tomorrow we need to add more cleanup code in those afterEach() hooks then I would need to revert those changes  and leave the code as it is now in this PR.

ibc: Also your suggested code is wrong @jmillan. If the worker.subprocessClosed is not true it doesn't mean that we have to await for an event to happen in that worker, because as explained above the worker MAY NOT exist.

jmillan: Good

PR #1308, opened 2024-01-09
clang-tidy C++

jmillan:
* Fixed the tidy `tasks.py` command.
* Added the needed env variable to `Build.md`.

TODO:
- [x] remove previous built in clang-tidy since it's 7 years old (TODO in this PR)

NOTES:
- clang tidy must be installed (See `Build.md`)

### DIFF
--- a/worker/include/Settings.hpp
+++ b/worker/include/Settings.hpp
@@ -55,8 +55,8 @@ public:
 	thread_local static struct Configuration configuration;
 
 private:
-	static absl::flat_hash_map<std::string, LogLevel> string2LogLevel;
-	static absl::flat_hash_map<LogLevel, std::string> logLevel2String;
+	static absl::flat_hash_map<std::string, LogLevel> String2LogLevel; // NOLINT(readability-identifier-naming)
+	static absl::flat_hash_map<LogLevel, std::string> LogLevel2String; // NOLINT(readability-identifier-naming)
 };
 
 #endif
--- a/worker/include/Utils.hpp
+++ b/worker/include/Utils.hpp
@@ -57,9 +57,7 @@ namespace Utils
 					  std::memcmp(
 					    std::addressof(reinterpret_cast<const struct sockaddr_in6*>(addr1)->sin6_addr),
 					    std::addressof(reinterpret_cast<const struct sockaddr_in6*>(addr2)->sin6_addr),
-					    16) == 0
-					    ? true
-					    : false);
+					    16) == 0);
 				}
 
 				default:
@@ -71,7 +69,9 @@ namespace Utils
 
 		static struct sockaddr_storage CopyAddress(const struct sockaddr* addr)
 		{
-			struct sockaddr_storage copiedAddr;
+			struct sockaddr_storage copiedAddr
+			{
+			};
 
 			switch (addr->sa_family)
 			{
@@ -125,7 +125,7 @@ namespace Utils
 			auto byte0 = data[i + 2]; // The less significant byte.
 
 			// Check bit 7 (sign).
-			uint8_t extension = byte2 & 0b10000000 ? 0b11111111 : 0b00000000;
+			const uint8_t extension = byte2 & 0b10000000 ? 0b11111111 : 0b00000000;
 
 			return int32_t{ byte0 } | (int32_t{ byte1 } << 8) | (int32_t{ byte2 } << 16) |
 			       (int32_t{ extension } << 24);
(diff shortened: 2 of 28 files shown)


worker/include/Settings.hpp
```diff
@@ -55,8 +55,8 @@ class Settings
 	thread_local static struct Configuration configuration;
 
 private:
-	static absl::flat_hash_map<std::string, LogLevel> string2LogLevel;
-	static absl::flat_hash_map<LogLevel, std::string> logLevel2String;
+	static absl::flat_hash_map<std::string, LogLevel> String2LogLevel; // NOLINT(readability-identifier-naming)
+	static absl::flat_hash_map<LogLevel, std::string> LogLevel2String; // NOLINT(readability-identifier-naming)
```
jmillan, line 59: I had to add the `NOLINT` directive, otherwise clang-tidy lowers the case of the valiables.

worker/src/RTC/Transport.cpp
```diff
@@ -2260,6 +2260,7 @@ namespace RTC
 				auto* consumer = it->second;
 				auto bweType   = this->tccClient->GetBweType();
 
+				// NOLINTNEXTLINE(bugprone-too-small-loop-variable)
```
jmillan, line 2263: `1u` in the second statement is considered of a bigger unit than the `uint8_t` defined in the first statement. Thus, the `NOLINTNEXTLINE` statement.

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -187,6 +187,7 @@ void TcpConnectionHandle::Start()
 		return;
 	}
 
+	// NOLINTNEXTLINE(misc-const-correctness)
```
jmillan, line 190: `err` is also used in case `MS_LIBURING_SUPPORTED` is set.

NOTE: We cannot run `clang-tidy` with `MS_LIBURING_SUPPORTED` in MacOS since it would just fail compiling.

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -66,11 +66,10 @@ UdpSocketHandle::UdpSocketHandle(uv_udp_t* uvHandle) : uvHandle(uvHandle)
 {
 	MS_TRACE();
 
-	int err;
-
 	this->uvHandle->data = static_cast<void*>(this);
 
-	err = uv_udp_recv_start(
+	// NOLINTNEXTLINE(misc-const-correctness)
+	int err = uv_udp_recv_start(
```
jmillan, line 72: same as in TCP socket handler.

worker/include/Settings.hpp
ibc: Why? They are supposed to be in CamelCase as defined in the conf file.

worker/include/Utils.hpp
```diff
@@ -266,10 +266,10 @@ namespace Utils
 
 			for (size_t i{ 0 }; i < len; ++i)
 			{
-				buffer[i] = chars[GetRandomUInt(0, sizeof(chars) - 1)];
+				buffer[i] = Chars[GetRandomUInt(0, sizeof(Chars) - 1)];
 			}
 
-			return std::string(buffer, len);
+			return { buffer, len };
```
ibc, line 272: What is this?

worker/include/Settings.hpp
jmillan: Yes, but clang-tidy is interpreting something wrong here. It's the only place where it's doing it wrong.

worker/include/handles/UdpSocketHandle.hpp
```diff
@@ -84,8 +84,10 @@ class UdpSocketHandle
 
 	/* Callbacks fired by UV events. */
 public:
+	// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
 	void OnUvRecvAlloc(size_t suggestedSize, uv_buf_t* buf);
```
ibc, line 88: Why is it ignored?

worker/include/handles/UdpSocketHandle.hpp
```diff
@@ -84,8 +84,10 @@ class UdpSocketHandle
...
 	void OnUvRecv(ssize_t nread, const uv_buf_t* buf, const struct sockaddr* addr, unsigned int flags);
+	// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
 	void OnUvSend(int status, UdpSocketHandle::onSendCallback* cb);
```
ibc, line 91: Why is it ignored?

worker/include/Settings.hpp
ibc: Perhaps because they are private?

ibc: They are not private, ignore me.

worker/include/Utils.hpp
jmillan: The method is returning a `string`, this is creating the `string` with the given initialization list.

worker/include/handles/UdpSocketHandle.hpp
jmillan: Because it can be made `static` but then we would have this and `OnUvSend` static while `OnUvRecv` non static. I  added the ignore for simply for consistency.

jmillan: But that was my personal view. I can remove just `NOLINT` directives so the methods are converted to static and then we move them to the corresponding place in the file.

worker/include/handles/UdpSocketHandle.hpp
jmillan: same as in `TcpSocketHandler.hpp`.

worker/include/Utils.hpp
ibc: Got it.

worker/include/handles/UdpSocketHandle.hpp
ibc: I'm fine with your current code.

ibc (approved):

PR #1310, opened 2024-01-10
CI codeql.yaml: use custom build instead of reyling on autodetected Autobuild ssytem

ibc:
Some docs: https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages

If CI Actions pass then we are fine.


ibc:
We are good.

PR #1312, opened 2024-01-14
Remove `no-async-promise-executor` linter disables

piranna:
Documentation of eslint [no-async-promise-executor](https://eslint.org/docs/latest/rules/no-async-promise-executor) rule says:

> If an async executor function throws an error, the error will be lost and won’t cause the newly-constructed Promise to reject. **This could make it difficult to debug and handle some errors.**

(Bold is mine)

And this is exactly what has happened to me, 16 hours catching a bug due to this :-)

This PR fix and/or replace all the places where the a Promise async executor is being defined, not only in the place where I was having problems. This way, now exceptions are not missed but instead the Promises are properly rejected when an error occurs.

The fixes in the tests that I have done to remove the `eslint-disable` have been:

- combine the waiting event and the method dispatching it with `Promise.all()`
- replace async executors for sync ones where `await` keyword was not being used
- swallow with `void` the returned Promise of an async function, that was being ignored anyway. This is not strictly necessary, but got used to it as a best practice to show [the Promise is intentionally not awaited](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md#ignorevoid)

### DIFF
--- a/node/src/test/test-Consumer.ts
+++ b/node/src/test/test-Consumer.ts
@@ -1,4 +1,4 @@
-import {once} from 'node:events';
+import { once } from 'node:events';
 import * as flatbuffers from 'flatbuffers';
 import * as mediasoup from '../';
 import { UnsupportedError } from '../errors';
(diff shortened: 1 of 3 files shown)


ibc (commented):
I'm be off for 2 weeks and I would like to review this. Please wait until I'm back. Thanks.

piranna:
> I'm be off for 2 weeks and I would like to review this. Please wait until I'm back. Thanks.

Ok, thank you for notifying in advance :-)

piranna:
Linting fixed, tests are passing.

jmillan (approved):
LGTM

piranna:
Cool, thank you for merging :-)

PR #1313, opened 2024-01-15
C++: Update libsrtp

jmillan:
Comes with performance enhancements on stream list traversal https://github.com/cisco/libsrtp/pull/674

Using the pre-v3 branch which contains little API enhancements.

### DIFF
--- a/worker/subprojects/libsrtp2.wrap
+++ b/worker/subprojects/libsrtp2.wrap
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = libsrtp-4c9f0956f2933ac0650208d69c8d897625ba6301
-source_url = https://github.com/versatica/libsrtp/archive/4c9f0956f2933ac0650208d69c8d897625ba6301.zip
-source_filename = libsrtp-4c9f0956f2933ac0650208d69c8d897625ba6301.zip
-source_hash = 4f3af61e26df398569605fc4bcf377587ca2d8bd34b2b4bf9cdb9590cadbd662
+directory = libsrtp-37d88ecac103f4e2a5336099c64b155ba4ca3b0c
+source_url = https://github.com/versatica/libsrtp/archive/37d88ecac103f4e2a5336099c64b155ba4ca3b0c.zip
+source_filename = libsrtp-37d88ecac103f4e2a5336099c64b155ba4ca3b0c.zip
+source_hash = edf8b2c7a27ed6b6aac64c8d6a165379d3a0042bd420706b4a1825acc182441d
 
 [provide]
 libsrtp2 = libsrtp2_dep
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### NEXT
+
+- C++: Update libsrtp to pre-3.0.
+
 ### 3.13.16
 
 - Node: Add new `worker.on('subprocessclose')` event ([PR #1307](https://github.com/versatica/mediasoup/pull/1307)).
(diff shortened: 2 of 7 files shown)


worker/subprojects/libsrtp2.wrap
```diff
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = libsrtp-4c9f0956f2933ac0650208d69c8d897625ba6301
-source_url = https://github.com/versatica/libsrtp/archive/4c9f0956f2933ac0650208d69c8d897625ba6301.zip
-source_filename = libsrtp-4c9f0956f2933ac0650208d69c8d897625ba6301.zip
-source_hash = 4f3af61e26df398569605fc4bcf377587ca2d8bd34b2b4bf9cdb9590cadbd662
+directory = libsrtp-37d88ecac103f4e2a5336099c64b155ba4ca3b0c
+source_url = https://github.com/versatica/libsrtp/archive/37d88ecac103f4e2a5336099c64b155ba4ca3b0c.zip
+source_filename = libsrtp-37d88ecac103f4e2a5336099c64b155ba4ca3b0c.zip
+source_hash = edf8b2c7a27ed6b6aac64c8d6a165379d3a0042bd420706b4a1825acc182441d
```
nazar-pc, line 5: Why pulling from a fork rather than upstream?

Also FYI GitHub doesn't guarantee the hash of ZIP files from random revisions to not change over time. It is only guaranteed for releases.

jmillan: > Why pulling from a fork rather than upstream?

This is a mistake, I thought I was taking it from libsrtp repo directly.

> Also FYI GitHub doesn't guarantee the hash of ZIP files from random revisions to not change over time. It is only guaranteed for releases.

That's right.., libsrtp version 3 is to be released sometime this year.., I'll ask for the timeline of the 2.5 release. I may create a PR for the enhancement to be included in 2.5.

jmillan:
I'm closing this PR for now.

worker/subprojects/libsrtp2.wrap
ibc: Why don't we create a GH relaase in our libsrtp fork and point to it in the wrap file?

nazar-pc: That would be a solution indeed

jmillan: > Also FYI GitHub doesn't guarantee the hash of ZIP files from random revisions to not change over time. It is only guaranteed for releases.

I can't read where it says that a zip file for a certain commit hash may change. Indeed the documentation says it won't.

https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#stability-of-source-code-archives

nazar-pc: > An archive of a commit ID will always have the same file contents whenever it's requested, assuming the commit ID is still in the repository and the repository's name has not changed.

Only contents because:
> The exact compression settings used to generate a zipball or tarball may change over time. The extracted contents won't change if the branch or tag doesn't change, but the outer compressed archive may have a different byte layout. GitHub will give at least six months' notice before changing compression settings.

ibc: Over a year ago GH changed the compression settings and half of the projects (including mediasoup) got into problems, so this is a real thing. However GH took the problem into consideration hence that new policy. Probably we are good whatever we do here but the only reliable way to go is by having a GH release.

nazar-pc: They have a note about 6 months warning, but I can imagine downstream projects might rely in older mediasoup releases that should ideally continue to compile in the future too.

jmillan: Nice, I'll do a release from our fork then as suggested :+1:

PR #1319, opened 2024-01-26
Fix prebuilt worker download

brynrichards:
Needed because the repository url in package.json now has a git+ prefix.


ibc (approved):
Very good catch. Thanks.

PR #1322, opened 2024-02-01
announced_ip can be a hostname as per spec

### LINKED ISSUE #1321: announced_ip can also be a hostname (not possible in Rust)
According to the [docs](https://mediasoup.org/documentation/v3/mediasoup/api/#TransportListenInfo), `announcedIp` (or `announced_ip` in Rust) in `TransportListenInfo` can be a IPv4 or a IPv6. However, it should be possible that it's also a hostname since that's a very common use case in WebRTC app deployments.

Funny thing is that it works perfect in Node if we pass `announcedIp: "some-domain.org"` (since nothing in Node or C++ validates whether it's a IP or not). However:

- Docs in the website are wrong (they state that it must be a IPv4 or IPv6).
- Rust is not ready for this since the type of `announced_ip` has type `Option<IpAddr>`.
  - @nazar-pc can we just make it a string? The worker and Rust side doesn't need to do absolutely anything with this value other than honoring it in `transport.ice_candidates`.
- Tests are needed in Node and Rust.

ibc:
Fixes #1321

As explained in the ticket, only Rust changes are needed since in Node (and Worker) `announcedIp` was already defined as optional string.

### DIFF
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -51,16 +51,17 @@ impl AppData {
 ///
 /// # Notes on usage
 /// If you use "0.0.0.0" or "::" as ip value, then you need to also provide `announced_ip`.
-#[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ListenInfo {
     /// Network protocol.
     pub protocol: Protocol,
     /// Listening IPv4 or IPv6.
     pub ip: IpAddr,
-    /// Announced IPv4 or IPv6 (useful when running mediasoup behind NAT with private IP).
+    /// Announced IPv4, IPv6 or hostname (useful when running mediasoup behind
+    /// NAT with private IP).
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub announced_ip: Option<IpAddr>,
+    pub announced_ip: Option<String>,
     /// Listening port.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub port: Option<u16>,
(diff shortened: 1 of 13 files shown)


rust/src/data_structures.rs
```diff
@@ -76,14 +77,14 @@ pub struct ListenInfo {
 }
 
 impl ListenInfo {
-    pub(crate) fn to_fbs(self) -> transport::ListenInfo {
+    pub(crate) fn to_fbs(&self) -> transport::ListenInfo {
```
ibc, line 80: Rust told me to change this to avoid confusing naming since `self` is passed as reference after removing `Copy` derive (or something like that).

nazar-pc (approved):
Since this is a breaking change for Rust API anyway, maybe rename `announced_ip` to `announced_hostname` since it is not guaranteed to always be an IP? Makes sense otherwise.

rust/src/messages.rs
```diff
@@ -644,7 +644,7 @@ impl RouterCreateWebrtcTransportListen {
                     web_rtc_transport::ListenIndividual {
                         listen_infos: listen_infos
                             .iter()
-                            .map(|listen_info| listen_info.to_fbs())
+                            .map(|listen_info| listen_info.clone().to_fbs())
```
ibc: Finally (after some other change) this is not needed, so I'm removing it.

ibc:
> Since this is a breaking change for Rust API anyway, maybe rename `announced_ip` to `announced_hostname` since it is not guaranteed to always be an IP? Makes sense otherwise.

It should be `announced_address` instead (to match the naming used in WebRTC). 

However I'd prefer if we don't rename it in Rust now. It would be a bit unexpected to have `announced_ip` in Node and `announced_address` in Rust. Do you mind if we do this in the future in a new version with breaking changes?

nazar-pc:
> However I'd prefer if we don't rename it in Rust now. It would be a bit unexpected to have announced_ip in Node and announced_address in Rust. Do you mind if we do this in the future in a new version with breaking changes?

We can also rename it in Node.js, but also keep the old field for backwards compatibility reasons. Ugly, I know, but we can get people upgrade to it already and maybe even print warnings in logs of those who didn't upgrade to the new field name.

Not critical or a blocker of course.

jmillan (approved):

ibc:
> We can also rename it in Node.js, but also keep the old field for backwards compatibility reasons. Ugly, I know, but we can get people upgrade to it already and maybe even print warnings in logs of those who didn't upgrade to the new field name.

This deserves a separate story since it's huger than expected:

https://github.com/versatica/mediasoup/issues/1323

PR #1324, opened 2024-02-01
Rename "announced ip" to "announced address"

### LINKED ISSUE #1323: Rename "announced ip" to "announced address"
As per discussion in https://github.com/versatica/mediasoup/pull/1322:

Tasks:

- [x] Rust: In `TransportListenInfo` rename `announced_ip` to `announced_address`.
- [x] Rust: In `IceCandidate` rename `ip` to `address`.
- [x] Rust: In `TransportTuple` rename `local_ip` to `local_address`.
- [x] Node: In `TransportListenInfo` rename `announcedIp` to `announcedAddress` and keep backwards compatibility.
- [x] Node: In `IceCandidate` rename `ip` to `address` and keep backwards compatibility.
- [x] Node: In `TransportTuple` rename `localIp` to `localAddress` and keep backwards compatibility.
- [x] Docs: Update.

ibc:
Fixes #1323

- Rust: In `TransportListenInfo` rename `announced_ip` to `announced_address`.
- Rust: In `IceCandidate` rename `ip` to `address`.
- Rust: In `TransportTuple` rename `local_ip` to `local_address`.
- Node: In `TransportListenInfo` rename `announcedIp` to `announcedAddress` and keep backwards compatibility.
- Node: In `IceCandidate` rename `ip` to `address` and keep backwards compatibility.
- Node: In `TransportTuple` rename `localIp` to `localAddress` and keep backwards compatibility.

**NOTE:** Why has been `TransportTuple.localIp` renamed to `TransportTuple.localAddress`? Because indeed its value is an announced address if a hostname was given in `announcedAddress` in a `TransportListenInfo`.

### DIFF
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -297,14 +297,14 @@ impl IceState {
 /// `PipeTransport`, or via dynamic detection as it happens in `WebRtcTransport` (in which the
 /// remote media address is detected by ICE means), or in `PlainTransport` (when using `comedia`
 /// mode).
-#[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
 #[serde(untagged)]
 pub enum TransportTuple {
     /// Transport tuple with remote endpoint info.
     #[serde(rename_all = "camelCase")]
     WithRemote {
         /// Local IP address or hostname.
-        local_address: IpAddr,
+        local_address: String,
         /// Local port.
         local_port: u16,
         /// Remote IP address.
@@ -318,7 +318,7 @@ pub enum TransportTuple {
     #[serde(rename_all = "camelCase")]
     LocalOnly {
         /// Local IP address or hostname.
-        local_address: IpAddr,
+        local_address: String,
         /// Local port.
         local_port: u16,
         /// Protocol
@@ -328,7 +328,7 @@ pub enum TransportTuple {
 
 impl TransportTuple {
     /// Local IP address or hostname.
-    pub fn local_address(&self) -> IpAddr {
+    pub fn local_address(&self) -> String {
         let (Self::WithRemote { local_address, .. } | Self::LocalOnly { local_address, .. }) = self;
         *local_address
     }
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -369,14 +369,14 @@ impl TransportTuple {
                 local_address: tuple
                     .local_address
                     .parse()
-                    .expect("Error parsing IP address"),
+                    .expect("Error parsing local address"),
                 local_port: tuple.local_port,
                 remote_ip: tuple
                     .remote_ip
                     .as_ref()
                     .unwrap()
                     .parse()
-                    .expect("Error parsing IP address"),
+                    .expect("Error parsing remote IP address"),
                 remote_port: tuple.remote_port,
                 protocol: Protocol::from_fbs(tuple.protocol),
             },
@@ -384,7 +384,7 @@ impl TransportTuple {
                 local_address: tuple
                     .local_address
                     .parse()
-                    .expect("Error parsing IP address"),
+                    .expect("Error parsing local address"),
                 local_port: tuple.local_port,
                 protocol: Protocol::from_fbs(tuple.protocol),
             },
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -328,9 +328,9 @@ pub enum TransportTuple {
 
 impl TransportTuple {
     /// Local IP address or hostname.
-    pub fn local_address(&self) -> String {
+    pub fn local_address(&self) -> &String {
         let (Self::WithRemote { local_address, .. } | Self::LocalOnly { local_address, .. }) = self;
-        *local_address
+        local_address
     }
 
     /// Local port.
--- a/rust/src/data_structures.rs
+++ b/rust/src/data_structures.rs
@@ -50,7 +50,8 @@ impl AppData {
 /// Listening protocol, IP and port for [`WebRtcServer`](crate::webrtc_server::WebRtcServer) to listen on.
 ///
 /// # Notes on usage
-/// If you use "0.0.0.0" or "::" as ip value, then you need to also provide `announced_ip`.
+/// If you use "0.0.0.0" or "::" as ip value, then you need to also provide
+/// `announced_address`.
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ListenInfo {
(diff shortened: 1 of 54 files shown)


ibc:
**TODO:** Once merged/released, docs in website must be updated.

ibc:
Missing thing, `local_address` in `TransportTuple` in Rust must not be an `IPAddr` but a `String`. On it.

rust/src/data_structures.rs
```diff
@@ -323,10 +327,10 @@ pub enum TransportTuple {
 }
 
 impl TransportTuple {
-    /// Local IP address.
-    pub fn local_ip(&self) -> IpAddr {
-        let (Self::WithRemote { local_ip, .. } | Self::LocalOnly { local_ip, .. }) = self;
-        *local_ip
+    /// Local IP address or hostname.
+    pub fn local_address(&self) -> String {
```
ibc: @nazar-pc should this return  `-> &String` instead (and hence replace `*local_address` with `local_address`? More fixed are needed no matter which way we go here, however I expect that returning a string reference is preferred over returning a copied one.

nazar-pc: I think `&str` makes most sense, for some use cases you don't need an owned value and will avoid allocation that way.

ibc:
I need help to resolve this:

```
error[E0507]: cannot move out of dereference of `parking_lot::lock_api::MutexGuard<'_, parking_lot::RawMutex, TransportTuple>`
   --> rust/src/router/pipe_transport.rs:727:9
    |
727 |         *self.inner.data.tuple.lock()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because value has type `TransportTuple`, which does not implement the `Copy` trait

error[E0382]: borrow of moved value: `tuple`
   --> rust/src/router/plain_transport.rs:691:56
    |
688 |                         Notification::Tuple { tuple } => {
    |                                               ----- move occurs because `tuple` has type `TransportTuple`, which does not implement the `Copy` trait
689 |                             *data.tuple.lock() = tuple;
    |                                                  ----- value moved here
690 |
691 |                             handlers.tuple.call_simple(&tuple);
    |                                                        ^^^^^^ value borrowed here after move
```

In the first error I've tried `*self.inner.data.tuple.clone().lock()` but:

```
error[E0599]: no method named `clone` found for struct `parking_lot::lock_api::Mutex` in the current scope
   --> rust/src/router/pipe_transport.rs:727:32
    |
727 |         *self.inner.data.tuple.clone().lock()
    |                                ^^^^^ method not found in `Mutex<RawMutex, TransportTuple>`
```

This mutex/lock thing is above my (increasing) little knowledge, so @nazar-pc help plz.

nazar-pc:
I think you want to clone inner data, in which case it should be `lock().clone()`

ibc:
> I think you want to clone inner data, in which case it should be `lock().clone()`

I tried it already but...

```
error[E0614]: type `TransportTuple` cannot be dereferenced
   --> rust/src/router/pipe_transport.rs:727:9
    |
727 |         *self.inner.data.tuple.lock().clone()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

nazar-pc:
`.lock()` method returns something like `MutexGruard`. This is a wrapper type that locks unlocks Mutex when dropped. This `MutexGuard` can be dereferenced to the type mutex contains. So `*mutex.lock()` is essentially similar to this:
```rust
let number = 5;
let number_copy = *&number;
```

Since the type you have is no longer `Copy`, you need to clone it explicitly.

So what was `*self.inner.data.tuple.lock()` before (can be also written as `self.inner.data.tuple.lock().clone()` explicitly) now needs to be written as `self.inner.data.tuple.lock().clone()` due to explicit cloning.

So essentially you'll do this:
```rust
let s = "Hello".to_string();
let s_copy = (&s).clone();
```

ibc:
So indeed replacing `*self.inner.data.tuple.lock()` with `self.inner.data.tuple.lock().clone()` works although I cannot understand how both things can return same type. AFAIU `xxxx` and `xxxx.clone()` return the same type, so how can `*xxxx` and `xxxx.clone()` return the same type as well?

ibc:
In addition I've also fixed another issue as follows:

```
error[E0382]: borrow of moved value: `tuple`
   --> rust/src/router/plain_transport.rs:691:56
    |
688 |                         Notification::Tuple { tuple } => {
    |                                               ----- move occurs because `tuple` has type `TransportTuple`, which does not implement the `Copy` trait
689 |                             *data.tuple.lock() = tuple;
    |                                                  ----- value moved here
690 |
691 |                             handlers.tuple.call_simple(&tuple);
    |                                                        ^^^^^^ value borrowed here after move
```

Solution in line 689:

```rust
*data.tuple.lock() = tuple.clone();
```

Hope it's ok. IMHO it is since it follows same rationale as discussed in above comments.

nazar-pc:
`MutexGuard` doesn't implement `Clone`, so compiler will try to call `Clone` on the type that it dereferences to, which is inner value you actually want.
https://doc.rust-lang.org/book/ch15-02-deref.html

ibc:
Ok this is ready for review. @nazar-pc please.

rust/src/data_structures.rs
```diff
@@ -84,7 +85,10 @@ impl ListenInfo {
                 Protocol::Udp => transport::Protocol::Udp,
             },
             ip: self.ip.to_string(),
-            announced_ip: self.announced_ip.clone().map(|ip| ip.to_string()),
+            announced_address: self
+                .announced_address
+                .clone()
+                .map(|address| address.to_string()),
```
nazar-pc: Two allocations look strange. I'm pretty sure this will also work:
```suggestion
            announced_address: self
                .announced_address
                .as_ref()
                .map(|address| address.to_string()),
```

Though this is not new, it was like this before, I just comment on what I noticed.

nazar-pc (commented):
I may not be looking carefully, but looks like TypeScript doesn't process `announcedIp` correctly now. It seems to send it to worker, but worker only cares about `announcedAddress` now. You should add something like `announcedAddress = announcedAddress || announcedIp` on TypeScript side and not send `announcedIp` to worker anymore.

ibc:
Moving back to draft PR because I forgot lot of things.

rust/src/data_structures.rs
ibc: You mean that this thing should also be done in other changes I did?

nazar-pc: Maybe, but I was commenting just about this one

ibc:
@nazar-pc, I should also update Changelog and versions in Rust, right?

rust/src/data_structures.rs
ibc: On it.

ibc: Nope, I don't see any other place in which `as_ref()` can be used instead of cloning.

nazar-pc:
> @nazar-pc, I should also update Changelog and versions in Rust, right?

Yes, please

ibc:
Rust versions bump done.

worker/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mediasoup-sys"
-version = "0.7.2"
+version = "0.7.3"
```
nazar-pc: This is a breaking API change between library and worker, please bump to 0.8.0 instead

ibc: Done.

nazar-pc (approved):
Makes sense to me

PR #1326, opened 2024-02-02
Cleanup remaining payload channel references

jmillan:


nazar-pc (approved):
This will require minor version bump for `mediasoup-sys` due to change in public API

ibc:
Rationale please?

nazar-pc:
Cargo will be happy to install compatible (according to semver) version of the package that is newer than technically required. Since arguments changed, older `mediasoup` will no longer be compatible with newer `mediasoup-sys`. So instead of increasing patch version we have to increase minor version to prevent breaking changes for people.

jmillan:
It can be merged since the minor version has been increased here https://github.com/versatica/mediasoup/pull/1324

PR #1328, opened 2024-02-05
CI: Add macos-14

ibc:
As per GH announcement: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

Also remove the NPM 'release:upload-mac-arm-prebuilt-worker' since it's no longer needed.

### DIFF
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -20,6 +20,8 @@ jobs:
           - os: ubuntu-22.04
             node: 20
           - os: macos-12
+            node: 18
+          - os: macos-14
             node: 20
           - os: windows-2022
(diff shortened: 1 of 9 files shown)


nazar-pc (approved):
I'd also remove macos-12, but up to you

ibc:
> I'd also remove macos-12, but up to you

Let's keep it for a while. Let's at least test latest 2 versions.

ibc:
TODO: If this works we should remove the `uploadMacArmPrebuiltWorker()` stuff in `npm-scripts` since it's no longer needed. But I'll do it later once we verify that.

UPDATE: No, I'll do it now.

jmillan (approved):

PR #1329, opened 2024-02-05
Add Send and Sync to dyn Errors

### LINKED ISSUE #1325: [Rust] `Box<dyn Error>` error types make mediasoup error handling annoying

PaulOlteanu:
Closes https://github.com/versatica/mediasoup/issues/1325


ibc:
Unclear why CI tests don't show up in this PR. GH is crazy...

ibc:
Now they do... running them.

@nazar-pc I assume this is ok as per previous discussions so we can safely merge it, right?

nazar-pc (approved):

PR #1088, opened 2023-05-26
 Make transport-cc feedback work similarly to libwebrtc

penguinol:
fix #1059

Issue #1059, opened 2023-04-11
TransportCC FeedBack implement is different from libwebrtc


penguinol:
Currently in mediasoup, a new transport-cc feedback will always start form last seq num in previous feedback, and this will result in a higher packet loss rate than actual when receving packet out of order.

For example:
We sent first:
100(R), 101(R), 102(NR), 103(R), 104(NR)

If we received 102 before sending next feedback, we should send：
102(R), 103(R), 104(NR), 105(R) 106(R), 107(R)

But currently we just send:
105(R) 106(R), 107(R)...

See follow:
https://webrtc.googlesource.com/src/+/refs/heads/main/modules/remote_bitrate_estimator/remote_estimator_proxy.cc#73
https://webrtc.googlesource.com/src/+/refs/heads/main/modules/remote_bitrate_estimator/remote_estimator_proxy.cc#113

jmillan:
Wow, I can't see such behaviour defined in the spec, but maybe lightly this paragraph:

https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01#section-3.1.1
```txt
Note: In the case the base sequence number is decreased, creating a
   window overlapping the previous feedback messages, the status for any
   packets previously reported as received must be marked as "Packet not
   received" and thus no delta included for that symbol.
```

We currently ignore those packets with lower wide seq num than the last seen: https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/RTCP/FeedbackRtpTransport.cpp#L304.

Implementing this would definitely require a sustancial rewrite of this RTCP packet implementation.

Can you @penguinol check which should be the correct behaviour in detail?

penguinol:
In webrtc they use `RemoteEstimatorProxy::packet_arrival_times_` to save packet arrival time in last 500ms.
And webrtc also does not observer the spec, they still mark those packets(which were reported previously as `Received`) as `Received`.
But I think it's ok, webrtc use `TransportFeedbackAdapter::history_` to save packet sending time, and once a packet is reported as `Received`, it will be removed form `history_`, and in the next feedback, it will be ignore because webrtc can't find it's `send_time` in history.
And the benefit of marking reported received packet as `Received` is if the first feedback is loss, we can still know it is received in the second feedback.

penguinol:
But there is also another issuse in webrtc, for example, if we received two feedbacks:
1: 100(R) 101(NR) 102(R) 103(R)
[...]

PR #1339, opened 2024-02-20
Fix regression (crash) in transport-cc feedback generation

### LINKED ISSUE #1336: (ABORT) RTC::RTCP::FeedbackRtpTransport::AddPacket() | failed assertion `baseSet': base not set
- Operating system: Linux Debian
- mediasoup version:3.13.19

### Issue description

Logs:

```
2024-02-20T07:08:16.181Z mediasoup:ERROR:Worker (stderr) (ABORT) RTC::RTCP::FeedbackRtpTransport::AddPacket() | failed assertion `baseSet': base not set
2024-02-20T07:08:16.569Z mediasoup:ERROR:Channel Producer Channel error: Error: read ECONNRESET
2024-02-20T07:08:16.574Z mediasoup-demo-server:WARN:Room _createDataConsumer() | failed:Error: request timeout at Timeout._onTimeout (/var/www/v3demo.mediasoup.org/node_modules/protoo-server/lib/Peer.js:156:14) at listOnTimeout (node:internal/timers:569:17) at process.processTimers (node:internal/timers:512:7)
2024-02-20T07:08:16.784Z mediasoup:ERROR:Worker worker process died unexpectedly [pid:1695953, code:null, signal:SIGABRT]
2024-02-20T07:08:16.823Z mediasoup-demo-server:ERROR mediasoup Worker died, exiting  in 2 seconds... [pid:1695953]
```

Core dump:

```
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
[Current thread is 1 (Thread 0x7f0774b3e440 (LWP 1695953))]
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007f0774bcad9f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007f0774b7bf32 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f0774b66472 in __GI_abort () at ./stdlib/abort.c:79
#4  0x000055b064e4bd6e in RTC::RTCP::FeedbackRtpTransportPacket::AddPacket(unsigned short, unsigned long, unsigned long) ()
#5  0x000055b064dd9ade in RTC::TransportCongestionControlServer::FillAndSendTransportCcFeedback() ()
#6  0x000055b0650dcf3e in uv.run_timers ()
#7  0x000055b0650e0c88 in uv_run ()
#8  0x000055b064c9a849 in DepLibUV::RunLoop() ()
#9  0x000055b064ca4d62 in Worker::Worker(Channel::ChannelSocket*) ()
#10 0x000055b064c998e4 in mediasoup_worker_run ()
#11 0x000055b064c98a2c in main ()
```

Clearly a regression in PR https://github.com/versatica/mediasoup/pull/1088 " Make transport-cc feedback work similarly to libwebrtc".

ibc:
- Fixes #1336

### Details

- Regression introduced in PR #1088, which has many issues.
- Basically there are cases in which `this->transportCcFeedbackPacket` is reset so it's a complete new packet and hence its `baseSet` is `false`.
- However in `TransportCongestionControlServer::FillAndSendTransportCcFeedback()` in which we just call `this->transportCcFeedbackPacket.SetBase()` once at the top, but then we enter a loop in which the packet maybe full so it's sent (or other things) and hence we **reset** the packet while in the loop. And then at the top of the loop `this->transportCcFeedbackPacket.AddPacket()` is called so the assertion fails because its `SetBase()` was not called on that fresh packet.
- So this PR calls `SetBase()` conditionally in every iteration of the loop by checking if current feedback has the base set or not.
- Also add a missing `break` in a `case` block in that loop.
- Also set proper packet count in all cases in which we reset `this->transportCcFeedbackPacket`.
- Fix crash scenario in which feedback-cc packet is full but cannot be sent: https://github.com/versatica/mediasoup/pull/1339/commits/b8a411b1d2b200040f053a1a79649b96f68d2085

### DIFF
--- a/worker/src/RTC/TransportCongestionControlServer.cpp
+++ b/worker/src/RTC/TransportCongestionControlServer.cpp
@@ -96,6 +96,9 @@ namespace RTC
 				// Create a new feedback packet.
 				this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(0u, 0u));
 
+				// Set current packet count.
+				this->transportCcFeedbackPacket->SetFeedbackPacketCount(this->transportCcFeedbackPacketCount);
+
 				break;
 			}
 
@@ -131,9 +134,10 @@ namespace RTC
 					break;
 				}
 
-				// We may receive packets with sequence number lower than the one in previous
-				// tcc feedback, these packets may have been reported as lost previously,
-				// therefore we need to reset the start sequence num for the next tcc feedback.
+				// We may receive packets with sequence number lower than the one in
+				// previous tcc feedback, these packets may have been reported as lost
+				// previously, therefore we need to reset the start sequence num for the
+				// next tcc feedback.
 				if (
 				  !this->transportWideSeqNumberReceived ||
 				  RTC::SeqManager<uint16_t>::IsSeqLowerThan(
@@ -192,13 +196,17 @@ namespace RTC
 			return;
 		}
 
+		auto baseTimestamp = it->second;
+
 		// Set base sequence num and reference time.
-		this->transportCcFeedbackPacket->SetBase(this->transportCcFeedbackWideSeqNumStart, it->second);
+		this->transportCcFeedbackPacket->SetBase(this->transportCcFeedbackWideSeqNumStart, baseTimestamp);
 
 		for (; it != this->mapPacketArrivalTimes.end(); ++it)
 		{
-			auto result =
-			  this->transportCcFeedbackPacket->AddPacket(it->first, it->second, this->maxRtcpPacketLen);
+			auto sequenceNumber = it->first;
+			auto timestamp      = it->second;
+			auto result         = this->transportCcFeedbackPacket->AddPacket(
+        sequenceNumber, timestamp, this->maxRtcpPacketLen);
 
 			switch (result)
 			{
@@ -223,6 +231,14 @@ namespace RTC
 					// Create a new feedback packet.
 					this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(
 					  this->transportCcFeedbackSenderSsrc, this->transportCcFeedbackMediaSsrc));
+
+					// Set current packet count.
+					// NOTE: Do not increment it since the previous ongoing feedback
+					// packet was not sent.
+					this->transportCcFeedbackPacket->SetFeedbackPacketCount(
+					  this->transportCcFeedbackPacketCount);
+
+					break;
 				}
 
 				case RTC::RTCP::FeedbackRtpTransportPacket::AddPacketResult::FATAL:
@@ -231,7 +247,7 @@ namespace RTC
 					this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(
 					  this->transportCcFeedbackSenderSsrc, this->transportCcFeedbackMediaSsrc));
 
-					// Use current packet count.
+					// Set current packet count.
 					// NOTE: Do not increment it since the previous ongoing feedback
 					// packet was not sent.
 					this->transportCcFeedbackPacket->SetFeedbackPacketCount(
@@ -242,6 +258,8 @@ namespace RTC
 			}
 		}
 
+		// It may happen that the packet is empty (no deltas) but in that case
+		// SendTransportCcFeedback() won't send it so we are safe.
 		SendTransportCcFeedback();
 	}
 
@@ -264,7 +282,7 @@ namespace RTC
 		}
 	}
 
-	inline void TransportCongestionControlServer::SendTransportCcFeedback()
+	void TransportCongestionControlServer::SendTransportCcFeedback()
 	{
 		MS_TRACE();
 
@@ -307,8 +325,7 @@ namespace RTC
 		this->transportCcFeedbackWideSeqNumStart = latestWideSeqNumber + 1;
 	}
 
-	inline void TransportCongestionControlServer::MayDropOldPacketArrivalTimes(
-	  uint16_t seqNum, uint64_t nowMs)
+	void TransportCongestionControlServer::MayDropOldPacketArrivalTimes(uint16_t seqNum, uint64_t nowMs)
 	{
 		MS_TRACE();
 
@@ -330,7 +347,7 @@ namespace RTC
 		}
 	}
 
-	inline void TransportCongestionControlServer::MaySendLimitationRembFeedback(uint64_t nowMs)
+	void TransportCongestionControlServer::MaySendLimitationRembFeedback(uint64_t nowMs)
 	{
 		MS_TRACE();
 
@@ -402,7 +419,7 @@ namespace RTC
 		this->packetLoss = totalPacketLoss / samples;
 	}
 
-	inline void TransportCongestionControlServer::OnRembServerAvailableBitrate(
+	void TransportCongestionControlServer::OnRembServerAvailableBitrate(
 	  const webrtc::RemoteBitrateEstimator* /*rembServer*/,
 	  const std::vector<uint32_t>& ssrcs,
 	  uint32_t availableBitrate)
@@ -440,7 +457,7 @@ namespace RTC
 		this->listener->OnTransportCongestionControlServerSendRtcpPacket(this, &packet);
 	}
 
-	inline void TransportCongestionControlServer::OnTimer(TimerHandle* timer)
+	void TransportCongestionControlServer::OnTimer(TimerHandle* timer)
 	{
 		MS_TRACE();
 
--- a/worker/src/RTC/TransportCongestionControlServer.cpp
+++ b/worker/src/RTC/TransportCongestionControlServer.cpp
@@ -35,9 +35,6 @@ namespace RTC
 				// Create a feedback packet.
 				this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(0u, 0u));
 
-				// Set initial packet count.
-				this->transportCcFeedbackPacket->SetFeedbackPacketCount(this->transportCcFeedbackPacketCount);
-
 				// Create the feedback send periodic timer.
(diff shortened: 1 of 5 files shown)


ibc:
This is a draft PR because I need help. See TODO above.

ibc:
I think I know how to solve it properly. Just a question:

Does somebody remember what is `feedbackPacketCount` in `FeedbackRtpTransportPacket`? It's obviously part of the packet fields but I don't remember its meaning. Specifically if an endpoint sends 2 sequential CC feedback packets, what should be the value of that field in those packets?

And why are we setting it when initiating transport server control?

```c++
	/* Instance methods. */

	TransportCongestionControlServer::TransportCongestionControlServer(
	  RTC::TransportCongestionControlServer::Listener* listener,
	  RTC::BweType bweType,
	  size_t maxRtcpPacketLen)
	  : listener(listener), bweType(bweType), maxRtcpPacketLen(maxRtcpPacketLen)
	{
		MS_TRACE();

		switch (this->bweType)
		{
			case RTC::BweType::TRANSPORT_CC:
			{
				// Create a feedback packet.
				this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(0u, 0u));

				// Set initial packet count.
				this->transportCcFeedbackPacket->SetFeedbackPacketCount(this->transportCcFeedbackPacketCount);
```

ibc:
Ah, is the number of CC feedback packet sent, right? so it's just a counter that must be incremented every time a feedback packet is sent. Right?

ibc:
Concerns resolved in theory. PR ready for review.

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -35,9 +35,6 @@ namespace RTC
 				// Create a feedback packet.
 				this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(0u, 0u));
 
-				// Set initial packet count.
-				this->transportCcFeedbackPacket->SetFeedbackPacketCount(this->transportCcFeedbackPacketCount);
```
ibc, line 39: This is not needed in the constructor because obviously the counter is zero.

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -264,7 +285,7 @@ namespace RTC
 		}
 	}
 
-	inline void TransportCongestionControlServer::SendTransportCcFeedback()
```
ibc, line 267: Unrelated by let's remove step by step those ugly `inline` usages. We are in 2024.

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -192,13 +193,23 @@ namespace RTC
...
-			auto result =
-			  this->transportCcFeedbackPacket->AddPacket(it->first, it->second, this->maxRtcpPacketLen);
+			auto sequenceNumber = it->first;
+			auto timestamp      = it->second;
+
+			// If the base is not set in this packet let's set it.
+			// NOTE: This maybe needed many times during this loop since the current
+			// feedback packet maybe a fresh new one if the previous one was full (so
+			// already sent) or failed to be built.
+			if (!this->transportCcFeedbackPacket->IsBaseSet())
```
jmillan, line 202: IMO this logic should be inside `this->transportCcFeedbackPacket->AddPacket()` itself. So It sets the base if not set. It's a `TransportCcFeedbackPacket` responsibility.

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -218,11 +229,19 @@ namespace RTC
...
+					MS_WARN_TAG(rtcp, "transport-cc feedback packet is exceeded");
 
 					// Create a new feedback packet.
 					this->transportCcFeedbackPacket.reset(new RTC::RTCP::FeedbackRtpTransportPacket(
 					  this->transportCcFeedbackSenderSsrc, this->transportCcFeedbackMediaSsrc));
+
+					// Set current packet count.
+					// NOTE: Do not increment it since the previous ongoing feedback
+					// packet was not sent.
+					this->transportCcFeedbackPacket->SetFeedbackPacketCount(
```
jmillan: Do we need to set this value? Doesn't already have the same value already?

jmillan (commented):
LGTM. Just a couple cosmetic suggestions

worker/src/RTC/TransportCongestionControlServer.cpp
ibc: This is exactly how it was before the PR that introduced this regression. Please look at it.

worker/src/RTC/TransportCongestionControlServer.cpp
ibc: How? There is a .reset() just above.

worker/src/RTC/TransportCongestionControlServer.cpp
ibc: There is a difference. In the original code before the regression PR (where there was no SetBase()) the first packet/info passed to AddPacket() was used to set the base fields and then it returned early. In the new code (before and after this PR) the same packet/info is used to set SetBase() and is later passed to AddPacket() where it does make that method return early. Not sure about the implications.

worker/src/RTC/TransportCongestionControlServer.cpp
jmillan: That's right :+1:

worker/src/RTC/TransportCongestionControlServer.cpp
ibc: So there is a difference. The AddPacket() in "modern" code will NOT "ignore" the packet that was used in `SetBase()` due to this:

```c++
// If the wide sequence number of the new packet is lower than the latest seen,
// ignore it.
// NOTE: Not very spec compliant but libwebrtc does it.
// Also ignore if the sequence number matches the latest seen.
if (!RTC::SeqManager<uint16_t>::IsSeqHigherThan(sequenceNumber, this->latestSequenceNumber))
{
	return AddPacketResult::SUCCESS;
}
```

because the value of `this->latestSequenceNumber` was set in `SetBase()` this way:

```c++
this->latestSequenceNumber = sequenceNumber - 1;
```

So the condition above doesn't match. In summary, in the "modern" code (after the addition of `SetBase()`) the first packet is used to set the base fields and it is ALSO processed meaning that it will generate a delta. I don't really know why this is needed.

ibc: Anyway I'm gonna get rid of that `SetBase()` and make the first packet givne to `AddPacket()` set the base (as before) and also be processed.

ibc: Nope, I am wrong. And honestly I don't understand it. In the modern code this is how it works:

- `SetBase()` is called with `seqNumber: this->transportCcFeedbackWideSeqNumStart` and `timestamp:` the value of the first entry in the `mapPacketArrivalTimes` map (after doing `lower_bound(this->transportCcFeedbackWideSeqNumStart)` on it).
- And then the first time `AddPacket()` is called its arguments are `seqNumber:` the key of the first entry in the `mapPacketArrivalTimes` (after doing `lower_bound(this->transportCcFeedbackWideSeqNumStart)` on it) and `timestamp:` the value of that entry.
- This is, `timestamp` has the same value in the argument passed to `SetBase()` and in the first call to `AddPacket()`.

I'm gonna assume that there is a good rationale for this change (read https://github.com/versatica/mediasoup/issues/1059 in which indeed there is complex but good rationale about this) so I won't change it.

ibc:
I've done this change to test:

```diff
uint16_t FeedbackRtpTransportPacket::maxMissingPackets{ (1 << 13) - 1 };
// uint16_t FeedbackRtpTransportPacket::maxPacketStatusCount{ (1 << 16) - 1 };
```

and it crashes:

```
(ABORT) RTC::RTCP::FeedbackRtpTransport::AddPacket() | failed assertion `!IsFull()': packet is full

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff80d79b1e2 __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff80d7d2ee6 pthread_kill + 263
2   libsystem_c.dylib             	    0x7ff80d6f9b45 abort + 123
3   mediasoup-worker              	       0x1095e6ea0 RTC::RTCP::FeedbackRtpTransportPacket::AddPacket(unsigned short, unsigned long long, unsigned long) (.cold.2) + 64
4   mediasoup-worker              	       0x1092b4c47 RTC::RTCP::FeedbackRtpTransportPacket::AddPacket(unsigned short, unsigned long long, unsigned long) + 231
5   mediasoup-worker              	       0x109274e22 RTC::TransportCongestionControlServer::FillAndSendTransportCcFeedback() + 210
```

Why? Because when `this->transportCcFeedbackPacket->IsFull()` we call `SendTransportCcFeedback()` which is supposed to send the feedback and reset `this->transportCcFeedbackPacket` so in the next iteration of the loop there is a fresh feedback packet which obviously is not full, however there are cases in which `SendTransportCcFeedback()` does NOT send the feedback and does NOT reset it:

```c++
if (!this->transportCcFeedbackPacket->IsSerializable())
{
	return;
}
```

So here another regression in that PR or perhaps

ibc:
Fixing it as follows:
CleanShot-2024-02-20-at-17 37 57@2x

jmillan (approved):

PR #1345, opened 2024-02-27
Fix DTLS packets do not honor configured DTLS MTU (attempt 3)

### LINKED ISSUE #1100: DTLS fragment exceeds MTU size if the certificate is large.

ibc:
- Fixes #1100
- Based on PR #1156 which was based on PR #1143

### Details

This PR is the same as PR #1156. However that PR introduced a memory leak (see issue #1340). This PR fixes that leak by following the discussion and research in associated issues and PRs, specially here: https://github.com/versatica/mediasoup/issues/1340#issuecomment-1965375082

To prove that this PR works at DTLS level, here the DTLS messages sent from mediasoup to browser while doing the DTLS handshake:

```
RTC::DtlsTransport::SendDtlsData() | 1349 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 1349 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 1349 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 1349 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 1349 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 242 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 67 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 129 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 385 bytes of DTLS data ready to be sent
RTC::DtlsTransport::SendDtlsData() | 45 bytes of DTLS data ready to be sent
```

which clearly corresponds to what we see in Wireshark (there is a huge certificate so it's splitted into N DTLS Certificate fragments):

CleanShot-2024-02-27-at-16 51 19@2x

Note that sizes of each message match the DTLS data sent by mediasoup plus UDP and IP encapsulation (32 bytes always).


ibc:
CC @gkrol-katmai @pnts-se

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -708,15 +734,19 @@ namespace RTC
...
-		// TODO: This is not honored, see issue:
-		//   https://github.com/versatica/mediasoup/issues/1100
 		SSL_set_mtu(this->ssl, DtlsMtu);
 		DTLS_set_link_mtu(this->ssl, DtlsMtu);
 
+		// We want to monitor OpenSSL write operations into our |sslBioToNetwork|
+		// buffer so we can immediately send those DTLS bytes (containing full DTLS
+		// messages, or valid DTLS fragment messages, or combination of them) to
+		// the endpoint, and hence we honor the configured DTLS MTU.
+		BIO_set_callback_ex(this->sslBioToNetwork, onSslBioOut);
```
jmillan, line 754: wasn't this an experimental function used to debug?

https://www.openssl.org/docs/man3.2/man3/BIO_set_callback_ex.html

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1022,9 +1047,24 @@ namespace RTC
...
+
+		MS_DEBUG_DEV("%zu bytes of DTLS data ready to be sent", len);
 
-		// Send data.
-		SendPendingOutgoingDtlsData();
+		// Notify the listener.
+		this->listener->OnDtlsTransportSendData(this, data, len);
+
+		// Clear the BIO buffer.
+		// NOTE: Here we now that OpenSSL called the onSslBioOut() callback with
```
jmillan:
```suggestion
		// NOTE: Here we now that OpenSSL called the onSslBioOut() callback
```

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1022,9 +1047,24 @@ namespace RTC
...
+		// with all the byte length that it previously wrote into the BIO mem
+		// buffer (this->sslBioToNetwork), so it's safe to clear the entire buffer
+		// after sending its entire DTLS data to the endpoint.
+		// NOTE: the (void) avoids the -Wunused-value warning.
+		(void)BIO_reset(this->sslBioToNetwork);
```
jmillan: `BIO_reset()` returns -1 on failure. It would be nice to log it in case of error.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -47,15 +47,41 @@ inline static void onSslInfo(const SSL* ssl, int where, int ret)
...
+  int ret,
+  size_t* /*processed*/)
+{
+	long resultOfcallback = (operationType == BIO_CB_RETURN) ? static_cast<long>(ret) : 1;
+
+	// This callback is called twice for write operations:
+	// - First one with operationType = BIO_CB_WRITE.
+	// - Second one with operationType = BIO_CB_RETURN | BIO_CB_WRITE.
+	// We only care about the former.
+	if (operationType == BIO_CB_WRITE && argp && len > 0)
```
jmillan: Cosmetic for readability:

```suggestion
	if ((operationType == BIO_CB_WRITE) && argp && len > 0)
```

jmillan (approved):
:+1: :+1:

worker/src/RTC/DtlsTransport.cpp
ibc: No.

> `BIO_set_callback_ex()` and `BIO_get_callback_ex()` set and retrieve the BIO callback. The callback is called during most high-level BIO operations. It can be used for debugging purposes to trace operations on a BIO or **to modify its operation**.

worker/src/RTC/DtlsTransport.cpp
ibc: No.

> `BIO_set_callback_ex()` and `BIO_get_callback_ex()` set and retrieve the BIO callback. The callback is called during most high-level BIO operations. It can be used for debugging purposes to trace operations on a BIO or **to modify its operation**.

worker/src/RTC/DtlsTransport.cpp
ibc: Fixed in local branch.

worker/src/RTC/DtlsTransport.cpp
ibc: It must return 1 on success on our BIO. Pushed.

worker/src/RTC/DtlsTransport.cpp
ibc: Pushed.

gkrol-katmai:
The code looks good to me! I think we now have a good grasp of what's going on. I know that adding the 'reset' in that function will fix the leak. I haven't tested this PR.

Not sure what the coding standards are for comments, but I would suggest we add comments to onSslBioOut and DtlsTransport::SendDtlsData (above the implementation). Those should describe the effect of the function and when/why it gets called. Preferably `/** */` style. We're doing some subtle things here so we need to be extra clear.

ibc:
> I know that adding the 'reset' in that function will fix the leak. I haven't tested this PR.

I think this is also proven here https://github.com/versatica/mediasoup/issues/1340#issuecomment-1966393672 where it's clear that the `onSslBioOut` is always called with `len` matching the previous `mem_write()` call.

> Not sure what the coding standards are for comments, but I would suggest we add comments to `onSslBioOut` and `DtlsTransport::SendDtlsData` (above the implementation). Those should describe the effect of the function and when/why it gets called. Preferably `/** */` style. We're doing some subtle things here so we need to be extra clear.

Done here: https://github.com/versatica/mediasoup/pull/1345/commits/e827248427956f53c442ae1a930700f213d9edb4

ibc:
BTW I've added some logs and Wireshark capture in the PR description above proving that everything makes sense.

ibc:
Let's merge and release. Thanks a lot.

PR #1348, opened 2024-02-29
Update FlatBuffers to 24.3.6-1 (fix cannot set temporal layer 0)

ibc:
- Due to an issue in Flatbuffers (https://github.com/google/flatbuffers/issues/8245) when we call `consumer.setPreferredLayers()` with `temporalLayer: 0` (or `transport.consume()` with `temporalLayer: 0` in `preferredLayers`), we are NOT passing `temporalLayer` to the worker so it selects the highest available temporal layer instead.
- This PR tests that use case in both Node and Rust, and tests should fail (only in Node) until the FlatBuffers issue is resolved.
- And the issue in FlatBuffers was fixed long ago (https://github.com/google/flatbuffers/issues/8245) but the release happened now. And here the new version in Meson WrapDB: https://github.com/mesonbuild/wrapdb/pull/1441


jmillan (approved):

ibc:
FlatBuffers 24.3.6 (with the fix) has been released. The NPM version is already available. Ongoing work in Meson WrapDB to update FlatBuffers there: https://github.com/mesonbuild/wrapdb/pull/1441

jmillan (approved):

PR #1362, opened 2024-03-31
Fix: modify comments about packet loss rate calculation

liuzhenyilang:
There are two errors in the comment on the calculation of packet loss rate for rtcp packet.


ibc:
Thanks.

PR #1363, opened 2024-04-02
add arm64 prebuilt workers for linux

arcinston:
Updates GH action to also build pre-built workers for linux instances running on arm64/aarch64

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -15,14 +15,19 @@ jobs:
   ci:
     strategy:
       matrix:
-        os:
-          - ubuntu-20.04
-          - ubuntu-22.04
-          - macos-12
-          - macos-14
-          - windows-2022
+        ci:
+          - os: ubuntu-20.04
+          - os: ubuntu-22.04
+          - os: ubuntu-20.04
+            arch: arm64
+          - os: ubuntu-22.04
+            arch: arm64
+          - os: macos-12
+          - os: macos-14
+          - os: windows-2022
 
-    runs-on: ${{ matrix.os }}
+
+    runs-on: ${{ matrix.ci.os }}
 
     env:
       KEEP_BUILD_ARTIFACTS: '1'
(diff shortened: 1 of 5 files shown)


ibc:
Before accepting this PR we need to fix this recent issue that didn't happen a few weeks ago:

https://github.com/versatica/mediasoup/actions/runs/8520123345/job/23337119268?pr=1363

arcinston:
even i was looking at this issue right now @ibc

ibc:
> even i was looking at this issue right now @ibc

I've fixed it and merged it right now.

ibc:
Please update your branch with v3 branch. CI should pass.

However, if we are building worker prebuilt images for Linux arm64, then we must also test Linux arm64 in other CI tasks, including worker CI, Rust CI, Node CI, fuzzer CI, etc.

arcinston:
got it updating

and looking into testing in other ci tasks @ibc

ibc:
Prettier errors (https://github.com/versatica/mediasoup/actions/runs/8521809219/job/23340709865?pr=1363).

Please run `npm run format:node`

.github/workflows/mediasoup-rust.yaml
```diff
@@ -37,7 +41,7 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ matrix.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ matrix.ci.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
```
ibc, line 44: 👍

ibc:
Thanks a lot, merging.

ibc:
@arcinston, I don't see that arm64 prebuilts are being generated, see assets in any recent version:

https://github.com/versatica/mediasoup/releases

It's even worse! The arm64 prebuilt images are replacing the x64 ones because they are generating the same name!!

CleanShot-2024-04-09-at-19 06 05@2x

I've created an issue https://github.com/versatica/mediasoup/issues/1372

arcinston:
I realised this too the reason is this

https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/

GitHub will launch its arm64 runners soon 
but it's not out yet it's in private beta 😞 

I did not know of this

I am so sorry

arcinston:
for now it should be fine
as the arm64 runner is actually Linux x64

so it should not cause trouble 

but as soon as the arm64 runners are out in public 

it should work as expected

arcinston:
cc @ibc

ibc:
Thanks. Let's please handle the discussion in https://github.com/versatica/mediasoup/issues/1372.

PR #1365, opened 2024-04-07
TransportListenInfo: Add portRange (deprecate worker port range)

ibc:
### Details

- New optional `portRange` field (an object with `min` and `max` keys) has been added in `TransportListenInfo`, so it's available in `router.createWebRtcTransport()`,  `router.createPlainTransport()`, `router.createPipeTransport()` and `worker.createWebRtcServer()` methods.
- If `portRange` is given, then a free random port in that range (and given `ip` and `protocol`) will be chosen.
- This feature deprecates worker ports range (`rtcMinPort` and `rtcMaxPort` in `WorkerSettings` given to `createWorker()` method). Rationale is that it doesn't make any sense to have a unique port range for different listening IPs (for example: an app may want to use different port ranges for public connections (public internet) and private connections (private network).

### TODO

- Documentation in website and proper announcement.


node/src/Channel.ts
```diff
@@ -1,5 +1,6 @@
 import * as os from 'node:os';
 import { Duplex } from 'node:stream';
+import { info, warn } from 'node:console';
```
ibc, line 3: This is to avoid that Jest mangles the output of `MS_DUMP()`.

node/src/test/test-WebRtcTransport.ts
```diff
@@ -197,6 +197,60 @@ test('router.createWebRtcTransport() with deprecated listenIps succeeds', async
...
+		listenInfos: [
+			// NOTE: udpReusePort flag will be ignored since protocol is TCP.
+			{ protocol: 'tcp', ip: '127.0.0.1', port, flags: { udpReusePort: true } },
+		],
+	});
+
+	expect(webRtcTransport.iceCandidates[0].port).toEqual(port);
+
+	webRtcTransport.close();
+}, 2000);
```
ibc: Moved to keep a proper order.

worker/include/RTC/PortManager.hpp
```diff
@@ -14,50 +13,72 @@ namespace RTC
 	class PortManager
 	{
 	private:
-		enum class Transport : uint8_t
+		enum class Protocol : uint8_t
```
ibc, line 16: consistency

worker/include/RTC/TransportTuple.hpp
```diff
@@ -163,77 +163,7 @@ namespace RTC
...
-		 * Hash for IPv6
-		 *
-		 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-		 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-		 |              PORT             | IP[0] ^  IP[1] ^ IP[2] ^ IP[3]|
-		 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-		 |IP[0] ^  IP[1] ^ IP[2] ^ IP[3] |          IP[0] >> 16      |F|P|
-		 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-		 */
-		void SetHash()
```
ibc, line 185: Moved to .cpp

worker/src/RTC/UdpSocket.cpp
```diff
@@ -10,20 +10,30 @@ namespace RTC
...
-	  Listener* listener, std::string& ip, uint16_t port, RTC::Transport::SocketFlags& flags)
+	  Listener* listener,
+	  std::string& ip,
+	  uint16_t minPort,
+	  uint16_t maxPort,
+	  RTC::Transport::SocketFlags& flags,
+	  uint64_t& portRangeHash)
 	  : // This may throw.
-	    ::UdpSocketHandle::UdpSocketHandle(PortManager::BindUdp(ip, port, flags)), listener(listener),
-	    fixedPort(true)
```
ibc, line 24: We had a critical BUG here! We marked UdpSocket with `fixedPort: true`, meaning that the port was never removed from `PortManager`'s maps when this UdpSocket was closed!!!

worker/src/RTC/UdpSocket.cpp
```diff
@@ -10,20 +10,30 @@ namespace RTC
...
 	}
 
 	UdpSocket::UdpSocket(
-	  Listener* listener, std::string& ip, uint16_t port, RTC::Transport::SocketFlags& flags)
+	  Listener* listener,
+	  std::string& ip,
+	  uint16_t minPort,
+	  uint16_t maxPort,
+	  RTC::Transport::SocketFlags& flags,
+	  uint64_t& portRangeHash)
```
ibc, line 28: Note that `portRangeHash` is passed as a reference to `uint64_t`, so the parent constructor (in the initializer list) calls ` RTC::PortManager::BindUdp()` and also passes `portRangeHash` to it as a reference to `uint64_t`, so `PortManager` will change the value of `portRangeHash` and the new value will be also visible here.

jmillan (approved):
Nice 👍 . If I had a comment to add is that it would be IMHO a bit cleaner adding a new type `portRange` rather than two `minPort` and `maxPort`. 

```ts
type PortRange = {
  min: number;  
  max: number;
};
```

rust/src/data_structures.rs
```diff
@@ -52,7 +52,7 @@ impl AppData {
 /// # Notes on usage
 /// If you use "0.0.0.0" or "::" as ip value, then you need to also provide
 /// `announced_address`.
-#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Deserialize, Serialize)]
```
ibc, line 55: I had to remove `Ord` and `PartialOrd` because `RangeInclusive` doesn't implement them. No downsides AFAIS.

rust/src/data_structures.rs
```diff
@@ -90,6 +93,18 @@ impl ListenInfo {
                 .as_ref()
                 .map(|address| address.to_string()),
             port: self.port.unwrap_or(0),
+            port_range: Box::new(transport::PortRange {
```
ibc, line 96: I'm not super convinced about this.

rust/src/data_structures.rs
```diff
@@ -90,6 +93,18 @@ impl ListenInfo {
...
+                    *self.port_range.as_ref().unwrap().start()
+                } else {
+                    0
+                },
+                max: if self.port_range.is_some() {
+                    *self.port_range.as_ref().unwrap().end()
+                } else {
+                    0
+                },
+            }),
```
ibc: I'm not super convinced about this.

ibc:
@jmillan, `minPort` and `maxPort` now moved to `portRange`.

rust/src/data_structures.rs
ibc: Improved thanks to @jmillan

CosmosisT:
Why not allowedPorts and provide ability for min/max and static ports. That way everyone is settled. I would assume if isolated processes are ran the min/max range would still be affective skipping used ports for an available one when initiating connections.

The parameter would look a little like this allowedPorts: {min: 20000, max: 30000, static:[443, 3923, 10293]}

This way no matter the setup it's simplified and in one spot and for people who want specifics between TCP/UDP connections this would suffice.

ibc:
We allow passing a static port (so the app must manage ports by itself) or a port range. IMHO is more than enough. In addition the PortRange C++ class is optimized to avoid trying already used ports. 

Also, it doesn't make sense that the app passes some static ports plus a port range. If the app passes a static port then the app is supposed to manage ports. mediasoup is not gonna try first your "maybe available or not" static ports and then, of all fail, choose a random one in the given range. It doesn't make sense.

Issue #1366, opened 2024-04-08
SeqManager<T>::SeqLowerThan() crashes when used as compare function in abseil set or map (only in debug mode)


shaymagsumov:
This happens very rarely. I can't reproduce it nor attach core dump for now. But i have located it at:

https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/NackGenerator.cpp#L119

I can repoduce similar crash separatedly from worker in this sample:

```
absl::btree_set<uint16_t, SeqManager<uint16_t>::SeqLowerThan> recoveredList;

recoveredList.insert(10000);
recoveredList.insert(40000);
recoveredList.insert(60000);
```

jmillan:
@shaymagsumov, that the provider snippet generate the crash?

shaymagsumov:
I mean that snippet reproduces the same crash as in NackGenerator. Same assert is failing in absl's btree container. Assert code in file `absl/container/internal/btree.h` is:

```
assert(
      iter.node_->is_ordered_correctly(static_cast<field_type>(iter.position_),
                                       original_key_compare(key_comp())) &&
      "If this assert fails, then either (1) the comparator may violate "
      "transitivity, i.e. comp(a,b) && comp(b,c) -> comp(a,c) (see "
      "https://en.cppreference.com/w/cpp/named_req/Compare), or (2) a "
      "key may have been mutated after it was inserted into the tree.");
```

ibc:
> comp(a,b) && comp(b,c) -> comp(a,c)

I feel that this may indeed happen due to `SeqManager<uint16_t>::SeqLowerThan` comparator...

ibc:
```c++
template<typename T, uint8_t N>
bool SeqManager<T, N>::SeqLowerThan::operator()(T lhs, T rhs) const
{
	return ((rhs > lhs) && (rhs - lhs <= MaxValue / 2)) ||
	       ((lhs > rhs) && (lhs - rhs > MaxValue / 2));
}

template<typename T, uint8_t N>
bool SeqManager<T, N>::IsSeqLowerThan(T lhs, T rhs)
{
	return isSeqLowerThan(lhs, rhs);
}
```


> absl::btree_set<uint16_t, SeqManager<uint16_t>::SeqLowerThan> recoveredList;
> 
> recoveredList.insert(10000);
> recoveredList.insert(40000);
> recoveredList.insert(60000);

> transitivity, i.e. comp(a,b) && comp(b,c) -> comp(a,c)

[...]

PR #1368, opened 2024-04-08
Add a test to check issue #1366 and use Debug mode

ibc:
Temporal PR just to test https://github.com/versatica/mediasoup/issues/1366 in different machines.

CC @jmillan @shaymagsumov

### Results using `absl::btree_set`

```c++
SECTION("absl::btree_set")
{
	absl::btree_set<uint16_t, RTC::SeqManager<uint16_t>::SeqLowerThan> recoveredList;

	recoveredList.insert(10000);
	recoveredList.insert(40000);
	recoveredList.insert(60000);

	REQUIRE(recoveredList.size() == 3);
}
```

Here we can see that, when compiled in debug mode, `mediasoup-worker` CI fail in all OS and archs and compilers (in all but Windows) due to the transitivity violation, i.e. `comp(a,b) && comp(b,c) -> comp(a,c)`.

All CI fail: https://github.com/versatica/mediasoup/actions/runs/8612477995/

### Results using `std::set`

```c++
SECTION("std::set")
{
	std::set<uint16_t, RTC::SeqManager<uint16_t>::SeqLowerThan> recoveredList;

	recoveredList.insert(10000);
	recoveredList.insert(40000);
	recoveredList.insert(60000);

	REQUIRE(recoveredList.size() == 3);
}
```

Despite C++ containers are supposed to also require transitivity (https://en.cppreference.com/w/cpp/container/set, https://en.cppreference.com/w/cpp/named_req/Compare), the same transitivity violation doesn't affect standard C++ STD containers when compiled in debug mode. Why? No idea, but it doesn't fail in any OS/arch/compiler.

All CI passes: https://github.com/versatica/mediasoup/actions/runs/8612301409/

### Conclusion

- Ok, we use abseil containers to gain more performance.
- But they are hell.
- We **SHOULD** be able to run mediasoup-worker in debug mode without crashes. Otherwise, what do we want the debug mode for? To detect problems? How are we gonna detect problems if we are literally introducing problems when we run mediasoup-worker in debug mode due to the required transitivity in abseil containers? It doesn't make any sense.
- So IMHO we **MUST NOT** use abseil conatiners with compare functions that do not honor the transitivity requirement,  and this is the task that must be done.

### DIFF
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -101,8 +101,11 @@ jobs:
         # TODO: Maybe fix this one day.
         if: runner.os != 'Windows'
 
-      - name: invoke -r worker mediasoup-worker
-        run: invoke -r worker mediasoup-worker
+      # - name: invoke -r worker mediasoup-worker
+      #   run: invoke -r worker mediasoup-worker
+
+      - name: MEDIASOUP_BUILDTYPE=Debug invoke -r worker mediasoup-worker
+        run: MEDIASOUP_BUILDTYPE=Debug invoke -r worker mediasoup-worker
 
       # - name: invoke -r worker test
       #   run: invoke -r worker test
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -11,6 +11,7 @@ concurrency:
 jobs:
   ci:
     strategy:
+      fail-fast: false
       matrix:
         build:
           - os: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


ibc:
Ok, not crashing but I think it will after my last commit.

jmillan:
I agree, let's not use abseil containers when the comparison function is needed.

ibc:
So I'm closing this temp PR and will make one with the agreed changes.

ibc:
PR done: https://github.com/versatica/mediasoup/pull/1369

PR #1373, opened 2024-04-09
CI: Disable machines with arch arm64

### LINKED ISSUE #1372: arm64 prebuilt binaries are replacing x64 ones
### Details

- This PR added worker prebuilt binaries for arm64: https://github.com/versatica/mediasoup/pull/1363
- And those prebuilt binaries are being generated, however they generate a name with `x64` rather than `arm64` so they are **replacing** the x64 binaries!!!  
  CleanShot-2024-04-09-at-19 06 05@2x
- See the resulting binaries: https://github.com/versatica/mediasoup/releases

This is the Node code we use to generate the file name of each worker prebuilt binary (in `npm-scripts.mjs`):

```ts
function getWorkerPrebuildTarName() {
	let name = `mediasoup-worker-${PKG.version}-${os.platform()}-${os.arch()}`;

	// In Linux we want to know about kernel version since kernel >= 6 supports
	// io-uring.
	if (os.platform() === 'linux') {
		const kernelMajorVersion = Number(os.release().split('.')[0]);

		name += `-kernel${kernelMajorVersion}`;
	}

	return `${name}.tgz`;
}
```

According to Node docs, [os.arch()](https://nodejs.org/api/os.html#osarch) should return one of these values:

> 'arm', 'arm64', 'ia32', 'loong64', 'mips', 'mipsel', 'ppc', 'ppc64', 'riscv64', 's390', 's390x', and 'x64'

However, in GitHub CI, despite we add `arch: arm64` in the matrix, the machines running the CI action report `os.arch() => 'x64'`.

ibc:
- Fixes issue #1372
- As told here https://github.com/versatica/mediasoup/issues/1372#issuecomment-2045722727, it turns out GitHub arm64 runners are still in private beta: https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/
- Let's remove them (to avoid running same CI actions basically twice) and will add them in the future when ARM64 support in CI is available and we know the syntax.

### DIFF
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -17,12 +17,12 @@ jobs:
             node: 18
           - os: ubuntu-22.04
             node: 20
-          - os: ubuntu-20.04
-            node: 18
-            arch: arm64
-          - os: ubuntu-22.04
-            node: 20
-            arch: arm64
+          # - os: ubuntu-20.04
+          #   node: 18
+          #   arch: arm64
+          # - os: ubuntu-22.04
+          #   node: 20
+          #   arch: arm64
           - os: macos-12
             node: 18
           - os: macos-14
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -17,14 +17,6 @@ jobs:
             node: 18
           - os: ubuntu-22.04
             node: 20
-          # NOTE: Disabled until CI arm64 runners are publicly available:
-          #   https://github.com/versatica/mediasoup/issues/1372
-          # - os: ubuntu-20.04
-          #   node: 18
-          #   arch: arm64
-          # - os: ubuntu-22.04
-          #   node: 20
-          #   arch: arm64
           - os: macos-12
             node: 18
           - os: macos-14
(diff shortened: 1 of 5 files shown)


jmillan (approved):

nazar-pc:
I'm reading `.github/workflows/mediasoup-node.yaml` and `arch: arm64` there is literally unused, it is not referenced anywhere and does literally nothing. Looks like the same is true for every other job.

How did anyone expect it to work? https://github.com/versatica/mediasoup/pull/1363 did nothing, it just compiles x86-64 build twice. If I read it correctly, it doesn't break anything and doesn't help with anything either.

ibc:
arm64 runners are only private beta yet, maybe that's why.

nazar-pc:
Even if it wasn't, I see nothing that would indicate we're trying to use it. If we actually tried we'd fail because as you say it is private.

ibc:
I assume that the author has access to the private beta and knew about it, but honestly no idea.

ibc:
Given that there is zero evidence that the syntax will be `arch: arm64` I've finally removed those entries entirely in all matrixes. Will be added back once ARM64 support becomes public in CI.

nazar-pc:
It is 100% not the syntax, it is just a matrix. You could have added `foo: bar` and it would make just as much difference until actually used for something.

ibc:
Yes I know. When I said "the syntax" I meant whatever that will be added to run arm64 hosts. I assume it will be something like "os: ubuntu-22.04-arm" or who knows.

PR #1377, opened 2024-04-16
fix: panic with normal errors in Rust

satoren:
Returns an error when an error can be returned

### DIFF
--- a/rust/src/router/consumer.rs
+++ b/rust/src/router/consumer.rs
@@ -1020,9 +1020,9 @@ impl Consumer {
             .inner
             .channel
             .request(self.id(), ConsumerGetStatsRequest {})
-            .await;
+            .await?;
 
-        if let Ok(response::Body::ConsumerGetStatsResponse(data)) = response {
+        if let response::Body::ConsumerGetStatsResponse(data) = response {
             match data.stats.len() {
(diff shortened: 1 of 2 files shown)


nazar-pc (commented):
Did you actually hit this? If so then under which condition?
This is the kind of error that we indeed do not expect unless I missed something.

satoren:
It is defined as a `RequestError`. Seems to be returning it except here. The type is also ` Result<ConsumerStats, RequestError>`. 

https://github.com/versatica/mediasoup/blob/d534be0cf3a90a7199435deb601dba73401bd767/rust/src/worker.rs#L46

nazar-pc:
IIRC `ConsumerGetStatsRequest` never errors, it is just ignored if consumer doesn't exist. No strong opinion though.

ibc:
So should we merge or not?

nazar-pc (approved):
Sure, I don't think it will ever happen, but it doesn't hurt either

satoren:
I have encountered the [L1042](https://github.com/versatica/mediasoup/blob/e4d90d9398f357cefac67e0616c9fa5cd9e28099/rust/src/router/consumer.rs#L1041) panic, but I don't know up to the exact cause because there is no other information available.

Maybe it is an error that the "Channel already closed".

PR #1378, opened 2024-04-19
Add optional sctpStreamId to SCTP-based data consumers

threema-lenny:
Second try (after #1111). This makes it possible to use a specific SCTP stream ID / data channel ID for forwarding data.

Could you give me feedback on whether this looks good for the Node API? Then I'll see if I can add the change to the Rust API, too.

## TODO

- [ ] Rust side and test.
- [ ] Document new `sctpStreamId` in `DataConsumerOptions` in the website. This task is for mediasoup authors once this PR is merged.


node/src/Transport.ts
```diff
@@ -1139,10 +1139,9 @@ export class Transport<
 			}
 
 			// This may throw.
-			sctpStreamId = this.getNextSctpStreamId();
-
-			this.#sctpStreamIds![sctpStreamId] = 1;
-			sctpStreamParameters.streamId = sctpStreamId;
+			sctpStreamParameters.streamId = this.getNextSctpStreamId(sctpStreamId);
+			ortc.validateSctpStreamParameters(sctpStreamParameters);
```
ibc, line 1143: Why this call to `ortc.validateSctpStreamParameters()`?

ibc (commented):
What is the real purpose of this PR?

jmillan:
@threema-lenny, how are you taking benefit of choosing the steam ID for a DataConsumer? Can you share a use case?

node/src/Transport.ts
threema-lenny: It seemed like an oversight as this call is also present for validating the SCTP stream parameters for the producer (https://github.com/versatica/mediasoup/pull/1378/files#diff-3253c176609c972f1279e2a6c6ca4c52cbfac8423efcc8b2891b9d8dbd92c358R1016).

threema-lenny:
> @threema-lenny, how are you taking benefit of choosing the steam ID for a DataConsumer? Can you share a use case?

Sure.

We're using OOB negotiated data channel IDs (using `pc.createDataChannel('foo', {negotiated: true, id: 0})`). This simplifies client code by making it possible to have a single bidirectional `RTCDataChannel` at creation of the peer connection without having to listen to an `ondatachannel` event and having to map data channels there (although I'm not even sure if the mediasoup SFU implements the data channel establishment protocol, haven't checked) nor is it necessary to signal it in another way because the ID is already known beforehand.

In our case, data is relayed via the SFU using this fixed SCTP stream ID which is why we set the corresponding SCTP stream ID for both producer and consumer.

ibc:
I get the point of this feature but need some time to review it.

node/src/Transport.ts
ibc: Please remove it. There is no need for this. These `sctpStreamParameters` were already validated in `transport.produceData()` and they are literally a copy of them.

node/src/test/test-DataConsumer.ts
```diff
@@ -107,6 +107,20 @@ test('transport.consumeData() succeeds', async () => {
 	});
 }, 2000);
 
+test('transport.consumeData() with already used sctpStreamId rejects with Error', async () => {
+	await ctx.webRtcTransport2!.consumeData({
+		dataProducerId: ctx.dataProducer!.id,
+		sctpStreamId: 123,
+	});
```
ibc, line 114: Please, create `const dataConsumer` here and assert that its `sctpStreamParameters.streamId` matches the given  `sctpStreamId`.

ibc (commented):
Other than the requested changes I think this PR goes in the correct direction, so please, make these minimal changes and write the Rust side as well.

node/src/Transport.ts
threema-lenny: You're right, only `streamId` is now user input but its value (besides type) is not validated in `validateSctpStreamParameters` anyways. I'll remove it.

ibc:
@threema-lenny are you working on this PR?

threema-lenny:
@ibc Not right now but I'll definitely need to come back to getting this done.

ibc:
This will be done once the new SCTP stack is merged: https://github.com/versatica/mediasoup/pull/1806/

lgrahl:
Checked a few months or so ago and this was still necessary with the new SCTP stack.

Anyways, my employment at Threema got terminated. I might still follow up for a personal project on this one.

ibc:
I'll come back to this once I finish other things

PR #1380, opened 2024-04-22
RtpPacket: Make (almost) all methods remove payload padding if present

### LINKED ISSUE #1379: RtpPacket::SetPayloadLength get wrong payloadLength
Hi,
this method padding length to 4 bytes, however it change ```real payloadLength```  and remove padding flags , is that the correct ?
I thing we should check padding length is the same as orig length, if padding happen then set padding flag to true and set padding byte to padding length.

```
void RtpPacket::SetPayloadLength(size_t length)
	{
		MS_TRACE();

		// Pad desired length to 4 bytes.
		length = static_cast<size_t>(Utils::Byte::PadTo4Bytes(static_cast<uint16_t>(length)));

		this->size -= this->payloadLength;
		this->size -= size_t{ this->payloadPadding };
		this->payloadLength  = length;
		this->payloadPadding = 0u;
		this->size += length;

		SetPayloadPaddingFlag(false);
}
```

[https://github.com/versatica/mediasoup/blob/d6373d540751c06b7fb96f0e7e26ad56f50f25bd/worker/src/RTC/RtpPacket.cpp#L653](url)

ibc:
Fixes #1379

### Details

Before this PR:

- We have inconsistent behavior: some methods in `RtpPacket` keep existing payload padding after mangling the payload others don't.
- Those methods that keep existing payload do it wrong (real bug here) since they do NOT write the number of padding bytes in the last byte of the padding.

After this PR:

- All methods (but `clone()`) remove payload padding if present. Period.
- There is no real need to keep padding (AKA be padded to 4 bytes). That's just for super legazy plain RTP engines.
- "And why don't we keep existing padding instead?". See section above "Before this PR". Lot of work for little/null benefit. I don't want to waste much energy on this topic given that (when time comes in) I'd like to focus on task https://github.com/versatica/mediasoup/issues/1233 instead.

### DIFF
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -549,14 +549,6 @@ namespace RTC
 		}
 
 		MS_ASSERT(ptr == this->payload, "wrong ptr calculation");
-
-		// TODO: Yes or not?
-		// If there is padding we have to ensure that last byte of it contains
-		// the number of bytes of padding.
-		// if (this->payloadPadding != 0u)
-		// {
-		// 	this->payload[this->payloadLength + this->payloadPadding - 1] = this->payloadPadding;
-		// }
 	}
 
 	void RtpPacket::UpdateMid(const std::string& mid)
@@ -668,7 +660,7 @@ namespace RTC
 		MS_TRACE();
 
 		this->size -= this->payloadLength;
-		this->payloadLength  = length;
+		this->payloadLength = length;
 		this->size += this->payloadLength;
 
 		// Remove padding if present.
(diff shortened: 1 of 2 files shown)


jmillan (approved):

worker/src/RTC/RtpPacket.cpp
```diff
@@ -650,20 +650,23 @@ namespace RTC
...
 		this->size -= this->payloadLength;
 		this->size -= size_t{ this->payloadPadding };
 		this->payloadLength  = length;
-		this->payloadPadding = 0u;
-		this->size += length;
+		this->payloadPadding = payloadPadding;
+		this->size += this->payloadLength + this->payloadPadding;
 
-		SetPayloadPaddingFlag(false);
+		SetPayloadPaddingFlag(this->payloadPadding > 0);
```
ibc: Oh this is not correct because we have to write number of padding bytes in the last byte of the padding right? I mean, SerPayloadPaddingFlag() doesn't do it AFAIR (I am out now).

jmillan (approved):

PR #1383, opened 2024-04-28
Worker: Fix sending callback leaks

### LINKED ISSUE #1382: DataConsumer::SendMessage cause onQueuedCallback* cb  leak and DATACONSUMER_SEND have no response
Hello, in some case I found DATACONSUMER_SEND will not get reasponse ,because  `onQueuedCallback* cb `  never called
Herre is the code: some case just return and cb not called and delete.   DirectTransport.cpp SendMessage have the problem too.

```
void DataConsumer::SendMessage(
	  const uint8_t* msg,
	  size_t len,
	  uint32_t ppid,
	  std::vector<uint16_t>& subchannels,
	  std::optional<uint16_t> requiredSubchannel,
	  onQueuedCallback* cb)
	{
		MS_TRACE();

		if (!IsActive())
		{
			return; // here is the problem
		}

		// If a required subchannel is given, verify that this data consumer is
		// subscribed to it.
		if (
		  requiredSubchannel.has_value() &&
		  this->subchannels.find(requiredSubchannel.value()) == this->subchannels.end())
		{
			return; // here is the problem
		}

		// If subchannels are given, verify that this data consumer is subscribed
		// to at least one of them.
		if (!subchannels.empty())
		{
			bool subchannelMatched{ false };

			for (const auto subchannel : subchannels)
			{
				if (this->subchannels.find(subchannel) != this->subchannels.end())
				{
					subchannelMatched = true;

					break;
				}
			}

			if (!subchannelMatched)
			{
				return;
			}
		}

```

https://github.com/versatica/mediasoup/blob/d4204c6e80132789b8b1b4828bd8679688dd6653/worker/src/RTC/DataConsumer.cpp#L518

https://github.com/versatica/mediasoup/blob/d4204c6e80132789b8b1b4828bd8679688dd6653/worker/src/RTC/DirectTransport.cpp#L214

ibc:
Fixes #1382

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Update worker subprojects ([PR #1376](https://github.com/versatica/mediasoup/pull/1376)).
 - OPUS: Fix DTX detection ([PR #1357](https://github.com/versatica/mediasoup/pull/1357)).
-- Worker: Fix sending callback leaks ([PR #1383](https://github.com/versatica/mediasoup/pull/1383)).
+- Worker: Fix sending callback leaks ([PR #1383](https://github.com/versatica/mediasoup/pull/1383), credits to @Lynnworld).
 
 ### 3.14.1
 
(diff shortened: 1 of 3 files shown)


jmillan (approved):

nazar-pc:
Looks important enough to make a Rust release too

PR #1384, opened 2024-04-29
Fix: call parent method explicitly

liuzhenyilang:
Call parent method explicitly, make the code easier to understand.


worker/src/RTC/RtpStreamRecv.cpp
```diff
@@ -843,7 +843,8 @@ namespace RTC
 		  score);
 #endif
 
-		RtpStream::UpdateScore(score);
+		// Call the parent method for update score
```
ibc:
```suggestion
		// Call the parent method for update score.
```

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -637,7 +637,8 @@ namespace RTC
 		  score);
 #endif
 
-		RtpStream::UpdateScore(score);
+		// Call the parent method for update score
```
ibc:
```suggestion
		// Call the parent method for update score.
```

ibc (commented):
Please add dots at the end of comments as everywhere else.

liuzhenyilang:
> Please add dots at the end of comments as everywhere else.

ok

ibc:
NOTE: Dots not added yet.

liuzhenyilang:
> NOTE: Dots not added yet.

Sorry, forget to commit.

Issue #1352, opened 2024-03-05
Multi-thread bug when using usrsctp in N Worker threads in Rust


ibc:
After investigating it in depth we have found a bad design in `DepUsrSCTP` class that allows a thread A make calls directly in thread B.

### Overview of the usrsctp library integration in mediasoup

[usrsctp](https://github.com/sctplab/usrsctp) was integrated in mediasoup once it merged the "Single thread" feature (PR https://github.com/sctplab/usrsctp/pull/339). That PR added the `usrsctp_init_nothreads()` API to run usrsctp in mono thread mode, meaning that usrsctp no longer runs its own timer into an internal thread but instead exposes an API for the application to be aware of when it should "ping" usrsctp so it can send pending SCTP messages (basically SCTP retransmissions due to non received SCTP ACK from the remote endpoint).

The way we integrated it was by wrapping the usrsctp API within the `DepUsrSCTP` and `SctpAssociation` class in mediasoup.

It's **important** to notice that this was designed having mediasoup for Node usage, this is: each mediasoup worker runs in a separate **process** with a **single thread**. From now on, this overview section describes the behavior in mediasoup  for Node:

1. The application creates a `Worker` so a new process is created with a single libuv loop.
2. During the `Worker` creation, `DepUsrSCTP::ClassInit()` is created which calls `usrsctp_init_nothreads()` by passing a static `onSendSctpData()` callback as argument, which will be later invoked by usrsctp when it needs to send SCTP retransmissions to remote endpoints.
3. The the `Worker` C++ constructor calls to `DepUsrSctp::CreateChecker()` which creates a singleton instance of the `DepUsrSCTP::Checker` class which is stored as a `static` member in `DepUsrSCTP`. Such a `Checker` inherits from `TimerHandler` since it's intended to "ping" usrsctp every 10ms so usrsctp can send pending SCTP retransmissions.
4. Of course, since mediasoup for Node launches a separate process for each mediasoup worker, there is no multi threading at all.
5. When the first `XxxxTransport` with SCTP support is created in the worker, its `SctpAssociation` instance calls to `DepUsrSCTP::RegisterSctpAssociation(this)` to register itself in `DepUsrSCTP`. If it's the first SCTP association then the `Checker` is started, meaning that it runs a periodic timer every 10ms.
[...]

PR #1394, opened 2024-05-09
Node WorkerChannel native addon

jmillan:
Substitute spawning a process with mediasoup for running mediasoup in a thread created by this addon.

The communication is now done via message passing rather than using unixsocket, avoiding the kernel networking stack for each and every message passed from Node to C++ and viceversa.

Performance gains are enhanced by approximately 8 times. Tested by sending a request from Node to C++ and waiting for the response N times in a loop.

Caveats:

* Since we don't use a separate process, a crash in mediasoup would crash Node, meaning it must be handled by node process signal handling.
  * At the same time, the coredumps will be named something-**node**-something
  * In order to debug coredumps the worker-channel.node binary can be used along with the coredump file and gdb/lldb 

TODO:

- [ ] Document removed 'died' and 'subprocessclose' Worker.ts events.
- [ ] Document removed 'died()' and 'subprocessClosed()' Worker.ts methods.
- [ ] Missing all the prebuilt binary upload/download stuff.


.gitignore
```diff
@@ -10,6 +10,12 @@
 # flatc generated files.
 /node/src/fbs
 
+## Node WorkerChannel addon.
+/node/workerChannel/node_modules
+/node/workerChannel/build
+/node/workerChannel/lib
```
ibc: Shouldn't we include this `lib` folder in "files" in main `package.json`? Or should it be managed by the `package.json` of the addon folder? What I mean is that:

- We only include JS transpiled files in the mediasoup NPM package and not TS files (AKA `node/src` folder).
- Is the addon properly built if we install mediasoup as a git+ssh dependency in the package.json of whatever app using mediasoup?

ibc:
`.prettierignore` and `.eslintignore` should include `/node/workerChannel/lib`.

node/src/Worker.ts
```diff
@@ -203,7 +202,6 @@ export type WorkerDump = {
 
 export type WorkerEvents = {
 	died: [Error];
-	subprocessclose: [];
```
ibc: This is a documented event, why is it being removed? I mean, we still may need to know when the native addon and its worker thread are closed, am I wrong?

node/src/Worker.ts
```diff
@@ -247,11 +245,13 @@ const workerLogger = new Logger('Worker');
...
-	// mediasoup-worker child process.
-	#child: ChildProcess;
+	/**
+	 * WorkerChannel instance.
+	 */
+	#workerChannel: WorkerChannel;
 
 	// Worker process PID.
-	readonly #pid: number;
+	readonly #pid: number = process.pid;
```
ibc, line 43: This amazes me. Does it mean that we only use the CPU core in which the Node process runs for **all the workers**? Ok, in theory those are native threads so they may use different CPU cores, is it?

node/src/Worker.ts
jmillan: They are never closed by themselves. The only way to close them is via `worker.close()` method in worker.ts, which sends a `WORKER_CLOSE` request to the worker. That should release all the libuv resources. Once that happens the thread running `mediasoup_worker_run()` will close, and when the JS `worker` instance is garbage collected, the native instance will be destroyed.

jmillan: I still have to do more testing on the closure part BTW.

node/src/Worker.ts
```diff
@@ -521,13 +398,6 @@ export class Worker<
 		return this.#died;
 	}
 
-	/**
-	 * Whether the Worker subprocess is closed.
-	 */
-	get subprocessClosed(): boolean {
-		return this.#subprocessClosed;
-	}
```
ibc: This requires changes in the public docs.

node/src/Worker.ts
jmillan: Correct, threads are not pinned to specific CPU.

node/workerChannel/src/workerChannel.hpp
ibc, line 1: Can we please organize public and private methods and member fields the same way we do in any other C++ class?

node/workerChannel/src/workerChannel.hpp
```diff
@@ -0,0 +1,34 @@
+#include "DepLibUV.hpp"
+// mediasoup header files.
+#include "common.hpp"
+#include "lib.hpp"
+
```
ibc: Remove blank line please, we don't do this anywhere.

node/workerChannel/src/workerChannel.hpp
jmillan: no problem

npm-scripts.mjs
```diff
@@ -128,6 +129,12 @@ async function run() {
 			break;
 		}
 
+		case 'worker:build-lib': {
+			buildWorkerLib();
+
+			break;
+		}
```
ibc: Is there any real reason to add this as a public script in `package.json`? I removed many of these in the past because it didn't make any sense to make them public. Does it make sense that we call `npm run worker:build-lib` directly? And why is this about "worker build lib"?

npm-scripts.mjs
```diff
@@ -315,6 +322,20 @@ function buildWorker() {
...
+	logInfo('buildWorkerLib()');
+
+	installInvoke();
+
+	executeCmd(`"${PYTHON}" -m invoke -r worker libmediasoup-worker`);
+
+	const buildType = process.env.MEDIASOUP_BUILDTYPE || 'Release';
+
+	executeCmd(
+		`cd node/workerChannel && GYP_DEFINES="mediasoup_build_type=${buildType}" npm i`
```
ibc: Can we use `npm ci` instead please? And copy `.npmrc` from root folder to `node/workerChannel`.

package.json
```diff
@@ -54,6 +54,7 @@
 		"typescript:build": "node npm-scripts.mjs typescript:build",
 		"typescript:watch": "node npm-scripts.mjs typescript:watch",
 		"worker:build": "node npm-scripts.mjs worker:build",
+		"worker:build-lib": "node npm-scripts.mjs worker:build-lib",
```
ibc: As expected, this is not needed at all, please remove.

jmillan: I've been using it while development, left over. Remember this is still a DRAFT.

npm-scripts.mjs
jmillan: Also, the prebuilt stuff is missing totally. We will need to upload the library now, not the binary.

npm-scripts.mjs
jmillan: leftover.

ibc:
Please merge v3 and check prettier and eslint path arrays in npm-scripts.mjs

.gitignore
ibc: Reminder

node/workerChannel/npm-scripts.mjs
ibc, line 1: lint and prettier missing, or better include this file in main npm-scripts.

node/workerChannel/src/index.ts
```diff
@@ -0,0 +1,36 @@
+import { EventEmitter } from 'events';
+
+const buildType = process.env.MEDIASOUP_BUILDTYPE ?? 'Release';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { WorkerChannel: NativeWorkerChannel } = require(
+	`../build/${buildType}/worker-channel.node`
+);
+
+export class WorkerChannel extends EventEmitter {
```
ibc: This should use our EnhancedEventEmitter with proper types.

node/workerChannel/src/workerChannel.cpp
```diff
@@ -0,0 +1,197 @@
+#include "../include/workerChannel.hpp"
+#include "napi.h"
+
```
ibc: Remove blank line.

node/workerChannel/src/workerChannel.cpp
```diff
@@ -0,0 +1,197 @@
...
+	auto params  = info[2].As<Napi::Array>();
+
+	this->emit = Napi::ThreadSafeFunction::New(env, cb, "WorkerChannel", 0, 1);
+
+	std::vector<std::string> args = { "" };
+
+	for (uint32_t i = 0; i < params.Length(); i++)
+	{
+		Napi::Value v = params[i];
+		if (!v.IsString())
```
ibc: Blank line.

node/workerChannel/src/workerChannel.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+		args.push_back(value.Utf8Value());
+	}
+
+	this->thread = std::thread(libmediasoup, this, version.Utf8Value(), args);
+}
+
+WorkerChannel::~WorkerChannel()
+{
+	std::cout << "~WorkerChannel()" << std::endl;
```
ibc: ????

jmillan: I need to do some tests with garbage collector, this log it temporal.

node/src/Worker.ts
jmillan: Added to the TODO list in the description.

node/src/Worker.ts
jmillan: Item added to the TODO list.

node/workerChannel/npm-scripts.mjs
jmillan: Do you mean not having an `npm-scripts` in node/workerChannel and do all related actions: build/lint/prettier, etc from the main `npm-scripts` file?

ibc: Probably it makes sense. It's part of the Node layer, it's not an optional one such as worker/scripts/ which is only needed for worker development.

jmillan: Lint and prettier are already done in main npm-scripts as I've added the given files path there, and in .eslintrc. But for building I think It's also nice to be able to build the addon in isolation, hence having a npm-scripts file there makes sense.

For example, you are changing the `WorkerChannel.cpp` and run `npm ci` in `node/workerChannel` which does not rebuild libmediasoup-worker but just rebuilds the addon.

What do you think?

ibc: ok

npm-scripts.mjs
```diff
@@ -441,7 +459,7 @@ function flatcWorker() {
 function testNode() {
 	logInfo('testNode()');
 
-	executeCmd(`jest --silent false --detectOpenHandles ${args}`);
+	executeCmd(`jest --silent false --detectOpenHandles --forceExit ${args}`);
```
ibc: Please don't

jmillan: Garbage collection on addons does not happen at our will. The native addon will be closed once the JS instance is garbage collected.

Within the tests, the addon is not garbage collected so the tests won't finish unless this option is set.

jmillan: Anyway I'm rechecking this because I need to see how it is garbage collected.

jmillan: Handled. It turns out that I need to set the JS instance to `undefined` to tell the GC that it is not longer used so it can clean it up, which will perform the destruction of the C++ class. I mean, if I don't do it then I assume GC would sometime in the future do it anyway but that would mean that the tests would not finish due to the open handles.

ibc: While it makes sense honestly I don't think that Node triggers garbage collection IMMEDIATELY as soon as an object is unreferenced. It may take some time depending on Node internal things and, if so, we will get random warnings here.

jmillan: Correct, it does not trigger GC at all. Setting a variable to  `undefined` just let the GC identify that the object is not used.

In this exact case, now the tests do finish while before they didn't unless the `--forceExit` option was set. This wouldn't be needed for real life, but we don't want the test to hang (CI would fail) nor use `--forceExit` since it could shadow real problems.

.gitignore
jmillan: Done and tested. package.json `files` are properly edited. adding mediasoup as a git+ssh dependency of a 3rd party app has also been tested.

jmillan:
Update:

This PR is almost ready to be shipped if we want to.

The are two main concerns already defined in the PR description:

* Since we don't use a separate process for the application and the worker, a crash in mediasoup would crash Node, meaning it must be handled by node process signal handling. This is true for Rust too.
  *  The coredumps will be named something-node-something
  *  In order to debug coredumps the worker-channel.node binary can be used along with the coredump file and gdb/lldb
* Since the worker is now used in multi-thread fashion, the very same concerns apply to Node as to Rust, such us https://github.com/versatica/mediasoup/issues/1352

Having said this, do we still want to go in this direction?. As indicated in the description:

> Performance gains are enhanced by approximately 8 times. Tested by sending a request from Node to C++ and waiting for the response N times in a loop.

But of course it comes with the mentioned caveats too.

What do you think @ibc, @nazar-pc?

nazar-pc:
If there use cases where the performance gap is that large, I think it is worth it. The proper solution to crashes is largely to write code that doesn't crash rather than running it in a separate process.

ibc:
I think that keeping the current design just to keep the "benefit" of not crashing the node process if the worker fails is not a good argument so I am ok but:

How does this change affect prebuilt worker binaries? Now they should not be "binaries" but "libraries" so more changes are needed specially in the naming of those assets and the function in npm-script that fetches them.

jmillan:
> How does this change affect prebuilt worker binaries? Now they should not be "binaries" but "libraries" so more changes are needed specially in the naming of those assets and the function in npm-script that fetches them.

Yes, there is a TODO for this.

:+1:,

jmillan:
I've updated the branch with 'v3'. Current status is that there are some linking issues when using `MESON_ARGS=-Db_sanitize=thread`. I guess the same options need to be provided to the addon.

jmillan:
@nazar-pc, this PR is up to date with v3, but I've commented out Windows usage on node CI actions because it [fails](https://github.com/versatica/mediasoup/actions/runs/18317003142/job/52159932229) when building the add-on. 

The .gyp file for the add-on is [this](https://github.com/versatica/mediasoup/blob/549d9b2e4508426f3f378ddceee1038729f345c4/node/src/workerChannel/binding.gyp). I'm quite sure it used to work before (months ago).

Since you have some other times fixed some other Windows issues, I wonder if you could take a look whenever you find some free slots.

nazar-pc:
Looks like gyp can't find visual studio. Maybe it doesn't support the latest MSVC version?

Specifically this line is concerning:
> subprocess.CalledProcessError: Command '['cl', '-dM', '-E', '-x', 'c', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpnwi1n4h7.c']' returned non-zero exit status 1.

Grok suggested to try this (yes, `cmd` instead of PowerShell) to set it up explicitly:
```
- name: Install dependencies
  shell: cmd
  run: |
    call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
    npm install
```

You might have to set up a Windows VM to debug what is going on there rather than pushing countless test commits here.

jmillan:
> You might have to set up a Windows VM to debug what is going on there rather than pushing countless test commits here

This is my dream 😊

ibc:
> Grok suggested to try this

OMG the Nazi AI of the Nazi Elon Musk 🥸

PR #1395, opened 2024-05-09
Get rid of .eslintignore and .prettierignore

ibc:
### Details

- Instead of ignoring complex paths, be explicit in what ESLint and Prettier must check.


npm-scripts.mjs
```diff
@@ -331,13 +353,15 @@ function cleanWorkerArtifacts() {
...
-	executeCmd('prettier . --check');
-
 	// Ensure there are no rules that are unnecessary or conflict with Prettier
 	// rules.
 	executeCmd('eslint-config-prettier .eslintrc.js');
 
-	executeCmd('eslint -c .eslintrc.js --max-warnings 0 .');
+	executeCmd(
+		`eslint -c .eslintrc.js --ext=ts,js,mjs --max-warnings 0 ${ESLINT_IGNORE_PATTERN_ARGS} ${ESLINT_PATHS}`
+	);
```
ibc, line 362: In order to tell ESLint to also check .mjs files it must be said via command line arg. Otherwise it will complain because `No files matching the pattern "worker/scripts" were found` because there are no .js or .ts files in worker/scripts (and if they were, then it would ignore worker/scripts/.mjs files).
See https://github.com/eslint/eslint/issues/9592

jmillan (approved):

PR #1398, opened 2024-05-13
feat: udp connect

o-u-p:
will be useful when binding connection to a specific thread

### DIFF
--- a/worker/include/RTC/TransportTuple.hpp
+++ b/worker/include/RTC/TransportTuple.hpp
@@ -79,7 +79,14 @@ namespace RTC
 		{
 			if (this->protocol == Protocol::UDP)
 			{
-				this->udpSocket->Send(data, len, this->udpRemoteAddr, cb);
+				if (this->udpSocket->IsConnected())
+				{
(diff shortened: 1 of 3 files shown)


ibc:
We already investigated the usage of connected UDP and it doesn't make sense for mediasoup. mediasoup must be ready to receive UD packets from different IP:port from same client (due to ICE nature) and hence we cannot assume that first one will be the chosen one. In addition we must be ready for clients to restart ICE from from a different network.

o-u-p:
No, I agree that " mediasoup must be ready to receive UD packets from different IP:port from same client". this feature will be used in muti-thread model. and when client use another Ip:port we can connect to the new tuple, and use consent logic to remove old one.

ibc:
This is not about client suddenly connecting to the new tuple. Client may sent UDP ICE packets from many different ip:ports at the same time at the very beginning and we must reply to all them.

Yes, we could eventually "connect" the current UDP tuple once ICE is established, and switch to another connected tuple later if client moves to another network, but that's huge work and honestly it doesn't make sense to just merge this partial work in v3 branch.

PR #1401, opened 2024-05-17
Fix rtcp_feedback for audio not written into flat buffers.

### LINKED ISSUE #1400: [Rust] Transport with only audio producer results in server-side transport cc is not created.
## Bug Report

`WebRtcTransport` with only audio producer with enabled transport congestion control results in transport CC is not actually enabled on server side, i.e. `Transport::tccServer` is `nullptr`

### Your environment

- Operating system: -
- Node version: -
- npm version: -
- gcc/clang version: -
- mediasoup version: v3.13.24
- mediasoup-client version: -

#### Steps to reproduce:
1. Run `rust/examples/echo`
2. Edit `rust/examples-frontend/echo/src/index.ts` to disable video and hence video producer
3. Run `rust/examples-frontend/echo`
4. Open echo app in browser `http://localhost:3001/`

#### Actual result:
Bitrate of sent audio is started at high values and drops to minimum allowed by webrtc (16Kb/s).
image

#### Expected result:
Bitrate of sent audio is depend on actual network bandwidth available.

mstyura:
Fixes #1400

### DIFF
--- a/rust/src/rtp_parameters.rs
+++ b/rust/src/rtp_parameters.rs
@@ -943,21 +943,19 @@ impl RtpParameters {
                             })
                             .collect(),
                     ),
-                    rtcp_feedback: match &codec {
-                        RtpCodecParameters::Audio { .. } => None,
-                        RtpCodecParameters::Video { rtcp_feedback, .. } => Some(
-                            rtcp_feedback
-                                .iter()
-                                .map(|rtcp_feedback| {
-                                    let (r#type, parameter) = rtcp_feedback.as_type_parameter();
-                                    rtp_parameters::RtcpFeedback {
-                                        type_: r#type.to_string(),
-                                        parameter: Some(parameter.to_string()),
-                                    }
-                                })
-                                .collect(),
-                        ),
-                    },
+                    rtcp_feedback: Some(
+                        codec
+                            .rtcp_feedback()
+                            .iter()
+                            .map(|rtcp_feedback| {
+                                let (r#type, parameter) = rtcp_feedback.as_type_parameter();
+                                rtp_parameters::RtcpFeedback {
+                                    type_: r#type.to_string(),
+                                    parameter: Some(parameter.to_string()),
+                                }
+                            })
+                            .collect(),
+                    ),
                 })
                 .collect(),
             header_extensions: self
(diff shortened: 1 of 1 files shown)


ibc:
Is this just a bug in Rust?

ibc:
Also please add "Fixes #XXXX" to link the issue.

mstyura:
> Is this just a bug in Rust?

I assume yes. And it was really hard to find out, because the transport CC was lost just before writing into flat buffers. In ortc objects it was reported that transport CC is active and enabled, while it was not. Thanks to example apps they dramatically helped to debug the problem. This is only reproduced with Rust, I've also checked JS demo app with disabled video - it does not have the problem.

ibc:
We only enable TCC server is a producer announces transport-cc support in its codec.rtcpFeedback (and if the clients support the corresponding wide-se-number RTP extension). libwebrtc includes it in OPUS codec, that's why it works.

ibc:
> Also please add "Fixes #XXXX" to link the issue.

Please add this in the PR description rather than in the title. That's what GH expects to link PR and issue.

nazar-pc (approved):
The change makes sense to me, but curious why @jmillan decided to explicitly skip audio when he introduced this code. This doesn't look like an accident.

mstyura:
> We only enable TCC server is a producer announces transport-cc support in its codec.rtcpFeedback (and if the clients support the corresponding wide-se-number RTP extension). libwebrtc includes it in OPUS codec, that's why it works.

Yes, exactly what I've managed to find out myself when debugging. But this information that opus producer supports transport CC were lost when rtp parameters were serialized into flat buffers.

ibc:
> The change makes sense to me, but curious why @jmillan decided to explicitly skip audio when he introduced this code. This doesn't look like an accident.

Probably because he assumed that only video codecs can include `transport-cc` feedback.

ibc (approved):

jmillan (approved):

jmillan:
> Probably because he assumed that only video codecs can include transport-cc feedback.

At that time I was doing quite a mechanic work porting types to FBS, so it could have been like that since the beginning, or I may have introduced it at that moment, I don't know.

PR #1402, opened 2024-05-17
fix: Windows compatibility

brhenrique:


nazar-pc (changes requested):
This seems like a random set of unclear changes. Could you describe what were the issues on Windows and how the changes you did address them?

jmillan:
@nazar-pc, this is Bruno helping me with Windows. We've done this PR p2p. I'll comment on this on Monday.

There was a problem because `workerChannel` had it's own tsconfig.json file and a source file was accessing the root typescript project. So now `workerChannel` is inside `node/src/` which also makes sense.

I'll re-review it on Monday.

ibc:
@nazar-pc, notice that this PR doesn't target v3 but @jmillan's ongoing branch.

Hi @brhenrique! Thanks for this. Some questions:

- Why are you moving node/workerChannel to node/src/workerChannel? This is literally producing a build/ folder with JS transpired files in node/src folder, which IMHO it's controversial.
- Can we remove all those things that are no longer needed instead of commenting them?

ibc:
> * Why are you moving node/workerChannel to node/src/workerChannel? This is literally producing a `lib/` folder with JS transpired files in node/src folder, which IMHO it's controversial.

I think I'm wrong. It's generating it in `node/lib/workerChannel` in theory. But, why do we need to include ""workerChannel/src" in "include" in `node/tsconfig.json`? https://github.com/versatica/mediasoup/pull/1402/files#diff-ebc4763df409943ccc3b18574546aba49d388b676eece10cd115b00239974761R13

That's wrong IMHO, such a path doesn't even exist.

jmillan:
> I think I'm wrong. It's generating it in node/lib/workerChannel in theory. But, why do we need to include ""workerChannel/src" in "include" in node/tsconfig.json? https://github.com/versatica/mediasoup/pull/1402/files#diff-ebc4763df409943ccc3b18574546aba49d388b676eece10cd115b00239974761R13

This PR moves `workerChannel` inside `node/src` rather than having it at the same level of `node`. `workerChannel` folder has the sources in `src/` folder so `workerChannel/src` is where the .ts files are. That is correct.

Anyway, let's try to keep `workerChannel` at `node` level as before. The problem was introduced when I used `../../enhancedEventEmitter` inside `workerChannel` as per your suggestion of using `EnhancedEventEmitter` @ibc. Windows is seeing that we are referencing a file from a different TS project and it's screwing the out dir for the .js files as seeing [here](https://github.com/versatica/mediasoup/actions/runs/9127507724/job/25097932964?pr=1394).

We can simply copy `enhancedEventEmitter.ts` file to `workerChannel/src/` so both TS projects remain independent.

I'm closing this PR for the given reasons.

PR #1403, opened 2024-05-19
Ignore RTP packets with empty payload

ibc:
### Details

- This PR replaces PR #793 (by @ggarber) with code updated to current v3 and addresses feedback.
- All `XxxxConsumer` classes drop empty RTP packets.
- `RtpStreamRecv` no longer considers empty RTP packets for counter increase and stream activation.
- A substancial change with original PR is that `RtpStreamRecv` accepts those empty packets (otherwise we would generate invalid NACKs), it's just that those don't account anymore for counter increase and stream activation.


ibc:
@ggarber can you review please?

jmillan (approved):

PR #1406, opened 2024-06-06
Replace support template message with a link

nazar-pc:
Because people don't read. I think this should be correct, though I don't have a direct preview.

Also someone didn't commit lock file changes last time, so I decided to include it too.


ibc (approved):

nazar-pc:
Both Node and Rust tests failed on Windows :thinking:

ibc:
> Both Node and Rust tests failed on Windows 🤔

Crazy. They fail consistently even in Node, which doesn't make sense. Basically all tests that depend on the worker fail.

ibc:
> > Both Node and Rust tests failed on Windows 🤔
> 
> Crazy. They fail consistently even in Node, which doesn't make sense. Basically all tests that depend on the worker fail.

Something happens in v3 branch (I've forced CI in branch v3) so this is not something related to this PR. So merging since the problem was already there anyway.

nazar-pc:
Okay, `https://mediasoup.org/documentation` didn't render as a link, but otherwise it worked!

ibc:
> Okay, `https://mediasoup.org/documentation` didn't render as a link, but otherwise it worked!

It's ok because users can just click on the "Open" button.

BTW issues created for those failing tests in Windows CI: https://github.com/versatica/mediasoup/issues/1407

PR #1411, opened 2024-06-19
fix asan error for new-delete-type-mismatch

satoren:
This handle seems to be of type `uv_pipe_t`.

### DIFF
--- a/worker/src/handles/UnixStreamSocketHandle.cpp
+++ b/worker/src/handles/UnixStreamSocketHandle.cpp
@@ -57,11 +57,6 @@ inline static void onClosePipe(uv_handle_t* handle)
 	delete reinterpret_cast<uv_pipe_t*>(handle);
 }
 
-inline static void onCloseShutdown(uv_handle_t* handle)
-{
-	delete reinterpret_cast<uv_shutdown_t*>(handle);
-}
-
 inline static void onShutdown(uv_shutdown_t* req, int /*status*/)
 {
 	auto* handle = req->handle;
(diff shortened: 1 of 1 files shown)


jmillan (changes requested):
indeed `handle` is of type `uv_stream_t*`.

ibc:
> indeed `handle` is of type `uv_stream_t*`.

Nope, it's `uv_pipe_t*`:

```c++
err = uv_shutdown(
      req, reinterpret_cast<uv_stream_t*>(this->uvHandle), static_cast<uv_shutdown_cb>(onShutdown));
```

In .hpp:

```hpp
uv_pipe_t* uvHandle{ nullptr };
```

So PR is ok IMHO.

jmillan:
> Nope, it's uv_pipe_t*:

I verified that `shutdown_t->handle` is a `uv_stream_t*`. Probably that's casted from `uv_pipe_t*`, but I'm not sure.

If ASAN is not complaining then :+1:

jmillan (approved):

ibc:
> > Nope, it's uv_pipe_t*:
> 
> 
> 
> I verified that `shutdown_t->handle` is a `uv_stream_t*`. Probably that's casted from `uv_pipe_t*`, but I'm not sure.
> 
> 
> 
> If ASAN is not complaining then :+1:

uv_pipe is a subtype of uv_stream. uv_tcp as well.

ibc:
Let's not merge yet, I want to check something.

ibc:
It's ok. Merging. Thanks @satoren. But the same should be done in `TcpConnectionHandle.cpp`, I will do it in a separate PR.

ibc:
> It's ok. Merging. Thanks @satoren. But the same should be done in `TcpConnectionHandle.cpp`, I will do it in a separate PR.

Done here: https://github.com/versatica/mediasoup/commit/08bdf564ed4210d14437d6102b879020265191c2

PR #1415, opened 2024-06-28
add tests with ASAN to node CI

satoren:


.github/workflows/mediasoup-node.yaml
```diff
@@ -11,12 +11,22 @@ concurrency:
...
           - os: ubuntu-20.04
             node: 18
           - os: ubuntu-22.04
             node: 20
+          - os: ubuntu-24.04
+            node: 20
+            meson_args: '-Db_sanitize=address'
+          - os: ubuntu-24.04
+            node: 20
+            meson_args: '-Db_sanitize=undefined'
```
ibc, line 26: Is `-Db_sanitize=undefined` the same as not passing `-Db_sanitize` argument?

.github/workflows/mediasoup-node.yaml
```diff
@@ -11,12 +11,22 @@ concurrency:
...
+          - os: ubuntu-24.04
+            node: 20
+            meson_args: '-Db_sanitize=thread'
```
ibc, line 29: What is the rationale for having 3 separate matrix entries with different `-Db_sanitize` options? Why not all together into a single one?

ibc (commented):
Looks good but I have questions.

.github/workflows/mediasoup-node.yaml
satoren: No, this is an undefined [behavior sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)

.github/workflows/mediasoup-node.yaml
satoren: Only one sanitizer can be enabled at a time.
For example, thread sanitizer and address sanitizer cannot be enabled at the same time.

.github/workflows/mediasoup-node.yaml
```diff
@@ -11,12 +11,22 @@ concurrency:
 jobs:
   ci:
     strategy:
+      fail-fast: false
```
ibc, line 14: While this is good while writing a PR we prefer not to have this once merged so I will remove this once we merge this PR.

ibc (approved):

.github/workflows/mediasoup-node.yaml
ibc: Then why this?

```
clang++ -fsanitize=signed-integer-overflow,null,alignment -fno-sanitize-recover=null -fsanitize-trap=alignment a.cc
```

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

ibc:
@satoren wanna help in this similar PR for worker tests? https://github.com/versatica/mediasoup/pull/1416

.github/workflows/mediasoup-node.yaml
ibc: I'm testing `-fsanitize='undefined,address,thread` and it definitely works. So honestly I don't think that we need to run just one at the same time.

ibc: ok, but the I read this:

https://stackoverflow.com/a/68907718

Issue #1417, opened 2024-07-01
AddressSanitizer issues


ibc:
This PR https://github.com/versatica/mediasoup/pull/1416 added AddressSanitizer tasks to worker CI. However some of them have been disabled since they fail. This ticket is to resolve them and re-enable those tasks in `mediasoup-worker.yaml` CI file.

Note that we just run AddressSanitizer in Linux (it works in Docker). So we can run them locally (in macOS) this way:

```bash
cd worker
make docker
make docker-run
```

Once in Docker:

```
cd worker
make clean-all
make test-asan-address
make clean-all
make test-asan-undefined
make clean-all
make test-asan-thread
```

Below in comments, the ASAN issues that show up until all them are fixed.

ibc:
### TODO 1

**UPDATE:** Fixed in https://github.com/versatica/mediasoup/pull/1419

`make test-asan-address`:

```
cd "/mediasoup/worker" && ASAN_OPTIONS=detect_leaks=1 "/mediasoup/worker/out/Release/build/mediasoup-worker-test-asan-address" --invisibles
Randomness seeded to: 3012346320
=================================================================
==24626==ERROR: AddressSanitizer: stack-buffer-overflow on address 0xfffff93fa0ec at pc 0xaaaabf80d778 bp 0xfffff93f9980 sp 0xfffff93f9170
READ of size 1488 at 0xfffff93fa0ec thread T0
    #0 0xaaaabf80d774 in __asan_memcpy (/mediasoup/worker/out/Release/build/mediasoup-worker-test-asan-address+0x3ad774) (BuildId: 29eb3dcfbb7682388ab3ac9b90247c7e46dec4d7)
    #1 0xaaaabf98b3d0 in RTC::RtpPacket::Clone() const (/mediasoup/worker/out/Release/build/mediasoup-worker-test-asan-address+0x52b3d0) (BuildId: 29eb3dcfbb7682388ab3ac9b90247c7e46dec4d7)
    #2 0xaaaabf98f720 in RTC::RtpRetransmissionBuffer::FillItem(RTC::RtpRetransmissionBuffer::Item*, RTC::RtpPacket*, std::shared_ptr<RTC::RtpPacket>&) (/mediasoup/worker/out/Release/build/mediasoup-worker-test-asan-address+0x52f720) (BuildId: 29eb3dcfbb7682388ab3ac9b90247c7e46dec4d7)
[...]

PR #1419, opened 2024-07-01
TestRtpStreamSend: ASAN, fix buffer overflow

jmillan:
Do not use a hardcoded length value when parsing a RTP packet

Fixes **TODO 1** in https://github.com/versatica/mediasoup/issues/1417#issuecomment-2199943723


ibc (approved):
Nice.

PR #1420, opened 2024-07-01
TestRtpStreamSend: ASAN, fix memory leaks

jmillan:
Use std::unique_pointer rather than raw pointers

By running `MEDIASOUP_BUILDTYPE=Debug make test-asan-address` most or all issues are related to not properly freeing instances in tests, so let's use `std::unique_ptr` everywhere rather than raw pointers.


ibc (approved):

ibc:
I'm merging this.

PR #1421, opened 2024-07-01
Worker tests: use unique pointers when possible

ibc:
(WIP)

### Details

- Avoid having allocated pointers outside `unique_ptr` pointers.
- So caution with all those mediasoup `Parse()` static methods that return an allocated pointer.
- `TestRtpStreamSend.cpp`: Use `unique_ptr` instead of `shared_ptr`.

### DIFF
--- a/worker/test/src/RTC/RTCP/TestXr.cpp
+++ b/worker/test/src/RTC/RTCP/TestXr.cpp
@@ -4,7 +4,6 @@
 #include "RTC/RTCP/XrReceiverReferenceTime.hpp"
 #include <catch2/catch_test_macros.hpp>
 #include <cstring> // std::memcmp(), std::memcpy()
-#include <memory>
 
 using namespace RTC::RTCP;
 
--- a/worker/test/src/RTC/RTCP/TestXr.cpp
+++ b/worker/test/src/RTC/RTCP/TestXr.cpp
@@ -28,7 +28,7 @@ SCENARIO("RTCP XR parsing", "[parser][rtcp][xr]")
 
 	SECTION("parse XR packet")
 	{
-		auto* packet = ExtendedReportPacket::Parse(buffer, sizeof(buffer));
+		std::unique_ptr<ExtendedReportPacket> packet(ExtendedReportPacket::Parse(buffer, sizeof(buffer)));
 
 		REQUIRE(packet);
 		// Despite total buffer size is 40 bytes, our GetSize() method doesn't not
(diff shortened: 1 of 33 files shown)


ibc:
### Questions

- Why does `TestRtpStreamSend.cpp` use `shared_ptr` instead of `unique_ptr`?
- Why does NOT `TestRtpStreamSend.cpp` use `xxx_ptr` in cases like this one?
  ```c++
  auto* packet = RtpPacket::Parse(rtpBuffer1, 1500);
  ```
- How is possible that ASAN doesn't report any problem in `TestRtpStreamSend.cpp` despite such a `packet` is not being deleted?

worker/test/src/RTC/RTCP/TestXr.cpp
```diff
@@ -133,7 +130,7 @@ SCENARIO("RTCP XrDelaySinceLastRt parsing", "[parser][rtcp][xr-dlrr]")
 	SECTION("create RRT")
 	{
 		// Create local report and check content.
-		auto* report1 = new ReceiverReferenceTime();
+		std::unique_ptr<ReceiverReferenceTime> report1(new ReceiverReferenceTime());
```
satoren: You can avoid using `new`.
https://stackoverflow.com/a/22571331
```suggestion
		auto report1 = std::make_unique<ReceiverReferenceTime>();
```

jmillan:
> Why does TestRtpStreamSend.cpp use shared_ptr instead of unique_ptr?

Because a `shared_ptr` is what the API expects.  Check the method being called.

> Why does NOT TestRtpStreamSend.cpp use xxx_ptr in cases like this one?

That pointer is owned by the caller of `CreateRtpPacket()`, which is handled by a `unique_ptr`

worker/test/src/RTC/RTCP/TestXr.cpp
jmillan: good point, done.

ibc:
> Why does TestRtpStreamSend.cpp use shared_ptr instead of unique_ptr?
> Because a shared_ptr is what the API expects. Check the method being called.

I don't understand. I literally mean that I replaced some shared_ptr with unique_ptr in that files and it compiles, tests pass and no ASAN warnings (in that test file).

ibc:
@jmillan, ongoing changes in `TestXr.cpp` in this PR are producing a SIGSEV when running worker tests (no need to run them with ASAN stuff):

```
-------------------------------------------------------------------------------
Scenario: RTCP XrDelaySinceLastRt parsing
  create RRT
-------------------------------------------------------------------------------
../../../test/src/RTC/RTCP/TestXr.cpp:131
...............................................................................

../../../test/src/RTC/RTCP/TestXr.cpp:131: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

test cases:     39 |     38 passed | 1 failed
assertions: 592664 | 592663 passed | 1 failed
```

My guess is that we are creating a XR packet and we are passing reports to it created with `unique_ptr`, however the XR packet will delete those pointers by itself when it's deleted. Does it make sense?

ibc:
> My guess is that we are creating a XR packet and we are passing reports to it created with `unique_ptr`, however the XR packet will delete those pointers by itself when it's deleted. Does it make sense?

Confirmed, and fixed here https://github.com/versatica/mediasoup/pull/1421/commits/281054c40476ab4fa0f91f68a566f7766e7eb752

jmillan:
@ibc, this PR is ready to merge, plus there are 0 issues when running `make test-asan-address`

ibc:
My aim in this PR is to use unique/share_ptr in all test files, not just in `TestXr.cpp`.

ibc:
@jmillan thanks a lot for your commits. Just one thing, we don't need to include `<memory>`, it's already done in `common.hpp`.

ibc:
This is done. I've enabled test-asan-address in CI (only for Linux) and will merge if CI passes.

PR #1422, opened 2024-07-04
flatbuffer builder reuse

Lynnworld:
use flatbuffer Builder Clear() , so it can reuse it's inner buf .  

```
void | Clear () // Reset all the state in this FlatBufferBuilder so it can be reused to construct another buffer.
```
`https://flatbuffers.dev/classflatbuffers_1_1_flat_buffer_builder.html#ae94b94ba71ea0aeb2d9a98c43b713412`

### DIFF
--- a/worker/include/Channel/ChannelRequest.hpp
+++ b/worker/include/Channel/ChannelRequest.hpp
@@ -22,6 +22,7 @@ namespace Channel
 
 	public:
 		static absl::flat_hash_map<FBS::Request::Method, const char*> method2String;
+		thread_local static flatbuffers::FlatBufferBuilder bufferBuilder;
 
 	public:
 		ChannelRequest(Channel::ChannelSocket* channel, const FBS::Request::Request* request);
@@ -29,7 +30,7 @@ namespace Channel
 
 		flatbuffers::FlatBufferBuilder& GetBufferBuilder()
 		{
-			return this->bufferBuilder;
+			return ChannelRequest::bufferBuilder;
 		}
 		void Accept();
 		template<class Body>
@@ -39,7 +40,7 @@ namespace Channel
 
 			this->replied = true;
 
-			auto& builder = this->bufferBuilder;
+			auto& builder = ChannelRequest::bufferBuilder;
 			auto response = FBS::Response::CreateResponse(builder, this->id, true, type, body.Union());
 
 			auto message =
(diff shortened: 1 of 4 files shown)


worker/src/Channel/ChannelRequest.cpp
```diff
@@ -83,7 +83,7 @@ namespace Channel
 		{ FBS::Request::Method::RTPOBSERVER_REMOVE_PRODUCER,                    "rtpObserver.removeProducer"                 },
 	};
 	// clang-format on
-
+	thread_local flatbuffers::FlatBufferBuilder ChannelRequest::bufferBuilder{};
```
ibc: This must be moved above to a new "Static variables." section at the top as done in many other cpp files. Also ensure you are using tab rather than spaces.

ibc (commented):
Thanks a lot. Just a cosmetic request.

jmillan:
Currently we use `Reset()` which calls `Clear()` **plus** deallocates the buffer. So indeed yes, calling `Clear()` is just enough and does not require to reallocate the buffer on the next call.

But this PR should only replace `Reset()` by `Clear()`. It's not necessary to move the builder from being an instance member to a thread local variable.

Lynnworld:
current Channel socket will `new ChannelRequest` every time ,  so try to  reuse request instead of thread_local builder ?

jmillan:
> current Channel socket will new ChannelRequest every time , so try to reuse request instead of thread_local builder ?

Correct, that's orthogonal to `Reset()` replacement described in the PR description, but it's true, we can reuse the `FlatBufferBuilder` for every request since all responses are synchronous at the moment.

jmillan:
Please run the linter `npm run lint:worker`

jmillan (approved):

worker/include/Channel/ChannelRequest.hpp
```diff
@@ -22,14 +22,15 @@ namespace Channel
...
+		thread_local static flatbuffers::FlatBufferBuilder bufferBuilder;
 
 	public:
 		ChannelRequest(Channel::ChannelSocket* channel, const FBS::Request::Request* request);
 		~ChannelRequest() = default;
 
 		flatbuffers::FlatBufferBuilder& GetBufferBuilder()
 		{
-			return this->bufferBuilder;
+			return ChannelRequest::bufferBuilder;
```
ibc, line 33: @jmillan you have approved this PR but here we are using a thread static member instead of a instance member, so I don't understand.

ibc:
Thanks, merging.

Lynnworld:
oops, I'll make a new PR later for `ChannelRequest` and `ChannelNotification` reuse , and remove `threal_local  static `. @jmillan @ibc

PR #1423, opened 2024-07-04
#1422 follow up, Reuse ChannelRequest and ChannelNotification

Lynnworld:
Try reuse `ChannelRequest` and `ChannelNotification`  , and no more thread_local static flatbuffer builder.

### DIFF
--- a/worker/include/Channel/ChannelNotification.hpp
+++ b/worker/include/Channel/ChannelNotification.hpp
@@ -17,8 +17,9 @@ namespace Channel
 		static absl::flat_hash_map<FBS::Notification::Event, const char*> event2String;
 
 	public:
-		explicit ChannelNotification(const FBS::Notification::Notification* notification);
+		ChannelNotification()  = default;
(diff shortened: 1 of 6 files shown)


jmillan:
I'll be few days away from the computer. Let me review this once I'm back after the weekend.

jmillan:
I don't like this approach. Each incoming request needs to create a new `Request` instance. I don't like the fact that we reuse the same request all over.

If you are doing this for performance purposes, please show us the penalties of creating a `Request` instance for each incoming request. We can sacrifice design for performance up to a point, and IMHO this is not justified.

ibc:
> I don't like this approach. Each incoming request needs to create a new `Request` instance. I don't like the fact that we reuse the same request all over.
> 
> If you are doing this for performance purposes, please show us the penalties of creating a `Request` instance for each incoming request. We can sacrifice design for performance up to a point, and IMHO this is not justified.

Agreed. Definitely having `this->request.Receive(xxx)` is not nice.

PR #1431, opened 2024-07-25
Fix frozen video in simulcast due to wrong dropping of padding only packets

### LINKED ISSUE #1429: Simulcast frozen video due to dropping RTP packets
## Bug Report

We recently upgraded mediasoup from version 3.12.13 to 3.14.7 and noticed that video freezing has become more frequent since the upgrade. After enabling additional mediasoup logging, we believe this issue is due to a regression introduced in [this pull request](https://github.com/versatica/mediasoup/pull/1403).

The problem occurs under the following conditions:

1. The producer has two simulcast streams (`SSRC-A` and `SSRC-B`).
2. The simulcast consumer is consuming `SSRC-B`.
3. `SSRC-A` has a packet that contains only padding. Let's assume this packet's sequence number is `SEQ-1`. This packet is dropped and stored in the `rtpSeqManager`
https://github.com/versatica/mediasoup/blob/06f66e62874ede5f4bb4a166ba2629280b9edada/worker/src/RTC/SimpleConsumer.cpp#L320
4. `SSRC-B` then sends an RTP packet with the same sequence number. Since this sequence number is in `rtpSeqManager's dropped` list, it is assigned a sequence number of 0.
https://github.com/versatica/mediasoup/blob/06f66e62874ede5f4bb4a166ba2629280b9edada/worker/src/RTC/SimpleConsumer.cpp#L394
5. The packet is then dropped because its sequence number is 0
https://github.com/versatica/mediasoup/blob/06f66e62874ede5f4bb4a166ba2629280b9edada/worker/src/RTC/SimpleConsumer.cpp#L432

Attached are logs showing that RTP packets with sequence numbers 42015-42134 were dropped due to this scenario.
[exportedLogRecords.CSV](https://github.com/user-attachments/files/16381309/exportedLogRecords.CSV)

### Your environment

- Operating system:
- Node version:
- npm version:
- gcc/clang version:
- mediasoup version: 3.14.7
- mediasoup-client version:

ibc:
Fixes #1429

Thanks to @quanli168 for the report.

### Details

- In `SimulcastConsumer` we must drop padding only packets **ONLY** if the packet belongs to the current simulcast stream being sent to the consuming endpoint.
- Changes in other `XxxxConsumer` classes are just cosmetic for consistency.

### DIFF
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -730,18 +730,6 @@ namespace RTC
 			return;
 		}
 
-		// Packets with only padding are not forwarded.
-		if (packet->GetPayloadLength() == 0)
-		{
-			this->rtpSeqManager.Drop(packet->GetSequenceNumber());
-
-#ifdef MS_RTC_LOGGER_RTP
-			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::EMPTY_PAYLOAD);
-#endif
-
-			return;
-		}
-
 		if (this->targetTemporalLayer == -1)
 		{
 #ifdef MS_RTC_LOGGER_RTP
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -810,6 +810,19 @@ namespace RTC
 			return;
 		}
 
+		// If the packet belongs to current spatial layer being sent and packet does
+		// not have payload other than padding, then drop it.
+		if (spatialLayer == this->currentSpatialLayer && packet->GetPayloadLength() == 0)
+		{
+			this->rtpSeqManager.Drop(packet->GetSequenceNumber());
+
+#ifdef MS_RTC_LOGGER_RTP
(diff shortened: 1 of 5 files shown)


worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -810,6 +810,19 @@ namespace RTC
...
+		{
+			this->rtpSeqManager.Drop(packet->GetSequenceNumber());
+
+#ifdef MS_RTC_LOGGER_RTP
+			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::EMPTY_PAYLOAD);
+#endif
+
+			return;
+		}
+
```
jmillan, line 813: This block should be removed https://github.com/versatica/mediasoup/blob/8b1ba663caec7b0a6909717e251f4a80e2f4397b/worker/src/RTC/SimulcastConsumer.cpp#L734

ibc: Upps, yes

jmillan:
:+1:

PR #1438, opened 2024-08-05
Mitigate libsrtp wraparound with loss decryption failure

### LINKED ISSUE #1437: libsrtp: wraparound with loss decryption failure
Rationale in the ticket: https://webrtc-review.googlesource.com/c/src/+/358360

Theoretical solution is that the first **output** seq number of each mediasoup `XxxxConsumer` is not 0 or 1 (to avoid that a previous packet arriving later triggers the bug in libsrtp) and it's smaller than 2^15.

ibc:
Fixes https://github.com/versatica/mediasoup/issues/1437

### Details

- Read the issue, please.
- So solution is that `SeqManager` now includes a second constructor with `initialOutput` and we use it in all `XxxxConsumer` classes.


ibc:
CC @fippo @lminiero

worker/src/RTC/PipeConsumer.cpp
```diff
@@ -680,7 +681,15 @@ namespace RTC
 			this->mapMappedSsrcSsrc[consumableEncoding.ssrc] = encoding.ssrc;
 			this->mapSsrcRtpStream[encoding.ssrc]            = rtpStream;
 			this->mapRtpStreamSyncRequired[rtpStream]        = false;
-			this->mapRtpStreamRtpSeqManager[rtpStream];
+
+			// Let's chosee an initial output seq number between 1000 and 32768 to avoid
```
jmillan:
```suggestion
			// Let's choose an initial output seq number between 1000 and 32768 to avoid
```

worker/include/RTC/SimpleConsumer.hpp
```diff
@@ -87,7 +87,7 @@ namespace RTC
 		RTC::RtpStreamRecv* producerRtpStream{ nullptr };
 		bool keyFrameSupported{ false };
 		bool syncRequired{ false };
-		RTC::SeqManager<uint16_t> rtpSeqManager;
+		std::unique_ptr<RTC::SeqManager<uint16_t>> rtpSeqManager;
```
jmillan, line 90: Why is this now a pointer?

jmillan (approved):

worker/include/RTC/SimpleConsumer.hpp
ibc: Because we no longer initialize it in constructor but later within some method and now we must pass an argument (the base offset or whatever).

PR #1439, opened 2024-08-06
add unit test for mapping packets with lower sequence number than the first mapped packet

fippo:
see #1437 and #1438. Works but why not have a test

### DIFF
--- a/worker/test/src/RTC/TestSeqManager.cpp
+++ b/worker/test/src/RTC/TestSeqManager.cpp
@@ -1400,4 +1400,19 @@ SCENARIO("SeqManager", "[rtc][SeqMananger]")
 		SeqManager<uint16_t> seqManager(/*initialOutput*/ 1000);
 		validate(seqManager, inputs);
 	}
+
+	SECTION("map packets prior to first mapped packet")
+	{
+		// clang-format off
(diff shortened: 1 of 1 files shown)


worker/test/src/RTC/TestSeqManager.cpp
```diff
@@ -1400,4 +1400,19 @@ SCENARIO("SeqManager", "[rtc][SeqMananger]")
 		SeqManager<uint16_t> seqManager(/*initialOutput*/ 1000);
 		validate(seqManager, inputs);
 	}
+
+	SECTION("map packets prior to first mapped packet")
```
ibc, line 1404: I fail to see the value  that this test adds to the already existing tests? Isn't this already tested in existing tests?

Also, what does "mapped packet" mean? We don't use that terminology anywhere in the source code and in existing tests.

fippo: Not covered (or at least not explicitily), does backward wrap-around as a bonus. What terminology do you suggest?

ibc:
Thanks. I've added the corresponding tests in my ongoing PR to test the new feature:

https://github.com/versatica/mediasoup/pull/1438/commits/ecbad4c823adfbce2577ec3a30bcfdccd7e6da6b

Let's close this PR.

Issue #1435, opened 2024-08-04
Docker and io_uring don't work together - add prebuilt binary without liburing


lowski:
## Motivation

When installing mediasoup in a Docker container (on the `linux/amd64` platform) the mediasoup worker will crash with this error `failure exit: io_uring_queue_init() failed: Operation not permitted`.

Containerd has removed support for `io_uring` as it has been deemed unsafe. The rationale for this decision is explained here: https://github.com/containerd/containerd/issues/9048

## Current Workarounds

There are two options to work around this currently: (1) force the local build of mediasoup without io_uring, and (2) override the default seccomp profile.

### Local build

This is the only viable option at the moment. We can force the npm postinstall to build the worker locally by adding these two environment variables before the `npm install` command:

```Dockerfile
# don't download the prebuilt binary
ENV MEDIASOUP_SKIP_WORKER_PREBUILT_DOWNLOAD="true"
# build without io_uring
ENV MESON_ARGS="-Dms_disable_liburing=true"
```

However, this obviously slows down the build process fairly significantly. But it works and has no security cost.

### Seccomp Profile (*BAD* idea)

We can manually override the seccomp profile by passing `--security-opt` with the `docker run` command. I have not been able to find the exact syscalls that cause the issue, but allowing all syscalls by default works (as mentioned here https://github.com/moby/moby/issues/39396#issuecomment-504642822).

This will make mediasoup work again, but at a great security cost -- the syscalls were disallowed for a reason.

*I DON'T THINK THIS IS A GOOD IDEA.* I'm just adding this in case someone stumbles across this solution as I did.

## Feature Request

Even though the first method works, it is less than ideal. However, I understand that `io_uring` brings some performance benefit which justifies its use in general. Therefore, another prebuilt binary should exist without `io_uring` when using Docker.

I don't know if it's possible to detect a docker build automatically and reliably (maybe [this](https://unix.stackexchange.com/questions/607695/how-to-check-if-its-docker-or-host-machine-inside-bash-script) helps?) but that would obviously be the best case. Otherwise an environment variable to switch between the version with and without `io_uring` would be the next best thing.

ibc:
Please add proper rationale. That commit doesn't explain anything to me.

lowski:
@ibc Sorry, the keybinding for submitting an issue without further confirmation got in the way -- I've updated the description!

ibc:
Ok, will think about this during this week.

amontalban:
[...]

PR #1442, opened 2024-08-09
WorkerSettings: Add disableLiburing option (enable_liburing in Rust)

ibc:
### Details

- Node: `createWorker({ disableLiburing: true })` disables liburing usage despite it's supported by the worker and current host.
- Rust: `create_worker({ enable_liburing: false })` disables liburing usage despite it's supported by the worker and current host.
- Related (still to be fixed) issue which brings lot of context: https://github.com/versatica/mediasoup/issues/1435

### DIFF
--- a/rust/src/router/tests.rs
+++ b/rust/src/router/tests.rs
@@ -16,11 +16,10 @@ async fn init() -> Worker {
     let worker_manager = WorkerManager::new();
 
     worker_manager
-        // .create_worker(WorkerSettings {
-        //     enable_liburing: false,
-        //     ..WorkerSettings::default()
-        // })
-        .create_worker(WorkerSettings::default())
+        .create_worker(WorkerSettings {
+            enable_liburing: false,
+            ..WorkerSettings::default()
+        })
         .await
         .expect("Failed to create worker")
 }
--- a/rust/src/router/tests.rs
+++ b/rust/src/router/tests.rs
@@ -16,11 +16,9 @@ async fn init() -> Worker {
     let worker_manager = WorkerManager::new();
 
     worker_manager
-        .create_worker({
-            let mut settings = WorkerSettings::default();
-            settings.enable_liburing = false;
-
-            settings
+        .create_worker(WorkerSettings {
+            enable_liburing: false,
+            ..WorkerSettings::default()
         })
         .await
         .expect("Failed to create worker")
--- a/rust/src/router/tests.rs
+++ b/rust/src/router/tests.rs
@@ -16,10 +16,11 @@ async fn init() -> Worker {
     let worker_manager = WorkerManager::new();
 
     worker_manager
-        .create_worker(WorkerSettings {
-            enable_liburing: false,
-            ..WorkerSettings::default()
-        })
+        // .create_worker(WorkerSettings {
+        //     enable_liburing: false,
+        //     ..WorkerSettings::default()
+        // })
+        .create_worker(WorkerSettings::default())
         .await
         .expect("Failed to create worker")
 }
--- a/rust/src/router/tests.rs
+++ b/rust/src/router/tests.rs
@@ -16,7 +16,12 @@ async fn init() -> Worker {
     let worker_manager = WorkerManager::new();
 
     worker_manager
-        .create_worker(WorkerSettings::default())
+        .create_worker({
+            let mut settings = WorkerSettings::default();
+            settings.enable_liburing = false;
+
+            settings
+        })
         .await
         .expect("Failed to create worker")
 }
@@ -28,12 +33,6 @@ fn worker_close_event() {
 
         let router = worker
             .create_router(RouterOptions::default())
-            .create_worker({
-                let mut settings = WorkerSettings::default();
-                settings.enable_liburing = false;
-
-                settings
-            })
             .await
             .expect("Failed to create router");
 
--- a/rust/src/router/tests.rs
+++ b/rust/src/router/tests.rs
@@ -28,6 +28,12 @@ fn worker_close_event() {
 
         let router = worker
             .create_router(RouterOptions::default())
+            .create_worker({
+                let mut settings = WorkerSettings::default();
+                settings.enable_liburing = false;
+
+                settings
+            })
             .await
             .expect("Failed to create router");
 
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -413,7 +413,7 @@ impl Inner {
             ));
         }
 
-        if enable_liburing {
+        if !enable_liburing {
             spawn_args.push("--disable_liburing".to_string());
         }
 
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -192,6 +192,8 @@ pub struct WorkerSettings {
     /// "WebRTC-Bwe-AlrLimitedBackoff/Enabled/".
     #[doc(hidden)]
     pub libwebrtc_field_trials: Option<String>,
+    /// Disable liburing (io_uring) despite it's supported in current host.
+    pub disable_liburing: Option<bool>,
     /// Function that will be called under worker thread before worker starts, can be used for
     /// pinning worker threads to CPU cores.
     pub thread_initializer: Option<Arc<dyn Fn() + Send + Sync>>,
@@ -221,6 +223,7 @@ impl Default for WorkerSettings {
             rtc_port_range: 10000..=59999,
             dtls_files: None,
             libwebrtc_field_trials: None,
+            disable_liburing: None,
             thread_initializer: None,
             app_data: AppData::default(),
         }
@@ -235,6 +238,7 @@ impl fmt::Debug for WorkerSettings {
             rtc_port_range,
             dtls_files,
             libwebrtc_field_trials,
+            disable_liburing,
             thread_initializer,
             app_data,
         } = self;
@@ -245,6 +249,7 @@ impl fmt::Debug for WorkerSettings {
             .field("rtc_port_range", &rtc_port_range)
             .field("dtls_files", &dtls_files)
             .field("libwebrtc_field_trials", &libwebrtc_field_trials)
+            .field("disable_liburing", &disable_liburing)
             .field(
                 "thread_initializer",
                 &thread_initializer.as_ref().map(|_| "ThreadInitializer"),
@@ -356,6 +361,7 @@ impl Inner {
             rtc_port_range,
             dtls_files,
             libwebrtc_field_trials,
+            disable_liburing,
             thread_initializer,
             app_data,
         }: WorkerSettings,
@@ -402,6 +408,10 @@ impl Inner {
                 "--libwebrtcFieldTrials={}",
                 libwebrtc_field_trials.as_str()
             ));
+        }
+
+        if let Some(disable_liburing) = disable_liburing {
+            spawn_args.push(format!("--disable_liburing"));
         }
 
         let id = WorkerId::new();
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -414,7 +414,7 @@ impl Inner {
         }
 
         if !enable_liburing {
-            spawn_args.push("--disableLiburing".to_string());
+            spawn_args.push("--disableLiburing=true".to_string());
         }
 
         let id = WorkerId::new();
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -414,7 +414,7 @@ impl Inner {
         }
 
         if !enable_liburing {
-            spawn_args.push("--disable_liburing".to_string());
+            spawn_args.push("--disableLiburing".to_string());
         }
 
         let id = WorkerId::new();
--- a/rust/src/worker.rs
+++ b/rust/src/worker.rs
@@ -192,8 +192,11 @@ pub struct WorkerSettings {
     /// "WebRTC-Bwe-AlrLimitedBackoff/Enabled/".
     #[doc(hidden)]
     pub libwebrtc_field_trials: Option<String>,
-    /// Disable liburing (io_uring) despite it's supported in current host.
-    pub disable_liburing: Option<bool>,
+    /// Enable liburing  This option is ignored if io_uring is not supported by
+    /// current host.
+    ///
+    /// Default `true`.
+    pub enable_liburing: bool,
     /// Function that will be called under worker thread before worker starts, can be used for
     /// pinning worker threads to CPU cores.
     pub thread_initializer: Option<Arc<dyn Fn() + Send + Sync>>,
@@ -223,7 +226,7 @@ impl Default for WorkerSettings {
             rtc_port_range: 10000..=59999,
             dtls_files: None,
             libwebrtc_field_trials: None,
-            disable_liburing: None,
+            enable_liburing: true,
             thread_initializer: None,
             app_data: AppData::default(),
         }
@@ -238,7 +241,7 @@ impl fmt::Debug for WorkerSettings {
             rtc_port_range,
             dtls_files,
             libwebrtc_field_trials,
-            disable_liburing,
+            enable_liburing,
             thread_initializer,
             app_data,
         } = self;
@@ -249,7 +252,7 @@ impl fmt::Debug for WorkerSettings {
             .field("rtc_port_range", &rtc_port_range)
             .field("dtls_files", &dtls_files)
             .field("libwebrtc_field_trials", &libwebrtc_field_trials)
-            .field("disable_liburing", &disable_liburing)
+            .field("enable_liburing", &enable_liburing)
             .field(
                 "thread_initializer",
                 &thread_initializer.as_ref().map(|_| "ThreadInitializer"),
@@ -361,7 +364,7 @@ impl Inner {
             rtc_port_range,
             dtls_files,
             libwebrtc_field_trials,
-            disable_liburing,
+            enable_liburing,
             thread_initializer,
             app_data,
         }: WorkerSettings,
@@ -410,8 +413,8 @@ impl Inner {
             ));
         }
 
-        if let Some(disable_liburing) = disable_liburing {
-            spawn_args.push(format!("--disable_liburing"));
+        if enable_liburing {
+            spawn_args.push("--disable_liburing".to_string());
         }
 
         let id = WorkerId::new();
--- a/worker/src/Settings.cpp
+++ b/worker/src/Settings.cpp
@@ -79,7 +79,7 @@ void Settings::SetConfiguration(int argc, char* argv[])
 	{
 		if (!optarg)
 		{
-			MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+			MS_THROW_TYPE_ERROR("missing value in command line argument in option '%c'", c);
 		}
 
 		switch (c)
--- a/worker/src/Settings.cpp
+++ b/worker/src/Settings.cpp
@@ -81,6 +81,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 		{
 			case 'l':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				stringValue = std::string(optarg);
 				SetLogLevel(stringValue);
 
@@ -89,6 +94,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 			case 't':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				stringValue = std::string(optarg);
 				logTags.push_back(stringValue);
 
@@ -97,6 +107,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 			case 'm':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				try
 				{
 					Settings::configuration.rtcMinPort = static_cast<uint16_t>(std::stoi(optarg));
@@ -111,6 +126,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 			case 'M':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				try
 				{
 					Settings::configuration.rtcMaxPort = static_cast<uint16_t>(std::stoi(optarg));
@@ -125,6 +145,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 			case 'c':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				stringValue                                 = std::string(optarg);
 				Settings::configuration.dtlsCertificateFile = stringValue;
 
@@ -133,6 +158,11 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 			case 'p':
 			{
+				if (!optarg)
+				{
+					MS_THROW_TYPE_ERROR("missing value in command line argument: %s", optarg);
+				}
+
 				stringValue                                = std::string(optarg);
 				Settings::configuration.dtlsPrivateKeyFile = stringValue;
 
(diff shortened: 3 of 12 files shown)


rust/src/worker.rs
```diff
@@ -404,6 +410,10 @@ impl Inner {
             ));
         }
 
+        if let Some(disable_liburing) = disable_liburing {
```
ibc: I assume here we are checking if user given `disable_liburing` is true.

rust/src/worker.rs
```diff
@@ -192,6 +192,8 @@ pub struct WorkerSettings {
     /// "WebRTC-Bwe-AlrLimitedBackoff/Enabled/".
     #[doc(hidden)]
     pub libwebrtc_field_trials: Option<String>,
+    /// Disable liburing (io_uring) despite it's supported in current host.
+    pub disable_liburing: Option<bool>,
```
nazar-pc: What about this instead?:
```suggestion
    pub disable_liburing: bool,
```

Or better yet (disable: false is much less natural):
```suggestion
    pub enable_liburing: bool,
```

ibc: No please, here we want to keep current behavior and parity with Node layer. Default behavior must be that liburing is enabled if supported by current host.

And why should `disable_liburing` be mandatory? I want it to be optional.

nazar-pc: Because by making it optional:
* you have tri-state boolean: Some(true), Some(false), None
* you don't let user know what the default value is

By having it as `bool` API is much clearer and self-documenting. Setting option to `false` is just as easy as setting it to `None` in `impl Default for WorkerSettings`.

On worker level it will be the same, but on Rust API level I would prefer `enable_io_uring` defaulting to `true`, which user can set to `false`. This is a clearer API, doesn't surface what library is used for `io_uring` and it is more natural to read `enable: true|false`.

rust/src/worker.rs
nazar-pc: No, you're checking that `disable_liburing` is `Some()`, for checking what is inside you'd also have to check the boolean `disable_liburing` itself. Can be simplified to `disable_liburing == Some(true)`. One more reason to not have tri-state boolean though.

rust/src/worker.rs
ibc: > on Rust API level I would prefer enable_io_uring defaulting to true

It's not defaulting to anything, you are making it mandatory for the user to pass true or false.

nazar-pc: You may be misunderstanding Rust API. It is not making anything mandatory, the field always has to be specified anyway and you're setting it to `None` in `impl Default for WorkerSettings`. You can't not specify fields like in JavaScript. So you just change `Option<bool>` to `bool` and set default value to `false` instead of `None`.

The struct is intentionally marked as `#[non_exhaustive]`, so user can't construct it directly anyway, the only way is to go through APIs provided by the library like `impl Default for WorkerSettings`. This is also the reason why this is actually **not** a breaking change for library API!

ibc: ok done: https://github.com/versatica/mediasoup/pull/1442/commits/0529246e7668f97542bf850e386434cd5ca6012b

ibc:
OMG Rust tests fail because of this:

```
mediasoup-worker::mediasoup_worker_run() | settings error: unknown long option given as argument
```

I **hate** that `getopt.h` ancient thing...

And BTW it fails due to this obviously wrong code that I'm fixing now, but it doesn't justify the problem:

```rs
if enable_liburing {
    spawn_args.push("--disable_liburing".to_string());
}
```

rust/src/worker.rs
```diff
@@ -404,6 +413,10 @@ impl Inner {
             ));
         }
 
+        if enable_liburing {
```
nazar-pc:
```suggestion
        if !enable_liburing {
```

ibc: yes yes yes, but see https://github.com/versatica/mediasoup/pull/1442#issuecomment-2278380704 please

rust/src/worker.rs
```diff
@@ -404,6 +413,10 @@ impl Inner {
             ));
         }
 
+        if !enable_liburing {
+            spawn_args.push("--disable_liburing".to_string());
```
nazar-pc: I believe it should have been the same as in settings file:
```suggestion
            spawn_args.push("--disableLiburing".to_string());
```

ibc: OMG....

ibc:
Ok, this is done.

nazar-pc (approved):
Assuming CI passes this looks fine to me

ibc:
Why is this happening in CI????

https://github.com/versatica/mediasoup/actions/runs/10323061408/job/28579657794?pr=1442

`cargo clippy` works in my machine and we force Rust version!!!!

ibc:
OMG I don't want to spend all the weekend with this.

ibc:
Ok, `cargo clippy --all-targets -- -D warnings` fails in local too.

nazar-pc:
I suspect you might be forcing a different version locally. It wants you to write this idiomatic Rust instead:
```rust
let settings = WorkerSettings {
    enable_liburing: false,
    ..WorkerSettings::default()
};
```

ibc:
Why does it complain about this?

```
error: field assignment outside of initializer for an instance created with Default::default()
  --> rust/src/router/tests.rs:21:13
   |
21 |             settings.enable_liburing = false;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: consider initializing the variable with `worker::WorkerSettings { enable_liburing: false, ..Default::default() }` and removing relevant reassignments
  --> rust/src/router/tests.rs:20:13
   |
20 |             let mut settings = WorkerSettings::default();
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This is the code:

```rust
worker_manager
        .create_worker({
            let mut settings = WorkerSettings::default();
            settings.enable_liburing = false;

            settings
        })
        .await
        .expect("Failed to create worker")
```

We do THE SAME in many other files such as here:

```rust
let worker = worker_manager
            .create_worker({
                let mut settings = WorkerSettings::default();

                settings.log_level = WorkerLogLevel::Debug;
                settings.log_tags = vec![WorkerLogTag::Info];
                settings.rtc_port_range = 0..=9999;
                settings.dtls_files = Some(WorkerDtlsFiles {
                    certificate: "tests/integration/data/dtls-cert.pem".into(),
                    private_key: "tests/integration/data/dtls-key.pem".into(),
                });
                settings.libwebrtc_field_trials =
                    Some("WebRTC-Bwe-AlrLimitedBackoff/Disabled/".to_string());
                settings.app_data = AppData::new(CustomAppData { bar: 456 });

                settings
            })
            .await
            .expect("Failed to create worker with custom settings");
```

nazar-pc:
I suspect it has a heuristic about number of fields or just didn't learn to handle more complex cases, not sure which one of those two

ibc:
Pushed, should be green now.

nazar-pc:
Generally what it suggests does look a bit nicer and even shorter, but not always

ibc:
OMG now tests failing...

ibc:
WTH???

https://github.com/versatica/mediasoup/actions/runs/10323816619/job/28581996102?pr=1442
```
  process didn't exit successfully: `/home/runner/work/mediasoup/mediasoup/target/debug/deps/mediasoup-8f5b5278ff2f31a1` (signal: 11, SIGSEGV: invalid memory reference)
```

nazar-pc:
Memory issues... again :confused:

ibc:
> Memory corruption... again 😕

When did this happen in a test recently? I don't remember.

ibc:
It fails consistently, this is depressing, no changes have been done that should affect this AT ALL:

- https://github.com/versatica/mediasoup/actions/runs/10323816619/job/28583849467?pr=1442
- https://github.com/versatica/mediasoup/actions/runs/10323816619/job/28600342140?pr=1442

ibc:
No idea why the issue happens consistently but indeed it's the already know issue in Rust due to the separate usrsctp thread. I worked on this some months ago and there is an unfinished and very complex PR for which honestly I don't have time to revive now. I think I will disable these failing tests because we do know that the underlying code can fail. Well, all this assuming that the failing test is using two workers with pipe transports. If not, ignore all I said here.

ibc:
Those are the ways Rust tests systematically fail. Since they run in parallel and taking into account that memory issues could make tests fail at any time once 

```
running 27 tests
test data_structures::tests::dtls_fingerprint ... ok
test ortc::tests::generate_router_rtp_capabilities_succeeds ... ok
test ortc::tests::generate_router_rtp_capabilities_too_many_codecs ... ok
test ortc::tests::generate_router_rtp_capabilities_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_get_consumable_rtp_parameters_get_consumer_rtp_parameters_get_pipe_consumer_rtp_parameters_succeeds ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_unsupported ... ok
test router::audio_level_observer::tests::router_close_event ... ok
test router::active_speaker_observer::tests::router_close_event ... ok
test router::consumer::tests::transport_close_event ... ok
test router::consumer::tests::producer_close_event ... ok
test router::direct_transport::tests::router_close_event ... ok
test router::data_producer::tests::transport_close_event ... ok
test router::data_consumer::tests::transport_close_event ... ok
test router::data_consumer::tests::data_producer_close_event ... ok
test router::pipe_transport::tests::data_producer_close_is_transmitted_to_pipe_data_consumer ... ok
test router::plain_transport::tests::router_close_event ... ok
test router::producer::tests::transport_close_event ... ok
error: test failed, to rerun pass `-p mediasoup --lib`

Caused by:
  process didn't exit successfully: `/home/runner/work/mediasoup/mediasoup/target/debug/deps/mediasoup-8f5b5278ff2f31a1` (signal: 11, SIGSEGV: invalid memory reference)
Error: Process completed with exit code 101.
##[debug]Finishing: cargo test
```

```
running 27 tests
test data_structures::tests::dtls_fingerprint ... ok
test ortc::tests::generate_router_rtp_capabilities_succeeds ... ok
test ortc::tests::generate_router_rtp_capabilities_too_many_codecs ... ok
test ortc::tests::generate_router_rtp_capabilities_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_get_consumable_rtp_parameters_get_consumer_rtp_parameters_get_pipe_consumer_rtp_parameters_succeeds ... ok
test router::audio_level_observer::tests::router_close_event ... ok
test router::active_speaker_observer::tests::router_close_event ... ok
test router::consumer::tests::producer_close_event ... ok
test router::data_consumer::tests::data_producer_close_event ... ok
test router::data_consumer::tests::transport_close_event ... ok
test router::consumer::tests::transport_close_event ... ok
test router::data_producer::tests::transport_close_event ... ok
test router::direct_transport::tests::router_close_event ... ok
test router::plain_transport::tests::router_close_event ... ok
test router::pipe_transport::tests::data_producer_close_is_transmitted_to_pipe_data_consumer ... ok
test router::producer::tests::transport_close_event ... ok
error: test failed, to rerun pass `-p mediasoup --lib`

Caused by:
  process didn't exit successfully: `/home/runner/work/mediasoup/mediasoup/target/debug/deps/mediasoup-8f5b5278ff2f31a1` (signal: 11, SIGSEGV: invalid memory reference)
Error: Process completed with exit code 101.
```

```
running 27 tests
test data_structures::tests::dtls_fingerprint ... ok
test ortc::tests::generate_router_rtp_capabilities_succeeds ... ok
test ortc::tests::generate_router_rtp_capabilities_too_many_codecs ... ok
test ortc::tests::generate_router_rtp_capabilities_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_get_consumable_rtp_parameters_get_consumer_rtp_parameters_get_pipe_consumer_rtp_parameters_succeeds ... ok
test router::active_speaker_observer::tests::router_close_event ... ok
test router::audio_level_observer::tests::router_close_event ... ok
test router::consumer::tests::transport_close_event ... ok
test router::consumer::tests::producer_close_event ... ok
test router::data_consumer::tests::transport_close_event ... ok
test router::direct_transport::tests::router_close_event ... ok
test router::data_producer::tests::transport_close_event ... ok
test router::data_consumer::tests::data_producer_close_event ... ok
test router::plain_transport::tests::router_close_event ... ok
error: test failed, to rerun pass `-p mediasoup --lib`

Caused by:
  process didn't exit successfully: `/home/runner/work/mediasoup/mediasoup/target/debug/deps/mediasoup-8f5b5278ff2f31a1` (signal: 11, SIGSEGV: invalid memory reference)
Error: Process completed with exit code 101.
```

I would like to say that this is pipe transport related but the third test attempt doesn't even execute any "pipe" test.

ibc:
This is completely amazing. Rust tests in v3 branch don't fail. But Rust tests in this PR branch fial consistently. Changes in this PR cannot be the culprit at all, they do not affect anything in Rust.

ibc:
@nazar-pc help please? I doubt that Rust tests are even failing due to this well know issue:

https://github.com/versatica/mediasoup/issues/1352

In a comment above you can see that sometimes pipe transports do not even run and tests still fail. Maybe I'm wrong but AFAIK we only run more than 1 workers in Rust tests in those pipe transport related tests, so I'm completely lost.

nazar-pc:
> Changes in this PR cannot be the culprit at all, they do not affect anything in Rust.

Either changes in this PR are the culprit (however unlikely that seems) or they trigger an issue that existed before as well, but was hard to reproduce.

> Maybe I'm wrong but AFAIK we only run more than 1 workers in Rust tests in those pipe transport related tests, so I'm completely lost.

There are as many tests running concurrently as CPU cores on the machine by default, so we have many workers with potentially different settings running at the same time and order is not deterministic.

worker/src/Settings.cpp
```diff
@@ -73,13 +74,9 @@ void Settings::SetConfiguration(int argc, char* argv[])
 
 	optind = 1; // Set explicitly, otherwise subsequent runs will fail.
 	opterr = 0; // Don't allow getopt to print error messages.
+
 	while ((c = getopt_long_only(argc, argv, "", options, &optionIdx)) != -1)
 	{
-		if (!optarg)
-		{
-			MS_THROW_TYPE_ERROR("unknown configuration parameter: %s", optarg);
-		}
```
nazar-pc: Why was this deleted? It seems like an important safety check that caller provided something meaningful as an input.

ibc: Because the new `--disableLiburing` command line argument doesn't have any value so such a check throws if present. I can check that `optargs` exist for all the other arguments but didn't consider it necessary.

nazar-pc: Well, but other options do require a value. Maybe we have a test that checks that and it causes memory corruption because you suddenly create a value out of null pointer?

ibc:
> > Maybe I'm wrong but AFAIK we only run more than 1 workers in Rust tests in those pipe transport related tests, so I'm completely lost.
> 
> There are as many tests running concurrently as CPU cores on the machine by default, so we have many workers with potentially different settings running at the same time and order is not deterministic.

Do you mean that different tests run in parallel in the same Rust process? I mean, of course! Ok, I assume there is no way in Rust to create a separate process for each test file, right?

nazar-pc:
> Do you mean that different tests run in parallel in the same Rust process? I mean, of course! Ok, I assume there is no way in Rust to create a separate process for each test file, right?

Yes, it does run in the same process by default. You could probably make it run in separate processes somehow, but we're dealing with memory safe language, so why, right? That is not a solution to the problem we have here anyway :slightly_smiling_face:

worker/src/Settings.cpp
ibc: I've made this change:

https://github.com/versatica/mediasoup/pull/1442/commits/db2252c4edc64a0f4be53f6d9d51622592ac25ba

Is it enough? What do you mean with "you suddenly create a value out of null pointer"? Command line arguments are created by Node and Rust layers in their `Worker` classes. Tests can not trigger wrong arguments passed to the worker.

Just wondering about this: In `worker/utils.rs`:

```rust
pub(super) fn run_worker_with_channels<OE>(
    id: WorkerId,
    thread_initializer: Option<Arc<dyn Fn() + Send + Sync>>,
    args: Vec<String>,
    worker_closed: Arc<AtomicBool>,
    on_exit: OE,
) -> WorkerRunResult
where
    OE: FnOnce(Result<(), ExitError>) + Send + 'static,
{
    let (channel, prepared_channel_read, prepared_channel_write) =
        Channel::new(Arc::clone(&worker_closed));
    let buffer_worker_messages_guard =
        channel.buffer_messages_for(SubscriptionTarget::String(std::process::id().to_string()));

    std::thread::Builder::new()
        .name(format!("mediasoup-worker-{id}"))
        .spawn(move || {
            if let Some(thread_initializer) = thread_initializer {
                thread_initializer();
            }
            let argc = args.len() as c_int;
            let args_cstring = args
                .into_iter()
                .map(|s| -> CString { CString::new(s).unwrap() })
                .collect::<Vec<_>>();
            let argv = args_cstring
                .iter()
                .map(|arg| arg.as_ptr().cast::<c_char>())
                .collect::<Vec<_>>();
            let version = CString::new(env!("CARGO_PKG_VERSION")).unwrap();

            let status_code = unsafe {
                let (channel_read_fn, channel_read_ctx, _channel_write_callback) =
                    prepared_channel_read.deconstruct();
                let (channel_write_fn, channel_write_ctx, _channel_read_callback) =
                    prepared_channel_write.deconstruct();

                mediasoup_sys::mediasoup_worker_run(
                    argc,
                    argv.as_ptr(),
                    version.as_ptr(),
                    0,
                    0,
                    channel_read_fn,
                    channel_read_ctx,
                    channel_write_fn,
                    channel_write_ctx,
                )
            };
```

Here `args` is a command line arguments string, something like:

```
"--logLevel=warn --disableLiburing"
```

Maybe something dangerous when doing this?:

```rust
let args_cstring = args
                .into_iter()
                .map(|s| -> CString { CString::new(s).unwrap() })
                .collect::<Vec<_>>();
            let argv = args_cstring
                .iter()
                .map(|arg| arg.as_ptr().cast::<c_char>())
                .collect::<Vec<_>>();
```

ibc: Nope, no problem there IMHO. It just splits the string into these strings:
- "--logLevel=warn"
- "--disableLiburing"

It doesn't do anything like assuming/expecting a "=" symbol, so no danger here.

ibc: I don't see why commit https://github.com/versatica/mediasoup/commit/db2252c4edc64a0f4be53f6d9d51622592ac25ba should fix this problem. It probably won't and, instead of wasting more time on this, I will change the new command line argument and add a value to it. No time to deal with ancient command line args stuff.

ibc:
> You could probably make it run in separate processes somehow

No, trust me, I couldn't.

> but we're dealing with memory safe language, so why, right?

Because there is also C code running its own separate thread: https://github.com/versatica/mediasoup/issues/1352

ibc:
Ok, some news:

- I can reproduce the Rust test failure in a real Linux host, no need to wait for CI.
- I've made command line argument mandatory here: https://github.com/versatica/mediasoup/pull/1442/commits/669ddca8a394bb985f191311689557990369ced3
- Now Rust tests fail even in macOS with this error:
  ```
  mediasoup-worker::mediasoup_worker_run() | settings error: missing value in command line 
  argument: (null)
  test router::data_producer::tests::transport_close_event ... ok
  test router::tests::worker_close_event ... FAILED
  ```

So it seems that we are sending some command line arg without value. I will print things to debug.

ibc:
> Ok, some news:
> 
> * I can reproduce the Rust test failure in a real Linux host, no need to wait for CI.
> * I've made command line argument mandatory here: [669ddca](https://github.com/versatica/mediasoup/commit/669ddca8a394bb985f191311689557990369ced3)
> * Now Rust tests fail even in macOS with this error:
>   ```
>   mediasoup-worker::mediasoup_worker_run() | settings error: missing value in command line 
>   argument: (null)
>   test router::data_producer::tests::transport_close_event ... ok
>   test router::tests::worker_close_event ... FAILED
>   ```
> 
> So it seems that we are sending some command line arg without value. I will print things to debug.

Ignore, I forgot to add `=true` in Rust. Now things should work.

ibc:
This is strongly depressing. Now WHAT????

```
running 27 tests
test data_structures::tests::dtls_fingerprint ... ok
test ortc::tests::generate_router_rtp_capabilities_succeeds ... ok
test ortc::tests::generate_router_rtp_capabilities_too_many_codecs ... ok
test ortc::tests::generate_router_rtp_capabilities_unsupported ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_get_consumable_rtp_parameters_get_consumer_rtp_parameters_get_pipe_consumer_rtp_parameters_succeeds ... ok
test ortc::tests::get_producer_rtp_parameters_mapping_unsupported ... ok
test router::audio_level_observer::tests::router_close_event ... ok
test router::active_speaker_observer::tests::router_close_event ... ok
test router::data_consumer::tests::transport_close_event ... ok
test router::consumer::tests::producer_close_event ... ok
test router::consumer::tests::transport_close_event ... ok
test router::data_consumer::tests::data_producer_close_event ... ok
test router::data_producer::tests::transport_close_event ... ok
error: test failed, to rerun pass `-p mediasoup --lib`

Caused by:
  process didn't exit successfully: `/home/deploy/src/mediasoup-v3-deploy/target/debug/deps/mediasoup-8f5b5278ff2f31a1` (signal: 11, SIGSEGV: invalid memory reference)
```

ibc:
I am 96% sure that the memory issue is here:

https://github.com/versatica/mediasoup/pull/1442/files#diff-5a9b1b4b01d5764dd21b1d1b7533d5795430dead60f84d450656db89da4e3930R18

ibc:
Ohhh, so we are considering that liburing is enabled or disabled for all workers, but Rust tests run in **parallel** in same process and in just **one** of those tests we were **disabling** liburing. In the other ones liburing is enabled. So here the problem:

```c++
void DepLibUring::ClassInit()
{
	const auto mayor = io_uring_major_version();
	const auto minor = io_uring_minor_version();

	MS_DEBUG_TAG(info, "liburing version: \"%i.%i\"", mayor, minor);

	if (Settings::configuration.liburingDisabled)
	{
		MS_DEBUG_TAG(info, "liburing disabled by user settings");

		return;
	}

	// This must be called first.
	DepLibUring::CheckRuntimeSupport();

	if (DepLibUring::IsEnabled())
	{
		DepLibUring::liburing = new LibUring();

		MS_DEBUG_TAG(info, "liburing enabled");
	}
	else
	{
		MS_DEBUG_TAG(info, "liburing not enabled");
	}
}
```

- `DepLibUring::liburing` singleton is `thread_local static`.
- However we have `static bool enabled`.

rust/src/router/tests.rs
```diff
@@ -16,7 +16,10 @@ async fn init() -> Worker {
     let worker_manager = WorkerManager::new();
 
     worker_manager
-        .create_worker(WorkerSettings::default())
+        .create_worker(WorkerSettings {
+            enable_liburing: false,
+            ..WorkerSettings::default()
+        })
```
ibc, line 22: @nazar-pc, despite this works at intended (default settings are used but `enable_liburing` which is set to `false` instead), is this correct idiomatic syntax in Rust? 

In Node the default object must be added first, then those fields whose value must be different:

```ts
options = { ...options, foo: false }
```

It surprises me that in Rust it works if the changes are added first before the object with default values.

nazar-pc: It is, I believe `..` should be used at the very end, I don't think it compiles otherwise at all

ibc: Ok. For me idiomatic would be "here some values and after them those that override them" but it's ok. Rust guys have their own anti idiomatic concept of what idiomatic should mean.

worker/src/Settings.cpp
nazar-pc: I know we're not calling it incorrectly right now, but we could. And that would blow up instead of crashing with a nice message. Do not trust input, at least not to the degree that impacts memory safety.

ibc: Now it's safe and we don't assume anything. Arg values are now mandatory. See latest changes.

ibc:
Finally green. Merging!

PR #1455, opened 2024-09-28
Bring back RTP and RTCP notifications

### LINKED ISSUE #1454: DirectTransport consumer not receiving RTP packets in Rust
### Your environment

- Operating system: Pop!_OS 22.04 LTS
- clang version: 14.0.0
- gcc version: 11.4.0
- mediasoup version: v3 (rust crate version 0.17.1)

### Issue description

The `.on_rtp` callback for a Consumer created from a DirectTransport never gets called.

```rs
let direct_transport = self.router
    .create_direct_transport(DirectTransportOptions::default())
    .await
    .unwrap();

let options = ConsumerOptions::new(self.producer_id, self.rtp_capabilities);

let consumer = direct_transport
    .consume(options)
    .await
    .unwrap();

// "on_rtp" never gets logged
consumer.on_rtp(|_| info!("on_rtp")).detach();

self.consumers.push(consumer);
```
Looking at the source, it looks like the RTP notification has been commented out: https://github.com/versatica/mediasoup/blob/v3/rust/src/router/consumer.rs#L846.

nazar-pc:
Looks like they were forgotten during conversion, should fix https://github.com/versatica/mediasoup/issues/1454 (though I didn't test it).

@sw-vish LMK if it does what you expect it to do.

P.S. Lock file was not committed after last change, so I had to include it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1354,7 +1354,7 @@ checksum = "b6e8aaa3f231bb4bd57b84b2d5dc3ae7f350265df8aa96492e0bc394a1571909"
 
 [[package]]
 name = "mediasoup"
-version = "0.17.0"
+version = "0.17.1"
 dependencies = [
  "actix",
  "actix-web",
@@ -1391,7 +1391,7 @@ dependencies = [
 
 [[package]]
 name = "mediasoup-sys"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "planus",
  "planus-codegen",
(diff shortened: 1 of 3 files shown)


sw-vish:
Yep, that works as expected. Thanks!

ibc (approved):

ibc:
Let's merge!

jmillan (commented):
👌

PR #1463, opened 2024-11-08
Add TypeScript interfaces for all exported classes

ibc:
## Details

- `mediasoup.types` namespace now exports `Worker`, `Router`, `Producer`, etc **types** instead of classes, and those types are TypeScript interfaces.
- New `XxxxxTypes.ts` files export the `Xxxxx` interface and related types. For example, `ProducerTypes.ts` exports the `Producer` interface and Producer related types.
- `Xxxxx.ts` files export the implementation classes with name `XxxxxImpl`. For example, `Producer.ts` exports the `ProducerImpl` class that implements the `Producer` interface declared in `ProducerTypes.ts`.
- Use the new interfaces everywhere.
- Split `RtpParameters.ts` into `rtpParametersTypes.ts` and `rtpParametersFbsUtils.ts` and same for SCTP and SRTP files.
- Split `RtpStream.ts` into `rtpStreamStatsTypes.ts` and `rtpStreamStatsFbsUtils`.
- Rename `scalabilityModes.ts` tp `scalabilityModesTypes.ts`
- Use `import type` when possible instead of `import`.
- The idea is that all files ending in `Types.ts` only import types and do not depend on FBS generated TS code/types.
- Add new `transport.type` getter than returns `'webrtc' | 'plain' | 'pipe' | 'direct'`.
- Add new `rtpObserver.type` getter than returns `'activespeaker' | 'audiolevel'`.
- Remove `Logger` from `EnhancedEventEmitter`.
- This makes it easier for applications to create mediasoup mocks.
- Also fix real random bugs such as some `dump()` methods missing in same classes and wrong tests.


jmillan (approved):
I like it :+1:

piranna:
> * This makes it easier for applications to create mediasoup mocks.

Definitelly yes, that will be useful, just only now I have a bit of work to update Mafalda... Have you though / do you mind if we move the interfaces to a types-only package that can be used by both projects as dependency?

And related to this, do you know of any other Mediasoup mock/API project?

ibc:
> Have you though / do you mind if we move the interfaces to a types-only package that can be used by both projects as dependency?

This will probably happen but definitely not yet. Will be done once we use and evaluate how easy and useful those interfaces are, and probably more changes/additions are needed (spoiler: they are).

piranna:
Yeah, I know they are useful, I think I asked for that interfaces several years ego for selfish reasons 😈 Just curious if it was already in the roadmap and how much soonish, or if I can help on that. I need first to fix the compatibility with 3.14.15 (I have been taking self-care and focused on my day job during last months and got totally disconnected of Mediasoup since at least May...), but after that I would need to have the types package, just only to know if do it as a nightly package that extract and publish them the same I'm doing with the [tests](https://www.npmjs.com/package/@mafalda-sfu/mediasoup-node-tests) to check for compatibility and other components, or do it by hand and make Mediasoup as dependent of it.

TL;DR: if you need help on moving out the interfaces, you can count on me.

ibc:
We will reach that moment in which it will make sense to talk about a separate types package.

PR #1470, opened 2024-12-10
patch mediasoup (always draft)

umegaya:
- remove libnv dependency from some of files
  - TimerHandle, UnixStreamSocket
- export Probator mid
- fix meson.build to add artifact that qrpc requires
- add accessor to RTC::Parameters
- add accessor and necessary method for RTC::DtlsTransport
- compatibility fix for use SctpAssociation in qrpc code
  - add necessary callback to implement qrpc c++ client
  - now important methods does not receive dataConsumer directly, because most of these code does not require it
- compatibility fix for use Router in qrpc code
  - WebRtcServer in Router.hpp is only used for naem, so move header file into Router.cpp
- make DtlsTransport::remoteFingerprint acceible from outside
- can check existence of consumer that has specified id without throwing error


ibc (commented):
Wrong target

umegaya:
oh sorry... I think I create it in my own repo.

PR #1473, opened 2024-12-27
Fix some typo

aggresss:


worker/include/RTC/RTCP/XR.hpp
```diff
@@ -72,7 +72,7 @@ namespace RTC
 			CommonHeader* header{ nullptr };
 
 		private:
-			uint8_t raw[Packet::CommonHeaderSize] = { 0 };
+			uint8_t raw[CommonHeaderSize] = { 0 };
```
ibc, line 75: Right. Despite value is 4 in both cases, here we want to reference the common header size of a extended report block.

ibc (approved):

ibc:
Thanks.

PR #1481, opened 2025-01-26
Add dependency-descriptor support

Lynnworld:
Add dependency-descriptor extension support. Tested on mediasoup-demo.
Note: Now mediasoup-client can't get dd extension showup by default , so I have hardcode for testing. See :https://github.com/Lynnworld/mediasoup-client/blob/5e8ff720bfae116e62989eae984ad3ccf2b11cd9/src/handlers/ortc/utils.ts#L25

image

### DIFF
--- a/node/src/rtpParametersFbsUtils.ts
+++ b/node/src/rtpParametersFbsUtils.ts
@@ -473,7 +473,7 @@ export function rtpHeaderExtensionUriToFbs(
 		}
 
 		case 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension': {
-			return FbsRtpHeaderExtensionUri.DependencyDescriptor
+			return FbsRtpHeaderExtensionUri.DependencyDescriptor;
 		}
 
 		default: {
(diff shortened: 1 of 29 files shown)


ibc:
I really appreciate this effort but definitely we don't want to depend on libwebrtc for things purely related to RTP parsing. In mediasoup we do RTP parsing from scratch, and that includes everything related to RTP header extensions and codec payloads.

Another reason is that we want to get rid of libwebrtc (once we have our own BWE code).

Said that, I will create a ticket to do this and reference this PR as documentation:

https://github.com/versatica/mediasoup/issues/1483

jmillan:
Hi @Lynnworld,

I've started playing with a self made dependency descriptor implementation and I'm testing it with H264. I see that NOT every package contains the header, but the ones that have it look good.

Did you experience the same?. Do not every H264 RTP packet contain the header extension?

Lynnworld:
try latest chrome version.old  version like  <m
120 has  a bug  for this extension.

jmillan:
> try latest chrome version.old version like <m120 has a bug for this extension.

Indeed I'm using the latest chrome version, but not all RTP packets come with the dependency descriptor extension.

ibc:
There are some header extensions that are not sent in all packets by design, for example the "video rotation". I don't know if it makes sense for payload descriptor extension to not be present in every packet but maybe it does. What would it mean? No idea. Doesn't the spec say anything about it? Otherwise I recommend asking in an appropriate forum/mailing-list or directly to people who has already implemented it.

jmillan:
> There are some header extensions that are not sent in all packets by design, for example the "video rotation". I don't know if it makes sense for payload descriptor extension to not be present in every packet but maybe it does. What would it mean? No idea. Doesn't the spec say anything about it? Otherwise I recommend asking in an appropriate forum/mailing-list or directly to people who has already implemented it.

Yes, I'm learning about it on the go.

PR #1485, opened 2025-01-28
Node: Expose `Index` interface

### LINKED ISSUE #1476: Add `Mediasoup` type interface for module itself
## Feature Request
Add a type definition with the interface of what a Mediasoup-compatible object (an object with the same fields and static functions of the `mediasoup` export, if we consider the module exports as a singleton object), so we can easily identify what's a compatible one, at least at API level (for behaviour level we would need to also export the tests, like I have done in https://github.com/Mafalda-SFU/Mediasoup-node-tests).

ibc:
## Details

- Fixes #1476
- Expose `types.indexTypes` or via `import { Index as MediasoupIndex } from 'mediasoup/lib/indexTypes'` ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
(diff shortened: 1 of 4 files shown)


jmillan (approved):

ibc:
Released in 3.15.3

PR #1486, opened 2025-01-28
Fix VP9 out of order packets forwarding

vpalmisano:
Hi :wave: 
When I run different tests with 1 VP9 video publisher with network packet loss (10-15%) and delay (100-500ms) at sender side, I always measure a low video framerate (~3-4fps at average) at receiver side.

Looking at the VP9 Process method, I see that when we receive an out-of-order packet, we forward it without checking if it satisfies the current spatial and temporal layer values. As opposite, dropping those out of order packets in some cases improves the measured fps at receiver side.

This PR tries uses an intermediate approach, saving the spatial/temporal and pictureID relation in a list when the current layers change. When an old packet arrives, we search for the spatial/temporal layer for the packets pictureID (getting the layer that was active for the given pictureID) and drop the packet if its layer is greater than that. This way we are sure that we forward the packet only if it would have been forwarded did it arrive in order.
This prevents forwarding packets that would later be discarded by the decoder, and that would potentially generate over-processing, overflowing the jitter buffer and hence discarding legitimate packets, etc.

Tests results in the same conditions show an increase of the measured fps from ~3-4 to 7-8fps.

Ref:
similar PR for VP8: https://github.com/versatica/mediasoup/pull/1009

### DIFF
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -2,6 +2,8 @@
 #define MS_RTC_CODECS_PAYLOAD_DESCRIPTOR_HANDLER_HPP
 
 #include "common.hpp"
+#include "RTC/SeqManager.hpp"
+#include <deque>
 
 namespace RTC
 {
@@ -12,6 +14,61 @@ namespace RTC
 		{
 			virtual ~PayloadDescriptor() = default;
 			virtual void Dump() const    = 0;
+		};
+
+		class PictureIdList
+		{
+			static constexpr uint16_t MaxCurrentLayerPictureIdNum{ 1000u };
+
+		public:
+			explicit PictureIdList()
+			{
+			}
+			~PictureIdList()
+			{
+				this->list.clear();
+			}
+
+			void Push(uint16_t pictureId, int16_t layer)
+			{
+				for (const auto& it : this->list)
+				{
+					// Layers can be changed only with ordered pictureId values.
+					// If pictureId is lower than the previous one, then it has rolled over the max value.
+					uint16_t diff = pictureId > it.first
+					                  ? pictureId - it.first
+					                  : pictureId + RTC::SeqManager<uint16_t, 15>::MaxValue - it.first;
+
+					if (diff > MaxCurrentLayerPictureIdNum)
+					{
+						this->list.pop_front();
+					}
+					else
+					{
+						break;
+					}
+				}
+				this->list.push_back({ pictureId, layer });
+			}
+
+			int16_t GetLayer(uint16_t pictureId) const
+			{
+				if (this->list.size() > 1)
+				{
+					for (auto it = std::next(this->list.begin()); it != this->list.end(); ++it)
+					{
+						if (RTC::SeqManager<uint16_t, 15>::IsSeqHigherThan(it->first, pictureId))
+						{
+							return std::prev(it)->second;
+						}
+					}
+				}
+
+				return -1;
+			}
+
+		private:
+			std::deque<std::pair<uint16_t, int16_t>> list;
 		};
 
 		// Encoding context used by PayloadDescriptorHandler to properly rewrite the
@@ -87,6 +144,49 @@ namespace RTC
 			}
 			virtual void SyncRequired() = 0;
 
+			void SetCurrentSpatialLayer(int16_t spatialLayer, uint16_t pictureId)
+			{
+				if (this->currentSpatialLayer == spatialLayer)
+				{
+					return;
+				}
+
+				this->spatialLayerPictureIdList.Push(pictureId, spatialLayer);
+
+				this->currentSpatialLayer = spatialLayer;
+			}
+			void SetCurrentTemporalLayer(int16_t temporalLayer, uint16_t pictureId)
+			{
+				if (this->currentTemporalLayer == temporalLayer)
+				{
+					return;
+				}
+
+				this->temporalLayerPictureIdList.Push(pictureId, temporalLayer);
+
+				this->currentTemporalLayer = temporalLayer;
+			}
+			int16_t GetCurrentSpatialLayer(uint16_t pictureId) const
+			{
+				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
+				if (layer > -1)
+				{
+					return layer;
+				}
+
+				return this->currentSpatialLayer;
+			}
+			int16_t GetCurrentTemporalLayer(uint16_t pictureId) const
+			{
+				int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);
+				if (layer > -1)
+				{
+					return layer;
+				}
+
+				return this->currentTemporalLayer;
+			}
+
 		private:
 			Params params;
 			int16_t targetSpatialLayer{ -1 };
@@ -94,6 +194,10 @@ namespace RTC
 			int16_t currentSpatialLayer{ -1 };
 			int16_t currentTemporalLayer{ -1 };
 			bool ignoreDtx{ false };
+
+		private:
+			PictureIdList spatialLayerPictureIdList;
+			PictureIdList temporalLayerPictureIdList;
 		};
 
 		class PayloadDescriptorHandler
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -166,7 +166,7 @@ namespace RTC
 
 				this->currentTemporalLayer = temporalLayer;
 			}
-			int16_t GetCurrentSpatialLayer(uint16_t pictureId) const
+			int16_t GetSpatialLayerForPictureId(uint16_t pictureId) const
 			{
 				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
 				if (layer > -1)
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -155,7 +155,6 @@ namespace RTC
 				}
 
 				this->spatialLayerPictureIdList.Push(pictureId, spatialLayer);
-
 				this->currentSpatialLayer = spatialLayer;
 			}
 			void SetCurrentTemporalLayer(int16_t temporalLayer, uint16_t pictureId)
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -24,6 +24,7 @@ namespace RTC
 			explicit PictureIdList()
 			{
 			}
+
 			~PictureIdList()
 			{
 				this->list.clear();
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -182,6 +182,7 @@ namespace RTC
 			int16_t GetTemporalLayerForPictureId(uint16_t pictureId) const
 			{
 				int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);
+
 				if (layer > -1)
 				{
 					return layer;
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -60,7 +60,7 @@ namespace RTC
 					return -1;
 				}
 
-				for (auto it = std::next(this->list.begin()); it != this->list.end(); ++it)
+				for (auto it = std::next(this->layerChanges.begin()); it != this->layerChanges.end(); ++it)
 				{
 					if (RTC::SeqManager<uint16_t, 15>::IsSeqHigherThan(it->first, pictureId))
 					{
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -48,6 +48,7 @@ namespace RTC
 						break;
 					}
 				}
+
 				this->list.push_back({ pictureId, layer });
 			}
 
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -55,18 +55,18 @@ namespace RTC
 
 			int16_t GetLayer(uint16_t pictureId) const
 			{
-				if (this->list.size() > 1)
+				if (this->list.empty())
 				{
-					for (auto it = std::next(this->list.begin()); it != this->list.end(); ++it)
-					{
-						if (RTC::SeqManager<uint16_t, 15>::IsSeqHigherThan(it->first, pictureId))
-						{
-							return std::prev(it)->second;
-						}
-					}
+					return -1;
 				}
 
-				return -1;
+				for (auto it = std::next(this->list.begin()); it != this->list.end(); ++it)
+				{
+					if (RTC::SeqManager<uint16_t, 15>::IsSeqHigherThan(it->first, pictureId))
+					{
+						return std::prev(it)->second;
+					}
+				}
 			}
 
 		private:
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -27,12 +27,12 @@ namespace RTC
 
 			~PictureIdList()
 			{
-				this->list.clear();
+				this->layerChanges.clear();
 			}
 
 			void Push(uint16_t pictureId, int16_t layer)
 			{
-				for (const auto& it : this->list)
+				for (const auto& it : this->layerChanges)
 				{
 					// Layers can be changed only with ordered pictureId values.
 					// If pictureId is lower than the previous one, then it has rolled over the max value.
@@ -42,7 +42,7 @@ namespace RTC
 
 					if (diff > MaxCurrentLayerPictureIdNum)
 					{
-						this->list.pop_front();
+						this->layerChanges.pop_front();
 					}
 					else
 					{
@@ -50,12 +50,12 @@ namespace RTC
 					}
 				}
 
-				this->list.push_back({ pictureId, layer });
+				this->layerChanges.push_back({ pictureId, layer });
 			}
 
 			int16_t GetLayer(uint16_t pictureId) const
 			{
-				if (this->list.size() <= 1)
+				if (this->layerChanges.size() <= 1)
 				{
 					return -1;
 				}
@@ -72,7 +72,9 @@ namespace RTC
 			}
 
 		private:
-			std::deque<std::pair<uint16_t, int16_t>> list;
+			// List populated with the spatial/temporal layer changes
+			// indexed by the corresponding pictureId.
+			std::deque<std::pair<uint16_t, int16_t>> layerChanges;
 		};
 
 		// Encoding context used by PayloadDescriptorHandler to properly rewrite the
@@ -199,6 +201,7 @@ namespace RTC
 			bool ignoreDtx{ false };
 
 		private:
+			// List of spatial/temporal layer changes indexed by the corresponding pictureId.
 			PictureIdList spatialLayerPictureIdList;
 			PictureIdList temporalLayerPictureIdList;
 		};
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -171,6 +171,7 @@ namespace RTC
 			int16_t GetSpatialLayerForPictureId(uint16_t pictureId) const
 			{
 				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
+
 				if (layer > -1)
 				{
 					return layer;
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -147,7 +147,6 @@ namespace RTC
 				this->ignoreDtx = ignoreDtx;
 			}
 			virtual void SyncRequired() = 0;
-
 			void SetCurrentSpatialLayer(int16_t spatialLayer, uint16_t pictureId)
 			{
 				if (this->currentSpatialLayer == spatialLayer)
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -165,7 +165,6 @@ namespace RTC
 				}
 
 				this->temporalLayerPictureIdList.Push(pictureId, temporalLayer);
-
 				this->currentTemporalLayer = temporalLayer;
 			}
 			int16_t GetSpatialLayerForPictureId(uint16_t pictureId) const
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -55,7 +55,7 @@ namespace RTC
 
 			int16_t GetLayer(uint16_t pictureId) const
 			{
-				if (this->list.empty())
+				if (this->list.size() <= 1)
 				{
 					return -1;
 				}
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -67,6 +67,8 @@ namespace RTC
 						return std::prev(it)->second;
 					}
 				}
+
+				return -1;
 			}
 
 		private:
--- a/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
+++ b/worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
@@ -176,7 +176,7 @@ namespace RTC
 
 				return this->currentSpatialLayer;
 			}
-			int16_t GetCurrentTemporalLayer(uint16_t pictureId) const
+			int16_t GetTemporalLayerForPictureId(uint16_t pictureId) const
 			{
 				int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);
 				if (layer > -1)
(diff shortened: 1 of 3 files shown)


worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -10,11 +10,11 @@ constexpr uint16_t MaxPictureId = (1 << 15) - 1;
 Codecs::VP9::PayloadDescriptor* CreateVP9Packet(
   uint8_t* buffer, size_t bufferLen, uint16_t pictureId, uint8_t tlIndex)
 {
-	buffer[0]             = 0xAD; // I and L bits
+	buffer[0]             = 0xAD; // I, L, B, E bits
 	uint16_t netPictureId = htons(pictureId);
 	std::memcpy(buffer + 1, &netPictureId, 2);
 	buffer[1] |= 0x80;
-	buffer[3] = tlIndex << 6;
+	buffer[3] = (tlIndex << 5) | (1 << 4); // tlIndex, switchingUpPoint
```
vpalmisano, line 17: The switchingUpPoint bit is 1, this way context.SetTargetTemporalLayer() will actually change the currentTemporaLayer value too.

ibc (commented):
Thanks. Please give us some days before we review (terribly busy these days). And will also review the VP8 related PR.

jmillan:
@vpalmisano, in #1009 we simply discard packets which temporal layer is higher than the current one. Why don't we follow the same approach here? I seems way more simple.

If the current spatial layer is N, then we know that the receiver is already ready to receive it, and receiving an N+1 old packet won't have any benefit as it will be dropped by the decoder.

vpalmisano:
> @vpalmisano, in #1009 we simply discard packets which temporal layer is higher than the current one. Why don't we follow the same approach here? I seems way more simple.
> 
> If the current spatial layer is N, then we know that the receiver is already ready to receive it, and receiving an N+1 old packet won't have any benefit as it will be dropped by the decoder.

Ideally yes, but the receiver could use the jitterbuffer the keep the latest N packets and in this case it would be better to keep receiving the correct layers instead of immediately switch the to the new layer.

jmillan:
> Ideally yes, but the receiver could use the jitterbuffer the keep the latest N packets and in this case it would be better to keep receiving the correct layers instead of immediately switch the to the new layer.

But since we are only changing layers with key frames, loss of previous packets won't generate issues because corresponding keyframe for the new layer is already in the jitterbuffer too. Ie:

The jitterBuffer in the client is as follows:
```ts
| layer-1 (seq-4) "keyframe (ksvc) or end of frame" |
| layer-2 (seq-2) |
| layer-2 (seq-1) |
```

The packet related to "layer-2 (seq-3)" is missing BUT "layer-1 (seq-4)" is a keyframe or an end of a frame, hence not receiving (seq-3) should not have a negative effect.

Can you please @vpalmisano try with this much simple approach and share the insigths?

vpalmisano:
> But since we are only changing layers with key frames, loss of previous packets won't generate issues because corresponding keyframe for the new layer is already in the jitterbuffer too. Ie:

Consider this scenario:
```
-> currentLayer=S1
pictureID_1: S1 -> forward
pictureID_1: S2 -> drop
pictureID_3: S1 -> forward
-> currentLayer=S2
pictureID_4: S2 -> forward
pictureID_2: S1 -> ?
pictureID_2: S2 -> ?
```
Here we can avoid dropping pictureID_2 and we can forward S1 layer, because if was the current selected layer since pictureID_1. If we drop pictureID-2 we will create an hole in the jitterbuffer, right?

jmillan:
> Here we can avoid dropping pictureID_2 and we can forward S1 layer, because if was the current selected layer since pictureID_1. If we drop pictureID-2 we will create an hole in the jitterbuffer, right?

We are talking about filtering layers greater than the current one, not the lower ones, hence both packets would be delivered in this case, right?

vpalmisano:
> We are talking about filtering layers greater than the current one, not the lower ones, hence both packets would be delivered in this case, right?

With the current code, old packets will be delivered at both S1 and S2 layers.
After the changes, only S1 will be delivered (I'm assuming that pictureID_2 is not a keyframe).

jmillan:
> Looking at the VP9 Process method, I see that when we receive an out-of-order packet, we forward it without checking if it satisfies the current spatial and temporal layer values. As opposite, dropping those out of order packets in some cases improves the measured fps at receiver side.

I fail to see why sending old packets would make the reception fps drop. Do you know a reason (in the codec spec or in the browser implementation) for this to happen?

vpalmisano:
> I fail to see why sending old packets would make the reception fps drop. Do you know a reason (in the codec spec or in the browser implementation) for this to happen?

I have only some hypothesis:
1. There is an high probability that the packet at S0 layer is both a start and an end of picture packet, so the decoder will prefer to decode it discarding the higher layer packets instead of waiting for them to form a complete picture.
2. Putting uncompleted pictures in the decoder buffer probably causes more RTX and PLIs requests.

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -75,4 +75,110 @@ SCENARIO("process VP9 payload descriptor", "[codecs][vp9]")
...
+			{ start + 2, 0, -1, true },
+			{ start + 2, 1, -1, false },
+			// targetTemporalLayer=1
+			{ start + 10, 0, 1, true },
+			{ start + 10, 1, -1, true },
+			{ start + 11, 0, -1, true },
+			{ start + 11, 1, -1, true },
+
+			{ start + 3, 0, -1, true }, // old packet
+			{ start + 3, 1, -1, false },
```
jmillan, line 139: @vpalmisano, While reading the test I realise I'm having a little hard time trying to understand the overall mechanism you're adding in this PR.

Can you please explain in detail what is the code doing? I may be a bit foggy as it's almost end of the week.

jmillan: Friendly ping @vpalmisano just in case you missed the comment.

vpalmisano: Sorry for the delay :(

Basically the new code saves the spatial/temporal layers changes and the pictureId when this happens. When we receive and out-of-order packet, we decide to forward it or not after looking for the previous pictureId in the table; this represents the desired layer value that we need to forward:

| pictureId | layer |
| 0 | 0 | <--- layer=0 should be forwarded for pictureIds from 0 to 9 
| 10 | 1 | <--- layer=1 for pictureIds=[10, 99]
| 100 | 0 |

In the test:
- pictureId=start + 3 
- temporalLayer=1
- new temporal layer value (-1 means that it doesn't change)
- if packet should be forwarded

here pictureID=`start + 3` arrived late, and the targetTemporalLayer was set to 0 from `start` up to `start + 9`, so the correct layer to be forwarded is 0 and not 1

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +144,60 @@ namespace RTC
...
+				if (this->currentTemporalLayer == temporalLayer)
+				{
+					return;
+				}
+
+				this->temporalLayerPictureIdList.Push(pictureId, temporalLayer);
+
+				this->currentTemporalLayer = temporalLayer;
+			}
+			int16_t GetCurrentSpatialLayer(uint16_t pictureId) const
```
jmillan:
```suggestion
			int16_t GetSpatialLayerForPictureId(uint16_t pictureId) const
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +144,60 @@ namespace RTC
...
+			{
+				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
+				if (layer > -1)
+				{
+					return layer;
+				}
+
+				return this->currentSpatialLayer;
+			}
+			int16_t GetCurrentTemporalLayer(uint16_t pictureId) const
```
jmillan:
```suggestion
			int16_t GetTemporalLayerForPictureId(uint16_t pictureId) const
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -242,78 +242,82 @@ namespace RTC
...
 					}
 				}
 			}
 
-			// Unless old packet filter spatial layers that are either
+			// Filter spatial layers that are either
 			// * higher than current one
 			// * different than the current one when KSVC is enabled and this is not a keyframe
 			// (interframe p bit = 1)
+			uint16_t tmpSpatialLayerCheck =
```
jmillan:
```suggestion
			uint16_t spatialLayerForPictureId =
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -369,12 +373,15 @@ namespace RTC
...
 				}
+			}
 
-				// Filter temporal layers higher than current one.
-				if (packetTemporalLayer > tmpTemporalLayer)
-				{
-					return false;
-				}
+			// Filter temporal layers higher than current one.
+			uint16_t tmpTemporalLayerCheck =
```
jmillan:
```suggestion
			uint16_t temporalLayerForPictureId =
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -242,78 +242,82 @@ namespace RTC
 			// clang-format on
 
 			// Upgrade current spatial layer if needed.
```
jmillan:
```suggestion
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -242,78 +242,82 @@ namespace RTC
...
-					tmpTemporalLayer = 0; // Just in case.
-				}
-			}
-			// Downgrade current spatial layer if needed.
-			else if (context->GetTargetSpatialLayer() < context->GetCurrentSpatialLayer())
+			if (!isOldPacket)
 			{
-				// In K-SVC we must wait for a keyframe.
-				if (context->IsKSvc())
+				if (context->GetTargetSpatialLayer() > context->GetCurrentSpatialLayer())
```
jmillan, line 247:
```suggestion
                                 // Upgrade current spatial layer if needed.
				if (context->GetTargetSpatialLayer() > context->GetCurrentSpatialLayer())
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -242,78 +242,82 @@ namespace RTC
...
-						  "downgrading tmpSpatialLayer from %" PRIu16 " to %" PRIu16 " (packet:%" PRIu8
-						  ":%" PRIu8 ") after keyframe (K-SVC)",
+						  "upgrading tmpSpatialLayer from %" PRIu16 " to %" PRIu16 " (packet:%" PRIu8 ":%" PRIu8
+						  ") old:%d",
 						  context->GetCurrentSpatialLayer(),
 						  context->GetTargetSpatialLayer(),
 						  packetSpatialLayer,
-						  packetTemporalLayer);
+						  packetTemporalLayer,
+						  isOldPacket);
```
jmillan: Not needed, this will always be `false`.

worker/src/RTC/Codecs/VP9.cpp
jmillan:
```c++
// Check and handle spatial layer (unless old packet).
```

worker/src/RTC/Codecs/VP9.cpp
jmillan: Please move this bock above the previous one ("// Check and handle spatial layer (unless old packet).") so we can follow the pattern of early returning.

worker/src/RTC/Codecs/VP9.cpp
jmillan: Please move this bock above the previous one ("// Check and handle temporal layer (unless old packet).") so we can follow the pattern of early returning.

jmillan (changes requested):
Some comments made. The idea looks fine @vpalmisano 👌 

I'll re-review after the requested changes are done

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,61 @@ namespace RTC
...
+					if (diff > MaxCurrentLayerPictureIdNum)
+					{
+						this->list.pop_front();
+					}
+					else
+					{
+						break;
+					}
+				}
+				this->list.push_back({ pictureId, layer });
```
ibc:
```suggestion
				}

				this->list.push_back({ pictureId, layer });
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,61 @@ namespace RTC
...
+			static constexpr uint16_t MaxCurrentLayerPictureIdNum{ 1000u };
+
+		public:
+			explicit PictureIdList()
+			{
+			}
+			~PictureIdList()
+			{
+				this->list.clear();
+			}
```
ibc:
```suggestion
			explicit PictureIdList()
			{
			}

			~PictureIdList()
			{
				this->list.clear();
			}
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
ibc:
```suggestion
			}

			int16_t GetCurrentSpatialLayer(uint16_t pictureId) const
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
ibc:
```suggestion
			}

			int16_t GetCurrentTemporalLayer(uint16_t pictureId) const
```

ibc (commented):
Some cosmetic changes requested.

jmillan:
@vpalmisano, for future reading of this PR, could you please update the PR summary block:

```txt
This PR tries to use a intermediate approach, saving the spatial/temporal layer changes with the corresponding pictureID in a list. When an old packet arrives, we decide to drop it or not searching the corresponding spatial/temporal layer value, this way we are sure that we forward the right layer as requested since that pictureID.
```

by:

```txt
This PR tries uses an intermediate approach, saving the spatial/temporal and pictureID relation in a list when the current layers change. When an old packet arrives, we search for the spatial/temporal layer for the packets pictureID (getting the layer that was active for the given pictureID) and drop the packet if its layer is greater than that. This way we are sure that we forward the packet only if it would have been forwarded did it arrive in order.

This prevents forwarding packets that would later be discarded by the decoder, and that would potentially generate over-processing, overflowing the jitter buffer and hence discarding legitimate packets, etc.
```

worker/src/RTC/Codecs/VP9.cpp
vpalmisano: But here we need to use the `tmpSpatialLayer` set in the previous block(`tmpSpatialLayer  = context->GetTargetSpatialLayer();`), so probably we cannot move it before.

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,63 @@ namespace RTC
...
+					for (auto it = std::next(this->list.begin()); it != this->list.end(); ++it)
+					{
+						if (RTC::SeqManager<uint16_t, 15>::IsSeqHigherThan(it->first, pictureId))
+						{
+							return std::prev(it)->second;
+						}
+					}
+				}
+
+				return -1;
```
jmillan, line 71: Please early return if the conditions are not met, for readability.

jmillan (approved):

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,65 @@ namespace RTC
 			virtual void Dump() const    = 0;
 		};
 
+		class PictureIdList
+		{
+			static constexpr uint16_t MaxCurrentLayerPictureIdNum{ 1000u };
```
jmillan, line 21: @vpalmisano I think a much lower value here is possible: ie 64 or 128. It'ts 64 or 128 changes or layers...

vpalmisano: But here we use that value to remove older pictureId values, it doesn't represent the number of layer changes.

jmillan: yep, but we just add new pictureId values when changing layers, don't we?

vpalmisano: Yes but MaxCurrentLayerPictureIdNum is used to evaluate the distance in pictureId units between the last and the first item in the list.

jmillan: The value `MaxCurrentLayerPictureIdNum` merely indicates how big the `std::deque` can be. We can make it smaller. I think 128 would be a good value, even 64, as older picture IDs will anyway be too old to consider.

vpalmisano: If we want to remove the older items using a counter we need to change the mechanism. Consider that after 64 or 128 layer changes the queue will hold 128 items even if they are very old.

jmillan: But why a 'counter'? My only concern is, can we make `MaxCurrentLayerPictureIdNum` smaller than the current `1000`? I mean just decrease that value.

vpalmisano: It we have a layer change every 1000 packets, we will hold only 1 item in the queue and it would be not useful.

jmillan: Current `MaxCurrentLayerPictureIdNum` is 1000. If we have a layer change every 1000 packets we will keep pictureId <-> layer relation for the last  `MaxCurrentLayerPictureIdNum` (1000) changes. This is, we will know if we must relay or drop the last 1M packets.

If we set `MaxCurrentLayerPictureIdNum` to 64 and we have a layer change every 1000 packets we will keep pictureId <-> layer relation for the last `MaxCurrentLayerPictureIdNum` (64) changes. This is, we will know if we must relay or drop the last  64000 packets.

That is correct, right?

If so, I consider that being able to relay or drop the last 1M packets is not needed, and a lower value of `MaxCurrentLayerPictureIdNum` would suffice.

vpalmisano: Why?
Every time we have a layer change at `pictureId` we put a new item in the queue and we remove the last one if the difference is larger than 1000 packets:

```
					uint16_t diff = pictureId - it.first
					if (diff > MaxCurrentLayerPictureIdNum)
					{
						this->list.pop_front();
					}
```
With a layer change every 1000 packets we will keep only 1 item in the queue. Right?

jmillan: > With a layer change every 1000 packets we will keep only 1 item in the queue. Right?

Correct, yes. So unless there is a layer change for every new pictureID, we will never get 1000 items in the queue...

A value of 1000 here doesn't seem that big as it does not mean there will be those many items in the queue. 

👌

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,65 @@ namespace RTC
...
+					else
+					{
+						break;
+					}
+				}
+
+				this->list.push_back({ pictureId, layer });
+			}
+
+			int16_t GetLayer(uint16_t pictureId) const
```
ibc, line 56: What does this `GetLayer()` method return? An spatial layer? A temporal layer? Or what? If a spatial or temporal layer then let's please make it explicit in the method name.

Ok ok, I see that this is `PictureIdList` class that later is used as follows:
```c++
int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
//
int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);
```

so it's ok.

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -14,6 +16,65 @@ namespace RTC
...
+					{
+						return std::prev(it)->second;
+					}
+				}
+
+				return -1;
+			}
+
+		private:
+			std::deque<std::pair<uint16_t, int16_t>> list;
```
ibc: Can we please add a comment here explaining what this `list` container contains and what its key is? And can we (probably) improve its name? Because `list` is super generic.

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
 			}
 			virtual void SyncRequired() = 0;
 
+			void SetCurrentSpatialLayer(int16_t spatialLayer, uint16_t pictureId)
```
ibc:
```suggestion
			virtual void SyncRequired() = 0;
			void SetCurrentSpatialLayer(int16_t spatialLayer, uint16_t pictureId)
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
...
 			Params params;
 			int16_t targetSpatialLayer{ -1 };
 			int16_t targetTemporalLayer{ -1 };
 			int16_t currentSpatialLayer{ -1 };
 			int16_t currentTemporalLayer{ -1 };
 			bool ignoreDtx{ false };
+
+		private:
+			PictureIdList spatialLayerPictureIdList;
+			PictureIdList temporalLayerPictureIdList;
```
ibc, line 206: Can we have comments here?

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
...
+			void SetCurrentSpatialLayer(int16_t spatialLayer, uint16_t pictureId)
+			{
+				if (this->currentSpatialLayer == spatialLayer)
+				{
+					return;
+				}
+
+				this->spatialLayerPictureIdList.Push(pictureId, spatialLayer);
+
+				this->currentSpatialLayer = spatialLayer;
```
ibc:
```suggestion
				this->spatialLayerPictureIdList.Push(pictureId, spatialLayer);
				this->currentSpatialLayer = spatialLayer;
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
...
+			void SetCurrentTemporalLayer(int16_t temporalLayer, uint16_t pictureId)
+			{
+				if (this->currentTemporalLayer == temporalLayer)
+				{
+					return;
+				}
+
+				this->temporalLayerPictureIdList.Push(pictureId, temporalLayer);
+
+				this->currentTemporalLayer = temporalLayer;
```
ibc:
```suggestion
				this->temporalLayerPictureIdList.Push(pictureId, temporalLayer);
				this->currentTemporalLayer = temporalLayer;
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
...
+			}
+			int16_t GetSpatialLayerForPictureId(uint16_t pictureId) const
+			{
+				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);
+				if (layer > -1)
```
ibc:
```suggestion
				int16_t layer = this->spatialLayerPictureIdList.GetLayer(pictureId);

				if (layer > -1)
```

worker/include/RTC/Codecs/PayloadDescriptorHandler.hpp
```diff
@@ -87,13 +148,60 @@ namespace RTC
...
+				{
+					return layer;
+				}
+
+				return this->currentSpatialLayer;
+			}
+			int16_t GetTemporalLayerForPictureId(uint16_t pictureId) const
+			{
+				int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);
+				if (layer > -1)
```
ibc:
```suggestion
				int16_t layer = this->temporalLayerPictureIdList.GetLayer(pictureId);

				if (layer > -1)
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -279,41 +260,63 @@ namespace RTC
...
-			// Unless old packet filter spatial layers that are either
+			// Filter spatial layers that are either
 			// * higher than current one
 			// * different than the current one when KSVC is enabled and this is not a keyframe
 			// (interframe p bit = 1)
+			uint16_t spatialLayerForPictureId =
+			  isOldPacket ? context->GetSpatialLayerForPictureId(this->payloadDescriptor->pictureId)
+			              : tmpSpatialLayer;
 			// clang-format off
 			if (
```
ibc:
```suggestion
			uint16_t spatialLayerForPictureId =
			  isOldPacket ? context->GetSpatialLayerForPictureId(this->payloadDescriptor->pictureId)
			              : tmpSpatialLayer;

			// clang-format off
			if (
```

worker/src/RTC/Codecs/VP9.cpp
```diff
@@ -369,12 +372,15 @@ namespace RTC
...
-				// Filter temporal layers higher than current one.
-				if (packetTemporalLayer > tmpTemporalLayer)
-				{
-					return false;
-				}
+			// Filter temporal layers higher than current one.
+			uint16_t temporalLayerForPictureId =
+			  isOldPacket ? context->GetTemporalLayerForPictureId(this->payloadDescriptor->pictureId)
+			              : tmpTemporalLayer;
+			if (packetTemporalLayer > temporalLayerForPictureId)
```
ibc: Please let's add a blank line after variable declarations and before and after condition blocks.
```suggestion
			// Filter temporal layers higher than current one.
			uint16_t temporalLayerForPictureId =
			  isOldPacket ? context->GetTemporalLayerForPictureId(this->payloadDescriptor->pictureId)
			              : tmpTemporalLayer;

			if (packetTemporalLayer > temporalLayerForPictureId)
```

worker/test/src/RTC/Codecs/TestVP9.cpp
```diff
@@ -75,4 +75,111 @@ SCENARIO("process VP9 payload descriptor", "[codecs][vp9]")
...
+			{ start + 1, 1, -1, false },
+			{ start + 2, 0, -1, true },
+			{ start + 2, 1, -1, false },
+			// targetTemporalLayer=1
+			{ start + 10, 0, 1, true },
+			{ start + 10, 1, -1, true },
+			{ start + 11, 0, -1, true },
+			{ start + 11, 1, -1, true },
+
+			{ start + 3, 0, -1, true }, // old packet
```
ibc: This blank line doesn't make sense IMHO.
```suggestion
			{ start + 11, 1, -1, true },
			{ start + 3, 0, -1, true }, // old packet
```

ibc:
Thanks @vpalmisano, merging!

PR #1492, opened 2025-02-20
Fix the problem of the TCC package being omitted from being sent

penguinol:
`FeedbackRtpTransportPacket::IsFull()` does not check whether the packet is smaller than MTU, so we need to handle `AddPacketResult::MAX_SIZE_EXCEEDED`.
Also make sure the tcc feedback packet smaller than `MTU - IPHeader - TCPHeader - SRTP trailer`.


jmillan (approved):

worker/src/RTC/TransportCongestionControlServer.cpp
```diff
@@ -233,14 +233,20 @@ namespace RTC
 
 				case RTC::RTCP::FeedbackRtpTransportPacket::AddPacketResult::MAX_SIZE_EXCEEDED:
 				{
-					// This should not happen.
-					MS_WARN_TAG(rtcp, "transport-cc feedback packet is exceeded");
+					// Send ongoing feedback packet
```
ibc:
```suggestion
					// Send ongoing feedback packet.
```

worker/include/RTC/RtpPacket.hpp
```diff
@@ -18,6 +18,14 @@ namespace RTC
...
 	// Max MTU size.
 	constexpr size_t MtuSize{ 1500u };
+	// Maximum size for a CompundPacket.
+	// * IPv4|Ipv6 header size: 20|40 bytes. IPv6 considered.
+	// * UDP|TCP header size:   8|20  bytes. TCP considered.
+	// * SRTP Encryption: 148 bytes.
+	//    SRTP_MAX_TRAILER_LEN+4 is the maximum number of octects that will be
+	//    added to an RTCP packet by srtp_protect_rtcp().
+	//    srtp.h: SRTP_MAX_TRAILER_LEN (SRTP_MAX_TAG_LEN + SRTP_MAX_MKI_LEN)
+	constexpr size_t MaxPacketSize{ RTC::MtuSize - 40 - 20 - 148u };
```
ibc, line 28: We should move this to a different file because this is `RtpPacket.hpp` and this `RTC::MaxPacketSize` is also being used in RTCP `CompoundPacket.cpp` which is not a RTP packet.

Can we have separate `MaxPacketSize` in both `RtpPacket.hpp` and `CompoundPacket.hpp` please? NOTE: I know that the problem is that the namespace would conflict since it's `RTC::MaxPacketSize` rather than `RTC::RtpPacket::MaxPacketSize`. Not sure how to solve it.

Ok, ignore this and I will fix later.

ibc (approved):

ibc:
Thanks a lot, @penguinol! Merging.

ibc:
Hi @penguinol, not sure if related to this PR, but here a regression: https://github.com/versatica/mediasoup/issues/1522

I'll investigate when possible.

PR #1494, opened 2025-02-24
Consumer: Fix sequence number gap

### LINKED ISSUE #1069: Large packet loss when estimated bandwidth is close to the lowest SVC layer

jmillan:
Drop the given sequence number if the RTP packet cannot be forwarded.

Fixes  #1069 

By not doing so, we are incorrectly generating non legitimate sequence number jumps which involves:

- The other end requesting NACKs for packets that do not exist, and us processing them.
- The other end waiting for those packets that will never arrive.
- Possible implications an jitter buffer and decoding level.

This problem is very apparent when using BWE and the available BW is low enough to discard a lot of packets, or when the outgoing bitrate is manually set low enough too.

### DIFF
--- a/worker/src/Channel/ChannelSocket.cpp
+++ b/worker/src/Channel/ChannelSocket.cpp
@@ -30,10 +30,12 @@ namespace Channel
 
 	/* Instance methods. */
 
+#ifdef MS_TEST
 	ChannelSocket::ChannelSocket()
 	{
 		MS_TRACE_STD();
 	}
+#endif
 
 	ChannelSocket::ChannelSocket(int consumerFd, int producerFd)
 	  : consumerSocket(new ConsumerSocket(consumerFd, MessageMaxLen, this)),
@@ -264,10 +266,6 @@ namespace Channel
 		else if (this->producerSocket)
 		{
 			this->producerSocket->Write(payload, payloadLen);
-		}
-		else
-		{
-			MS_DEBUG_DEV("sending Channel message: %s", payload);
 		}
 	}
 
--- a/worker/src/Channel/ChannelSocket.cpp
+++ b/worker/src/Channel/ChannelSocket.cpp
@@ -30,6 +30,11 @@ namespace Channel
 
 	/* Instance methods. */
 
+	ChannelSocket::ChannelSocket()
+	{
+		MS_TRACE_STD();
+	}
+
 	ChannelSocket::ChannelSocket(int consumerFd, int producerFd)
 	  : consumerSocket(new ConsumerSocket(consumerFd, MessageMaxLen, this)),
 	    producerSocket(new ProducerSocket(producerFd, MessageMaxLen))
@@ -256,9 +261,13 @@ namespace Channel
 		{
 			this->channelWriteFn(payload, payloadLen, this->channelWriteCtx);
 		}
-		else
+		else if (this->producerSocket)
 		{
 			this->producerSocket->Write(payload, payloadLen);
+		}
+		else
+		{
+			MS_DEBUG_DEV("sending Channel message: %s", payload);
 		}
 	}
 
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -736,6 +736,8 @@ namespace RTC
 			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::CONSUMER_INACTIVE);
 #endif
 
+			this->rtpSeqManager->Drop(packet->GetSequenceNumber());
+
 			return;
 		}
 
@@ -744,6 +746,8 @@ namespace RTC
 #ifdef MS_RTC_LOGGER_RTP
 			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::INVALID_TARGET_LAYER);
 #endif
+
+			this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 
 			return;
 		}
@@ -777,6 +781,7 @@ namespace RTC
 				packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::NOT_A_KEYFRAME);
 #endif
 
+				this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 				return;
 			}
 
@@ -804,6 +809,7 @@ namespace RTC
 			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::NOT_A_KEYFRAME);
 #endif
 
+			this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 			return;
 		}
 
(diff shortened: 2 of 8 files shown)


worker/src/Channel/ChannelSocket.cpp
```diff
@@ -256,10 +261,14 @@ namespace Channel
...
 			this->channelWriteFn(payload, payloadLen, this->channelWriteCtx);
 		}
-		else
+		else if (this->producerSocket)
 		{
 			this->producerSocket->Write(payload, payloadLen);
 		}
+		else
+		{
+			MS_DEBUG_DEV("sending Channel message: %s", payload);
```
ibc: What is the purpose of this log that says something that in fact is not happening?

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -777,6 +781,7 @@ namespace RTC
 				packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::NOT_A_KEYFRAME);
 #endif
 
+				this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 				return;
```
```suggestion
				this->rtpSeqManager->Drop(packet->GetSequenceNumber());

				return;
```

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -804,19 +809,20 @@ namespace RTC
 			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::NOT_A_KEYFRAME);
 #endif
 
+			this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 			return;
```
ibc, line 814:
```suggestion
			this->rtpSeqManager->Drop(packet->GetSequenceNumber());

			return;
```

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -935,6 +941,7 @@ namespace RTC
 					packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::TOO_HIGH_TIMESTAMP_EXTRA_NEEDED);
 #endif
 
+					this->rtpSeqManager->Drop(packet->GetSequenceNumber());
 					return;
```
```suggestion
					this->rtpSeqManager->Drop(packet->GetSequenceNumber());

					return;
```

worker/test/src/RTC/TestSimpleConsumer.cpp
```diff
@@ -0,0 +1,283 @@
...
+        0x00, 0x00, 0x00, 0x05
+    };
+	// clang-format on
+
+	SECTION("RTP packets are not forwarded when the consumer is not active")
+	{
+		auto* listener = new ConsumerListener();
+		auto* consumer = CreateConsumer(listener);
+
+		auto* rtpStream = CreateRtpStreamRecv();
```
ibc:
```suggestion
		auto* listener  = new ConsumerListener();
		auto* consumer  = CreateConsumer(listener);
		auto* rtpStream = CreateRtpStreamRecv();
```

ibc (commented):
I'm amazed that this fix looks too obvious to never have realized of it. I mean, we were literally dropping packets in MANY SCENARIOS without dropping then in SeqManager. How is possible that we never noticed it?

Please add entry in CHANGELOG.

Also minimal cosmetic changes.

And ASAN checks are failing in CI.

ibc:
> Fixes #1069

Fixes #1069 or fixes #1070 or both?

ibc:
I'm updating the branch with pending changes in v3.

ibc:
> And ASAN checks are failing in CI.

All those are caused by the new `TestSimpleConsumer.cpp` file.

worker/src/Channel/ChannelSocket.cpp
jmillan: I've created a new `ChannelSocket` constructor that does not received callbacks (RUST) nor file descriptors (Node). Just for testing purposes. So when sending something out it will just log it, nothing else.

ibc: That's what I meant: why is it logging that it is sending a message if it's not doing it?

worker/test/src/RTC/TestSimpleConsumer.cpp
```diff
@@ -10,7 +10,11 @@
 
 using namespace RTC;
 
-const uint8_t PayloadType = 111;
+const uint8_t PayloadType       = 111;
+auto* channelMessageRegistrator = new ChannelMessageRegistrator();
```
ibc, line 14: const?

worker/test/src/RTC/TestSimpleConsumer.cpp
```diff
@@ -10,7 +10,11 @@
...
+auto* channelSocket             = new Channel::ChannelSocket();
+auto* channelNotifier           = new Channel::ChannelNotifier(channelSocket);
+auto shared                     = Shared(channelMessageRegistrator, channelNotifier);
```
ibc, line 17: const?

ibc (approved):

worker/src/Channel/ChannelSocket.cpp
jmillan: It's like a mock. Shall I change the text to "TEST -- sending Channel message..." or alike?

Give me some time and I'll deal with it later with a better approach.

ibc: We have condition blocks above in the same if else section in which we do send a message but don't log anything and here we are adding a block that doesn't send a message and logs that it has sent a message. I would just remove that log and that's all.

Said that, I don't remember if we have some macro/define that just exists during worker tests but if so, such a new constructor could be wrapped into a macro condition that checks whether such a define exists.

jmillan: Fixed in the lat commit. Such constructor is only compiled if `MS_TEST` define is enabled.

worker/test/src/RTC/TestSimpleConsumer.cpp
jmillan: Nop, as this is provided to `SimpleConsumer` which expects a non const `Shared`. We can probably change that and pass a const `Shared` instead.

Let's please use another PR to do so.

ibc (approved):

PR #1500, opened 2025-03-07
CI: Remove deprecated ubuntu-20.04 host and add windows-2025, ubuntu-22.04-arm and ubuntu-24.04-arm

ibc:
### Details

- https://github.com/actions/runner-images/issues/11101
- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -56,11 +56,13 @@ jobs:
             node: 22
             cc: gcc
             cxx: g++
-          - os: ubuntu-24.04-arm
-            node: 22
-            cc: gcc
-            cxx: g++
-            meson_args: '-Db_sanitize=address'
+          # TOOD: Uncomment when this issue is fixed:
+          # https://github.com/versatica/mediasoup/issues/1502
+          # - os: ubuntu-24.04-arm
+          #   node: 22
+          #   cc: gcc
+          #   cxx: g++
+          #   meson_args: '-Db_sanitize=address'
           - os: ubuntu-24.04-arm
             node: 22
             cc: clang
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -21,7 +21,15 @@ jobs:
             node: 18
             cc: gcc
             cxx: g++
+          - os: ubuntu-22.04-arm
+            node: 18
+            cc: gcc
+            cxx: g++
           - os: ubuntu-24.04
+            node: 20
+            cc: gcc
+            cxx: g++
+          - os: ubuntu-24.04-arm
             node: 20
             cc: gcc
             cxx: g++
@@ -40,6 +48,25 @@ jobs:
             cxx: clang++
             meson_args: '-Db_sanitize=undefined'
           - os: ubuntu-24.04
+            node: 22
+            cc: gcc
+            cxx: g++
+            meson_args: '-Db_sanitize=thread'
+          - os: ubuntu-24.04-arm
+            node: 22
+            cc: gcc
+            cxx: g++
+          - os: ubuntu-24.04-arm
+            node: 22
+            cc: gcc
+            cxx: g++
+            meson_args: '-Db_sanitize=address'
+          - os: ubuntu-24.04-arm
+            node: 22
+            cc: clang
+            cxx: clang++
+            meson_args: '-Db_sanitize=undefined'
+          - os: ubuntu-24.04-arm
             node: 22
             cc: gcc
             cxx: g++
--- a/.github/workflows/mediasoup-node.yaml
+++ b/.github/workflows/mediasoup-node.yaml
@@ -2,7 +2,7 @@ name: mediasoup-node
 
 on:
   push:
-    branches: [ v3 ]
+    branches: [v3]
   pull_request:
   workflow_dispatch:
 
@@ -17,15 +17,15 @@ jobs:
     strategy:
       matrix:
         build:
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             node: 18
             cc: gcc
             cxx: g++
-          - os: ubuntu-22.04
+          - os: ubuntu-24.04
             node: 20
             cc: gcc
             cxx: g++
-          - os: ubuntu-22.04
+          - os: ubuntu-24.04
             node: 22
             cc: gcc
             cxx: g++
@@ -61,6 +61,10 @@ jobs:
             cc: cl
             cxx: cl
           - os: windows-2022
+            node: 22
+            cc: cl
+            cxx: cl
+          - os: windows-2025
             node: 22
             cc: cl
             cxx: cl
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -21,7 +21,9 @@ jobs:
       matrix:
         build:
           - os: ubuntu-22.04
+          - os: ubuntu-22.04-arm
           - os: ubuntu-24.04
+          - os: ubuntu-24.04-arm
           - os: macos-13
           - os: macos-14
           - os: macos-15
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -2,7 +2,7 @@ name: mediasoup-rust
 
 on:
   push:
-    branches: [ v3 ]
+    branches: [v3]
   pull_request:
   workflow_dispatch:
 
@@ -20,13 +20,13 @@ jobs:
     strategy:
       matrix:
         build:
-          - os: ubuntu-20.04
           - os: ubuntu-22.04
           - os: ubuntu-24.04
           - os: macos-13
           - os: macos-14
           - os: macos-15
           - os: windows-2022
+          - os: windows-2025
 
     runs-on: ${{ matrix.build.os }}
 
--- a/.github/workflows/mediasoup-worker-fuzzer.yaml
+++ b/.github/workflows/mediasoup-worker-fuzzer.yaml
@@ -2,7 +2,7 @@ name: mediasoup-worker-fuzzer
 
 on:
   push:
-    branches: [ v3 ]
+    branches: [v3]
   pull_request:
   workflow_dispatch:
 
@@ -18,6 +18,10 @@ jobs:
       matrix:
         build:
           - os: ubuntu-24.04
+            cc: clang
+            cxx: clang++
+            pip-break-system-packages: true
+          - os: ubuntu-24.04-arm
             cc: clang
             cxx: clang++
             pip-break-system-packages: true
--- a/.github/workflows/mediasoup-worker-prebuild.yaml
+++ b/.github/workflows/mediasoup-worker-prebuild.yaml
@@ -3,24 +3,18 @@ name: mediasoup-worker-prebuild
 # Only trigger on GitHub releases.
 on:
   release:
-    types: [ published ]
+    types: [published]
 
 jobs:
   ci:
     strategy:
       matrix:
         build:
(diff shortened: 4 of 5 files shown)


nazar-pc (approved):
FYI there are also aarch64 versions of Ubuntu runners (yes, the naming is bad):
* ubuntu-22.04-arm
* ubuntu-24.04-arm

Might be worth testing on those too

ibc:
> FYI there are also aarch64 versions of Ubuntu runners (yes, the naming is bad):
> 
> * ubuntu-22.04-arm
> * ubuntu-24.04-arm
> 
> Might be worth testing on those too

Will do next.

jmillan (approved):

ibc:
@nazar-pc I've also added ubuntu-22/24.04-arm and it's failing A LOT XD

ibc:
> @nazar-pc I've also added ubuntu-20/24.04-arm and it's failing A LOT XD

Why? Because there is no arm support for clang-format: https://github.com/versatica/mediasoup/actions/runs/13720571403/job/38374956016?pr=1500

nazar-pc:
That is some really comprehensive set of tests. I'm wondering if all of those permutations on the Node.js side are needed if Rust/C++ worker tests are passing.

Also formatting can be skipped since it is not platform-specific.

ibc:
> That is some really comprehensive set of tests. I'm wondering if all of those permutations on the Node.js side are needed if Rust/C++ worker tests are passing.

I don't understand. Node tests are a thing, Worker tests are another thing and Rust tests are another thing.

> Also formatting can be skipped since it is not platform-specific.

Pushed now. Worker formating check disabled in Windows and Ubuntu ARM.

nazar-pc:
What I'm saying is that Node.js is cross-platform, whatever you're writing in TS/JS will work exactly the same way on x86-64 and aarch64, which is not necessarily the case in Rust/C++ with various pointers and other low-level stuff.

So testing all versions of distributions and all versions of Node.js in all possible combinations is probably unnecessary. It should be sufficient to have one for Linux and one for Windows in case of Node.js and focus on testing lower-level stuff on different platforms instead.

There are 78 jobs right now, that'll take a lot of time to run and hardly provide a lot of useful insights.

nazar-pc:
I also see macOS 13/14 can probably be removed to make testing faster. They are only used for development anyway.

ibc:
> What I'm saying is that Node.js is cross-platform, whatever you're writing in TS/JS will work exactly the same way on x86-64 and aarch64, which is not necessarily the case in Rust/C++ with various pointers and other low-level stuff.

I don't agree because Node may behave different depending on OS. For example, remember that we had to split the UnixSocket channel into two (one for sending from Node to Worker and one for receiving in Node from Worker) due to a bug of Node (some old version) in Windows.

In `mediasoup-node.yaml` we are not just testing the TS/JS layer of mediasoup but also the worker since TS tests rely on the Worker.

ibc:
> I also see macOS 13/14 can probably be removed to make testing faster. They are only used for development anyway.

We are including them in the worker prebuild CI, we need to test them:

```yaml
# Worker prebuild for Linux with kernel version 6 Ubuntu (22.04).
# Let's use Ubuntu 22.04 instead of 24.04 so that it builds the
# mediasoup-worker binary using an old version of GLib that will work
# on Linux hosts running more modern GLib versions.
# See https://github.com/versatica/mediasoup/issues/1089.
- os: ubuntu-22.04
  cc: gcc
  cxx: g++
- os: ubuntu-22.04-arm
  cc: gcc
  cxx: g++
- os: macos-13
  cc: clang
  cxx: clang++
- os: macos-14
  cc: clang
  cxx: clang++
- os: macos-15
  cc: clang
  cxx: clang++
- os: windows-2022
  cc: cl
  cxx: cl
- os: windows-2025
  cc: cl
  cxx: cl
```

ibc:
I'm happy that `mediasoup-worker-fuzzer` CI passes in Ubuntu ARM machines.

nazar-pc:
> I don't agree because Node may behave different depending on OS. For example, remember that we had to split the UnixSocket channel into two (one for sending from Node to Worker and one for receiving in Node from Worker) due to a bug of Node (some old version) in Windows.

Sure, run one on Linux, one on macOS and one on Windows. But I don't see a lot of sense running more than one of each. You can run a few versions of Node.js on Linux too to make sure Node.js versions that are supported are covered too. But I don't think all permutations are really necessary.

ibc:
> > I don't agree because Node may behave different depending on OS. For example, remember that we had to split the UnixSocket channel into two (one for sending from Node to Worker and one for receiving in Node from Worker) due to a bug of Node (some old version) in Windows.
> 
> Sure, run one on Linux, one on macOS and one on Windows. But I don't see a lot of sense running more than one of each. You can run a few versions of Node.js on Linux too to make sure Node.js versions that are supported are covered too. But I don't think all permutations are really necessary.

Imagine that the Node bug in Windows I told above only happens in certain version of Windows, then what?

nazar-pc:
Then there will be a bug, but as I said, the probability of that in Node.js specifically is quite low and it is likely that there would be nothing we could do about that other than recommend users to upgrade/downgrade Node.js anyway

ibc:
> Then there will be a bug, but as I said, the probability of that in Node.js specifically is quite low and it is likely that there would be nothing we could do about that other than recommend users to upgrade/downgrade Node.js anyway

Ok, let me please merge this PR first and then I will work on removing entries from CI on next week.

ibc:
Worried about these Node tests failing in ubuntu-24.04-arm: https://github.com/versatica/mediasoup/actions/runs/13721389481/job/38377476129?pr=1500

ibc:
Ok, this action never ends: https://github.com/versatica/mediasoup/actions/runs/13721389481/job/38377476129?pr=1500

ibc:
> Ok, this action never ends: https://github.com/versatica/mediasoup/actions/runs/13721389481/job/38377476129?pr=1500

Hell, this is systematically failing:

https://github.com/versatica/mediasoup/actions/runs/13721389481/job/38383101706?pr=1500

nazar-pc:
Let's disable it for now then and debug later

ibc:
Issue created here: https://github.com/versatica/mediasoup/issues/1502

I will comment that job for this PR to pass and be merged.

Bessonov:
I'm not sure if this was intended, but after this PR/since v3.15.6, there are no more kernel5 worker binaries. However, it seems they are still needed for Ubuntu 22.04:
```
docker run --rm ubuntu:jammy-20250404 uname -a
Linux a9f3d4b47bb7 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:25:09 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

docker run --rm -it ubuntu:jammy-20250404 cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.5 LTS"
```

nazar-pc:
No builds are needed (strictly speaking), they just help to avoid building the worker, improving installation time. But it will install and work just fine if you have satisfied the dependencies.

Bessonov:
@nazar-pc Thanks for your quick response! I believe you mean that the worker will be built if it can't be downloaded and the dependencies are satisfied. However, they aren't. I would still love to use the prebuilt workers rather than installing the entire compile stack on every machine and building one.

ibc:
> Thanks for your quick response! I believe you mean that the worker will be built if it can't be downloaded and the dependencies are satisfied. However, they aren't.

What does it mean that "they aren't"? That's literally how it's designed: if the installation process cannot fetch a suitable worker binary for current arch then it builds it locally.


> I would still love to use the prebuilt workers rather than installing the entire compile stack on every machine and building one.

I'm sorry but GitHub has removed kernel 5 hosts from CI so we cannot build kernel5 based worker binaries. This is not up to us.

nazar-pc:
> I'm sorry but GitHub has removed kernel 5 hosts from CI so we cannot build kernel5 based worker binaries. This is not up to us.

I'm pretty sure you can still build it with some flags or worse case you can run a VM with an old kernel, but that kind of complexity is hard to justify with things working fine as is. Want prebuilt workers? I guess it is time to upgrade. On Ubuntu it is typically as easy as installing HWE kernel.

ibc:
It's not a good idea to build kernel dependant binaries on VMs since the kernel they run is usually constrained by the kernel of the host.

But I agree. Nothing actionable here. We do our best and we'll not do magic just to support worker binaries in old kernels.

nazar-pc:
> It's not a good idea to build kernel dependant binaries on VMs since the kernel they run is usually constrained by the kernel of the host.

If it is an actual VM, you can run any older kernel or newer, but it will be a mess to implement, I'm not looking forward to it.

Bessonov:
> What does it mean that "they aren't"?

It means that dependencies like g++ aren't installed. This is the whole point of using prebuilt binaries: to save space, time, bandwidth, and hassle.

> I'm sorry but GitHub has removed kernel 5 hosts from CI so we cannot build kernel5 based worker binaries. This is not up to us.

That's really pity. I found this one: https://github.com/tatry/kernel-version-demo . However, I think it should also be doable with Docker.

> I guess it is time to upgrade.

It seems like Kernel 5 will be used quite a lot over the next few years:
```
docker run --rm ubuntu:24.04 uname -a
Linux e8b790844cd1 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:25:09 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
So even upgrading to 24.04 wouldn't help - at least not with the official images.

> On Ubuntu it is typically as easy as installing HWE kernel.

Thanks for pointing that out. However - oh boy:
```
docker run --rm ubuntu:22.04 bash -c "apt-get update && apt install linux-generic-hwe-22.04"
[...]
Need to get 503 MB of archives.
After this operation, 2054 MB of additional disk space will be used.
```

I think I'll stick with v3.15.5 for a while. I also see that I can set the `MEDIASOUP_WORKER_PREBUILT_DOWNLOAD_BASE_URL` variable - maybe that'll help avoid the hassle.

Thanks for hearing me out! :)

nazar-pc:
> It means that dependencies like g++ aren't installed

Those are listed as requirements on the official website, it is not guaranteed to work if you don't have them. Prebuilt binaries are an optional optimization.

> It seems like Kernel 5 will be used quite a lot over the next few years:

As I said, you can install much newer official kernel, even on Ubuntu 22.04, see https://ubuntu.com/kernel/lifecycle
On Ubuntu 22.04.5 HWE kernel is 6.8.0.
Simply switch to HWE kernel and reboot, done!

ibc:
> It seems like Kernel 5 will be used quite a lot over the next few years:
> 
> docker run --rm ubuntu:24.04 uname -a
> Linux e8b790844cd1 5.15.0-78-gen

The Docker image runs the kernel of the host. So what you are showing in  that command only means that your host has kernel 5. However Ubuntu 24.04 uses kernel 6. In fact we create worker binaries using 22.04 and 24.04 and both use kernel 6 as commented in the code:

https://github.com/versatica/mediasoup/blob/v3/.github/workflows/mediasoup-worker-prebuild.yaml

ibc:
So in summary you should update the Linux version you use to run Docker.

nazar-pc:
> The Docker image runs the kernel of the host

Yes, but Docker creates containers (with the default runtime anyway), not VMs

Bessonov:
OMG, that's pretty embarrassing 🙀 I've been using Docker for nearly ten years and never thought about that! 😄 Thanks and sorry!

nazar-pc:
I mean it is running under Linux VM on macOS and Windows (when Linux containers are used), but generally it is a different technology from virtualization (though there are runtimes that allow running Docker containers in VMs too, it is just not the default). Things are complicated :sweat_smile:

PR #1506, opened 2025-03-11
CI: Remove redundant hosts `macos-14` and `windows-2022` from `mediasoup-worker-prebuild` job

ibc:
## Details

- Remove `windows-2022` host because we also have `windows-2025` and both produce the same binary name so they override one to each other: `mediasoup-worker-3.15.6-win32-x64.tgz`
- Remove `macos-14` host because we also have `macos-15` (same reason).

### DIFF
--- a/.github/workflows/mediasoup-worker-prebuild.yaml
+++ b/.github/workflows/mediasoup-worker-prebuild.yaml
@@ -24,15 +24,9 @@ jobs:
           - os: macos-13
             cc: clang
             cxx: clang++
-          - os: macos-14
-            cc: clang
-            cxx: clang++
           - os: macos-15
             cc: clang
             cxx: clang++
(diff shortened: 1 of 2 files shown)


nazar-pc (approved):
Makes sense. What about macOS 13? It is x86-64, but I doubt many people care about it and the probability of it not working on x86-64 specifically is kind of limited too. Also it is quite old.

ibc:
I was using a Mac Intel until January this year and I promise it did work 😀

nazar-pc:
I believe you, it works on way more systems than we have in CI, the question is whether it is necessary in CI. But I guess it doesn't matter as much as long as it costs nothing to support.

ibc:
We can remove macOS-13 in next months in all CI jobs, but I'd feel sad if I did it now.

PR #1508, opened 2025-03-18
AV1 Codec

jmillan:
Implements Dependency Descriptor (DD) RTP header extension and AV1 codec handler.

* DD Header extension is set as `recvonly` as we do not override it.

- [x] blocker https://github.com/versatica/mediasoup-client/pull/320


ibc:
> ```diff
> -                       pc.addTransceiver('video');
> +                       // Create video transceiver with simulcast in order to retrieve
> +                       // dependency descriptor header extension too.
> +                       pc.addTransceiver('video', {
> +                               sendEncodings: [{ rid: '0' }, { rid: 1 }],
> +                       });
> ```

Better than this, in Chrome 111 we can add a single encoding with `scalabilityMode`.

worker/src/RTC/Codecs/BitStream.cpp
```diff
@@ -0,0 +1,52 @@
+#define MS_CLASS "RTC::Codecs::BitStream"
+
+#include "RTC/Codecs/BitStream.hpp"
+#include "Logger.hpp"
+
+namespace RTC
+{
+	namespace Codecs
+	{
+		BitStream::BitStream(const uint8_t* data, size_t len) : data(data), len(len)
```
ibc: Does it make sense that this class is inside Codecs? We may need it to parse whatever.

worker/src/RTC/RtpDictionaries/RtpHeaderExtensionUri.cpp
ibc, line 1: Aren't you removing frame-marking stuff in this PR?

worker/src/RTC/RtpPacket.cpp
```diff
@@ -395,16 +395,17 @@ namespace RTC
...
-		this->ssrcAudioLevelExtensionId    = 0u;
-		this->videoOrientationExtensionId  = 0u;
-		this->playoutDelayExtensionId      = 0u;
+		this->midExtensionId                  = 0u;
+		this->ridExtensionId                  = 0u;
+		this->rridExtensionId                 = 0u;
+		this->absSendTimeExtensionId          = 0u;
+		this->transportWideCc01ExtensionId    = 0u;
+		this->frameMarking07ExtensionId       = 0u;
+		this->frameMarkingExtensionId         = 0u;
```
ibc, line 404: Don't we want to remove everything related to frame-marking?

worker/src/RTC/RtpDictionaries/RtpCodecMimeType.cpp
```diff
@@ -35,6 +35,7 @@ namespace RTC
 		// Video codecs:
 		{ "vp8",             RtpCodecMimeType::Subtype::VP8             },
 		{ "vp9",             RtpCodecMimeType::Subtype::VP9             },
+		{ "av1",             RtpCodecMimeType::Subtype::AV1             },
```
ibc: Can we add it after h265 for consistency with `supportedRtpCapabilities.ts` and because this is the last added codec?

worker/src/RTC/RtpDictionaries/RtpCodecMimeType.cpp
```diff
@@ -63,6 +64,7 @@ namespace RTC
 		// Video codecs:
 		{ RtpCodecMimeType::Subtype::VP8,             "VP8"             },
 		{ RtpCodecMimeType::Subtype::VP9,             "VP9"             },
+		{ RtpCodecMimeType::Subtype::AV1,             "AV1"             },
```
ibc: Same.

ibc (commented):
Partial super cosmetic review done for now.

worker/src/RTC/Codecs/BitStream.cpp
jmillan: Yep, I was definitely thinking of moving it out. Maybe under `RTC` directly?

worker/src/RTC/RtpDictionaries/RtpHeaderExtensionUri.cpp
jmillan: I will, yes

worker/src/RTC/RtpPacket.cpp
jmillan: yes, yes

worker/src/RTC/Codecs/BitStream.cpp
ibc: I clearly see it in Utils namespace which is outside RTC.

worker/src/RTC/RtpPacket.cpp
jmillan: I'll remove frame marking in another PR as it's an orthogonal change that can be made even if this PR is not done.

jmillan:
This PR is ready for review. Provides DD support and AV1 codec.

worker/src/RTC/Codecs/AV1.cpp
```diff
@@ -0,0 +1,223 @@
...
+			MS_TRACE();
+
+			std::unique_ptr<PayloadDescriptor> payloadDescriptor(new PayloadDescriptor());
+
+			// Read fields.
+			payloadDescriptor->startOfFrame  = dependencyDescriptor->startOfFrame;
+			payloadDescriptor->endOfFrame    = dependencyDescriptor->endOfFrame;
+			payloadDescriptor->spatialLayer  = dependencyDescriptor->spatialLayer;
+			payloadDescriptor->temporalLayer = dependencyDescriptor->temporalLayer;
+
```
ibc, line 22: Remove blank line?

worker/test/src/RTC/Codecs/TestDependencyDescriptor.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <catch2/catch_test_macros.hpp>
+
+using namespace RTC;
+
+SCENARIO("parse Dependency Descriptor", "[codecs][DD]")
+{
+	SECTION("parse Dependency Descriptor")
+	{
+		/**
+		 * Taken from https://issues.webrtc.org/issues/42225660.
```
ibc, line 12: Can we please use /** */ comment syntax with all the lines in the comment prefixed with *?

ibc (approved):

ibc:
In fuzzer we must also check the AV1 codec, not just the dependency descriptor.

jmillan:
> In fuzzer we must also check the AV1 codec, not just the dependency descriptor.

I'll add it, but it does literally just assign the values from DD. There is no actual parsing.

jmillan:
> I'll add it, but it does literally just assign the values from DD. There is no actual parsing.

Done

ibc (approved):

Lynnworld:
note:  Dependency Descriptor  may use two byte extensions because it's too long. For example L3T3 mode

ibc:
> note:  Dependency Descriptor  may use two byte extensions because it's too long. For example L3T3 mode

This is already covered in mediasoup (2 byte extensions are supported since ever) and in mediasoup-client last version which negotiates it properly the way Chrome likes.

Lynnworld:
> > note:  Dependency Descriptor  may use two byte extensions because it's too long. For example L3T3 mode
> 
> This is already covered in mediasoup (2 byte extensions are supported since ever) and in mediasoup-client last version which negotiates it properly the way Chrome likes.

you missed something :( https://github.com/versatica/mediasoup/blob/09a23107e3e1020e4649ef95327902ab59c6d1fb/worker/src/RTC/Producer.cpp#L1423

ibc:
> you missed something :( https://github.com/versatica/mediasoup/blob/09a23107e3e1020e4649ef95327902ab59c6d1fb/worker/src/RTC/Producer.cpp#L1423

mediasoup doesn't add the Payload Descriptor header on packets it sends to clients. It only parses it from received packets.

Lynnworld:
For pipe consumer 😂

ibc:
>  For pipe consumer 😂

Excellent point

PR #1509, opened 2025-03-20
fix typecheck error when use node20&alpine build

lrobot:
fix typecheck error when use node20&alpine build


ibc:
> Sorry that I notice this test issue has already been fixed in latest commit. So Just close this pull request

Ok thanks

PR #1511, opened 2025-03-21
Rust: fix PipeConsumer get_stats()

### LINKED ISSUE #1510: Panic at consumer.rs:getStats() when cascade simulcast h264 stream
## Bug Report

**IMPORTANT:** We primarily use GitHub as an issue tracker. Just open an issue here if you have encountered a bug in mediasoup.

If you have questions or doubts about mediasoup or need support, please use the mediasoup Discourse Group instead:

https://mediasoup.discourse.group

If you got a crash in mediasoup, please try to provide a core dump into the issue report:

https://mediasoup.org/support/#crashes-in-mediasoup-get-a-core-dump

### Your environment

- Operating system: Centos7.9
- Node version:
- npm version:
- gcc/clang version:
- mediasoup version: rust-0.17.1
- mediasoup-client version:

### Issue description

When publishing a Simulcast media stream using H.264, multiple RTP streams exist. When using a PipeTransport to send the stream to another mediasoup server, calling consumer::getStats returns two SendStats. However, when encountering the following code:

```
let consumer_stat = ConsumerStat::from_fbs(&data.stats[0]);
let producer_stat = ProducerStat::from_fbs(&data.stats[1]);

```
a panic occurs with the error: RecvStats data: SendStats.

jmillan:
Fixes #1510

Consider PipeConsumer which stats contains multiple send streams.

### DIFF
--- a/rust/src/router/consumer.rs
+++ b/rust/src/router/consumer.rs
@@ -1033,14 +1033,25 @@ impl Consumer {
             .await?;
 
         if let response::Body::ConsumerGetStatsResponse(data) = response {
-            match data.stats.len() {
-                0 => panic!("Empty stats response from worker"),
-                1 => {
-                    let consumer_stat = ConsumerStat::from_fbs(&data.stats[0]);
+            match self.r#type() {
+                ConsumerType::Simple | ConsumerType::Simulcast | ConsumerType::Svc => {
+                    match data.stats.len() {
+                        0 => panic!("Empty stats response from worker"),
+                        1 => {
+                            let consumer_stat = ConsumerStat::from_fbs(&data.stats[0]);
 
-                    Ok(ConsumerStats::JustConsumer((consumer_stat,)))
+                            Ok(ConsumerStats::JustConsumer((consumer_stat,)))
+                        }
+                        2 => {
+                            let consumer_stat = ConsumerStat::from_fbs(&data.stats[0]);
+                            let producer_stat = ProducerStat::from_fbs(&data.stats[1]);
+
+                            Ok(ConsumerStats::WithProducer((consumer_stat, producer_stat)))
+                        }
+                        _ => panic!("More than two stats response from worker"),
+                    }
                 }
-                _ => {
+                ConsumerType::Pipe => {
                     let mut stats = Vec::<ConsumerStat>::with_capacity(data.stats.len());
                     for stat in data.stats {
                         stats.push(ConsumerStat::from_fbs(&stat));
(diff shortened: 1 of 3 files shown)


rust/src/router/consumer.rs
```diff
@@ -442,6 +444,7 @@ impl ConsumerStats {
         match self {
             ConsumerStats::JustConsumer((consumer_stat,)) => consumer_stat,
             ConsumerStats::WithProducer((consumer_stat, _)) => consumer_stat,
+            ConsumerStats::MultipleConsumers(consumer_stats) => &consumer_stats[0],
```
jmillan, line 447: `consumer_stats()` returns a single consumer stat by definition, hence here it's returning the first one. All consumer stats are accessible from the `MultipleConsumers` variant directly.

nazar-pc: Is `ConsumerStats::WithProducer` used anywhere now? Two entries meant consumer+producer before :thinking:

jmillan: Yes, `SimpleConsumer` and `SimulcastConsumer` returns the pair of recv-send streams stats.

nazar-pc: But with this PR there is nothing that ever constructs `ConsumerStats::WithProducer`. Should you check the kind of consumer and then produce either `ConsumerStats::WithProducer` or `ConsumerStats::MultipleConsumers` maybe?

ibc:
Entry in rust changelog missing and after this we should release a rust version.

rust/src/router/consumer.rs
jmillan: `MultipleConsumer` must remain. Yes, I'll create one or another based on the consumer type.

jmillan: Done

ibc (approved):
So the issue is fixed by this PR despite the ongoing discussion in the issue?

rust/tests/integration/consumer.rs
```diff
@@ -679,6 +679,10 @@ fn consume_succeeds() {
                 video_pipe_consumer.app_data().downcast_ref::<()>().unwrap(),
                 &(),
             );
+            video_pipe_consumer
+                .get_stats()
+                .await
+                .expect("Failed to get consumer stats");
```
nazar-pc, line 685: Assuming this failed before the fix, looks good to me

nazar-pc (approved):

rust/tests/integration/consumer.rs
jmillan: yes, it failed before.

nazar-pc:
Adding enum variant is a breaking change. It is best to yank version 0.17.2 from crates.io and publish 0.18.0 to prevent breaking things for users. Also I'm wondering if new worker version should be releases too (likely also with a minor version bump) with some changes that happened recently.

nazar-pc:
@jmillan in case you missed above message

jmillan:
Done,

mediasoup-sys and mediasoup crates published. 0.17.2 version has been yanked.

PR #1516, opened 2025-03-24
Fix wrong SCTP stream parameters in SCTP `DataConsumer` that consumes from a direct `DataProducer`

### LINKED ISSUE #1515: DataConsumers that consume from a DataProducer created on a DirectTransport have ordered: false
In `SctpAssociation.cpp`:

```c++
void SctpAssociation::SendSctpMessage(
  RTC::DataConsumer* dataConsumer, const uint8_t* msg, size_t len, uint32_t ppid, onQueuedCallback* cb)
{
	// [...]

	const auto& parameters = dataConsumer->GetSctpStreamParameters();

	// Fill sctp_sendv_spa.
	struct sctp_sendv_spa spa
	{
	}; // NOLINT(cppcoreguidelines-pro-type-member-init)

	std::memset(&spa, 0, sizeof(spa));
	spa.sendv_flags             = SCTP_SEND_SNDINFO_VALID;
	spa.sendv_sndinfo.snd_sid   = parameters.streamId;
	spa.sendv_sndinfo.snd_ppid  = htonl(ppid);
	spa.sendv_sndinfo.snd_flags = SCTP_EOR;

	// If ordered it must be reliable.
	if (parameters.ordered)
	{
		MS_DUMP_STD("------- parameters.ordered");
		spa.sendv_prinfo.pr_policy = SCTP_PR_SCTP_NONE;
		spa.sendv_prinfo.pr_value  = 0;
	}
	// Configure reliability: https://tools.ietf.org/html/rfc3758
	else
	{
		MS_DUMP_STD("------- !!!!! NO parameters.ordered NO");
		spa.sendv_flags |= SCTP_SEND_PRINFO_VALID;
		spa.sendv_sndinfo.snd_flags |= SCTP_UNORDERED;

```

Let's assume that:

```ts
const dataProducer = await directTransport.produceData(...);

const dataConsumer = await webRtcTransport.consumeData({
  dataProducerId: dataProducer.id,
  ...
});
```

In this scenario, `dataProducer.dump()` returns `sctpStreamParameters: undefined` (as expected since this `DataProducer` was created on a `DirectTransport` so no SCTP here. And `dataConsumer.dump()` returns `sctpParameters` with `ordered: false`:

```
dataProducer.dump(): {
  id: 'ff43b64c-7c95-4b59-9533-d45b11d916bd',
  type: 'direct',
  sctpStreamParameters: undefined,
  label: 'bot',
  protocol: '',
  paused: false
}

dataConsumer.dump(): {
  id: '4f1f9db8-fdf2-4b59-8f91-098279d192f2',
  dataProducerId: 'ff43b64c-7c95-4b59-9533-d45b11d916bd',
  type: 'sctp',
  sctpStreamParameters: {
    streamId: 0,
    ordered: false,
    maxPacketLifeTime: undefined,
    maxRetransmits: undefined
  },
  label: 'bot',
  protocol: '',
  bufferedAmountLowThreshold: 0,
  paused: false,
  dataProducerPaused: false,
  subchannels: []
}
```

This is wrong. We know that every `DataConsumer` that consumes from a `DataProducer` of type "direct" (so created on a `DirectTransport`) is reliable by nature (we don't implement unreliable messages on `DirectTransport`). So this `DataConsumer` has `ordered: false` which means that when we send a message to the corresponding browser, the message is sent in unreliable mode (see `SctpAssociation::SendSctpMessage()` above).

ibc:
Fixes #1515

## Details

As exposed in the ticket #1515, let's assume this scenario:

- Alice and Bob create a WebRTC `dataProducer` with `ordered: true` (so a reliable DataChannel).
- In mediasoup side, there is a `directTransport` and a `directDataConsumer` which consumes those WebRTC `dataProducer` of Alice and Bob.
- The server runs some logic with messages received from Alice and Bob and then delivers those messages to Alice or Bob by using a `directDataProducer` created in that `directTransport`. 
- For that to work, Alice and Bob create a WebRTC `dataConsumer` that consumes messages from that `directDataProducer`.

**So what was the problem?**

- `dataProducers` created on a `directTransport` are always reliable. Why? Because they send messages through the mediasoup UnixSocket which is a reliable transport.
- So it makes sense that when we create a WebRTC `dataConsumer` in a `webRtcTransport` that consumes that `directDataProducer`, such a `dataConsumer` should be reliable (it should have `directDataConsumer.sctpParameters.ordered === true`.
- **However this was not true!** Or in other words, this was only true if `webRtcTransport.consumeData()` was called with `ordered: true` **explicitly**. But the point is that by default it should be reliable (so `ordered: true`) because we are consuming from a `directDataProducer` that is reliable/ordered by design.
- So at the end, assuming we don't include `ordered: true` in `webRtcTransport.consumeData()`, the resulting `dataConsumer` (which is the one that will deliver messages to the browser) is **unreliable** (it has `dataConsumer.sctpParameters.ordered === false`) so messages sent to the browser could reach the browser out or order **and be dispatched by the browser WebRTC `PeerConnection` out of order**, and some messages could be lost!
- So imagine that Alice sends messages "A", "B" and "C" to mediasoup, those messages travel through the `directTransport` blablabla and finally they are sent to Bob via WebRTC `dataConsumer`. It could happen that Bob receives "C" and "B" (in that order). Ok, that can always happen due to network conditions, but the problem is that, since the `dataConsumer` is configured as unreliable (it has `ordered: false`) the `dataConsumer` in Bob's browser will dispatch "message" events in that wrong order, this is: "C" and "B".
- Note that if the `dataConsumer` was reliable, then the WebRTC engine of the browser would buffer "C" and "B" and wouldn't dispatch them to the app until "A" is received, and it would dispatch them in order: "A", "B", "C".
- And even worse: mediasoup would try to retransmit "A" N times (depending on `sctpParameters.maxRetransmits`) and if none of the retransmissions of "A" reaches Bob's browser, the `webRtcTransport` of Bob in mediasoup side **won't emit "sctpstatechange" with `state: 'closed'`**!

**Good news**

This is fixed in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - CI: Remove redundant hosts `macos-14` and `windows-2022` from `mediasoup-worker-prebuild` job ([PR #1506](https://github.com/versatica/mediasoup/pull/1506)).
 - Node: Modernize code ([PR #1513](https://github.com/versatica/mediasoup/pull/1513)).
-- Fix wrong SCTP stream parameters in SCTP `DataConsumer` that consume from direct `DataProducers` ([PR #1516](https://github.com/versatica/mediasoup/pull/1516)).
+- Fix wrong SCTP stream parameters in SCTP `DataConsumer` that consumes from a direct `DataProducer` ([PR #1516](https://github.com/versatica/mediasoup/pull/1516)).
 
 ### 3.15.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - CI: Remove redundant hosts `macos-14` and `windows-2022` from `mediasoup-worker-prebuild` job ([PR #1506](https://github.com/versatica/mediasoup/pull/1506)).
 - Node: Modernize code ([PR #1513](https://github.com/versatica/mediasoup/pull/1513)).
+- Fix wrong SCTP stream parameters in SCTP `DataConsumer` that consume from direct `DataProducers` ([PR #1516](https://github.com/versatica/mediasoup/pull/1516)).
 
 ### 3.15.6
 
--- a/node/src/Transport.ts
+++ b/node/src/Transport.ts
@@ -864,26 +864,25 @@ export abstract class TransportImpl<
 			this.#sctpStreamIds![sctpStreamId] = 1;
 			sctpStreamParameters.streamId = sctpStreamId;
 
-			// Override if given.
 			if (ordered !== undefined) {
 				sctpStreamParameters.ordered = ordered;
 
 				if (ordered) {
-					maxPacketLifeTime = undefined;
-					maxRetransmits = undefined;
 					sctpStreamParameters.maxPacketLifeTime = undefined;
 					sctpStreamParameters.maxRetransmits = undefined;
 				}
 			}
 
-			if (maxPacketLifeTime !== undefined) {
-				sctpStreamParameters.ordered = undefined;
-				sctpStreamParameters.maxPacketLifeTime = maxPacketLifeTime;
-			}
+			if (!ordered) {
+				if (maxPacketLifeTime !== undefined) {
+					sctpStreamParameters.ordered = false;
+					sctpStreamParameters.maxPacketLifeTime = maxPacketLifeTime;
+				}
 
-			if (maxRetransmits !== undefined) {
-				sctpStreamParameters.ordered = undefined;
-				sctpStreamParameters.maxRetransmits = maxRetransmits;
+				if (maxRetransmits !== undefined) {
+					sctpStreamParameters.ordered = false;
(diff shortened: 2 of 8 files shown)


jmillan (dismissed):

ibc:
@jmillan please don't approve this yet, I'm still working on the Rust changes.

jmillan:
> @jmillan please don't approve this yet, I'm still working on the Rust changes.

I wonder why GH allows approving a draft PR. Please, reassign it to me when done.

ibc:
> > @jmillan please don't approve this yet, I'm still working on the Rust changes.
> 
> I wonder why GH allows approving a draft PR. Please, reassign it to me when done.

PR ready.

jmillan (approved):

ibc:
Released in mediasoup 3.15.7.

PR #1517, opened 2025-03-27
worker: H264, enable Dependency Descriptor

jmillan:
This PR base branch is `dependency_descriptor`

Uses Dependency Descriptor header extension rather than the obsolete FrameMarking in order to retrieve the temporal layers.

### DIFF
--- a/worker/src/RTC/Codecs/H264.cpp
+++ b/worker/src/RTC/Codecs/H264.cpp
@@ -12,7 +12,31 @@ namespace RTC
 		/* Class methods. */
 
 		H264::PayloadDescriptor* H264::Parse(
-		  const uint8_t* data, size_t len, RTC::RtpPacket::FrameMarking* frameMarking, uint8_t frameMarkingLen)
+		  const uint8_t* data, size_t len, Codecs::DependencyDescriptor* dependencyDescriptor)
+		{
+			MS_TRACE();
+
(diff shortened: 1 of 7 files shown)


ibc:
I am a bit lost. You have added fuzzer for DD in the AV1 codec instead than here.

worker/src/RTC/Codecs/H264.cpp
```diff
@@ -12,140 +12,119 @@ namespace RTC
 		/* Class methods. */
 
 		H264::PayloadDescriptor* H264::Parse(
-		  const uint8_t* data, size_t len, RTC::RtpPacket::FrameMarking* frameMarking, uint8_t frameMarkingLen)
+		  const uint8_t* data, size_t len, Codecs::DependencyDescriptor* dependencyDescriptor)
 		{
 			MS_TRACE();
 
-			if (len < 2)
```
ibc, line 19: Is this safe?

jmillan:
> I am a bit lost. You have added fuzzer for DD in the AV1 codec instead than here.

DD is implemented in the DD & AV1 PR. This is why DD fuzzer is implemented there.

worker/src/RTC/Codecs/H264.cpp
jmillan: We are not touching `data` here, so yes. We are passing `data` and `len` to `IsKeyFrame()` method where the len is verified.

ibc (approved):

Issue #1521, opened 2025-04-08
Panic at consumer.rs when casecade simulcast rtp stream


bertonwalker-cn:
## Bug Report

**IMPORTANT:** We primarily use GitHub as an issue tracker. Just open an issue here if you have encountered a bug in mediasoup.

If you have questions or doubts about mediasoup or need support, please use the mediasoup Discourse Group instead:

https://mediasoup.discourse.group

If you got a crash in mediasoup, please try to provide a core dump into the issue report:

https://mediasoup.org/support/#crashes-in-mediasoup-get-a-core-dump

### Your environment

- Operating system:
- Node version:
- npm version:
- gcc/clang version:
- mediasoup version: 0.18.0
- mediasoup-client version:

### Issue description

The bug is from fixing  [#1511 ] (https://github.com/versatica/mediasoup/issues/1510) .
Getstats from simulcast consumer.cpp returns SendStats and RecvStats, not multi SendStat . 
So panic in consumer.rs 
```
let mut stats = Vec::<ConsumerStat>::with_capacity(data.stats.len());
                    for stat in data.stats {
                        stats.push(ConsumerStat::from_fbs(&stat));
                    }
```

ConsumerStat::from_fbs only receive SendStats. So panic with RecvStats

```
panicked at crates/mediasoup/rust/src/router/consumer.rs:413:13:\nWrong message from worker
```

jmillan:
It does not make much sense:

The given panicked line is this: https://github.com/versatica/mediasoup/blob/v3/rust/src/router/consumer.rs#L413 

On the other hand the snippet you shared is for pipe consumer. Is there a reliable way to test it?

jmillan:
I've created a PR that will help with this issue https://github.com/versatica/mediasoup/pull/1523

jmillan:
I'm testing the mediasoup rust example and I'm getting a proper simulcast consumer stats:

```txt
[...]

PR #1523, opened 2025-04-08
Rust: FBS, provide proper data upon panic 

jmillan:
Show the data that generated the panic.

Will help with https://github.com/versatica/mediasoup/issues/1521


ibc (approved):

nazar-pc (approved):
At some point we should consider transitioning from panics to error handling in these cases even though we do not really expect these bugs to happen.

jmillan:
> At some point we should consider transitioning from panics to error handling in these cases even though we do not really expect these bugs to happen.

Totally agreed.

PR #1525, opened 2025-04-23
Create our own SCTP stack - Part 1: SCTP parsers and factories

ibc:
## Details

- Add classes for SCTP packet, chunks, chunk parameters and error causes.
- Implement parsers and factories for all them.
- Add tests for all them.
- All chunks, chunk parameters and error causes of [RFC 9260](https://datatracker.ietf.org/doc/html/rfc9260) are implemented (more will be added in future PRs, see https://github.com/versatica/mediasoup/wiki/SCTP).
- Cosmetic changes in other files, specially in worker `Utils` module.

### The new `Serializable` class

- See `worker/RTC/Serializable.hpp` (and its `.cpp` file). This is the base class of SCTP `Packet`, `Chunk`, `ChunkParameter` and `ErrorCause`.
- In the future we should migrate STUN, RTP and SRTP to it.
- See also the new `MS_DUMP_CLEAN(indentation, desc, ...)` macro in `Logger.hpp` which is extensively used in all SCTP classes for a much better dump feature.


worker/include/RTC/SCTP/chunks/ShutdownChunk.hpp
```diff
@@ -0,0 +1,106 @@
...
+		// Forward declaration.
+		class Packet;
+
+		class ShutdownChunk : public Chunk
+		{
+			// We need that Packet calls protected and private methods in this class.
+			friend class Packet;
+
+		public:
+			static const size_t ShutdownChunkHeaderLength{ 8 };
```
jmillan, line 42: Considering we are inside `ShutdownChunk` class, this could be named `HeaderLength` as it's going to be accessed as `ShutdownChunk::HeaderLength`.

worker/meson.build
```diff
@@ -173,6 +180,45 @@ common_sources = [
   'src/RTC/RTCP/XR.cpp',
   'src/RTC/RTCP/XrDelaySinceLastRr.cpp',
   'src/RTC/RTCP/XrReceiverReferenceTime.cpp',
+  'src/RTC/SCTP/Packet.cpp',
```
jmillan, line 183: What about avoiding compiling if the flag is not set for the time being?

worker/src/RTC/SCTP/PacketItemBase.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+		/* Instance methods. */
+
+		PacketItemBase::PacketItemBase(uint8_t* buffer, size_t bufferLength)
+		  : Serializable(buffer, bufferLength)
+		{
+			MS_TRACE();
+		}
+
+		PacketItemBase::~PacketItemBase()
```
jmillan, line 70: Is this intended while development is ongoing? Otherwise why this rather than using a `=default`. Is it just for debugging?

worker/src/RTC/SCTP/chunkParameters/CookiePreservativeChunkParameter.cpp
```diff
@@ -0,0 +1,148 @@
...
+				MS_WARN_DEV("invalid Chunk Parameter type");
+
+				return nullptr;
+			}
+
+			return CookiePreservativeChunkParameter::ParseStrict(
+			  buffer, bufferLength, parameterLength, padding);
+		}
+
+		CookiePreservativeChunkParameter* CookiePreservativeChunkParameter::ParseStrict(
```
jmillan, line 64: I'm trying to understand the rationale for this name. Is the process always `Parse()` calling `ParseStrict()`?

worker/src/RTC/SCTP/chunkParameters/CookiePreservativeChunkParameter.cpp
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		CookiePreservativeChunkParameter* CookiePreservativeChunkParameter::Factory(
+		  uint8_t* buffer, size_t bufferLength)
+		{
+			MS_TRACE();
+
+			if (bufferLength < CookiePreservativeChunkParameter::CookiePreservativeChunkParameterLength)
+			{
+				MS_THROW_TYPE_ERROR("too small buffer");
```
jmillan: @Cosmetic, this message resonated in my head as I would expect 'buffer too small'. As said, this is cosmetic, just a comment. Not asking to change.

jmillan (approved):
Bravo 👏 , just few comments made.

worker/meson.build
ibc: The flag is to enable or disable the custom SCTP stack in runtime. But I want to test and fuzz it despite it's not enabled in runtime. There is zero benefit in hiding those SCTP files and I want to keep working on SCTP and run tests and fuzzer in CI despite the SCTP stack is not enabled/used in production/runtime.

worker/src/RTC/SCTP/PacketItemBase.cpp
ibc: For debugging purposes and that saved my life many times.

worker/src/RTC/SCTP/chunkParameters/CookiePreservativeChunkParameter.cpp
ibc: `Xxxxx::Parse()` is intended for directly parsing a buffer containing `Xxxxx` packet/chunk/parameter/error-cause. We don't really need it in production runtime since what we parse is only received full SCTP `Packets`. But anyway those `Xxxxx::Parse()` are useful for testing and maybe needed in the future in case we need to parse full chunks or chunk parameters contained in the body of some SCTP error causes, etc.

`Xxxxx::ParseStrict()` is used within `Packet::Parse()` to parse chunks, chunk parameters and error causes within a SCTP buffer that contains a SCTP packet. It's more efficient the `Xxxxx::Parse()`. In fact, `Xxxxx::Parse()` internally calls `Xxxxx::ParseStrict()`. See any class to understand the difference.

worker/src/RTC/SCTP/chunkParameters/CookiePreservativeChunkParameter.cpp
ibc: I've changed it.

worker/meson.build
ibc: I mean, yes, I can check the `ms_sctp_stack` option in `meson_options.txt` and then exclude SCTP files from `meson.build` and from `tests.cpp` in worker tests and from `fuzzer.cpp` in worker fuzzer, and then next week when I write another SCTP related PR I would need to edit CI yaml files to add that meson build option so SCTP is tested and fuzzed, etc etc. And all that complexity just to save ~8 seconds of build time.

worker/include/RTC/SCTP/chunks/ShutdownChunk.hpp
ibc: If I forget to re-declare `HeaderLength` in a `FooChunk` class whose header length is not the default one and then I use `FooChunk::HeaderLength` is will give me the value of `Chunk::HeaderLength` which is 4 and things would fail, so I prefer to be explicit and prefix everything with the full subclass name.

worker/include/RTC/SCTP/Packet.hpp
```diff
@@ -0,0 +1,236 @@
...
+		private:
+			/**
+			 * Constructor is private because we only want to create Packet instances
+			 * via Parse() and Factory().
+			 */
+			Packet(uint8_t* buffer, size_t bufferLength);
+
+		public:
+			~Packet() override;
+			;
```
jmillan: Ups,

```suggestion
```

worker/src/RTC/SCTP/chunkParameters/CookiePreservativeChunkParameter.cpp
jmillan: IMHO we could name both methods `Parse()`, since they have different arguments we would just use function overloading. I personally don't see a reason behind the name `strict`.

ibc: After long call, we will keep `ParseStrict()` naming becase it's an "internal" static method and the alternative could be `UnsafeParse()` which sounds terrifying.

ibc: Note, I've made `ParseStrict()` static methods private because friend class can access them.

ibc:
Ok, I'm merging this.

PR #1527, opened 2025-04-29
Worker: Fix, encode retransmitted packets with the corresponding data

jmillan:
NOTE: Only affects VP8.

Each VP8 RTP packet leaves the worker with its own encoded payload, as we are rewriting the `pictureId` and `tl0PictureIndex` per Consumer. When multiple Consumers are conusming the same Producer, we were not honouring that data on retransmissions.

This change should make mediasoup more resilient on lossy networks.

### DIFF
--- a/worker/test/src/RTC/Codecs/TestVP8.cpp
+++ b/worker/test/src/RTC/Codecs/TestVP8.cpp
@@ -154,6 +154,67 @@ SCENARIO("parse VP8 payload descriptor", "[codecs][vp8]")
 		}
 	};
 
+	SECTION("parse payload descriptor, encode")
+	{
+		/** VP8 Payload Descriptor
+		 *
+		 * 1 = X bit: Extended control bits present (I L T K)
+		 * 1 = R bit: Reserved for future use (Error should be zero)
+		 * 0 = N bit: Reference frame
+		 * 1 = S bit: Start of VP8 partition
+		 * Part Id: 0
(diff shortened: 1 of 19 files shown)


ibc (approved):
CHANGELOG please.

worker/test/src/RTC/Codecs/TestVP8.cpp
```diff
@@ -387,3 +448,88 @@ SCENARIO("process VP8 payload descriptor", "[codecs][vp8]")
...
+	 */
+
+	// clang-format off
+	uint8_t buffer[] =
+	{
+		0x80, 0xe0, 0x01, 0x01,
+		0xe8, 0x40, 0x7a, 0xd8
+	};
+
+	// clang-format on
```
ggoldens: Double comment here?

ibc: And/or wrong blank line before the clang-format comment.

jmillan:
NOTE: I'll add more tests implying the processing of NACK and the check of different retransmitted payloads for the same received RtpPacket.

worker/test/src/RTC/Codecs/TestVP8.cpp
ibc: @jmillan ☝🏻

jmillan:
Tests added. I'll edit the CHANGELOG after all CI is finished and merge.

PR #1534, opened 2025-05-16
Node: close worker process by sending `WORKER_CLOSE` through the channel instead of by sending a SIGINT signal

ibc:
# Details

- I have no idea why (before this PR) `worker.close()` sends a `SIGINT` signal to the worker process instead of sending a `WORKER_CLOSE` request via the channel. In fact `WORKER_CLOSE` request was **not** used (at least in mediasoup Node).
- The problem with that design is that a signal can reach the worker at arbitrary time. For instance, it can reach the worker process sooner than a channel request that was sent before. Or the signal could reach the worker process later than a channel request that was sent later. Meaning that the signal could interrupt legitimate channel requests sent before.
- If instead, we send a close request through the channel, then the close request will reach the worker immediately after previously sent requests (guaranteed since the channel is a `UnixSocket` in stream mode, so order is guaranteed).
- So I've changed it.
- I've also made `Worker.cpp` reply (with success) to the `WORKER_CLOSE` request so mediasoup Node knows it and can close its `this.#channel` when appropiate (and give a proper chance to incoming logs to reach Node land).
- Also improve `Worker.ts` state (`this.subprocessClosed` was not properly set in all cases, or not when it should be set).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Node: Make `worker.close()` close the worker process by sending a `WORKER_CLOSE` request through the channel instead of by sending a SIGINT signal ([PR #1528](https://github.com/versatica/mediasoup/pull/1528)).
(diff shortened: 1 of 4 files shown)


jmillan (approved):
:+1:

PR #1535, opened 2025-05-24
Fix KSVC bitrate in IncreaseLayer

vpalmisano:
In the K-SVC configuration, when we calculate the amount of bitrate required to increase the spatial layer from S1 to S2, we need to subtract the bitrate of the current used layer (https://github.com/versatica/mediasoup/pull/1036). In the current code instead we use `GetBitrate` that outputs the total bitrate used by the current plus the lower layer (S0+S1). The correct getter should be instead `GetSpatialLayerBitrate` that gets the total bitrate of the current spatial layer.


ibc (commented):
Seems like a good catch but let me properly review this on next week.

jmillan (approved):

ibc (approved):
Thanks. Merging.

PR #1539, opened 2025-06-04
Return status code 134 instead of abort when built as a library

satoren:
When built as a library, the program no longer aborts but returns status code 134 instead. This is because environments using the library may not be able to handle signals properly.

We encountered the following error in our environment, which caused server to shut down:
```
(ABORT) RTC::SimulcastConsumer::SendRtpPacket() | failed assertion `producerTsReferenceRtpStream->GetSenderReportNtpMs()': no Sender Report for TS reference RTP stream
```

This change modifies the behavior so that when built as a library, only the worker thread stops instead of shutting down the entire process.

This behavioral change also affects https://github.com/versatica/mediasoup/pull/1394, and I believe this new behavior is more desirable.

### DIFF
--- a/worker/src/lib.cpp
+++ b/worker/src/lib.cpp
@@ -167,10 +167,12 @@ extern "C" int mediasoup_worker_run(
 		// 40 is a custom exit code to notify "unknown error" to the Node library.
 		return 40;
 	}
+#ifndef MS_EXECUTABLE
 	catch (const std::runtime_error& error)
 	{
 		return 134; // 134 is the exit code for SIGABRT.
 	}
+#endif
 }
 
 void IgnoreSignals()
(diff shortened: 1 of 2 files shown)


jmillan:
Makes sense to me. Let's wait few days so @ibc can also review.

ibc (approved):
LGTM. Have you tested it by forcing somehow a crash?

satoren:
I wasn’t sure how to forcibly crash the program with the existing code, so I added code to directly call MS_ABORT and tested its behavior.

worker/src/lib.cpp
```diff
@@ -167,6 +167,10 @@ extern "C" int mediasoup_worker_run(
 		// 40 is a custom exit code to notify "unknown error" to the Node library.
 		return 40;
 	}
+	catch (const std::runtime_error& error)
```
jmillan, line 171: This block does happen for both executable and lib modes, but it' only desired for lib mode.

ibc: Agreed

satoren: ok

ibc:
Merged. Thanks.

PR #1548, opened 2025-06-25
Consumer: add target layer retransmission buffer

### LINKED ISSUE #1547: Potential unrecoverable RTP packet upon un-order reception
From the [forum](https://mediasoup.discourse.group/t/potential-key-frame-corruption-in-simulcastconsumer-due-to-packet-reordering/6732/8?u=ibc)

Imagine this scenario:

1. VP8 keyframe takes two RTP packets with seq numbers 1001 and 1002.
2. Producer device sends both but 1002 arrives to mediasoup first.
3. mediasoup sends NACK requesting retransmission of 1001 but anyway it will arrive later.
4. Packet 1002 is not a keyframe so the SimulcastConsumer that was waiting for a keyframe of that ssrc will ignore it and WONT store it in the retransmission buffer of the consumer.
5. Later packet 1001 arrives and the SimulcastConsumer allows it to pass and it switches to that ssrc stream, so the consumer device receives that packet 1001 which is the first half of the keyframe.
6. Later the producer sends packet 1003 and the consumer device receives it.
7. Then the consumer device sends NACK requesting packet 1002.
8. But 1002 is not in the retransmission buffer of the SimulcastConsumer so mediasoup will not retransmit it.
9. Result is that 1002 will never be sent to the consumer device and hence, after some terrible seconds the consumer device will send PLI requesting a new keyframe.

ibc:
## Details

- Fixes #1547.

## TODO

1. In `SimulcastConsumer` we should only add packets to the new buffer if it's belong to the target spatial layer.
2. Implement it in `PipeConsumer`.

### DIFF
--- a/worker/src/RTC/SimpleConsumer.cpp
+++ b/worker/src/RTC/SimpleConsumer.cpp
@@ -11,6 +11,10 @@
 
 namespace RTC
 {
+	/* Static. */
+
+	static constexpr size_t TargetLayerRetransmissionBufferSize{ 20u };
+
 	/* Instance methods. */
 
 	SimpleConsumer::SimpleConsumer(
@@ -74,6 +78,7 @@ namespace RTC
 		this->shared->channelMessageRegistrator->UnregisterHandler(this->id);
 
 		delete this->rtpStream;
+		this->targetLayerRetransmissionBuffer.clear();
 	}
 
 	flatbuffers::Offset<FBS::Consumer::DumpResponse> SimpleConsumer::FillBuffer(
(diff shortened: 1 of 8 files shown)


worker/src/RTC/SimpleConsumer.cpp
```diff
@@ -457,6 +469,28 @@ namespace RTC
 		// Restore packet fields.
 		packet->SetSsrc(origSsrc);
 		packet->SetSequenceNumber(origSeq);
+
+		// If sent packet was the first packet of a key frame, we may need to send
+		// buffered packets belonging to the same key frame that arrived earlier due
+		// to packet misorder.
+		if (sendPacketsInTargetLayerRetransmissionBuffer)
+		{
+			for (auto& pair : this->targetLayerRetransmissionBuffer)
```
jmillan, line 482: Only those which seq number is higher than the key frame, right?

ibc: Sure, I forgot. Done here: https://github.com/versatica/mediasoup/pull/1548/commits/38298856c6558515c830efc9c7f929cff6be0e42

ibc:
Done in `SimulcastConsumer`: https://github.com/versatica/mediasoup/pull/1548/commits/11680ad530a4cc7cbcbb7a2fbaf4314f758f8464

jmillan (approved):

ibc:
Closed in favour of https://github.com/versatica/mediasoup/pull/1550.

PR #1549, opened 2025-06-25
Consumer: Only drop packets in RTP sequence manager when they belong to current spatial layer

ibc:
## Details

- Do not call `rtpSeqManager->Drop(seq)` if the packet doesn't belong to the current spatial layer. Why? Because if we call `Drop(seq)` in `SimulcastConsumer` no matter the spatial layer (AKA SSRC stream) we are polluting our single `rtpSeqManager` with drops of unrelated seq numbers, and this could perfectly break the ongoing continuous RTP stream.
- This also affects other classes such as `SimpleConsumer` etc. it's not just about simulcast.
- Also move some blocks/checks to have more consistent code across different `XxxxConsumer` classes.
- Also complete many missing things in `PipeConsumer.cpp`.

### Bonus Tracks

- Rename `RtcLogger::Drop()` to `RtcLogger::Discard()` and `DropReason` to `DiscardReason`. This is because we want to only use "drop" term when we also call `Drop()` on the RTP sequence manager while there are packets that are discarded without involving any action in the RTP sequence manager.
- Remove some `inline` keywords in `XxxxConsumer.cpp` files because they are completely useless in 2025.

### DIFF
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -1678,7 +1678,7 @@ namespace RTC
 		}
 	}
 
-	inline bool SimulcastConsumer::CanSwitchToSpatialLayer(int16_t spatialLayer) const
+	bool SimulcastConsumer::CanSwitchToSpatialLayer(int16_t spatialLayer) const
 	{
 		MS_TRACE();
 
@@ -1704,7 +1704,7 @@ namespace RTC
 		// clang-format on
 	}
 
-	inline void SimulcastConsumer::EmitScore() const
+	void SimulcastConsumer::EmitScore() const
 	{
 		MS_TRACE();
 
@@ -1720,7 +1720,7 @@ namespace RTC
 		  notificationOffset);
 	}
 
-	inline void SimulcastConsumer::EmitLayersChange() const
+	void SimulcastConsumer::EmitLayersChange() const
 	{
 		MS_TRACE();
 
@@ -1750,7 +1750,7 @@ namespace RTC
 		  notificationOffset);
 	}
 
-	inline RTC::RtpStreamRecv* SimulcastConsumer::GetProducerCurrentRtpStream() const
+	RTC::RtpStreamRecv* SimulcastConsumer::GetProducerCurrentRtpStream() const
 	{
 		MS_TRACE();
 
@@ -1763,7 +1763,7 @@ namespace RTC
 		return this->producerRtpStreams.at(this->currentSpatialLayer);
 	}
 
-	inline RTC::RtpStreamRecv* SimulcastConsumer::GetProducerTargetRtpStream() const
+	RTC::RtpStreamRecv* SimulcastConsumer::GetProducerTargetRtpStream() const
 	{
 		MS_TRACE();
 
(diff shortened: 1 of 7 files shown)


worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -941,11 +966,8 @@ namespace RTC
 					this->syncRequired       = false;
 					this->spatialLayerToSync = -1;
 
-#ifdef MS_RTC_LOGGER_RTP
-					packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::TOO_HIGH_TIMESTAMP_EXTRA_NEEDED);
```
jmillan, line 945: `logger.Dropped()` reflects when a packet is `discarded` by mediasoup. It is not tied with `SeqManager` by any means.

This call to (and any other that results in not forwarding the packet) must be added. That's the whole purpose of the `RtcLogger`: To know which packets are forwarded and which are discarded, with the corresponding reason.

We can rename `Dropped` to `Discarded` to remove any ambiguity, but those calls should remain.

worker/src/RTC/SimulcastConsumer.cpp
```diff
@@ -1112,10 +1139,6 @@ namespace RTC
 			  origSsrc,
 			  origSeq,
 			  origTimestamp);
-
-#ifdef MS_RTC_LOGGER_RTP
-			packet->logger.Dropped(RtcLogger::RtpPacket::DropReason::SEND_RTP_STREAM_DISCARDED);
```
jmillan, line 1117: Same, this packet is discarded, hence `packet->logger.Dropped()` must be called

jmillan (changes requested):
PR is fine. Just please check my comments and put the discard logs back in place. Feel free to rename `RtcLogger.Dropped()` by `RtcLogger.Discarded()`.

worker/src/RTC/SimulcastConsumer.cpp
ibc: That log was only present in SimulcastConsumer and not in the very same code block in other Consumers. Why?

worker/src/RTC/SimulcastConsumer.cpp
ibc: Should we also log all discarded packets when consumer is not active?

Why should we also log all discarded packets that are discarded because the consumer is consuming a different stream/spatial-layer or temporal layer?

jmillan: > Should we also log all discarded packets when consumer is not active?

I don't even think there is an actual DropReason for that.

> Why should we also log all discarded packets that are discarded because the consumer is consuming a different stream/spatial-layer or temporal layer?

I'm fine if we only log discarded packets from current layer, and I'm fine logging all too. This is a tool for troubleshooting only enabled by a macro. We want to know when and why a packet is discarded  or not when we are troubleshooting.

ibc: In the code above this thread the keyframe of the target layer is being dropped. I agree it makes sense to log it despite that packet obviously doesn't belong to the current spatial layer. I'm adding it back.

ibc: > > Should we also log all discarded packets when consumer is not active?
> 
> I don't even think there is an actual DropReason for that.

Yes we had and we have: `CONSUMER_INACTIVE`.

ibc: I've added back those logs and renamed things:

https://github.com/versatica/mediasoup/pull/1549/commits/0aa4ac0b632f07cdee8d0f3a76cd125efc6366ad

jmillan (approved):

PR #1553, opened 2025-06-26
RtcLogger: Use MS_DUMP() instead of std::cout()

### LINKED ISSUE #1551: `RtcLogger`: Use `MS_DUMP_CLEAN()` instead of `std::cout()`
## Issue

In `RtcLogger::RtpPacket::Log()` we are using std::cout() which means that it's directly written to stdout. When in Node, it means that `Worker.ts` captures it and prints it **ONLY** if `DEBUG` env contains "Worker", and it prefixes the log with "(stdout)".

We should migrate to `MS_DUMP_CLEAN()` which is **always** printed no matter the value of `DEBUG` env.

_NOTE:_ Let's use `MS_DUMP_CLEAN()` instead of `MS_DUMP()` (which should be deprecated soon) because `MS_DUMP_CLEAN()` logs without any prefix and allows indentation.

jmillan:
Closes #1551

Output looks like:

```log
RTC::RtcLogger::Log() | {"timestamp": 4518452978, "recvTransportId": "fe2bf4c2-f939-4d51-b8fd-73ba30b3a196", "producerId": "3d47591c-1ffc-40ce-87c9-b2636ee39df9", "recvRtpTimestamp": 1962299018, "sendRtpTimestamp": 0, "recvSeqNumber": 12595, "sendSeqNumber": 0, "discarded": true, "discardReason": 'RecvRtpStreamNotFound'}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: RtcLogger: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
+- RtcLogger: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
 
 ### 3.16.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changelog
 
-### NEXT
-
-- `RtcLogger`: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
-
 ### 3.16.1
 
 - libuv: Update to v1.51.0 ([PR #1543](https://github.com/versatica/mediasoup/pull/1543)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: RtcLogger: Use MS_DUMP() instead of std::cout() ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
+- Worker: RtcLogger: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
 
 ### 3.16.1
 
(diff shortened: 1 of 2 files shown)


worker/src/RTC/RtcLogger.cpp
```diff
@@ -52,42 +52,41 @@ namespace RTC
 		{
 			MS_TRACE();
 
-			// TODO: Here we are using std::cout() which means that it's directly
-			// written to stdout. When in Node, it means that Worker.ts captures it
-			// and prints it ONLY if DEBUG env contains "Worker", and it prefixes the
-			// log with "(stdout)". We should move to MS_DUMP() or MS_DUMP_CLEAN().
+			std::stringstream ss;
```
ibc, line 56: Please include `<sstream>` header.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### NEXT
+
+- Worker: RtcLogger: Use MS_DUMP() instead of std::cout() ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
```
```suggestion
- Worker: RtcLogger: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
```

ibc (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### NEXT
+
+- Worker: RtcLogger: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
```
ibc: Being a bit pedantic I don't get this syntax:

```
Foo: Bar: xxx
```

```suggestion
- `RtcLogger`: Use `MS_DUMP()` instead of `std::cout()` ([PR #1553](https://github.com/versatica/mediasoup/pull/1553)).
```

No need to mention Worker. Everything related to RTC is in the worker already.
Said that, IMHO we don't need an entry in CHANGELOG for this change. This is not public API and it's only intended for pure debugging.

ibc (approved):

CHANGELOG.md
jmillan: > IMHO we don't need an entry in CHANGELOG for this change

You're totally right :-)

ibc (approved):

PR #1560, opened 2025-07-04
Rust: Don't log error if `xxx.close()` fails due to channel closed

### LINKED ISSUE #1396: [Rust] Error log “XXX closing failed on drop: Channel already closed”
### Your environment

- mediasoup version: f1973891e792eab40ebda54deed163d04aa1ead4

### Issue description
When dynamically creating/deleting Workers, I get a lot of “XXXXXX closing failed on drop: Channel already closed” error logs.
I think this is useless with no way to fix it.

The test code also receives many error logs.
`RUST_LOG=error cargo test -- --nocapture`

If you're open for this I'd be happy to file a PR

ibc:
Fixes #1396

### DIFF
--- a/rust/src/router.rs
+++ b/rust/src/router.rs
@@ -404,10 +404,17 @@ impl Inner {
             {
                 let channel = self.channel.clone();
                 let request = RouterCloseRequest { router_id: self.id };
+
                 self.executor
                     .spawn(async move {
-                        if let Err(error) = channel.request("", request).await {
-                            error!("router closing failed on drop: {}", error);
+                        match channel.request("", request).await {
+                            Err(RequestError::ChannelClosed) => {
+                                println!("router closing failed on drop: Channel already closed");
+                            }
+                            Err(error) => {
+                                error!("router closing failed on drop: {}", error);
+                            }
+                            Ok(_) => {}
                         }
                     })
                     .detach();
--- a/rust/src/router.rs
+++ b/rust/src/router.rs
@@ -409,7 +409,7 @@ impl Inner {
                     .spawn(async move {
                         match channel.request("", request).await {
                             Err(RequestError::ChannelClosed) => {
-                                println!("router closing failed on drop: Channel already closed");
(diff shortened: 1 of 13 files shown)


jmillan (approved):

rust/src/router.rs
```diff
@@ -404,10 +404,17 @@ impl Inner {
...
                 let channel = self.channel.clone();
                 let request = RouterCloseRequest { router_id: self.id };
+
                 self.executor
                     .spawn(async move {
-                        if let Err(error) = channel.request("", request).await {
-                            error!("router closing failed on drop: {}", error);
+                        match channel.request("", request).await {
+                            Err(RequestError::ChannelClosed) => {
+                                println!("router closing failed on drop: Channel already closed");
```
nazar-pc: Did you mean `error!()` or `warn!()` or other log level?

nazar-pc (changes requested):
This doesn't fix anything, it just replaces `error!()` log level with `println!()` for some reason for one of the error cases. What did you actually mean to do here?

rust/src/router.rs
nazar-pc: Or maybe `debug!()` or `trace!()`? `println!()` will simply write it to standard output, which is even more annoying than going through logger.

ibc: Just run `RUST_LOG=error cargo test -- --nocapture` and you will see tons of those false errors. Calling `xxxx.close()` on something that is already closed should never log an **error**.

ibc: What I want is to NOT use `error!()` or `warn!()` when there is no real error.

nazar-pc: I see, make it `debug!()` then, so it can still be seeing if desired, but not shown with a typical info log level.

ibc: Done here: https://github.com/versatica/mediasoup/pull/1560/commits/69d8fc6d3cf3b067b63dbc4196d2f0287ed5400c

nazar-pc (approved):

PR #1561, opened 2025-07-04
mediasoup crate 0.18.2 and mediasoup-sys crate 0.9.3

ibc:


nazar-pc (approved):
Assuming `0.9.3` worker works fine with older library this looks good to me

ibc:
> Assuming `0.9.3` worker works fine with older library this looks good to me

I want to believe we'll know that once Rust CI jobs complete: https://github.com/versatica/mediasoup/actions/runs/16074657241/

nazar-pc:
I don't think that CI job has anything to do with that. What I meant is that by publishing a patch version, it is possible that someone will have `0.18.1` of the library and `0.9.3` of the worker and since they have a compatible semver version, it should still work. Assuming the worker API didn't change it should be the case, if not then a minor version would need to increase instead.

ibc:
> I don't think that CI job has anything to do with that. What I meant is that by publishing a patch version, it is possible that someone will have `0.18.1` of the library and `0.9.3` of the worker and since they have a compatible semver version, it should still work. Assuming the worker API didn't change it should be the case, if not then a minor version would need to increase instead.

In this PR I'm updating patch version of both `mediasoup` and `mediasoup-sys` crates and I'm making `mediasoup` crate depend on the new `mediasoup-sys` crate so I understand we are good.

nazar-pc:
Right, I'm talking about already published `mediasoup` crate. It has `mediasoup-sys = "0.9.1"`, which means it can pull `0.9.2` or `0.9.3` and expects those are fully compatible (meaning only contain fixes/improvements not visible in API).

PR #1562, opened 2025-07-04
Docs: add "Rust crates"

ibc:
## Details

- Add docs about how to pubish Rust crates.
- Remove some old and outdated docs.


nazar-pc (approved):
Makes sense to me. Might add `cargo package` to Extras too, as a risk-free way to check whether the publish will succeed without actually publishing. I think publish should also have `--dry-run` option or something like that.

jmillan (approved):

ibc:
> Makes sense to me. Might add `cargo package` to Extras too, as a risk-free way to check whether the publish will succeed without actually publishing. I think publish should also have `--dry-run` option or something like that.

Done here: https://github.com/versatica/mediasoup/pull/1562/commits/260eb37407fe26962ef3dd01a9007845fb41bd5e

doc/Rust-crates.md
```diff
@@ -36,10 +36,13 @@ cargo publish
 
 - `cargo publish` will create the crate package, check if all necessary dependencies are already present on [crates.io](https://crates.io/), will then compile the package (to ensure that you don't publish a broken version) and will upload it to [crates.io](https://crates.io/).
 - Never publish from random branches or local state that is not on GitHub. If you have local files modified Cargo will refuse to publish until you commit all the changes.
-- If you want to do everything except publishing itself, `cargo package` command exists.
 
 ## Extras
 
+### Check crate without publishing
+
+If you want to do everything except publishing itself, `cargo package` command exists. You can also run `cargo package --dry-run` to avoid package generation.
```
nazar-pc, line 44: And you can probably do `cargo publish --dry-run` to do everything except publishing to `crates.io`.

ibc: Ok ok, but if `cargo publish --dry-run` then `cargo publish` will fail too and we will figure it out :)

PR #1565, opened 2025-07-07
Remove H265 codec and deprecated frame-marking RTP extension

ibc:
## Details

- Remove H265 video codec because we cannot even detect a keyframe on it. It's never been maintained.
- Remove frame-marking RTP extension because it is deprecated and replaced by dependency-descriptor in libwebrtc, meaning that nobody implements frame-marking anymore.


nazar-pc:
Title says H265, description H264, but what do you actually mean here?

ibc:
> Title says H265, description H264, but what do you actually mean here?

H265. Fixed.

jmillan (approved):

PR #1567, opened 2025-07-08
`SimulcastConsumer`: Fix selecting spatial layer higher than preferred one

### LINKED ISSUE #1460: consumer.setPreferredLayer(spatial: 1) makes mediasoup select layer 2 if layer 1 doesn't exist
# Bug Report

## Your environment

- mediasoup version: 3.14.16

## Issue description

If the `Producer` has 3 encodings (so 3 simulcast spatial layers 0,1,2) and the encoding in the middle (encoding 1) has `active: false`, then when the consumer calls `setPreferredLayers({ spatial: 1 })` mediasoup selects layer 2 instead of 0.

CleanShot 2024-10-18 at 17 51 33

This can be tested by doing this in the demo app:

```diff
diff --git a/app/lib/RoomClient.js b/app/lib/RoomClient.js
index 85794c0..ce20ffa 100644
--- a/app/lib/RoomClient.js
+++ b/app/lib/RoomClient.js
@@ -1096,7 +1096,8 @@ export default class RoomClient
 							{
 								scaleResolutionDownBy : 2,
 								maxBitrate            : 1000000,
-								scalabilityMode       : this._webcamScalabilityMode || 'L1T3'
+								scalabilityMode       : this._webcamScalabilityMode || 'L1T3',
+								active                : false
 							}
 						);
 					}
```

## Notes

We have a similar bug with temporal layers in `SvcConsumer`: https://github.com/versatica/mediasoup/issues/1552

ibc:
## Details

- Fixes #1460

## Video

Compare this video to the one in the issue ticket (and take into account the changes made in demo client to trigger it).

CleanShot 2025-07-08 at 13 44 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### NEXT
 
 - Remove H265 codec and deprecated frame-marking RTP extension ([PR #1564](https://github.com/versatica/mediasoup/pull/1564)).
+- `SimulcastConsumer`: Fix selecting spatial layer higher than preferred one ([PR #1565](https://github.com/versatica/mediasoup/pull/1565)).
 
 ### 3.16.5
 
--- a/worker/src/RTC/SimulcastConsumer.cpp
+++ b/worker/src/RTC/SimulcastConsumer.cpp
@@ -464,8 +464,9 @@ namespace RTC
 		{
 			spatialLayer = static_cast<int16_t>(sIdx);
 
-			// If this is higher than current spatial layer and we moved to to current spatial
-			// layer due to BWE limitations, check how much it has elapsed since then.
+			// If this is higher than current spatial layer and we moved to to current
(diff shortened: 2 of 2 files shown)


jmillan (approved):

Lynnworld:
what if  active layer is higher than prefer layer ? eg:  L0:inactive  L1:active , prefer layer is L0
Question is  if the only active layer is higher than we need, shall we choose it ?

ibc:
> what if  active layer is higher than prefer layer ? eg:  L0:inactive  L1:active , prefer layer is L0
> 
> Question is  if the only active layer is higher than we need, shall we choose it ? 

Consumer has info and events about the scores of each stream in the producer. If consumer selects spatial layer 0 as preferred then we move to 0 no matter 0 is inactive. The consumer has the info to decide.

PR #1568, opened 2025-07-09
Remove H264-SVC codec

ibc:
## Details

- `H264-SVC` codec processing in mediasoup is not maintained (it was an external contribution).
- H.264-SVC codec is basically academic and marginal, and just a very specific hardware devices and service providers support (or supported) it:
  - Old models of Cisco TelePresence.
  - Polycom VoIP phones.
  - Vidyo (known for using H.264 SVC in its own stack before being acquired by Enghouse).
  - VLC and ffmpeg can decode certain H.264 SVC profiles, but cannot encode it properly.
  - x264, the most common H.264 encoder, does not implement H.264 SVC.
  - No browser, mobile device, or commercial GPU supports it natively.
- Also, after removing Frame-Marking RTP extension, many chunks of logic in `H264_SVC.cpp` have been removed.
- Also remove `X-H264UC` (it was only defined in the `RtpCodecMimeType` enum, not used at all) which is a Microsoft proprietary stuff nobody cares about anymore.


jmillan (approved):
🧹 👍

PR #1571, opened 2025-07-14
`Router`: Add `updateMediaCodecs()` method to dynamically change Router's RTP capabilities

### LINKED ISSUE #1564: Appending mediaCodecs to an existing Router
# Feature Request
In my application, when a new client connects, it informs the server of all it’s supported capabilities; the server then finds the best match across all clients in one room and broadcasts the new selected codec. 
While the majority of cases can be easily covered by specifying most browser defaults ahead of time, there is still a gap in the event that all clients support, for example, a higher H264 profile, but the router was not configured to handle it.
Therefore, I would like to request a way to append or modify the supported codecs of an existing Router, ensuring the preferred available codec is always supported.
This also ties in together with #1126 as it would not only prevent the need to reconfigure clients in the event an existing payload type changes, but also avoid the exhaustion of dynamic range payload types.

ibc:
## Details

- Added `router.updateMediaCodecs(mediaCodecs)` method that recomputes Router's RTP capabilities.
- Fixes #1564

### DIFF
--- a/rust/src/router.rs
+++ b/rust/src/router.rs
@@ -67,7 +67,6 @@ use std::fmt;
 use std::net::{IpAddr, Ipv4Addr};
 use std::ops::Deref;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Mutex as SyncMutex;
 use std::sync::{Arc, Weak};
 use thiserror::Error;
 
@@ -382,7 +381,7 @@ struct Handlers {
 struct Inner {
     id: RouterId,
     executor: Arc<Executor<'static>>,
-    rtp_capabilities: SyncMutex<RtpCapabilitiesFinalized>,
+    rtp_capabilities: Arc<RwLock<RtpCapabilitiesFinalized>>,
     channel: Channel,
     handlers: Arc<Handlers>,
     app_data: AppData,
@@ -493,7 +492,7 @@ impl Router {
         let inner = Arc::new(Inner {
             id,
             executor,
-            rtp_capabilities: SyncMutex::new(rtp_capabilities),
+            rtp_capabilities: Arc::new(RwLock::new(rtp_capabilities)),
             channel,
             handlers,
             producers,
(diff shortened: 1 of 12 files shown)


rust/src/router.rs
```diff
@@ -533,8 +542,8 @@ impl Router {
     /// * See also how to [filter these RTP capabilities](https://mediasoup.org/documentation/v3/tricks/#rtp-capabilities-filtering)
     ///   before using them into a client.
     #[must_use]
-    pub fn rtp_capabilities(&self) -> &RtpCapabilitiesFinalized {
-        &self.inner.rtp_capabilities
+    pub fn rtp_capabilities(&self) -> RtpCapabilitiesFinalized {
+        self.inner.rtp_capabilities.lock().unwrap().clone()
```
ibc: I don't understand why I cannot have a mutable internal property and a method that returns its reference. Or maybe I'm wrong and it's possible.

nazar-pc: You absolutely can, but you don't have a mutable reference. What you have is a `Mutex`, which wraps the value. You can lock a `Mutex` to get a `MutexGuard` and temporary look at the internals, when `MutexGuard` is dropped, `Mutex` is unlocked automatically and some else can lock it now.

The reason you can work with it as a reference is that `MutexGuard` implements `Deref` and `DerefMut` to `Mutex`'s internal contents. You can't drop `MutexGuard` in the method and return a dereference to `Mutex`'s contents at the same time, that would be unsound.

ibc: So then is it ok as it is now or not?

ibc: Note that the downside is the extra clone() in the getter.

nazar-pc: The way it is right now if fine. Depending on how "hot" access is, it might make sense to wrap the value in `Arc`, which will make cloning much cheaper, but I don't think that complexity is needed here.

ibc: But that would change the API since the getter would now return an Arc

nazar-pc: Of course. Well, the API has changed already due to return of an owned struct, but it is not really avoidable here.

ibc:
@nazar-pc what do you thing about this?

https://github.com/versatica/mediasoup/pull/1571/commits/eea7922cde20c060df8db194ad9439e1314630f1

jmillan (approved):

nazar-pc:
> @nazar-pc what do you thing about this?

Honestly, I don't like it. It exposes internals and is really awkward to use. Yes, you can potentially save clones in some cases, but as you can see from the PR itself, you're just making ergonomics much worse and end up cloning in most cases anyway.

ibc:
> > @nazar-pc what do you thing about this?
> 
> Honestly, I don't like it. It exposes internals and is really awkward to use. Yes, you can potentially save clones in some cases, but as you can see from the PR itself, you're just making ergonomics much worse and end up cloning in most cases anyway.

Fair, reverted here: https://github.com/versatica/mediasoup/pull/1573

PR #1572, opened 2025-07-15
mediasoup-types

jmillan:
New crate holding mediasoup types.

In order to be used by other crates to signal data to mediasoup crate and vice versa.

This crate permits not having to depend on the mediasoup crate which implies the C++ worker compilation.

This crate exports:

```
pub mod data_structures;
pub mod rtp_parameters;
pub mod scalability_modes;
pub mod sctp_parameters;
pub mod srtp_parameters;
```

The FBS conversion has been completely moved out of this types definition and it's implemented in the corresponding mediasoup files `data_structures.rs` and `rtp_parameters.rs`, etc, via the corresponding new FBS conversion traits defined in fbs.rs.

The API does not change because such types are still being exported by `mediasoup`, although marked as deperecated, plus all types are re-exported by `mediasoup` under the name `types`:

````
pub use mediasoup_types as types;
#[deprecated(note = "Deprecated in favour of types::data_structures")]
pub mod data_structures;
#[deprecated(note = "Deprecated in favour of types::rtp_parameters")]
pub mod rtp_parameters;
#[deprecated(note = "Deprecated in favour of types::scalability_modes")]
pub mod scalability_modes;
#[deprecated(note = "Deprecated in favour of types::sctp_parameters")]
pub mod sctp_parameters;
#[deprecated(note = "Deprecated in favour of types::srtp_parameters")]
pub mod srtp_parameters;
``` 

Closes #1498


rust/src/data_structures.rs
ibc, line 1: This file now only contains `FromFbs()` and `ToFbs()` functions. Does it make sense that it's called `data_structures.rs`? I don't think so. Maybe they could be moved to the new `rust/src/fbs.rs` file.

jmillan: `fbs.rs` defines the traits to convert to/from FBS.

`data_structures.rs` holds the trait implementations for `data_structure.rs` in `mediasoup_types`.
`rtp_parameters.rs` holds the trait implementations for `rtp_parameters.rs` in `mediasoup_types`.

Other `mediasoup` code _do also_ implement the traits defined in `fbs.rs`. It's not a good idea putting in that file all implementations of the traits. The traits need to be implemented wherever the types are defined and no outside. `data_structure.rs` and `rtp_parameters.rs` are an exception because they are not defined in `mediasoup` crate anymore.

ibc: Not sure if I understand. The only thing I say is that IMHO it doesn't make sense to have a file named `data_structures.rs` if in only exposes `FromFbs()` and `ToFbs()` functions. And I mean the `rust/src/data_structures.rs`, not the new one in `fbs` folder.

jmillan: Let me rewrite my previos comment with full paths:

`fbs.rs` in `mediasoup` defines the traits to convert to/from FBS.

`data_structures.rs` in `mediasoup` holds the trait implementations for `data_structure.rs` in `mediasoup_types`.
`rtp_parameters.rs` in `mediasoup` holds the trait implementations for `rtp_parameters.rs` in `mediasoup_types`.

Other `mediasoup` code _do also_ implement the traits defined in `fbs.rs`. It's not a good idea putting in that file all implementations of the traits. The traits need to be implemented wherever the types are defined and no outside. `data_structure.rs` and `rtp_parameters.rs` in `mediasoup` are an exception because those types are not defined in `mediasoup` crate anymore.

Ie: `active_speaker_observer.rs` in `mediasoup` does also implement one of such traits. We DON'T want to move that implementation to `fbs.rs` but keep it in `active_speaker_observer.rs`.

In a future more `mediasoup` types may implement those traits rather than just having a method named `from_fbs`, `to_fbs`, but it's a lot of manual work that I'm not about to do right now.

ibc (approved):

rust/types/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![warn(rust_2018_idioms, missing_debug_implementations)]
+//! MediaSoup Rust Types
+//!
+//! This library provides Rust types for interacting with [mediasoup](https://docs.rs/mediasoup) crate.
+//!
```
nazar-pc:
```suggestion
```

rust/types/Cargo.toml
nazar-pc, line 1: Not a fan of packages in random subdirectories. What do you think about creating `rust/mediasoup` and `rust/mediasoup-types` instead?

rust/src/ortc.rs
```diff
@@ -566,6 +572,7 @@ pub(crate) fn get_consumable_rtp_parameters(
                     rtcp_feedback: rtcp_feedback.clone(),
                 }
             }
+            _ => panic!("Unexpected codec type"),
```
jmillan: @nazar-pc, this is set here because `RtpCodecCapabilityFinalized`, which is now defined in `mediasoup-types` crate, is `non_exhaustive`.

rust/src/fbs.rs
```diff
@@ -0,0 +1,26 @@
+//! Traits for converting between Rust and FlatBuffers data structures.
+
+pub(crate) trait TryFromFbs<'a>: Sized {
+    type FbsType;
+    type Error;
+
+    fn try_from_fbs(fbs: Self::FbsType) -> Result<Self, Self::Error>;
+}
+
+pub(crate) trait FromFbs: Sized {
```
jmillan, line 10: These traits take the argument via value or reference. There is no strong reason for each as both have their pros and cons. Any consideration here @nazar-pc ?

rust/src/router/pipe_transport.rs
```diff
@@ -159,10 +160,7 @@ impl PipeTransportDump {
                 .sctp_parameters
                 .as_ref()
                 .map(|parameters| SctpParameters::from_fbs(parameters.as_ref())),
-            sctp_state: dump
-                .base
-                .sctp_state
-                .map(|state| SctpState::from_fbs(&state)),
+            sctp_state: dump.base.sctp_state.map(SctpState::from_fbs),
```
nazar-pc: BTW I think you may also be able implement additional helper impls this (pseudo-code, may not compile):
```rust
impl<Source, Target> FromFbs for Option<Target>
where
    Source: FromFbs,
{
    type FbsType = Option<Source>;

    fn from_fbs(value: Self::FbsType) -> Self {
       value.map(Self::FbsType::from_fbs)
    }
}
```

Then also helper for inverse conversion like `From`/`Into` symmetry and many cases could be simplified:
```suggestion
            sctp_state: dump.base.sctp_state.to_fbs_source(),
```

rust/src/router/pipe_transport.rs
```diff
@@ -173,7 +171,7 @@ impl PipeTransportDump {
                 .map(TransportTraceEventType::from_fbs)
                 .collect(),
             // PipeTransport specific.
-            tuple: TransportTuple::from_fbs(dump.tuple.as_ref()),
+            tuple: TransportTuple::from_fbs(*dump.tuple),
```
nazar-pc: IIRC Flatbuffers requires allocations anyway, in which case can't we continue using references everywhere?

nazar-pc (commented):
A few high-level comments, overall seems like a lot of tedious moving code around. Thanks for doing this!

rust/types/Cargo.toml
jmillan: yes, I like it better :+1:

rust/src/lib.rs
```diff
@@ -49,6 +49,7 @@
 //! and are a good place to start until we have demo apps built in Rust).
 
 pub mod data_structures;
+pub(crate) mod fbs;
```
nazar-pc, line 54: Consider also re-exporting `mediasoup-type`:
```rust
pub use mediasoup_types as types;
```

Otherwise it'll be quite annoying to use because any usage of the type not re-exported through prelude will require adding a separate dependency.

rust/Cargo.toml
nazar-pc, line 1: This required minor version bump since a bunch of data structures disappeared from corresponding modules and now live in a separate crate.

You can make temporary compatibility via re-exports, but I don't think it is worth it.

rust/types/Cargo.toml
jmillan: This would require moving around all existing rust/ code to rust/mediasoup. I'll do it *after* merging this PR to keep things clear.

rust/src/ortc.rs
nazar-pc: What is the point of making it `non_exhaustive` though? I don't think we'll have anything other than audio/video any time soon

rust/src/fbs.rs
nazar-pc: I'd say use a reference if you can get away with it and do not reuse allocation anyway

rust/src/lib.rs
jmillan: The way I've done it is for `mediasoup` users to keep using `rtp_parameters` and `data_structures` from `mediasoup` and those being taken from `mediasoup-types`. Like [here](https://github.com/versatica/mediasoup/blob/0954609ee506710f488d03e3f9ba632cc15169cf/rust/src/data_structures.rs#L6)

I'll add your suggestion too.

nazar-pc: I see. That makes sense to me as a compatibility layer, though I'd probably purposefully re-export just the required types and added `#[deprecated(note = "Explanation")]` attribute on it, so folks can upgrade properly. And a TODO to remove in the next breaking release.

rust/src/router/pipe_transport.rs
jmillan: Do you mean, to avoid re-allocating when we have a FBS reference?

nazar-pc: Yes. I don't think it makes a huge difference here, but generally if we do not consume input type, the argument should be a reference.

rust/src/router/pipe_transport.rs
jmillan: This implementation would take an `Option` of a type implementing `FromFbs`:

```rust
impl<T> FromFbs for Option<T>
where
    T: FromFbs,
{
    type FbsType = Option<T::FbsType>;

    fn from_fbs(value: Self::FbsType) -> Self {
        value.map(T::from_fbs)
    }
}
```

This can later be used as:

```rust
sctp_state: FromFbs::from_fbs(dump.base.sctp_state) // dump.base.sctp_state is of type Option<fbs::SctpState>
```

nazar-pc: Yes

rust/src/router/pipe_transport.rs
jmillan: Done.

rust/src/router/pipe_transport.rs
jmillan: I've already done this. Since there are still `mediasoup` (no `mediasoup-types`) types that do have their own `to_fbs` and `from_fbs` rather than implementing the new traits, there is place for enhancement but I'm not going to implement all those traits in this PR.

rust/src/fbs.rs
jmillan: Done

rust/Cargo.toml
jmillan: Considering that we are doing both, keeping current types (by re-exporting from the mediasoup files) and re-exporting all mediasoup-types. Plus we are adding a deprecated flag, do we still need a minor version bump?

rust/src/ortc.rs
jmillan: Yep, it was like that (you make it :-)). But I don't see how we are going to have more types. Removing...

rust/Cargo.toml
nazar-pc: Assuming everything still works just like it did before (same types, same API, etc.) then no

rust/src/ortc.rs
nazar-pc: I think it was done so you can't construct it from externally, only using provided APIs :thinking:

jmillan:
@nazar-pc, @ibc , I've added all types to the new crate (please check the updated PR description) and handled all the comments. ([still one unclosed](https://github.com/versatica/mediasoup/pull/1572#discussion_r2209935580)).

Please have a final review at your convenience.

rust/src/ortc.rs
```diff
@@ -1,13 +1,16 @@
...
-    RtpEncodingParameters, RtpEncodingParametersRtx, RtpHeaderExtensionDirection,
-    RtpHeaderExtensionParameters, RtpHeaderExtensionUri, RtpParameters,
+    RtpCapabilitiesFinalized, RtpCodecCapability, RtpCodecCapabilityFinalized, RtpCodecParameters,
+    RtpCodecParametersParameters, RtpCodecParametersParametersValue, RtpEncodingParameters,
+    RtpEncodingParametersRtx, RtpHeaderExtensionDirection, RtpHeaderExtensionParameters,
+    RtpHeaderExtensionUri, RtpParameters,
 };
-use crate::scalability_modes::ScalabilityMode;
 use crate::supported_rtp_capabilities;
+
```
nazar-pc:
```suggestion
```

rust/src/ortc.rs
```diff
@@ -1,13 +1,16 @@
...
     MediaKind, MimeType, MimeTypeAudio, MimeTypeVideo, RtcpFeedback, RtcpParameters,
-    RtpCapabilities, RtpCapabilitiesFinalized, RtpCodecCapability, RtpCodecCapabilityFinalized,
-    RtpCodecParameters, RtpCodecParametersParameters, RtpCodecParametersParametersValue,
-    RtpEncodingParameters, RtpEncodingParametersRtx, RtpHeaderExtensionDirection,
-    RtpHeaderExtensionParameters, RtpHeaderExtensionUri, RtpParameters,
+    RtpCapabilitiesFinalized, RtpCodecCapability, RtpCodecCapabilityFinalized, RtpCodecParameters,
+    RtpCodecParametersParameters, RtpCodecParametersParametersValue, RtpEncodingParameters,
+    RtpEncodingParametersRtx, RtpHeaderExtensionDirection, RtpHeaderExtensionParameters,
+    RtpHeaderExtensionUri, RtpParameters,
 };
```
nazar-pc: Shouldn't these be imported from `mediasoup_types` directly?

ibc:
Issue https://github.com/versatica/mediasoup/pull/1498 is NOT linked to this PR.

Please add a proper `Fixes #1498` comment in the PR description (not within a code block).

**UPDATE:** Ok, it doesn't close #1498 because it's a PR XD, ignore please. But that PR must be closed.

ibc (approved):

rust/src/ortc.rs
jmillan: yes, fixed. I'll check for other imports like that.

jmillan: All types are now imported from `mediasoup-types` directly.

nazar-pc (approved):
I didn't check every line, but high level makes sense

rust/src/ortc.rs
jmillan: ahh, I see. Well, members are private so they cannot be constructed externally anyway.

rust/Cargo.toml
jmillan: Yep, well, integration tests are working fine so we should be good.

rust/src/ortc.rs
nazar-pc: Which members are private? All enum fields have the same visibility as the enum iself.

To hide them you'd have to do something like `RtpCodecCapabilityFinalized::Video(RtpCodecCapabilityFinalizedVideo {})`. Then `RtpCodecCapabilityFinalizedVideo` could have private fields or be `#[non_exhaustive]`.

jmillan: Oh, I see. 

I think we can live without non_exhaustive as there is no fore coming intention to add a new type.

jmillan: > Which members are private? 

I thought for a moment that this was a struct rather than an enum.

jmillan:
@nazar-pc , the `deprecated` tag does not have any effect [*] (I get no warnings from using it). I would say because in the end it's using the re-exported types from `mediasoup-types`. Not a problem IMO.

[*]: Using `mediasoup::data_structures` in integration tests generates no warning.

nazar-pc:
I think that is fine, just need to not forget to drop it once we introduce breaking changes.

There are tools that can check for backwards compatibility (they pull compatible version from crates.io and compare APIs IIRC), but I have not personally used them.

nazar:
WRONG @nazar !!

nazar-pc:
Sorry :see_no_evil:

ibc:
Hi @jose

jmillan:
> WRONG @nazar !!

Sorry !!

PR #1576, opened 2025-07-17
mediasoup-types: Fix manifest documentation URL

jmillan:


rust/types/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Type definitions and shared data structures for the mediasoup crate"
 edition = "2021"
 license = "ISC"
-documentation = "https://docs.rs/mediasoup-types"
+documentation = "https://docs.rs/mediasoup_types"
```
nazar-pc, line 7: Are you sure? I think `-` should be supported as well and in fact is the correct one

jmillan: Yesterday just upon releas only the '_' worked. Today both work. I'll update it.

PR #1578, opened 2025-07-17
don't skip counting padding-only packets

### LINKED ISSUE #1577: Simulcast consumer is deactivated due to padding-only packets on Producer side
## Your environment

- mediasoup version: 3.16.6

## Issue description

We are using VP8 Simulcast in our application and have occasionally observed that the video freezes for a few seconds.

After examining RTP traces and mediasoup debug logs, we believe the issue is caused by the handling of padding-only packets from producing browser.

1. Padding-only packets are discarded in `RtpStreamRecv:ReceivePacket`, so the packet count in the `mediaTransmissionCounter` is not updated.
2. When the score is calculated in `RtpStreamRecv:UpdateScore`, `getExpectedPackets` (based on sequence numbers) will count the padding packets, but `mediaTransmissiongCounter.getPacketCount` does not. The difference is incorrectly treated as packet loss.
3. As a result the score drops to zero
4. In `SimulcastConsumer:RecalculateTargetLayers`,  any producer stream with a score of 0 is considered inactive and is ignored. If there is no other Simulcast stream, `newTarget[Spatial|Temoral]Layer` is set to {-1, -1}
5.Similarly `SimulcastConsumer::ProducerRtpStreamScore` calls `UpdateTargetLayers(-1, -1)` because the `isActive()` check fails (again due to producer score being 0)

Attached is a log showing 5 out of 28 packets are non-padding only, causing layer change to {-1, -1}

[padding_only_packets.txt](https://github.com/user-attachments/files/21302378/padding_only_packets.txt)

>   mediasoup:Channel [pid:33245] RTC::RtpStreamRecv::UpdateScore() | [totalExpected:28, totalReceived:5, totalRepaired:0 +113ms
>   mediasoup:Channel [pid:33245] RTC::RtpStreamRecv::UpdateScore() | fixed values [expected:28, received:5, lost:5, repaired:0, retransmitted:0 +0ms
>   mediasoup:Channel [pid:33245] RTC::RtpStreamRecv::UpdateScore() | [deliveredRatio:0.000000, repairedRatio:0.000000, repairedWeight:1.000000, new lost:5, score:0] +0ms
>   mediasoup:Channel [pid:33245] RTC::RtpStream::UpdateScore() | [added score:0, previous computed score:10, new computed score:0] (calling listener) +0ms
>   mediasoup:Channel [pid:33245] RTC::SimulcastConsumer::UpdateTargetLayers() | target layers changed [spatial:-1, temporal:-1, consumerId:2f6a7206-d497-436e-823b-0559ad5bf92b] +0ms

quanli168:
This is to address [Issue 1577](https://github.com/versatica/mediasoup/issues/1577)
Fixes #1577

Padding-only RTP packets should be included in the received packet count to ensure accurate producer score calculation.

### DIFF
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -303,18 +303,17 @@ namespace RTC
 		// Calculate Jitter.
 		CalculateJitter(packet->GetTimestamp());
 
-		// Padding only packet, do not consider it for counter increase nor
-		// stream activation.
-		if (packet->GetPayloadLength() == 0)
-		{
-			return true;
-		}
-
 		// Increase transmission counter.
 		this->transmissionCounter.Update(packet);
 
 		// Increase media transmission counter.
 		this->mediaTransmissionCounter.Update(packet);
+
+		// Padding only packet, do not consider it for stream activation.
+		if (packet->GetPayloadLength() == 0)
+		{
+			return true;
+		}
 
 		// Not inactive anymore.
 		if (this->inactive)


ibc:
Looks legit. Thanks a lot. Let's review and merge/release next week.

jmillan (commented):
`mediaTransmissionCounter` must not be increased, as this packet contains no media.

ibc:
> `mediaTransmissionCounter` must not be increased, as this packet contains no media.

I haven't checked it but, are you 100% that skipping this media transmission counter will not affect Consumers the way is described in the issue this PR fixes?

ibc:
> > `mediaTransmissionCounter` must not be increased, as this packet contains no media.
> 
> I haven't checked it but, are you 100% that skipping this media transmission counter will not affect Consumers the way is described in the issue this PR fixes?

For example, in `RtpStreamRecv::GetRtcpReceiverReport()` we take into account `this->mediaTransmissionCounter` to compute packet lost stats. Why should we discard it if the packet is padding only?

```c++
RTC::RTCP::ReceiverReport* RtpStreamRecv::GetRtcpReceiverReport()
{
	MS_TRACE();

	uint8_t worstRemoteFractionLost{ 0 };

	if (this->params.useInBandFec)
	{
		// Notify the listener so we'll get the worst remote fraction lost.
		static_cast<RTC::RtpStreamRecv::Listener*>(this->listener)
		  ->OnRtpStreamNeedWorstRemoteFractionLost(this, worstRemoteFractionLost);

		if (worstRemoteFractionLost > 0)
		{
			MS_DEBUG_TAG(rtcp, "using worst remote fraction lost:%" PRIu8, worstRemoteFractionLost);
		}
	}

	auto* report = new RTC::RTCP::ReceiverReport();

	report->SetSsrc(GetSsrc());

	const uint32_t prevPacketsLost = this->packetsLost;

	// Calculate Packets Expected and Lost.
	auto expected = GetExpectedPackets();

	if (expected > this->mediaTransmissionCounter.GetPacketCount())
	{
		this->packetsLost = expected - this->mediaTransmissionCounter.GetPacketCount();
	}
	else
	{
		this->packetsLost = 0u;
	}
```

And also in `RtpStreamRecv::UpdateScore()`:

```c++
void RtpStreamRecv::UpdateScore()
{
	MS_TRACE();

	// Calculate number of packets expected in this interval.
	const auto totalExpected = GetExpectedPackets();
	const uint32_t expected  = totalExpected - this->expectedPriorScore;

	this->expectedPriorScore = totalExpected;

	// Calculate number of packets received in this interval.
	const auto totalReceived = this->mediaTransmissionCounter.GetPacketCount();
	const uint32_t received  = totalReceived - this->receivedPriorScore;

	this->receivedPriorScore = totalReceived;

	// Calculate number of packets lost in this interval.
	uint32_t lost;

	if (expected < received)
	{
		lost = 0;
	}
	else
	{
		lost = expected - received;
	}
```


Another story is: why are we doing `if (expected > this->mediaTransmissionCounter.GetPacketCount())` instead of `if (expected > this->transmissionCounter.GetPacketCount())`? And why are we doing `const auto totalReceived = this->mediaTransmissionCounter.GetPacketCount()` instead of `const auto totalReceived = this->transmissionCounter.GetPacketCount()`?

**UPDATE:** Auto-reply: because `mediaTransmissionCounter` counts packets sent on this SSRC stream while `transmissionCounter` considers both packets sent over the main RTP stream and also packets sent over the RTX stream.

But then, @jmillan, if we follow your concern above:

> mediaTransmissionCounter must not be increased, as this packet contains no media.

Then we are NOT fixing issue #1577 because the score depends on the `mediaTransmissionCounter`.

ibc:
## TODO

Despite the ongoing discussion in comments above, this PR should also address the potentially same problem in `RtpStreamRecv::ReceiveRtxPacket()` in which we do something similar:

```c++
// Pass the packet to the NackGenerator and return true just if this was a
// NACKed packet.
if (this->nackGenerator->ReceivePacket(packet, /*isRecovered*/ true))
{
	// Padding only packet, do not consider it for counter increase nor
	// stream activation.
	if (packet->GetPayloadLength() == 0)
	{
		return true;
	}

	// Mark the packet as repaired.
	RTC::RtpStream::PacketRepaired(packet);

	// Increase transmission counter.
	this->transmissionCounter.Update(packet);
```

ibc:
I think we did it very wrong when we started ignoring padding packets. See this (for example among many other concerns said above):

- In `RtpStream.hpp` we have:
  ```c++
  uint32_t GetExpectedPackets() const
  {
  	return (this->cycles + this->maxSeq) - this->baseSeq + 1;
  }
  ```
- Such a `this->maxSeq` is computed/updated in `RtpStream::UpdateSeq(packet)` in `RtpStream.cpp` (which knows NOTHING about padding packets) and `RtpStream::UpdateSeq(packet)` is called from `RtpStream::ReceiveStreamPacket()` which is called **always** (no matter it's a padding only packet) from `RtpStreamRecv::ReceivePacket()`.
- And then in `RtpStreamRecv::UpdateScore()` we have:
  ```c++
  void RtpStreamRecv::UpdateScore()
  {
  	MS_TRACE();
  
  	// Calculate number of packets expected in this interval.
  	const auto totalExpected = GetExpectedPackets();
  	const uint32_t expected  = totalExpected - this->expectedPriorScore;
  
  	this->expectedPriorScore = totalExpected;
  
  	// Calculate number of packets received in this interval.
  	const auto totalReceived = this->mediaTransmissionCounter.GetPacketCount();
  	const uint32_t received  = totalReceived - this->receivedPriorScore;
  
  	this->receivedPriorScore = totalReceived;
  
  	// Calculate number of packets lost in this interval.
  	uint32_t lost;
  
  	if (expected < received)
  	{
  		lost = 0;
  	}
  	else
  	{
  		lost = expected - received;
  	}
  ```
- Note that there we are comparing `GetExpectedPackets()` (which knows nothing about padding only packets) with `this->mediaTransmissionCounter.GetPacketCount()` (which ignores padding only packets).

So there is a real and complex bug here. We cannot just ignore padding packets the way we are ignoring them. Honestly I think that the proper way to ignore them should be in each `XxxxConsumer` but it wouldn't be good because then it would mean that when we estimate the bitrate of the `rtpStreamRecv` we are getting higher values than the real bitrate that the `Consumer` will really send to the consuming endpoint.

ibc:
Maybe we should have an additional `this->paddingOnlyTransmissionCounter` and take it into account where needed?

jmillan:
Indeed I think this PR is just fine:

```hpp
		// Valid media + valid RTX.
		TransmissionCounter transmissionCounter;
		// Just valid media.
		RTC::RtpDataCounter mediaTransmissionCounter;

```

A padding only packet is neither of them, so it's fine not setting them, indeed it would be incorrect setting any of them.

jmillan:
After a call with @ibc, we came to the realisation of not having to update the rate in [RateCalculator::Update()](https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/RateCalculator.cpp#L199) for padding only packets.

ibc:
I'm closing this PR in favour of PR https://github.com/versatica/mediasoup/pull/1580.

Thanks a lot @quanli168.

PR #1583, opened 2025-07-28
`ListenInfo`: Add `exposeInternalIp` field

ibc:
## Details

- Add `ListenInfo.exposeInternalIp` boolean (Defaults to `false`).
- If set to `true` and `announcedAddress` is set, it exposes an additional ICE candidate in `WebRtcTransport` whose IP is `listenInfo.ip` rather than `listenInfo.announcedAddress`.

### What???

Imagine this not that crazy use-case:

- You deploy your SFU in AWS so the host has a private IP 192.168.1.101 and a public IP 9.8.7.6 is assigned and routed to the host.
- You want to have a single `WebRtcServer` that listens on 192.168.1.101:55555.
- Users connect from Internet, so you need to expose IP 9.8.7.6 in the ICE candidates.
- For a local RTP recorder that implements WebRTC you need local network communication, so you need to expose IP 192.168.1.101 in the ICE candidates.

You may want to do this:

```ts
const webRtcServer = await router.createWebRtcServer({
  listenInfos: [
    {
      protocol: 'udp',
      ip: '192.168.1.101',
      announcedAddress: '9.8.7.6',
      port: 55555,
    },
    {
      protocol: 'udp',
      ip: '192.168.1.101',
      port: 55555,
    },
  ],
});
```

However that's gonna fail because you are telling mediasoup to bind on UDP 192.168.1.101:55555 twice, so it will fail with "address already in use" error.

Now, with this feature you can do this:

```ts
const webRtcServer = await router.createWebRtcServer({
  listenInfos: [
    {
      protocol: 'udp',
      ip: '192.168.1.101',
      announcedAddress: '9.8.7.6',
      exposeInternalIp: true,
      port: 55555,
    },
  ],
});
```

And that's all. When later you create a `WebRtcTransport` (by providing it with the above `WebRtcServer` to it) and you check its `iceCandidates`, there will be 2 ICE candidates:

1. `IP: 9.8.7.6, port: 55555, protocol: UDP`
2. `IP: 192.168.1.101, port: 55555, protocol: UDP`

## TODO

- [x] Document it in the website.


jmillan (approved):

nazar-pc (commented):
Why does it need to be specified in `listenInfos` though? You can already use whatever IP you want to establish the connection, including internal IP. Yes, you'll only get candidates with public IP, but it should be easy to modify the list of candidates after it was generated if the application requires it.

ibc:
> Why does it need to be specified in `listenInfos` though? You can already use whatever IP you want to establish the connection, including internal IP. Yes, you'll only get candidates with public IP

That approach would make sense if we took it from the very beginning. However we didn't. We already generate ICE candidates with announced IP instead of binding IP. This is: we moved the ICE candidate generation logic to the worker already (this includes priority value etc) and hence it makes sense to also put this logic into the worker (so in ListenInfo). Opinionable decision, I know, but this is what we chose already.

> but it should be easy to modify the list of candidates after it was generated if the application requires it.

It should have been the other way around: you only pass bindable IP and then in Node or Rust you replace the Candidate IP with the public one (announced IP), but we didn't do that. So despite I agree with you, this feature is consistent with what we had already.

nazar-pc:
I guess it makes sense then

ibc:
> I guess it makes sense then

Alternatively it doesn't make any sense but it's consistent.

PR #1584, opened 2025-07-30
Make `RtpCodecCapability.preferredPayloadType` mandatory

ibc:
## Details

- Make `RtpCodecCapability.preferredPayloadType` mandatory. This is because it was mandatory in the already deprecated ORTC spec and because it's also mandatory in latest `mediasoup-client` 3.14.0, and this is because it makes sense that it's mandatory.
- The only case in which `preferredPayloadType` is optional is in the `mediaCodecs` given to `worker.createRouter()` and this is because, if `preferredPayloadType` is not given in a codec then mediasoup assigns a random one to it. So that's an obvious exception and not the general rule.
- So also add a new `RouterRtpCodecCapability` type in which `preferredPayloadType` is optional, and use it in `mediaCodecs` options in `worker.createRouter()` and also in `router.updateMediaCodecs()`.
- So also add a new `RouterRtpCapabilities` type in which `codecs` array has type `RouterRtpCodecCapability[]`, and use it in `supportedCapabilities.ts`. 
- And make `codecs` field in `RtpCapabilities` be `(RtpCodecCapability | RouterRtpCodecCapability)[]`.

### Note

This is a breaking change in the sense that if an application declares a variable `mediaCodecs: RtpCodecCapability[]` and introduces entries without `preferredPayloadType` then TS is gonna fail. But still this change makes sense.

## TODO

- [x] Document the new type in the website and update types in existing API (`worker.createRouter()`  and `router.updateMediaCodecs()`).
- [x] Announce the change in the forum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Node: Make `RtpCodecCapability.preferredPayloadType` mandatory and add `RouterRtpCodecCapability` in which `preferredPayloadType` is optional ([PR #15XX](https://github.com/versatica/mediasoup/pull/15XX)).
+
 ### 3.17.1
 
 - `WebRtcServer`: Remove the limit of 8 `listenInfos`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Node: Make `RtpCodecCapability.preferredPayloadType` mandatory and add `RouterRtpCodecCapability` in which `preferredPayloadType` is optional ([PR #15XX](https://github.com/versatica/mediasoup/pull/15XX)).
+- Node: Make `RtpCodecCapability.preferredPayloadType` mandatory and add `RouterRtpCodecCapability` type (in which `preferredPayloadType` is optional) and `RouterRtpCapabilities` type ([PR #15XX](https://github.com/versatica/mediasoup/pull/15XX)).
 
 ### 3.17.1
 
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -63,7 +63,10 @@ import type {
 	AudioLevelObserverOptions,
 } from './AudioLevelObserverTypes';
 import { AudioLevelObserverImpl } from './AudioLevelObserver';
-import type { RtpCapabilities, RtpCodecCapability } from './rtpParametersTypes';
+import type {
+	RtpCapabilities,
+	RouterRtpCodecCapability,
+} from './rtpParametersTypes';
 import { cryptoSuiteToFbs } from './srtpParametersFbsUtils';
 import type { AppData } from './types';
 import * as utils from './utils';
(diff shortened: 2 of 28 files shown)


jmillan:
Two comments.

#### 1
Rather than having `RtpCodecCapability` a mandatory `preferredPayloadType` and then have a `RouterRtpCodecCapability` omitting this field and adding a new one with the same name mandatory, both could use a base type which does not include `preferredPayloadType` at all, and later add it to their needs.

```ts
// Private, no export.
type BaseRtpCodecCapability = {
	kind: MediaKind;
	mimeType: string;
	clockRate: number;
	channels?: number;
	parameters?: Record<string, unknown>;
	rtcpFeedback?: RtcpFeedback[];
}

export type RtpCodecCapability = BaseRtpCodecCapability & {
        preferredPayloadType: number
}


export type RouterRtpCodecCapability = BaseRtpCodecCapability & {
	preferredPayloadType?: number;
};
```

#### 2
If `preferredPayloadType` is now mandatory in `RtpCodecCapability` , shouldn't it be called just `payloadType`?

```ts
// Private, no export.
type BaseRtpCodecCapability = {
	kind: MediaKind;
	mimeType: string;
	clockRate: number;
	channels?: number;
	parameters?: Record<string, unknown>;
	rtcpFeedback?: RtcpFeedback[];
}

export type RtpCodecCapability = BaseRtpCodecCapability & {
        payloadType: number
}


export type RouterRtpCodecCapability = BaseRtpCodecCapability & {
	preferredPayloadType?: number;
};
```

jmillan:
Another thing, just a reminder that if this is going to create breaking changes then we can also remove  (does not need to be in this exact PR) other old fields that where deprecated when we introduced `listenInfo`, namely `listenIp`, `port`, `rtcpPort`,  `TransportListenIp` etc. Those are marked as deprecated in code as well as in the doc.

jmillan (approved):

ibc:
> Rather than having `RtpCodecCapability` a mandatory `preferredPayloadType` and then have a `RouterRtpCodecCapability` omitting this field and adding a new one with the same name mandatory, both could use a base type which does not include `preferredPayloadType` at all, and later add it to their needs.

Honestly I don't want to go that way. `RtpCapabilities` and `RtpCodecCapability` mimic **standard** WebRTC types (well, quite similars). They are not an invent in mediasoup ecosystem, and what it's an invent are the `RouterXxxxx` versions of those types which are just the exception. I wouldn't like to put both at the same level of relevance.

ibc:
> If `preferredPayloadType` is now mandatory in `RtpCodecCapability` , shouldn't it be called just `payloadType`?

No, `RtpCodecCapability` comes from ORTC spec. It's different in WebRTC (less granular and more useless type) but still we follow what was done for ORTC which perfectly suits our needs in mediasoup since ever. It's `preferredPayloadType` because that's what it is. The preferred payload type and not the payload type of a codec. It's used to kindly tell to the remote about our local preferences.

ibc:
> Another thing, just a reminder that if this is going to create breaking changes 

Breaking changes **already** exist if users have both `mediasoup` and `mediasoup-client` deps in a Node project and try to correlate types from both. this PR fixes that.

> then we can also remove (does not need to be in this exact PR) other old fields that where deprecated when we introduced `listenInfo`, namely `listenIp`, `port`, `rtcpPort`, `TransportListenIp` etc. Those are marked as deprecated in code as well as in the doc.

Definitely not related to this PR but I agree. Although unfortunately too many people still rely on deprecated `listenIps`, `port`, etc.

jmillan:
> Breaking changes already exist if users have both mediasoup and mediasoup-client deps in a Node project and try to correlate types from both. this PR fixes that.

I was referring to the breaking change this PR introduces itself, which you pointed out:

> This is a breaking change in the sense that if an application declares a variable mediaCodecs: RtpCodecCapability[] and introduces entries without preferredPayloadType

> Although unfortunately too many people still rely on deprecated listenIps, port, etc.

Yes. The deprecated fields are marked as deprecated since they become so, meaning users are advised. Increasing a major version only for that seems too much though.

ibc:
@jmillan i am a bit lost. What is your concern after comments above? Anything I can do to improve things?

jmillan:
>  @jmillan i am a bit lost. What is your concern after comments above? Anything I can do to improve things?

Nothing that I'm aware of. All is good :+1:

PR #1592, opened 2025-08-22
Fix Path Normalization in `worker/tasks.py`

### LINKED ISSUE #1591: `worker/tasks.py` Uses Double Quotes To Wrap Paths, which is Problematic on Windows
# Bug Report

Worker build script uses `f'"{WORKER_DIR}"'` to format WORKER_DIR, which causes a parsing issue on Windows (i.e. PowerShell) if the path contains a whitespace.

## Your environment

- Operating system: Windows 11 Pro 24H2
- Node version: v20.11.1
- npm version: 10.7.0
- gcc/clang version: 
- mediasoup version: 0.19.0
- mediasoup-client version: n/a

## Issue description

On Windows, my user name always had a space in it, which makes all paths to my user files have a space (I know, I know, sorry fellas...).

This means that, given my user is `My User`, the correct path to the cargo registry as an example would be `C:\Users\My User\.cargo\registry\`.

When running `cargo check` on a project including `mediasoup`, I get the following error:

```
cd "C:\Users\My\ User\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\mediasoup-sys-0.10.0" && "C:/Users/My User/Documents/Projects/path-to-project/target/debug/build/mediasoup-sys-60af2f1db5e39515/out/out/pip_meson_ninja/bin/meson" setup --prefix "C:/Users/My User/path-to-project/target/debug/build/mediasoup-sys-60af2f1db5e39515/out" --bindir "" --libdir "" --buildtype release -Db_ndebug=true --vsenv "C:/Users/My User/path-to-project/target/debug/build/mediasoup-sys-60af2f1db5e39515/out/build"
    --- stderr
    The system cannot find the path specified.
```

After some debugging I believe the issue comes from [task.py](https://github.com/versatica/mediasoup/blob/2194eedf7e9ebc3b54dc5babd6d4755ada211a28/worker/tasks.py#L142) (and other similar code snippets within the same file), which uses `f'"{WORKER_DIR}"'` to format paths. Unfortunately `subprocess.run` doesn't like that for PowerShell and escapes whitespaces when it shouldn't, as can be shown by the following minimal Python example when running on Windows:

```python
import subprocess
import os


# Sample WORKER_DIR with a space in it
WORKER_DIR = r".\Hello World"
os.makedirs(WORKER_DIR, exist_ok=True)

print("WORKER_DIR=", WORKER_DIR)
print("Running cd with WORKER_DIR:")
subprocess.run(["cmd", "/c", f"cd {WORKER_DIR} && cd"])

bad_worker_dir = f'"{WORKER_DIR}"'
print("\nf'\"{WORKER_DIR}\"' =", bad_worker_dir)
print("Running cd with f'\"{WORKER_DIR}\"':")
subprocess.run(["cmd", "/c", f"cd {bad_worker_dir} && cd"])  # The system cannot find the path specified.
```

The solution to this should be quite simple, just replace all `\` with `/`, which we're already doing in [build.rs](https://github.com/versatica/mediasoup/blob/2194eedf7e9ebc3b54dc5babd6d4755ada211a28/worker/build.rs#L147) anyways, unfortunately it wasn't kept consistent. I'm happy to propose a PR for it if maintainers are too busy for it.

ShuzhaoFeng:
Fixes #1591

Windows specific issue, see #1591 for details.

As mentioned in the issue, the fix is to handle Window's case separately and replace `\` by `/`.

I tested on Windows & Linux (through WSL Ubuntu). Since I don't own a Mac, I haven't tested it on MacOS yet, but in theory there's no behavior change as long as it's not Windows platform.

### DIFF
--- a/worker/tasks.py
+++ b/worker/tasks.py
@@ -85,9 +85,10 @@ else:
 # normalize direct uses of WORKER_DIR
 if os.name == 'nt':
     # NOTE: this is necessary for Windows to avoid whitespace within the path breaking the command
-    WORKER_DIR = WORKER_DIR.replace('\\', '/');
+    WORKER_DIR = os.path.normpath(WORKER_DIR);
 else:
-    WORKER_DIR = f'"{WORKER_DIR}"'
+    # NOTE: this is also necessary, because Linux cd command does not accept multiple arguments
+    WORKER_DIR = f'"{WORKER_DIR}"';
 
 # Instruct Python where to look for modules it needs, such that meson actually
 # runs from installed location.
--- a/worker/tasks.py
+++ b/worker/tasks.py
@@ -82,6 +82,13 @@ if os.name == 'nt':
 else:
     os.environ['NINJA'] = f'{PIP_MESON_NINJA_DIR}/bin/ninja';
 
+# normalize direct uses of WORKER_DIR
+if os.name == 'nt':
+    # NOTE: this is necessary for Windows to avoid whitespace within the path breaking the command
+    WORKER_DIR = WORKER_DIR.replace('\\', '/');
+else:
+    WORKER_DIR = f'"{WORKER_DIR}"'
+
 # Instruct Python where to look for modules it needs, such that meson actually
 # runs from installed location.
 # NOTE: On Windows we must use ; instead of : to separate paths.
@@ -139,7 +146,7 @@ def setup(ctx, meson_args=MESON_ARGS):
     Run meson setup
     """
     if MEDIASOUP_BUILDTYPE == 'Release':
-        with ctx.cd(f'"{WORKER_DIR}"'):
+        with ctx.cd(WORKER_DIR):
             ctx.run(
                 f'"{MESON}" setup --prefix "{MEDIASOUP_INSTALL_DIR}" --bindir "" --libdir "" --buildtype release -Db_ndebug=true {meson_args} "{BUILD_DIR}"',
                 echo=True,
@@ -147,7 +154,7 @@ def setup(ctx, meson_args=MESON_ARGS):
                 shell=SHELL
             );
     elif MEDIASOUP_BUILDTYPE == 'Debug':
-        with ctx.cd(f'"{WORKER_DIR}"'):
+        with ctx.cd(WORKER_DIR):
             ctx.run(
                 f'"{MESON}" setup --prefix "{MEDIASOUP_INSTALL_DIR}" --bindir "" --libdir "" --buildtype debug {meson_args} "{BUILD_DIR}"',
                 echo=True,
@@ -155,7 +162,7 @@ def setup(ctx, meson_args=MESON_ARGS):
                 shell=SHELL
             );
     else:
-        with ctx.cd(f'"{WORKER_DIR}"'):
+        with ctx.cd(WORKER_DIR):
             ctx.run(
                 f'"{MESON}" setup --prefix "{MEDIASOUP_INSTALL_DIR}" --bindir "" --libdir "" --buildtype {MEDIASOUP_BUILDTYPE} -Db_ndebug=if-release {meson_args} "{BUILD_DIR}"',
                 echo=True,
@@ -194,7 +201,7 @@ def clean_subprojects(ctx):
     """
     Clean meson subprojects
     """
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         ctx.run(
             f'"{MESON}" subprojects purge --include-cache --confirm',
             echo=True,
@@ -208,7 +215,7 @@ def clean_all(ctx):
     """
     Clean meson subprojects and all installed/built artificats
     """
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         try:
             ctx.run(
                 f'"{MESON}" subprojects purge --include-cache --confirm',
@@ -228,7 +235,7 @@ def update_wrap_file(ctx, subproject):
     """
     Update the wrap file of a subproject
     """
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         ctx.run(
             f'"{MESON}" subprojects update --reset {subproject}',
             echo=True,
@@ -242,7 +249,7 @@ def flatc(ctx):
     """
     Compile FlatBuffers FBS files
     """
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         ctx.run(
             f'"{MESON}" compile -C "{BUILD_DIR}" flatbuffers-generator',
             echo=True,
@@ -260,14 +267,14 @@ def mediasoup_worker(ctx):
         print('skipping mediasoup-worker compilation due to the existence of the MEDIASOUP_WORKER_BIN environment variable');
         return;
 
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         ctx.run(
             f'"{MESON}" compile -C "{BUILD_DIR}" -j {NUM_CORES} mediasoup-worker',
             echo=True,
             pty=PTY_SUPPORTED,
             shell=SHELL
         );
-    with ctx.cd(f'"{WORKER_DIR}"'):
+    with ctx.cd(WORKER_DIR):
         ctx.run(
             f'"{MESON}" install -C "{BUILD_DIR}" --no-rebuild --tags mediasoup-worker',
             echo=True,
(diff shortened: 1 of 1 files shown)


worker/tasks.py
```diff
@@ -82,6 +82,13 @@
 else:
     os.environ['NINJA'] = f'{PIP_MESON_NINJA_DIR}/bin/ninja';
 
+# normalize direct uses of WORKER_DIR
```
nazar-pc, line 85: Does Python maybe have a proper "path" wrapper type, like https://doc.rust-lang.org/stable/std/path/struct.Path.html in Rust, which correctly handles escaping of characters?

These things shouldn't really be strings in the first place if we want to handle platform-specific special cases properly.

ShuzhaoFeng: Actually yes, you can use [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) to normalize a path, which does exactly this among others and would probably be the preferred approach.

That said, I've received instructions to make sure the changes does not break the existing pipeline (which to be fair makes 100% sense), and since the script does not have a unit test suite, my initial plan was to keep my change minimal to fix the issue to avoid problems as much as possible.

If we'd want to switch to path handling using `normpath`, that's a behavior change across all environments and we need some more thorough testing and possibly start building a unit test suite, which is beyond my current concern. Also at that point, there's probably more than the path that needs some care in this script :D

I advocate for my fix to be merged and deployed to at least fix the issue, then we can prioritize standardizing this script and build more thorough validation around it. I'm happy to help with the process if you need me.

nazar-pc: I think that would be preferred. I am not sure what behavior change you mean, but we have CI that checks all 3 major platforms, so if anything is completely broken we'll catch it (it should also run in CI in your fork if you want to experiment before pushing it here).

nazar-pc (commented):
Also do you have an explanation why it helps? `\` is the native path separator on Windows, so should generally be preferred. It does support both though, even when mixed together. But I don't understand how it affects whitespaces either way.

worker/tasks.py
ShuzhaoFeng: Sure, I'm also OK with that. I'll let you know when it's done, and then please do some manual testing in case I missed something.

ShuzhaoFeng: Sure, I'm also OK with that. I'll let you know when it's done, and then please do some manual testing in case I missed something.

Edit: See below comment, I'll wait for the review before making the changes.

ibc: What I meant with not breaking things was that current v3 branch allows installing mediasoup on a path with spaces in folder names under Linux/macOS. We don't test that in CI.

I will review this PR on Monday. Thanks a lot.

ShuzhaoFeng:
> Also do you have an explanation why it helps? \ is the native path separator on Windows, so should generally be preferred. It does support both though, even when mixed together. But I don't understand how it affects whitespaces either way.

How [Python's `subprocess` module](https://github.com/python/cpython/blob/b9bcd02e9fe976e0a6a8e8a0b336d598b1d73b13/Lib/subprocess.py#L581) process string is that, when your argument contains wither a whitespace ` ` or a tab `\t`, the functions marks the argument as "needquote" and will automatically surround your argument with quote and escapes things within it.

So, if your "Python command" looks like `['cmd', '/c', 'cd', 'Hello World']`, this is actually a great feature, because it will parse it to the following PowerShell command:

```powershell
cmd /c cd "Hello World"
```

And preserve your intention, which is to execute `cd "Hello World"` using `cmd`.

However, when you have backslash in your Python string, for example: `['cmd', '/c', 'cd', 'C:\Hello World']`

Given backslash is a reserved character in Python when using formatting and Python is a interpreted language, what it does is it double-escapes your escape characters (`\` -> `\\`) within the string, and that happens before sending the string to `subprocess`, which will then parse `['cmd', '/c', 'cd', 'C:\\Hello World']` into `cmd /c cd \"C:\\Hello World\"` (because double quotes also need to be escaped). Then, PowerShell, the root of all evils here, decides that double quotes should not be escaped with `\`, but they should be escaped with `` ` ``. So, while the correct PowerShell command should be:

```powershell
cmd /c cd `"C:\\Hello World`"
```

What PowerShell ends up seeing is:

```powershell
cmd /c cd \"C:\\Hello World\"
```

Which is totally broken. However, if your original string didn't contain any space, the function will not mark your argument as "needquote" and add double quotes around it. And `cmd /c cd "C:\\HelloWorld"` is still a valid expression (because when you use f-string the double quotes don't get escaped), hence why it has been working when there's no whitespace.

So, by replacing `\` with `/`, we're basically avoiding to use Python's reserved character for string formatting, rather than this is being a PowerShell-targeted fix (but I won't argue that PowerShell is the root of evil here 😅).

worker/tasks.py
ibc: @ShuzhaoFeng if you wish you can push an extra commit switching to that python normpath utility or create a second PR with that change and then on next week we will decide which one to merge.

worker/tasks.py
```diff
@@ -82,6 +82,14 @@
...
+if os.name == 'nt':
+    # NOTE: this is necessary for Windows to avoid whitespace within the path breaking the command
+    WORKER_DIR = os.path.normpath(WORKER_DIR);
+else:
+    # NOTE: this is also necessary, because Linux cd command does not accept multiple arguments
```
nazar-pc, line 90: Really? We can't just use `os.path.normpath()` on all platforms? We're not really having multiple arguments, this is a directory path.

ShuzhaoFeng: I'm sorry, but unfortunately we can't because if a path contains a whitespace on Linux (I can't speak for MacOS just because I didn't test), it must be double-quoted because `cd` doesn't take multiple arguments (since `/this path/` unquoted will be parsed into two arguments, but PowerShell cd actually accepts it).

And as I explained the double quote is the root of evil on Windows for as long as there are backslashes, so 🤷.

nazar-pc: I'm not sure what the code does exactly, but what does Linux or Windows `cd` command has to do with any of this? Usually when you spawn a process or something you can set a working directory, so I expected there would be a type like `Path` in Rust, which you use for its argument and it wouldn't run a script with `cd` being the first command in the script, it'd instead use OS APIs to set the path properly from the start.

ShuzhaoFeng: > I'm not sure what the code does exactly, but what does Linux or Windows cd command has to do with any of this?

The `cd` commands used matters because the code will call `ctx.cd(WORKER_DIR)`, which is a wrapper of `subprocess.run` for the `cd` command. So the result highly depends on the `cd` command being called. AFAIK this is the standard process suggested by `pyinvoke` to set up the working directory.

> Usually when you spawn a process or something you can set a working directory, so I expected there would be a type like Path in Rust, which you use for its argument and it wouldn't run a script with cd being the first command in the script, it'd instead use OS APIs to set the path properly from the start.

I see what you mean, I wasn't on that page earlier in our discussion.

There isn't a `Path` wrapper in Python like in Rust, but you can again use the built-in OS command `os.chdir` to set a working directory. So the bottom-line answer is that it can be done.

However, we're no longer talking about a path issue, but about reworking how the script executes commands. Although I'm familiar with Python, I'm less familiar with Python's `Invoke` library and MediaSoup's build scripts in general, so it will take me time to walk through this. I can't guarantee that I'll have time to get this done over this weekend.

For that reason I won't work on `os.chdir` for this PR specifically, and will focus on just addressing the path issue. However, I'm happy to help with the script refactoring over a longer development period if you need me.

nazar-pc: I'm not really working with Python myself, hence the questions. Appreciate you finding and fixing the issue.

I think none of us (maintainers) have extensive production experience with Python, so if you can improve things I think it'll be a very welcome change. Doesn't have to be done in this PR necessarily, but I'd certainly prefer a more robust solution than calling `cd` shell command and `os.chdir` you mentioned sounds much more appropriate.

Thanks for providing all this context!

ibc:
Sorry guys but the problem is huger than expected. It's also failing in v3 branch in macOS. I mean, everything is failing. And this PR is not fixing it AFAIS. See details in a new ticket here:

https://github.com/versatica/mediasoup/issues/1595

nazar-pc:
Then I think we need to go `os.chdir` route as suggested in above threads, rather than messing with escape characters in shell commands (which are prone to shell command injection among other things anyway).

ShuzhaoFeng:
We need to refactor `tasks.py`, and this time offering some unit tests around its features, and potentially adding a new PR workflow on it.

Also as @ibc mentioned, the problem seems larger than just `tasks.py`, so we need to also investigate elsewhere.

I'm happy to work on this if you can expect around 1 month development time (I'll be really busy and traveling for the next two weeks). Will let you know when it's done.

ibc:
@ShuzhaoFeng thanks for your offering. I've written this PR that, at least, fixes the issue in macOS and Docker (Linux container on macOS host). I've tested `npm` commands, Python `invoke` and `make` in both environments and AFAIS it works. But I cannot test in Windows. Could you please take a look and test it in your local setup using Windows?

https://github.com/versatica/mediasoup/pull/1596

ibc:
> I'm happy to work on this if you can expect around 1 month development time (I'll be really busy and traveling for the next two weeks)

BTW don't worry if you are busy. Just take your time, please. I think that this issue is not urgent since in production environments probably nobody uses paths with spaces in folder names.

ibc:
Closing this PR in favour of https://github.com/versatica/mediasoup/pull/1596.

PR #1594, opened 2025-08-25
Rust: Remove unused `Notification` enum

ibc:
## Details

- Fixes #1593
- Such a `Notification` enum is not used anywhere.
- AFAIS this "running" notification was intended to mimic the behavior in Node where the worker sends a "running" notification to Node layer once it's ready. But AFAIK that has never been needed in mediasoup Rust.


nazar-pc (approved):
Fine if really unused. Probably it was used earlier when worker communication happened in JSON, but not anymore.

PR #1597, opened 2025-08-26
[Rust] Make `parameters` and `rtcp_feedback` optional in `RtpCodecParameters` and `RtpCodecCapability` during deserialization

### LINKED ISSUE #1589: Rust types: RtpCodecParametrers.parameters should be optional
In `rust/types/src/rtp_parameters.rs`, the struct `RtpCodecParameters` and `RtpCodecCapability` have a mandatory `parameters` of type `RtpCodecParametersParameters` that should be optional. Same applies to `rtcpFeedback` array.

Otherwise, if we receive a serialized `RtpParameters` object generated by a client whose `codecs` don't contain `parameters` or `rtcpFeedback` fields, automatic deserialization will fail because those fields are mandatory in the Rust types. And it shouldn't be like that.

ibc:
## Details

- Fixes #1589

## TODO

- Should I also add `#[serde(default)]` above `parameters` and `rtcp_feedback` fields in `RtpCodecCapabilityFinalized` type?

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # NEXT
 
+- Make `parameters` and `rtcp_feedback` optional in `RtpCodecParameters` and `RtpCodecCapability` during deserialization.
+
 # 0.19.1
 
 - Fix installation in paths with spaces (PR #1596).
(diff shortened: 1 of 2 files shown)


ibc:
NOTE: My changes work. Tested.

nazar-pc (approved):
This looks reasonable.

> Should I also add `#[serde(default)]` above `parameters` and `rtcp_feedback` fields in `RtpCodecCapabilityFinalized` type?

I don't think finalized version is created by the client, so there should generally be no need to de-serialize it such that it is missing the field.

PR #1599, opened 2025-08-26
[Rust] Make codec `mime_type` case insensitive when it comes to deserialization

### LINKED ISSUE #1590: Rust types: RtpCodecParameters.mimeType must be case insensitive
# Bug Report

```
"mimeType": "video/vp8"
```

This is invalid because in Rust it should be `video/VP8`. I understand rationale here, but we have a problem when we try to convert RTP parameters generated by other clients (such as `mediasoup-client` JS library) in which `mimeType` is case insensitive.

What are we supposed to do in a Rust application if we receive RTP parameters via WebSocket (for example)?

ibc:
Fixes #1590

Tested (locally by running another project that uses `mediasoup-types` crate).

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 # NEXT
 
-- Make `parameters` and `rtcp_feedback` optional in `RtpCodecParameters` and `RtpCodecCapability` during deserialization.
-- Make codec `mime_type` case insensitive.
+- Make `parameters` and `rtcp_feedback` optional in `RtpCodecParameters` and `RtpCodecCapability` during deserialization (PR #1589).
+- Make codec `mime_type` case insensitive during deserialization (PR #1590).
 
 # 0.19.1
 
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -3,6 +3,7 @@
 # NEXT
 
 - Make `parameters` and `rtcp_feedback` optional in `RtpCodecParameters` and `RtpCodecCapability` during deserialization.
+- Make codec `mime_type` case insensitive.
 
 # 0.19.1
 
--- a/rust/types/src/rtp_parameters.rs
+++ b/rust/types/src/rtp_parameters.rs
@@ -247,7 +247,7 @@ impl MimeType {
 
 /// Known Audio MIME types.
 #[allow(non_camel_case_types)]
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize)]
 pub enum MimeTypeAudio {
     /// Opus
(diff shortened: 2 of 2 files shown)


nazar-pc (approved):
This should work fine, but it will allocate. Would have been nice to use https://doc.rust-lang.org/stable/std/primitive.str.html#method.eq_ignore_ascii_case, but they `match` will stop working and will have to be replaced with a bunch of `if`/`else`.

ibc:
Yes, `s.to_ascii_lowercase()` allocates a new string in the heap. It's a little price to pay compared to using [eq_ignore_ascii_case](https://doc.rust-lang.org/stable/std/primitive.str.html#method.eq_ignore_ascii_case) with tons of if else if else if else if else.

ibc:
> Yes, `s.to_ascii_lowercase()` allocates a new string in the heap. It's a little price to pay compared to using [eq_ignore_ascii_case](https://doc.rust-lang.org/stable/std/primitive.str.html#method.eq_ignore_ascii_case) with tons of if else if else if else if else.

Also, I think that using `if else` so many times will be much less performant than using `match` which probably doesn't perform one-by-one full string matching.

nazar-pc:
> Also, I think that using `if else` so many times will be much less performant than using `match` which probably doesn't perform one-by-one full string matching.

I didn't look at the assembly, but I suspect it might actually optimize very similarly.

You could also create a small stack buffer (since string length is limited), copy the bytes there and convert to lowercase in-place. But it will again make the code much less readable.

ibc:
> Also, I think that using `if else` so many times will be much less performant than using `match` which probably doesn't perform one-by-one full string matching.

However I am wrong and `match` in Rust is just a wrapper of if-else without any kind of hashing or similar optimizations. So maybe I should switch to your proposal.

ibc:
Ok, at some point we should assume that having cool and readable code is better than saving a few bytes of memory by writing awful code. I will merge this PR as it is.

nazar-pc:
Another idea: try to match on things like `audio/PCMU` and `audio/pcmu` (you can use `"audio/PCMU" | "audio/pcmu"` with `match`) and only make it lowercase as a fallback if none of the options match.

This way it will not allocate most of the time and still reasonably readable.

ibc:
> Another idea: try to match on things like `audio/PCMU` and `audio/pcmu` (you can use `"audio/PCMU" | "audio/pcmu"` with `match`) and only make it lowercase as a fallback if none of the options match.
> 
> This way it will not allocate most of the time and still reasonably readable.

I don't understand. How will that match **valid** things like "video/Vp8"?

nazar-pc:
You'd have `match` the way I described above, and then when going into `s => Err(if s.starts_with("audio/")` at the end where nothing matched, you do another `match` the way you did it in this PR.

So initially we check for the typical/common case MIME type is written, but in case it is something unusual, you try to normalize it by lower-casing. But you only pay for allocation when you have to rather than always.

ibc:
I think that would save a few bytes of memory while making the code really ugly

nazar-pc:
It is not critical for sure

PR #1603, opened 2025-08-27
[Node] Improve worker location detection

### LINKED ISSUE #1601: When bundling mediasoup `__dirname` no longer points to `mediasoup-worker` executable
## Description of the issue

- `node/src/Worker.ts` in mediasoup is transpiled to `node/lib/Worker.js` in build time (the mediasoup NPM library only contains `node/lib` folder).
- `Worker.js` contains this:
  ```js
  exports.workerBin = process.env['MEDIASOUP_WORKER_BIN']
      ? process.env['MEDIASOUP_WORKER_BIN']
      : process.env['MEDIASOUP_BUILDTYPE'] === 'Debug'
          ? path.join(__dirname, '..', '..', 'worker', 'out', 'Debug', 'mediasoup-worker')
          : path.join(__dirname, '..', '..', 'worker', 'out', 'Release', 'mediasoup-worker');
  ```
- When using webpack or rollup or similar tools to bundle application + dependencies into a single bundle file, the above code fails because `__dirname` no longer points to the `app/node_modules/mediasoup` folder (now it points to the current root folder), so `mediasoup.createWorker()` fails with this error:
  ```
  mediasoup createWorker() +0ms
  mediasoup:Worker constructor() +0ms
  mediasoup:Worker spawning worker process: /private/tmp/worker/out/Release/mediasoup-worker --logLevel=debug --rtcMinPort=10000 --rtcMaxPort=59999 +1ms
  mediasoup:Channel constructor() +0ms
  mediasoup:ERROR:Worker worker process failed [pid:undefined]: spawn /private/tmp/worker/out/Release/mediasoup-worker ENOENT
  ```

## STR

This test project reproduces the problem:

**package.json**
```json
{
  "scripts": {
    "start": "webpack && DEBUG='*' node dist/bundle.js"
  },
  "dependencies": {
    "mediasoup": "^3.18.1",
    "webpack": "^5.101.3",
    "webpack-cli": "^6.0.1"
  }
}
```

**webpack.config.js**
```js
const path = require('path');

module.exports = {
  mode: 'development',
  target: 'node',
  entry: './index.js',
  output: {
    filename: 'bundle.js',
    path: path.resolve(__dirname, 'dist')
  },
  module: {
    rules: [
      {
        test: /node_modules\/mediasoup\/node\/lib\/.*\.js$/,
        type: 'javascript/auto' // fuerza CJS real para mediasoup
      }
    ]
  },
  resolve: {
    extensions: ['.js']
  }
};
```

**index.js**
```js
const mediasoup = require('mediasoup');

run();

async function run() {
  const worker = await mediasoup.createWorker({ logLevel: 'debug' });

  console.log("Worker PID:", worker.pid);

  const dump = await worker.dump();

  console.log("Worker dump:", dump);
}
```

Run it:

```bash
npm start
```

The above error happens.

## Note

Unclear whether we should address this problem. Bundling a mediasoup based application is probably out of the scope of this project. If the user does it then it's his responsibility to "fix" the problem by setting `MEDIASOUP_WORKER_BIN` env when running their application.

## Related issues

- https://github.com/versatica/mediasoup/issues/152

ibc:
## Details

- Fixes #1601
- Use `require.resolve('mediasoup')` to get the path of mediasoup module and, from there, compute the location of the worker binary.
- This prevents mediasoup from failing to spawn the worker binary when the Node application uses webpack or rollup and bundles the source code (including mediasoup dependency) into a `dist/` folder.
- As fallback mechanism, use the previous `__dirname` solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Node: Improve worker binary location detection ([PR #1603](https://github.com/versatica/mediasoup/pull/1603)).
+
 ### 3.18.1
 
 - `TransportTuple`: Generate hash based not only on remote IP:port but also on local IP:port ([PR #1586](https://github.com/versatica/mediasoup/pull/1586)).
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -650,9 +650,8 @@ function getWorkerBin(): string {
 		);
 	} catch (error) {
 		logger.warn(
-			`getWorkerBin() | require.resolve('mediasoup') failed: ${error}`
+			`getWorkerBin() | require.resolve('mediasoup') failed, using __dirname: ${error}`
 		);
-		logger.debug('getWorkerBin() | using __dirname');
 
 		// mediasoup module path is two folders above this file.
 		mediasoupModulePath = path.join(__dirname, '..', '..');
--- a/node/src/Worker.ts
+++ b/node/src/Worker.ts
@@ -32,33 +32,10 @@ import * as FbsWorker from './fbs/worker';
 import * as FbsTransport from './fbs/transport';
 import { Protocol as FbsTransportProtocol } from './fbs/transport/protocol';
 
-// If env MEDIASOUP_WORKER_BIN is given, use it as worker binary.
-// Otherwise if env MEDIASOUP_BUILDTYPE is 'Debug' use the Debug binary.
-// Otherwise use the Release binary.
-export const workerBin = process.env['MEDIASOUP_WORKER_BIN']
-	? process.env['MEDIASOUP_WORKER_BIN']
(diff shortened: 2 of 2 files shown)


Bessonov:
Thanks, @ibc ! It looks good! Give me some time to try it out. I think I can get to it this weekend 👍

ibc:
> Thanks, @ibc ! It looks good! Give me some time to try it out. I think I can get to it this weekend 👍 

I want it merged before weekend so probably I will merge it tomorrow. Said that I'd appreciate if you later confirm that it works for you.

ibc:
Released in mediasoup 3.19.0

Bessonov:
@ibc I'm using a slightly outdated version, but I was able to patch it with your changes, and it works! Thanks!

ibc:
Great

PR #1604, opened 2025-08-27
`router.pipeToRouter()` can now connect two `Routers` in the same `Worker`

### LINKED ISSUE #1602: Critical router.pipeToRouter() error with Channel request handler
Hello Mediasoup Community,

I am experiencing a critical issue with `router.pipeToRouter()` that has persisted despite numerous attempts to resolve it.

**The Problem:**

I have a single Mediasoup Worker with two Routers. My goal is to pipe a video and audio stream from `Router 1` (Station Router) to `Router 2` (Manager Router). However, when the `pipeToRouter()` method is called for the video stream, it consistently fails with the following error:

`Error: Channel request handler with ID [original_producer_id] already exists [method:transport.produce]`

**What I Have Tried:**

- I am using `mediasoup@3.11.4` on the backend and `mediasoup-client@3.9.8` on the frontend. I've updated both to the latest versions, but the error remains.
- I've implemented a sequential, two-step piping process. The video pipe operation completes, and only then does a separate function attempt to pipe the audio stream.
- I've implemented a locking mechanism using a `Set` to prevent simultaneous `pipeToRouter()` calls.
- I've also tried a client-side approach where I only send the `start-piping` event to the server after both video and audio producers are successfully created.

**Context:**

- The error occurs specifically on the `pipeToRouter()` call, during the internal `transport.produce` method.
- The ID in the error message is the same as the original producer ID, suggesting that the method is attempting to create a new producer with a non-unique ID, which should not happen.

This issue seems to be a deep-rooted race condition or an internal bug within the `pipeToRouter` method itself.

Has anyone encountered this specific error before, or do you have any suggestions on how to work around this issue? Your help would be greatly appreciated.

Thank you.

ibc:
## Details

- Fixes #1602
- Remove the limitation by which `router1.pipeToRouter({ router: router2, etc })` throws if both `router1` and `router2` are in the same `Worker`.
- The solution is very simple: When we `produce()` the internally created `pipeConsumer` into the target `router`, pass a random `producerId` instead of keeping the same `id` that the original `producer` has.
- This may break some applications if they assume that the generated `pipeProducer` has same `id` than the `original` producer. The application should obtain the new `id` from the `{ pipeProducer, pipeDataProducer }` returned by `pipeToRouter()` method.
- **In order to avoid breaking existing applications that assume that the `pipeProducer.id` is the same as the original `producer.id`**, this new behavior is disabled by default. To enable it, `keepId: false` must be passed to `pipeToRouter()`.

## TODO

- [x] Document the new `keepId` field and new behavior in the docs.
- [x] Remove or clarify the note in the docs: https://mediasoup.org/documentation/v3/mediasoup/api/#router-pipeToRouter
- [x] Make public announcement.

### DIFF
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -1088,7 +1088,8 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 				});
 
 				pipeProducer = await remotePipeTransport!.produce({
-					id: producer.id,
+					// Generate a new id for the pipeProducer.
+					id: utils.generateUUIDv4(),
 					kind: pipeConsumer.kind,
 					rtpParameters: pipeConsumer.rtpParameters,
 					paused: pipeConsumer.producerPaused,
@@ -1145,7 +1146,8 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 				});
 
 				pipeDataProducer = await remotePipeTransport!.produceData({
-					id: dataProducer.id,
+					// Generate a new id for the pipeDataProducer.
+					id: utils.generateUUIDv4(),
 					sctpStreamParameters: pipeDataConsumer.sctpStreamParameters,
 					label: pipeDataConsumer.label,
 					protocol: pipeDataConsumer.protocol,
--- a/node/src/Router.ts
+++ b/node/src/Router.ts
@@ -924,6 +924,7 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 		producerId,
 		dataProducerId,
 		router,
+		keepId = true,
 		listenInfo,
 		listenIp,
 		enableSctp = true,
@@ -1088,8 +1089,8 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 				});
 
 				pipeProducer = await remotePipeTransport!.produce({
-					// Generate a new id for the pipeProducer.
-					id: utils.generateUUIDv4(),
+					// If requested, generate a new id for the pipeProducer.
+					id: keepId ? producer.id : utils.generateUUIDv4(),
 					kind: pipeConsumer.kind,
 					rtpParameters: pipeConsumer.rtpParameters,
 					paused: pipeConsumer.producerPaused,
@@ -1146,8 +1147,8 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 				});
 
 				pipeDataProducer = await remotePipeTransport!.produceData({
-					// Generate a new id for the pipeDataProducer.
-					id: utils.generateUUIDv4(),
+					// If requested, generate a new id for the pipeDataProducer.
+					id: keepId ? dataProducer.id : utils.generateUUIDv4(),
 					sctpStreamParameters: pipeDataConsumer.sctpStreamParameters,
 					label: pipeDataConsumer.label,
 					protocol: pipeDataConsumer.protocol,
--- a/rust/src/router.rs
+++ b/rust/src/router.rs
@@ -184,9 +184,6 @@ pub struct PipeProducerToRouterPair {
 /// Error that caused [`Router::pipe_producer_to_router()`] to fail.
 #[derive(Debug, Error)]
 pub enum PipeProducerToRouterError {
-    /// Destination router must be different
-    #[error("Destination router must be different")]
-    SameRouter,
     /// Producer with specified id not found
     #[error("Producer with id \"{0}\" not found")]
     ProducerNotFound(ProducerId),
@@ -240,9 +237,6 @@ pub struct PipeDataProducerToRouterPair {
 /// Error that caused [`Router::pipe_data_producer_to_router()`] to fail.
 #[derive(Debug, Error)]
 pub enum PipeDataProducerToRouterError {
-    /// Destination router must be different
-    #[error("Destination router must be different")]
-    SameRouter,
     /// Data producer with specified id not found
     #[error("Data producer with id \"{0}\" not found")]
     DataProducerNotFound(DataProducerId),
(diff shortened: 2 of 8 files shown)


node/src/Router.ts
```diff
@@ -1088,7 +1088,8 @@ export class RouterImpl<RouterAppData extends AppData = AppData>
 				});
 
 				pipeProducer = await remotePipeTransport!.produce({
-					id: producer.id,
+					// Generate a new id for the pipeProducer.
+					id: utils.generateUUIDv4(),
```
nazar-pc: This will certainly break applications that used the feature in the past. I think there should be some other solution, like having an option to reuse producer ID or to generate a new one. I wrote such applications in the past myself.

Moreover, it is much harder to debug this way.

ibc: Yes, there should be another option, but I already spent tons of hours some months ago trying to figure out one and it doesn't exist. Current behavior in v3 is that `pipeToRouter()` generates pipe Producers with same `id` so the app doesn't check the generated `pipeProducer.id` and they assume they can safely `consume()` from the target router using the original `producer.id`. So no, there is no other solution.

Said that, I could add an option in `pipeToRouter()` to make this new behavior optional, but I will make it default to enabled.

ibc: Ok, I will make it disabled by default.

nazar-pc: There is an option to keep ID if router is different, but change if it is the same, but that would be confusing. I also didn't find a real need for this behavior to be completely honest.

ibc: There have been many many reports or complains or feature requests or questions about people that, for whatever reason, want to connect routers no matter they run in the same worker.

ibc: Ok, `keepId: boolean` (default `true`) added:

https://github.com/versatica/mediasoup/pull/1604/commits/2316ec4ec19cc90b9872ad3865b27aa5c4e6e438

rust/src/router.rs
```diff
@@ -162,6 +170,12 @@ impl PipeToRouterOptions {
             enable_srtp: false,
         }
     }
+
+    /// Set `keep_id`.
+    pub fn with_keep_id(mut self, keep: bool) -> Self {
+        self.keep_id = keep;
+        self
+    }
```
nazar-pc: Not sure this method is needed. We didn't have other `.with_*` methods before.

rust/src/router.rs
```diff
@@ -1083,7 +1097,9 @@ impl Router {
     ) -> Result<PipeProducerToRouterPair, PipeProducerToRouterError> {
         debug!("pipe_producer_to_router()");
 
-        if pipe_to_router_options.router.id() == self.id() {
+        let PipeToRouterOptions { keep_id, .. } = pipe_to_router_options;
+
+        if keep_id && pipe_to_router_options.router.id() == self.id() {
```
nazar-pc: This is inconsistent. Let's either do this:
```suggestion
        if pipe_to_router_options.keep_id && pipe_to_router_options.router.id() == self.id() {
```
Or if you want to make the line shorter then this:
```suggestion
        let PipeToRouterOptions { router, keep_id, .. } = &pipe_to_router_options;

        if *keep_id && router.id() == self.id() {
```

nazar-pc (commented):
Overall makes sense, just some code style nits.

Note that this is a backwards compatible change for Rust because one of the fields of `PipeToRouterOptions` was already private.

rust/src/router.rs
ibc: Then what do you propose?

nazar-pc: For this method - just remove it. You can change the field directly just like with any other.

rust/src/router.rs
ibc: We **need** to extract `keep_id` **before** moving `pipe_to_router_options` to `self.get_or_create_pipe_transport_pair()` so your proposal doesn't compile.

rust/src/router.rs
ibc: No, I can't XD.

ibc:
```rust
impl PipeToRouterOptions {
    /// Crate pipe options for piping into given local router.
    #[must_use]
    pub fn new(router: Router) -> Self {
```

Only `router` can be set when creating a `PipeToRouterOptions`.

And in the tests, for whatever reasons, this is done:

```rust
router1.pipe_producer_to_router(
  video_producer.id(),
    PipeToRouterOptions::new(router1bis.clone()),
  )
  .await;
```

nazar-pc: That is not true:
```rust
{
    let mut options = PipeToRouterOptions::new(router1bis.clone());
    options.keep_id = false;
    options
}
```

rust/src/router.rs
nazar-pc: I don't understand what doesn't compile. Which option have you tried and what was the compiler error?

ibc: I've done it here: https://github.com/versatica/mediasoup/pull/1604/commits/d736b90ff67a7fbb45a13114fa80d48483c4b124

nazar-pc: I never liked `ref`, but sure, that works too.

rust/src/router.rs
ibc: Oh, I didn't know that syntax... done here https://github.com/versatica/mediasoup/pull/1604/commits/5174cc1c98bea57e38079e92c98744df7d00f894

nazar-pc (approved):

rust/src/router.rs
nazar-pc: It is just a variable and the fact that blocks are expressions

ibc: and the block doesn't not end with a explicit `return options` because just ending with `options` is more obvious (it's not).

ibc:
Released in mediasoup 3.19.0

PR #1605, opened 2025-09-01
Rust, deadlock problem caused by 'paused' and 'producer_paused' in 'consumer.rs'

gnauhtt:
[Rust] Fixed the deadlock problem caused by inconsistent acquisition order of 'paused' and 'producer_paused', and released early before the 'call_simple()'.

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - `router.pipe_producer_to_router()` and `router.pipe_data_producer_to_router()` can now connect two `Routers` in the same `Worker` if `keep_id` is set to `false` (PR #1604).
 - Updates from mediasoup TypeScript `3.18.1.=3.19.0`.
+- Ensure that the order of acquiring the `paused` and `producer_paused` locks in `consumer.rs` is consistent at all times to avoid deadlock.
 
 # 0.20.0
 
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - `router.pipe_producer_to_router()` and `router.pipe_data_producer_to_router()` can now connect two `Routers` in the same `Worker` if `keep_id` is set to `false` (PR #1604).
 - Updates from mediasoup TypeScript `3.18.1.=3.19.0`.
-- Ensure that the order of acquiring the `paused` and `producer_paused` locks in `consumer.rs` is consistent at all times to avoid deadlock.
+- Ensure that the order of acquiring the `paused` and `producer_paused` locks in `consumer.rs` is consistent at all times to avoid deadlock (PR #1605).
 
 # 0.20.0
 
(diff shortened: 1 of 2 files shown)


jmillan:
@gnauhtt, please add some description about the issue and how the change fixes it.

gnauhtt:
I have a producer and a consumer of this producer. If i run this, 
```rs
tokio::spawn(producer.resume());
tokio::spawn(consumer.resume());
```

The mediasoup worker thread will execute the consumer's Notification::ProducerResume. If it obtains the 'producer_paused' Mutex, and consume.resume() may get the 'paused' Mutex. Deadlock occurs.

```rs
Notification::ProducerResume => {
    let mut producer_paused = producer_paused.lock(); // lock 'producer_paused' first.
    let paused = *paused.lock(); // wait for 'paused' unlocked
    *producer_paused = false;

    handlers.producer_resume.call_simple();

    if !paused {
        handlers.resume.call_simple();
    }
}
```
```rs
pub async fn resume(&self) -> Result<(), RequestError> {
    debug!("resume()");

    self.inner
        .channel
        .request(self.id(), ConsumerResumeRequest {})
        .await?;

    let mut paused = self.inner.paused.lock(); // lock 'paused' second.
    let was_paused = *paused || *self.inner.producer_paused.lock(); // wait for 'producer_paused' unlocked
    *paused = false;

    if was_paused {
        self.inner.handlers.resume.call_simple();
    }

    Ok(())
}
```

jmillan (approved):
So basically we were acquiring `producer_paused` lock first and `paused` lock later in some places, and `paused` first and `producer_paused` later in other places, which could lead to the  deadlock.

This change always acquires the locks in the same order avoiding the issue.

nazar-pc (approved):
Makes sense to me. That said, I think `producer_paused` specifically might also be okay to change to `AtomicBool` rather than `Mutex`, I don't think we need to hold it as a lock anywhere for synchronization purposes. But it can be done in a separate PR.

Thanks for detailed explanation and a fix!

nazar-pc:
Could you update the changelog before we merge this, please?

gnauhtt:
Sure. Thank you for your patient review.

rust/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - `router.pipe_producer_to_router()` and `router.pipe_data_producer_to_router()` can now connect two `Routers` in the same `Worker` if `keep_id` is set to `false` (PR #1604).
 - Updates from mediasoup TypeScript `3.18.1.=3.19.0`.
+- Ensure that the order of acquiring the `paused` and `producer_paused` locks in `consumer.rs` is consistent at all times to avoid deadlock.
```
nazar-pc:
```suggestion
- Ensure that the order of acquiring the `paused` and `producer_paused` locks in `consumer.rs` is consistent at all times to avoid deadlock (PR #1605).
```

ibc:
@nazar-pc a new release would be great ☺️

nazar-pc:
Could you publish it, please? I have not pushed any releases for some time anyway.

Issue #1536, opened 2025-05-26
Investigate AV1 issues with downlink packet loss


jmillan:
When applying downlink packet loss on the consuming endpoint (ie: 10%), there are a lot of video freezes.

Setting the Dependency Description extension to `recvonly` seems to fix the problem. However, setting `recvonly` direction does not allow properly handling AV1 SVC within pipe transports.

STR:

- Set `sendrecv` direction for the extension header in [supportedRtpCapabilities.ts](https://github.com/versatica/mediasoup/blob/v3/node/src/supportedRtpCapabilities.ts#L314). 
- Start a chrome tab sending a AV1 stream (in mediasoup demo: `?forceAV1=true`).
- Start another chrome tab (in another machine) receiving such AV1 stream.
- Apply download network packet loss on the machine receiving the AV1 stream (ie: 10%).

Related [webrtc issue](https://issues.webrtc.org/issues/417857735).

Due to this issue, the header extension must for now be set as `recvonly` and hence pipe producer won't receive it. This means that AV1 won't work properly within pipe transports.

We need to discard whether the indicated is a Chrome issue or it's somehow? related to us.

### Try it from https://test.mediasoup.org/

Change `testAV1` with any other name.

* Open a tab with: https://test.mediasoup.org/?roomId=testAV1&consume=false&forceAV1=true 
* Open a tab with: https://test.mediasoup.org/?roomId=testAV1&produce=false&forceAV1=true

With Network Link Conditioner or another tool, apply 10% of packet loss. The image on the second tab will be almost immediately froze.

## Notes

- Maybe related issue using VP9 SVC: https://github.com/versatica/mediasoup/issues/1616

jiyeyuran:
What's the difference between our AV1 implementation and LiveKit's? LiveKit doesn't seem to have this issue: https://github.com/livekit/livekit/issues/942

ibc:
> LiveKit doesn't seem to have this issue

LiveKit probably has the very same issue but they don't know it, or they haven't tested how it works when there is packet loss, or they haven't documented the problem in that PR.

jiyeyuran:
> > LiveKit doesn't seem to have this issue
> 
> LiveKit probably has the very same issue but they don't know it, or they haven't tested how it works when there is packet loss, or they haven't documented the problem in that PR.

[...]

PR #1606, opened 2025-09-02
Worker: Fix retransmissions, set proper marker bit

jmillan:
Retransmissions must contain the marker bit value of the originally sent packet.

Changes only applies to SVC as corresponding codecs are the only ones modifying it.

This greatly enhances AV1 issue described [here](https://github.com/versatica/mediasoup/issues/1536), maybe even fixing it.

NOTE: The mentioned issue is not fixed with this PR. But the quality of the rendered video is enhanced when DD RTP extension is not sent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### NEXT
 
 Worker: Worker: Fix retransmissions, set proper marker bit
-([PR #N](https://github.com/versatica/mediasoup/pull/N)).
+([PR #1606](https://github.com/versatica/mediasoup/pull/1606)).
 
 ### 3.19.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-Worker: Worker: Fix retransmissions, set proper marker bit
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### NEXT
 
+Worker: Worker: Fix retransmissions, set proper marker bit
```
ibc: Typo

ibc (approved):
Good catch but please explain to me how it is possible that this issue only affects SVC and why just now in AV1.

jmillan:
> Good catch but please explain to me how it is possible that this issue only affects SVC and why just now in AV1.

The problem exists regardless of which codecs affects, this is, the `marker` bit in the restransmissions must set to the same value as the original packet, per spec.

Only SVC codecs do alter `marker` bit, hence this is only related to SVC. Check VP8 or H264 code and see how they completely ignore the `marker` referenece provided. They never change, and even more, they are unaware of its existence.

We narrow then to VP9 and AV1. Either VP9 does not need to check the marker and only checks the payload or it's more resilient to it being wrongly set. AV1 does seem to be affected by it, at least in my tests I feel the rendered video smoother.

PR #1610, opened 2025-09-12
AV1 issues investigation

jmillan:
- [x] Do not rewrite frame number in DD as the frame dependency template does reference the original ones.
- [x] Include Active Decode Targets bitmask in each RTP packet as commented [here](https://github.com/versatica/mediasoup/issues/1536#issuecomment-3283553329) . 
- [x] Include DD on `RtpPacket->Dump()`
- [x] Implement logic to encode and decode payload (for retransmissions), adding the correct Active Decode Targets bitmask for each case.

### Resources

https://aomediacodec.github.io/av1-rtp-spec


worker/src/RTC/SvcConsumer.cpp
```diff
@@ -1,12 +1,14 @@
...
 #define MS_CLASS "RTC::SvcConsumer"
 // #define MS_LOG_DEV_LEVEL 3
 
-#include "RTC/SvcConsumer.hpp"
 #include "DepLibUV.hpp"
 #include "Logger.hpp"
 #include "MediaSoupErrors.hpp"
 #include "Utils.hpp"
 #include "RTC/Codecs/Tools.hpp"
+#include "RTC/SvcConsumer.hpp"
```
ibc: Run make format. This change is not right.

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -57,6 +57,8 @@ namespace RTC
 				return nullptr;
 			}
 
+			dependencyDescriptor->Dump();
```
ibc: Add TODO: Remove comment

ibc (dismissed):

worker/src/RTC/SvcConsumer.cpp
jmillan: This is just an investigation branch while troubleshooting.  It's a draft, not a PR. I'll care about cosmetics later :+1:

worker/include/RTC/RtpPacket.hpp
```diff
@@ -648,6 +648,10 @@ namespace RTC
 	private:
 		void ParseExtensions();
 
+		/* Pure virtual methods inherited from RTC::RtpStreamRecv::Listener. */
+	public:
+		void OnDependencyDescriptorUpdated(uint8_t* data, size_t len);
```
ibc: Please add `UpdateDependencyDescriptorExtension()` method so we have all together instead of moving similar code into a listener. I mean, make `OnDependencyDescriptorUpdated` listener call `UpdateDependencyDescriptorExtension()`.

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -240,5 +443,127 @@ namespace RTC
...
 			return true;
 		}
+
+		const uint8_t* DependencyDescriptor::Serialize(uint8_t& len)
+		{
+			MS_TRACE();
+
+			this->bitStream.Reset();
+			this->WriteMandatoryDescriptorFields();
+			this->WriteExtendedDescriptorFields();
```
ibc: Please no this-> when calling methods.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -53,7 +87,12 @@ namespace RTC
 			bool ReadExtendedDescriptorFields();
 			bool ReadTemplateDependencyStructure();
 			bool ReadTemplateLayers();
+			bool ReadTemplateDecodeTargetIndications();
+			bool ReadTemplateFrameDiffs();
+			bool ReadTemplateFrameDiffChains();
 			bool ReadFrameDependencyDefinition();
+			bool WriteMandatoryDescriptorFields();
+			bool WriteExtendedDescriptorFields();
```
ibc, line 102: Those methods should be private AFAIS

jmillan: They are private.

worker/include/RTC/RtpPacket.hpp
jmillan: done

worker/fuzzer/src/RTC/Codecs/FuzzerAV1.cpp
```diff
@@ -1,17 +1,24 @@
...
+
 void Fuzzer::RTC::Codecs::AV1::Fuzz(const uint8_t* data, size_t len)
 {
+	Listener listener;
 	std::unique_ptr<::RTC::Codecs::DependencyDescriptor::TemplateDependencyStructure> templateDependencyStructure;
 
-	auto* dependencyDescriptor =
-	  ::RTC::Codecs::DependencyDescriptor::Parse(data, len, templateDependencyStructure);
+	auto dependencyDescriptor = std::unique_ptr<::RTC::Codecs::DependencyDescriptor>(
+	  ::RTC::Codecs::DependencyDescriptor::Parse(data, len, &listener, templateDependencyStructure));
```
ibc: Not critical, but I always try to use `std::addressof(listener)` instead than `&listener`.

worker/fuzzer/src/RTC/Codecs/FuzzerDependencyDescriptor.cpp
```diff
@@ -1,12 +1,22 @@
...
+
 void Fuzzer::RTC::Codecs::DependencyDescriptor::Fuzz(const uint8_t* data, size_t len)
 {
 	std::unique_ptr<::RTC::Codecs::DependencyDescriptor::TemplateDependencyStructure> templateDependencyStructure;
 
+	Listener listener;
+
 	auto* descriptor =
-	  ::RTC::Codecs::DependencyDescriptor::Parse(data, len, templateDependencyStructure);
+	  ::RTC::Codecs::DependencyDescriptor::Parse(data, len, &listener, templateDependencyStructure);
```
ibc: same

worker/include/RTC/Codecs/AV1.hpp
```diff
@@ -65,14 +63,17 @@ namespace RTC
 				uint8_t spatialLayer{ 0 };
 				uint8_t temporalLayer{ 0 };
 
+				std::unique_ptr<Codecs::DependencyDescriptor> dependencyDescriptor{ nullptr };
+
 				// Parsed values.
 				bool isKeyFrame{ false };
```
ibc: any reason for these estrange blank lines here?

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -1,49 +1,86 @@
...
+			class Listener
+			{
+			public:
+				virtual ~Listener() = default;
+
+			public:
+				virtual void OnDependencyDescriptorUpdated(const uint8_t* data, size_t len) = 0;
+			};
+
+			enum class DecodeTargetIndication : uint8_t
```
ibc, line 22: Please add public or private before this as we always do.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -1,49 +1,86 @@
...
+			{
+				NOT_PRESENT = 0,
+				DISCARDABLE = 1,
+				SWITCH      = 2,
+				REQUIRED    = 3
+			};
+
+			static std::unordered_map<DecodeTargetIndication, std::string> DtiToString;
```
ibc, line 31: Please add public or private before this as we always do.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -1,49 +1,86 @@
...
+
 			struct FameDependencyTemplate
```
ibc, line 34: Please add public or private before this as we always do.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -1,49 +1,86 @@
...
 			{
 				uint32_t spatialLayer;
 				uint32_t temporalLayer;
+				std::vector<DecodeTargetIndication> decodeTargetIndications;
+				std::vector<uint8_t> frameDiffs;
+				std::vector<uint8_t> frameDiffChains;
 			};
 
 			struct TemplateDependencyStructure
```
ibc, line 44: Please add public or private before this as we always do.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -1,49 +1,86 @@
...
+			std::vector<uint8_t> decodeTargetProtectedBy;
+			std::optional<uint32_t> activeDecodeTargetsBitmask{ std::nullopt };
 			// Calculated.
 			uint8_t temporalLayer{ 0 };
 			uint8_t spatialLayer{ 0 };
 			// Whether the frame is a key frame. Set to true if the descriptor contains template layers.
 			bool isKeyFrame{ false };
 
+		private:
+			DependencyDescriptor::Listener* listener;
```
ibc:
```suggestion
			DependencyDescriptor::Listener* listener{ nullptr };
```

worker/src/RTC/Codecs/AV1.cpp
```diff
@@ -63,62 +62,51 @@ namespace RTC
...
-			// We are overriding a value, we don't need the real buffer length.
-			static const uint8_t len = 64;
-
-			Utils::BitStream bitStream{ data, len };
-
-			static unsigned maxFrameNumber = std::numeric_limits<uint16_t>::max();
+			if (!this->encoder.has_value())
+			{
+				MS_WARN_DEV("there is no encoder present");
+			}
```
ibc, line 87: just a warning? shouldn't this do a return?

worker/src/RTC/Codecs/AV1.cpp
```diff
@@ -281,15 +254,14 @@ namespace RTC
...
-			uint16_t frameNumber;
-
-			context->frameNumberManager.Input(this->payloadDescriptor->frameNumber, frameNumber);
-
 			// Store the encoding data for retransmissions.
-			this->payloadDescriptor->CreateEncoder({ frameNumber });
-
-			uint8_t len;
-			this->payloadDescriptor->Encode(packet->GetDependencyDescriptionExtension(len));
+			// clang-format off
```
ibc, line 260: do we need clang-format off here?

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -101,18 +116,67 @@ namespace RTC
...
 					MS_DUMP_CLEAN(indentation + 3, "</FrameDependencyTemplate>");
 				}
 				MS_DUMP_CLEAN(indentation + 2, "</TemplateLayers>");
-				MS_DUMP_CLEAN(indentation + 1, "  </TemplateDependencyStructure>");
+				MS_DUMP_CLEAN(indentation + 1, "</TemplateDependencyStructure>");
 			}
 
+			if (this->activeDecodeTargetsBitmask.has_value())
+			{
+				MS_DUMP_CLEAN(indentation, "  <ActiveDecodeTargets>");
```
ibc: this is not ok, there shouldn't be spaces before `<ActiveDecoderTarget>`. It should be controlled with `indentantion + N` instead.

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -101,18 +116,67 @@ namespace RTC
...
+				MS_DUMP_CLEAN(
+				  indentation + 1,
+				  "  %s",
+				  std::bitset<32>(this->activeDecodeTargetsBitmask.value()).to_string().c_str());
+				MS_DUMP_CLEAN(indentation, "  </ActiveDecodeTargets>");
```
ibc: Same.

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -217,11 +315,109 @@ namespace RTC
...
+		bool DependencyDescriptor::ReadTemplateDecodeTargetIndications()
+		{
+			MS_TRACE();
+
+			auto templateCount = this->templateDependencyStructure->templateLayers.size();
+
+			for (size_t templateIndex = 0; templateIndex < templateCount; templateIndex++)
+			{
+				for (uint8_t dtIndex = 0; dtIndex < this->templateDependencyStructure->decodeTargetCount;
+				     dtIndex++)
```
ibc:
```suggestion
				     ++dtIndex)
```

consistency

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -217,11 +315,109 @@ namespace RTC
...
+			}
+
+			return true;
+		}
+
+		bool DependencyDescriptor::ReadTemplateFrameDiffChains()
+		{
+			MS_TRACE();
+
+			// TODO: Check left bits.
```
ibc: how important is this TODO and its implications?

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -240,5 +436,122 @@ namespace RTC
 
 			return true;
 		}
+
+		const uint8_t* DependencyDescriptor::Serialize(uint8_t& len)
+		{
+			MS_TRACE();
+
+			MS_ASSERT(!this->isKeyFrame, "Serialization of key frames is not supported");
```
ibc:
```suggestion
			MS_ASSERT(!this->isKeyFrame, "serialization of key frames is not supported");
```

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -240,5 +436,122 @@ namespace RTC
...
+			}
+
+			auto availableSpatialLayers  = this->templateDependencyStructure->spatialLayers;
+			auto availableTemporalLayers = this->templateDependencyStructure->temporalLayers;
+
+			MS_DUMP(
+			  "availableSpatialLayers: %d, maxSpatialLayer: %d", availableSpatialLayers, maxSpatialLayer);
+			MS_DUMP(
+			  "availableTemporalLayers: %d, maxTemporalLayer: %d", availableTemporalLayers, maxTemporalLayer);
+			;
```
ibc: remove

worker/src/RTC/Codecs/DependencyDescriptor.cpp
```diff
@@ -240,5 +436,122 @@ namespace RTC
...
+			MS_ASSERT(
+			  maxSpatialLayer <= availableSpatialLayers,
+			  "maxSpatialLayer must be less than or equal to availableSpatialLayers");
+			MS_ASSERT(
+			  maxTemporalLayer <= availableTemporalLayers,
+			  "maxTemporalLayer must be less than or equal to availableTemporalLayers");
+
+			uint32_t activeDecodeTargetsBitmask = 0;
+			size_t bitIndex                     = 0;
+			for (uint32_t spatialLayer = 0; spatialLayer <= maxSpatialLayer; ++spatialLayer)
```
ibc, line 539:
```suggestion
			uint32_t activeDecodeTargetsBitmask = 0;
			size_t bitIndex                     = 0;

			for (uint32_t spatialLayer = 0; spatialLayer <= maxSpatialLayer; ++spatialLayer)
```

worker/src/RTC/Producer.cpp
```diff
@@ -1356,6 +1356,23 @@ namespace RTC
 				{
 					std::memcpy(bufferPtr, extenValue, extenLen);
 
+					// Make place for the active decode target bitmask.
+					// clang-format off
+					if (
+					  (packet->HasOneByteExtensions() &&
+					   extenLen + 5 <= RTC::Consts::OneByteRtpExtensionMaxLength) ||
+					  (packet->HasTwoBytesExtensions() &&
```
ibc, line 1364: missing indentation space here. clang-format off is hiding it

worker/src/RTC/RtpPacket.cpp
```diff
@@ -597,6 +619,30 @@ namespace RTC
...
+	void RtpPacket::UpdateDependencyDescriptor(const uint8_t* data, size_t len)
+	{
+		MS_TRACE();
+
+		uint8_t extenLen;
+		uint8_t* extenValue = GetExtension(this->dependencyDescriptorExtensionId, extenLen);
+
+		if (!extenValue)
+		{
+			MS_ERROR("Dependency description not found");
```
ibc: please, no uppercase logs... 
Also, do we really need a MS_ERROR? It's not an error, it's a warning because it depends on what the client sends, right? we almost never use MS_ERROR

worker/src/RTC/RtpPacket.cpp
```diff
@@ -597,6 +619,30 @@ namespace RTC
...
+			return;
+		}
+
+		std::memcpy(extenValue, data, len);
+
+		auto result = SetExtensionLength(this->dependencyDescriptorExtensionId, len);
+
+		if (!result)
+		{
+			MS_ERROR("Failed to set extension length");
```
ibc: Same.

worker/src/RTC/SvcConsumer.cpp
```diff
@@ -842,7 +842,7 @@ namespace RTC
 		packet->SetMarker(origMarker);
 
 		// Restore the original payload if needed.
-		packet->RestorePayload();
+		// packet->RestorePayload();
```
ibc: What is going on here? No need to call `packet->RestorePayload();` anymore? If so remove the commented code.

worker/src/Utils/BitStream.cpp
```diff
@@ -57,6 +84,34 @@ namespace Utils
 		this->offset += count;
 	}
 
+	/*
+	 * non-symmetric unsigned encoded integer with maximum
+	 * number of values n (i.e., output in range 0..n-1).
+	 */
+	uint32_t BitStream::ReadNs(uint32_t n)
+	{
```
ibc, line 99: MS_TRACE() missing. Perhaps also in other functions.

worker/test/src/RTC/Codecs/TestDependencyDescriptor.cpp
```diff
@@ -19,6 +28,7 @@ SCENARIO("parse Dependency Descriptor", "[codecs][DD]")
 		 *         "templateIdOffset" : 0,
 		 *         "templateInfo" : {
 		 *           "0" : {
+
```
ibc: this blank line doens't look good

worker/test/src/RTC/Codecs/TestDependencyDescriptor.cpp
```diff
@@ -112,12 +182,119 @@ SCENARIO("parse Dependency Descriptor", "[codecs][DD]")
...
-		  Codecs::DependencyDescriptor::Parse(data, sizeof(data), templateDependencyStructure));
+		auto dependencyDescriptor =
+		  std::unique_ptr<Codecs::DependencyDescriptor>(Codecs::DependencyDescriptor::Parse(
+		    data, sizeof(data), &listener, templateDependencyStructure));
 
 		REQUIRE(dependencyDescriptor);
 		REQUIRE(dependencyDescriptor->frameNumber == 303);
 	}
+
+	SECTION("Serialize")
```
ibc: please use lowercase everywhere
```suggestion
	SECTION("serialize")
```

ibc (dismissed):
Mostly cosmetic comments but some of them are requested changes.

ibc:
Please add CHANGELOG entry.

worker/src/RTC/Codecs/AV1.cpp
jmillan: totally correct :+1:

worker/src/RTC/Codecs/AV1.cpp
jmillan: yes, otherwise we get the following block:

```c++
			this->payloadDescriptor->CreateEncoder(
			  { static_cast<uint32_t>(context->GetCurrentSpatialLayer()),
			    static_cast<uint32_t>(context->GetCurrentTemporalLayer()) });

```

ibc:
also just in case, please merge master branch

worker/src/RTC/Codecs/DependencyDescriptor.cpp
jmillan: Correct, I've modified MS_DUMP_CLEAN to support up to 4 intents.

worker/src/RTC/RtpPacket.cpp
jmillan: Right.., leftovers from developement..

worker/src/RTC/SvcConsumer.cpp
jmillan: Right, this was a leftover. In case of AV1 it does nothing. Nor it's needed for VP9 which is not rewritten. Left there in case we add other SVC codecs in the future.

I've left the line uncommented.

worker/src/RTC/Codecs/DependencyDescriptor.cpp
jmillan: Missing actions added.

jmillan:
This PR is ready. Forwarding DD is now provides at least the same quality as not forwarding it.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- AV1: Add support for DD extension header forwarding ([#1610](https://github.com/versatica/mediasoup/pull/1610))
```
ibc: You have started a dark battle against dots at the end of sentences and I won't give up.


```suggestion
- AV1: Add support for DD extension header forwarding ([#1610](https://github.com/versatica/mediasoup/pull/1610)).
```

ibc (approved):

vpalmisano:
@jmillan After this change the piped AV1 consumers are still not working :(

jmillan:
Not at all? I'll test it as soon as I can.

vpalmisano:
Yep, no packets are delivered to the consumer on the piped router.
I'm trying to reproduce it with mediasoup-demo.

ibc:
Just in case, are we negotiating DD extension in router.pipeToRouter()?

vpalmisano:
It should, we use the `sendrecv` mode.

vpalmisano:
Before the change made by @jmillan , I've proposed this one https://github.com/versatica/mediasoup/pull/1617 that works but probably it forwards everything from the producer to the receivers.

ibc:
packet->Dump() in shows if DD header is present

ibc:
And there is a C define in RtpPacket.cpp to make Dump() method print all the details of DD.

jmillan:
Ok, here the summary of actions and rationale behind this PR.

Firstly AV1 capability was added [here](https://github.com/versatica/mediasoup/pull/1508). It consisted of minimum DD parsing in order to get the spatial and temporal layers and frame number, basically. The RTP header extension was being explicitly left to `recvonly`, this is, mediasoup did not forward it.

This lead to the limitation of AV1 + DD not being able to be used in pipe transports.

In a second phase, as I was testing forwarding the DD, I realised of the problem with the freezes when forwarding the DD and that is why I created the corresponding [mediasoup](https://github.com/versatica/mediasoup/issues/1536) and [chrome](https://issues.webrtc.org/issues/417857735) issues.

Upon investigating such issue, we starting rewriting the `frameNumber` of the DD to provide a continuous `frameNumber` to the receiver. Which I later realized it was wrong. We cannot alter the `frameNumber` because DD internally does contain a inter layer dependency relation based on the original `frameNumber`, so changing that would make all dependency structure incorrect. We also realized that webkit does not rewrite it.

To this point, `frameNumber` rewrite was removed, and following the AV1 payload spec we added to each packet the `Active Decode Target Bitmask` which basically indicates which layers are being sent, so the decoder knows at any time if a certain layer is no longer being sent.

Apart from writing such bitmask, we don't manipulate the packet, so if simply forwarding the packet works with pipe transports and doing that with the updated packets does not, then there must be obviously a problem there.

I would suggest we check that. This patch would remove the edition of the packet for AV1 + DD:

```diff
diff --git a/worker/src/RTC/Codecs/AV1.cpp b/worker/src/RTC/Codecs/AV1.cpp
index c3fbb7c10..da28d47a3 100644
--- a/worker/src/RTC/Codecs/AV1.cpp
+++ b/worker/src/RTC/Codecs/AV1.cpp
@@ -256,15 +256,6 @@ namespace RTC
                                context->SetCurrentTemporalLayer(tmpTemporalLayer);
                        }

-                       // Store the encoding data for retransmissions.
-                       // clang-format off
-                       this->payloadDescriptor->CreateEncoder({
-                         static_cast<uint32_t>(context->GetCurrentSpatialLayer()),
-                         static_cast<uint32_t>(context->GetCurrentTemporalLayer())
-                       });
-                       // clang-format on
-                       this->payloadDescriptor->Encode();
-
                        return true;
                }

diff --git a/worker/src/RTC/Producer.cpp b/worker/src/RTC/Producer.cpp
index 6dc6d6867..05e7a581e 100644
--- a/worker/src/RTC/Producer.cpp
+++ b/worker/src/RTC/Producer.cpp
@@ -1356,23 +1356,6 @@ namespace RTC
                                {
                                        std::memcpy(bufferPtr, extenValue, extenLen);

-                                       // Make place for the active decode target bitmask.
-                                       // clang-format off
-                                       if (
-                                         (packet->HasOneByteExtensions() &&
-                                               extenLen + 5 <= RTC::Consts::OneByteRtpExtensionMaxLength) ||
-                                         (packet->HasTwoBytesExtensions() &&
-                                               extenLen + 5 <= RTC::Consts::TwoBytesRtpExtensionMaxLength)
-                                       )
-                                       // clang-format on
-                                       {
-                                               extenLen += 5;
-                                       }
-                                       else
-                                       {
-                                               MS_WARN_DEV("cannot increase DD extension header length, current length %zu", extenLen);
-                                       }
-
                                        extensions.emplace_back(
                                          static_cast<uint8_t>(RTC::RtpHeaderExtensionUri::Type::DEPENDENCY_DESCRIPTOR),
                                          extenLen,
```

Now I have several important questions I need to answer myself after all, because we need a proper summary of the current state:

1. Do the freezes happen upon packet loss when DD extension is `recvonly`? This is crucial to know as it would completely turn the ongoing investigation. At this point I'm not completely sure of this, and must be assured.
2. Do the rendering work with pipe transports if the patch above is applied? If so, then there is obviously a problem with the packet rewrite. If not, then there is a more deep problem as we are merely passing the packet down the pipe as with the rest of codecs.
3. LiveKit apparently does not have the freezes when negotiating AV1 + DD and applying downlink packet loss. I've used [this link](https://meet.livekit.io/rooms/whln-z760?codec=av1) few times and the number of freezes was lower, but this needs to be revisited. We need to know which scalability mode is used there, as we know that L1T3 also works in mediasoup. 

Now, we need to implement pipe transport usage in `mediasoup-demo` in order to test this and pipe transports in general, in order to be able to test this reliably. Believe it or not, up to this date, there is no such implementation in the demo.

ibc:
I will implement pipe transport in the new demo server next week. Said that, what about moving your previous comment into a new separate issue @jmillan ?

vpalmisano:
> 3\. LiveKit apparently does not have the freezes when negotiating AV1 + DD and applying downlink packet loss. I've used [this link](https://meet.livekit.io/rooms/whln-z760?codec=av1) few times and the number of freezes was lower, but this needs to be revisited. We need to know which scalability mode is used there, as we know that L1T3 also works in mediasoup.

They use "L3T3_KEY":

```
[
    {
        "active": true,
        "adaptivePtime": false,
        "maxBitrate": 1190000,
        "maxFramerate": 30,
        "networkPriority": "low",
        "priority": "low",
        "scalabilityMode": "L3T3_KEY"
    }
]
```

vpalmisano:
> I will implement pipe transport in the new demo server next week. Said that, what about moving your previous comment into a new separate issue @jmillan ?

I'm running my test with these changes https://github.com/versatica/mediasoup-demo/pull/149

jmillan:
NOTE: I'll create a proper issue with all this info.

> 1. Do the freezes happen upon packet loss when DD extension is recvonly? This is crucial to know as it would completely turn the ongoing investigation. At this point I'm not completely sure of this, and must be assured.

No, setting the DD extension to `recvonly` there are virtually no freezes (`freezeCount`, `totalFreezesDuration`) compared to using `sendrecv`.

jmillan:
NOTE: There are as much freezes if we modify DD (to add the Active Dependency Targets bitmask) or if we don't touch it, by applying [this patch](https://github.com/versatica/mediasoup/pull/1610#issuecomment-3414251698).

jmillan:
So as for now we have these important points:
* There is not known implementation working with AV1 S3T3. mediasoup and google meet work with S1T3. So maybe we are chasing ghosts here.
* We'll need to try S3T3_KEY, as at least we know livekit does use it.
* We need to investigate why pipe transports do not work when using `sendrecv` on DD.

ibc:
> * We'll need to try S3T3_KEY, as at least we know livekit does use it.

This can be tested already using the demo with with `&webcamScalabilityMode=S3T3_KEY`

PR #1614, opened 2025-09-24
Require Node >= 22

jmillan:
- [x] package.json: engines field.
- [x] Installation and Requirements in the website.

> - Maybe also in README of the project (not sure).

It's not there and I've never missed it anyway. IMO we don't need to put it there.

> Remove the --force in npm ci in mediasoup-node.yaml (not sure if in there or in other projects, if so there is a self-explanatory note on the line).

It was already removed.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -41,6 +41,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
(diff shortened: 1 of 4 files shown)


.github/workflows/codeql.yaml
```diff
@@ -41,6 +41,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
```
ibc: Please node 24

ibc:
> * Installation and Requirements in the website.

Pushed: https://mediasoup.org/documentation/v3/mediasoup/installation/

ibc (approved):

ibc:
Let's merge.

PR #1617, opened 2025-10-06
Make dependency descriptor header sendrecv

vpalmisano:
We need to make the dependency descriptor header sendrecv, otherwise producers created inside piped transports will not forward the required header and the consumers will not receive any AV1 packet.

### DIFF
--- a/node/src/supportedRtpCapabilities.ts
+++ b/node/src/supportedRtpCapabilities.ts
@@ -266,7 +266,7 @@ const supportedRtpCapabilities: RouterRtpCapabilities = {
 			uri: 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension',
 			preferredId: 8,
 			preferredEncrypt: false,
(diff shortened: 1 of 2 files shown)


jmillan:
Not that easy :-), there is a long story behind that. I've just merged it https://github.com/versatica/mediasoup/pull/1610

jmillan:
Closing as it does not apply.

ibc:
@jmillan I'm a bit confused.

I'm checking and I don't see any open ticket that mentions this problem we have with DD and pipe transports. But I think we had it... Do I miss something?

jmillan:
We could not use AV1 with pipe tra sport because we could not enable sendrecv for DD due to the issue handled in the comitted PR. The whole reason for not enabling sendrecv in DD was that doing so would generate total freeze in receiver upon packet loss.

jmillan:
> I'm checking and I don't see any open ticket that mentions this problem we have with DD and pipe transports. But I think we had it... Do I miss something?

Check the description of #1536

PR #1618, opened 2025-10-08
RtpPacket: update Dependency Descriptor listener on Clone()

jmillan:
The outgoing RTP packet is not the original packet which parsed the Dependency Decriptor, hence the cloned packet (the packet being sent) needs to be set as the new listener
 of Dependency Descriptor.


ibc (approved):

ibc:
CHANGELOG please.

jmillan:
Changelog done, I need to fix the tests.

ibc:
👏🏻

PR #1620, opened 2025-10-10
Fix VP9 freezes due to wrong marker bit setting

### LINKED ISSUE #1616: Unexpected freezes with VP9 SVC and packetloss
# Bug Report

## Your environment

- Chrome 140, 
- MacOS 15.6.1
- Mediasoup online demo

## Issue description

We see lots of freezes when introducing packetloss on reception of VP9 SVC. Seems to only happen when using SVC - not K-SVC. 

Reproducible with mediasoup demo when forcing VP9 and overriding default scalabilitymode:
https://v3demo.mediasoup.org/?roomId=vp9svcpacketloss&forceVP9=true&webcamScalabilityMode=L3T3

I think this might be a browser/lib issue, so I have reported it there:
https://issues.webrtc.org/issues/449408585
But if i'm wrong and it is a mediasoup issue, hopefully you can use this info to track it down

I narrowed it down to the freezes happening when dropping a single packet and then dropping the first re-transmit of that packet. Somehow we only receive a new nack (re-nack same seq number) if it is the first packet in a picture (first packet after a packet with marker bit being sent). Since it doesn't get the re-nack, the receiver never gets what it needs and gives up after a while and sends a PLI which unfreezes the image. 

I can reproduce by:
- running latest release and testing on localhost (no network lag or packetloss)
- dropping every Nth (i.e 200th) packet in WebRtcTransport (by not calling `this->iceServer->GetSelectedTuple()->Send(data, len, cb)`;)
- drop the first RTX packet after this

Logging packets i see it only freezes when not re-nacking the packet i stopped retransmission of. And it only does that when it isn't the first packet of picture.

Maybe related to:
https://github.com/versatica/mediasoup/issues/1536

jmillan:
Fixes #1616

Even though the spec says:

"This bit MUST be set to one for the final packet of the highest spatial-layer frame (the final packet of the picture) "

Since the receiver endpoint does not know which is the highest spatial-layer we are sending, marker flag must be set for every final packet of frame.

Tests demonstrate this reduces freezes massively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - AV1: Add support for DD extension header forwarding ([#1610](https://github.com/versatica/mediasoup/pull/1610)).
 - DependencyDescriptor: Update listener on RtpPacket clone ([#1618](https://github.com/versatica/mediasoup/pull/1618)).
+- VP9: Fix VP9 freezes due to wrong marker bit setting ([#1620](https://github.com/versatica/mediasoup/pull/1620)), thanks to @geirbakke
+  for reporting.
 
 ### 3.19.3
 
--- a/worker/src/RTC/Codecs/VP9.cpp
+++ b/worker/src/RTC/Codecs/VP9.cpp
@@ -381,7 +381,7 @@ namespace RTC
 			}
 
 			// Set marker bit if needed.
-			if (packetSpatialLayer == tmpSpatialLayer && this->payloadDescriptor->e)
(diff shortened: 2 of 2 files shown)


jmillan:
Merging as this PR clearly reduces highly the freezes.

jmillan:
To put some numbers, in a minute under the same packet loss:

- Before: freezeCount:12, TotalFreezesDuration 60
- After: freezeCount:5, TotalFreezesDuration 1,2

PR #1622, opened 2025-10-14
Revert marker changes vp9 av1

jmillan:
The changes were wrong. By adding marker to every end of frame the receiver can only decode the lowest spatial layer...


ibc (approved):
Please also reopen the corresponding issues.

PR #1627, opened 2025-10-20
RtpPacket: add uniqueId

ibc:
The problem is that **obviously** such an `uniqueId` is lost when sending the RTP packet through a pipe transport, which was the purpose of this PR... but anyway...

Bonus track: Improve order of items in `RtpPacket::Dump()`


ibc:
Let's close this.

PR #1629, opened 2025-10-20
RtpPacket: add abs-capture-time RTP extension API

ibc:
Because of course we accepted an external PR that didn't do it: https://github.com/versatica/mediasoup/pull/761

Result:
```
<RtpPacket>
  packet size: 645 bytes
  sequence number: 24612
  timestamp: 2233432389
  marker: false
  payload type: 101
  ssrc: 447540820
  csrc count: 0
  padding: false
  header extension: id:48862, length:36
  RFC5285 ext style: One-Byte Header
  RFC5285 ext ids: 1,4,5,13
  mid: extId:1, value:'0'
  absSendTime: extId:4
  transportWideCc01: extId:5, value:0
  absCaptureTime: extId:13, absCaptureTimestamp:17050879188517931008, estimatedCaptureClockOffset:0
  payload size: 593 bytes
  spatial layer: 0
  temporal layer: 2
</RtpPacket>
```

### DIFF
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -1,3 +1,4 @@
+#include <cstdint>
 #define MS_CLASS "RTC::RtpPacket"
 // #define MS_LOG_DEV_LEVEL 3
 // #define DUMP_PAYLOAD_DESCRIPTOR 1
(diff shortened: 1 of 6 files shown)


jmillan (approved):

worker/src/RTC/RtpPacket.cpp
```diff
@@ -312,11 +349,12 @@ namespace RTC
 				MS_DUMP_CLEAN(
 				  indentation,
 				  "  playoutDelay: extId:%" PRIu8 ", minDelay:%" PRIu16 ", maxDelay:%" PRIu16,
-				  this->videoOrientationExtensionId,
+				  this->playoutDelayExtensionId,
```
ibc, line 352: Yes, this was a bug.

worker/src/RTC/RtpPacket.cpp
```diff
@@ -1,3 +1,4 @@
+#include <cstdint>
```
jmillan, line 1: Does not belong here

ibc: It was some plugin I hate in my editor. Removed in v3.

PR #1630, opened 2025-10-20
Windows ssize fix and compile fix

CosmosisT:
Fix code and compiler issues with Windows 10.


CosmosisT:
Wasn't suppose to push yet,

PR #1631, opened 2025-10-21
Add custom 'urn:mediasoup:params:rtp-hdrext:packet-id' (mediasoup-packet-id) header extension

ibc:
# Details

- `transport.produce()` gets a new boolean option `enableMediasoupPacketIdHeaderExtension`.
- If set, then when a RTP packet is received in the mediasoup worker `Producer`, a custom header extension 'urn:mediasoup:params:rtp-hdrext:packet-id' (mediasoup-packet-id) will be added to the packet. Its value is unique (incremetal) per worker/thread.
- Such a header extension is negotiated when using `router.pipeToRouter()`, meaning that the header extension is propagated through pipe transports and accesible in the target worker/router.
- Its value is printed in `RtpPacket::Dump()` method. However, as usual, take into account that we should never call `packet->Dump()` before `Producer::ReceiveRtpPacket()` method has completed. Once in consumer classes, we can call `packet->Dump()` at any time.
- **IMPORTANT:** Ids of extensions have been modified. Nobody should rely on their static values anyway.
- Bonus track: Order header extensions following the same order everywhere (because it was HELL).

## Example

Here we are running the mediasoup-demo with pipe transports enabled.

```
// RTP packet in the original WebRTC Producer (at the end of `Producer::MangleRtpPacket()` method):

<RtpPacket>
  packet size: 162 bytes
  sequence number: 4878
  timestamp: 2704768489
  marker: false
  payload type: 100
  ssrc: 519582781
  csrc count: 0
  padding: false
  header extension: id:48862, length:16
  RFC5285 ext style: One-Byte Header
  RFC5285 ext ids: 1,6,12
  mid: extId:1, value:''
  absSendTime: extId:4
  ssrcAudioLevel: extId:6, volume:56, voice:true
  mediasoupPacketId: extId:12, mediasoupPacketId:69736763
  payload size: 130 bytes
  spatial layer: 0
  temporal layer: 0
</RtpPacket>

// RTP packet in the Producer of the PipeTransport (at the end of `Producer::MangleRtpPacket()` method):

<RtpPacket>
  packet size: 162 bytes
  sequence number: 22774
  timestamp: 2704768489
  marker: false
  payload type: 100
  ssrc: 207610826
  csrc count: 0
  padding: false
  header extension: id:48862, length:16
  RFC5285 ext style: One-Byte Header
  RFC5285 ext ids: 1,6,12
  mid: extId:1, value:''
  absSendTime: extId:4
  ssrcAudioLevel: extId:6, volume:56, voice:true
  mediasoupPacketId: extId:12, mediasoupPacketId:69736763
  payload size: 130 bytes
  spatial layer: 0
  temporal layer: 0
</RtpPacket>

// RTP packet in the final WebRTC Consumer (at the beginning of `SimpleConsumer::SendRtpPacket()` method):

<RtpPacket>
  packet size: 162 bytes
  sequence number: 22774
  timestamp: 2704768489
  marker: false
  payload type: 100
  ssrc: 207610826
  csrc count: 0
  padding: false
  header extension: id:48862, length:16
  RFC5285 ext style: One-Byte Header
  RFC5285 ext ids: 1,6,12
  mid: extId:1, value:'0'
  absSendTime: extId:4
  ssrcAudioLevel: extId:6, volume:56, voice:true
  mediasoupPacketId: extId:12, mediasoupPacketId:69736763
  payload size: 130 bytes
  spatial layer: 0
  temporal layer: 0
</RtpPacket>
```


pnts-se-whereby:
I'm sorry if the answer to this question is too obvious, but I'll ask anyway.

To me it looks like `enableMediasoupPacketIdHeaderExtension` will make this change backwards compatible with older versions, i.e. in this PR we do not both add the extension and make it mandatory (assume other routers we might pipe to supports it). Am I correct?

ibc:
@pnts-se-whereby this is an optional feature intended for development/debugging that will never be mandatory and nobody should use it other than us.

worker/include/RTC/RtpPacket.hpp
```diff
@@ -5,6 +5,7 @@
 #include "Utils.hpp"
 #include "FBS/rtpPacket.h"
 #include "RTC/Codecs/PayloadDescriptorHandler.hpp"
+#include <cstdint>
```
jmillan: This is present in 'common.hpp' already. I've seen this header file included in other file/s in this PR.

worker/src/RTC/Producer.cpp
```diff
@@ -1,13 +1,15 @@
+#include "RTC/RtpPacket.hpp"
+#include <cstdint>
```
jmillan: hello

worker/include/RTC/RtpPacket.hpp
ibc: Ok some C++ plugin I added in my editor. Will remove.

Said that there are/were many C++ files in mediasoup that use uint8_t etc types without including out common.hpp header.

worker/src/RTC/Producer.cpp
```diff
@@ -1331,73 +1355,127 @@ namespace RTC
...
+					highestExtenLen = GetHighestExtenLen(highestExtenLen, extenLen);
 
 					// Not needed since this is the latest added extension.
 					// bufferPtr += extenLen;
 				}
 			}
 
 			// Set the new extensions into the packet.
-			packet->SetExtensions(packet->HasTwoBytesExtensions() ? 2 : 1, extensions);
+			// Use 1-byte or 2-bytes type depending on the highest extension id and
```
jmillan, line 1445: Why don't we calculate the `highestExtenId` and `highestExtenLen` here, by just iterating the `extensions` vector, rather than calculating it every time an extension is added? I find it more clean and concise doing it on it's own block, and it would reduce the scope of the `highestExtenId` and `highestExtenLen` variables.

worker/src/RTC/SimpleConsumer.cpp
```diff
@@ -319,6 +319,8 @@ namespace RTC
 	{
 		MS_TRACE();
 
+		packet->Dump();
```
jmillan: Leftover

jmillan (approved):
Just few comments added.

worker/src/RTC/Producer.cpp
ibc: Done here and verified in local testing: https://github.com/versatica/mediasoup/pull/1631/commits/8a82168cadc63515c0ffc5543a6c5c18987de519

PR #1632, opened 2025-10-22
AV1: do not forward Dependency Descriptor header extension

jmillan:
Do not forward to the endpoints but do forward it to the pipe producers so pipe SvcConsumers can read it.

### DIFF
--- a/node/src/ortc.ts
+++ b/node/src/ortc.ts
@@ -40,6 +40,18 @@ const DynamicPayloadTypes = [
 	99,
 ];
 
+// TODO: Remove this if we switch to 'sendrecv' in Dependency-Descriptor header
+// extension.
+//
+// This is an object where we store some objects we may later need.
+type Cache = {
+	dependencyDescriptorHeaderExtensionParametersForPipeConsumer?: RtpHeaderExtensionParameters;
+};
+
+const cache: Cache = {
+	dependencyDescriptorHeaderExtensionParametersForPipeConsumer: undefined,
+};
+
 /**
  * Validates RtpCapabilities. It may modify given data by adding missing
  * fields with default values.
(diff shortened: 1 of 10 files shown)


ibc:
DD is now propagated to piped Producer (and ofc not propagated to final SVC consumer). Commit: https://github.com/versatica/mediasoup/pull/1632/commits/63951942f7c6ce996666e069528e09d2a21cdebd

rust/src/ortc.rs
```diff
@@ -380,6 +397,26 @@ pub(crate) fn generate_router_rtp_capabilities(
...
+    // extension.
+    //
+    // We need to create and store this Dependency-Descriptor header extension to
+    // leter be used by `getPipeConsumerRtpParameters()` function.
+    let dependency_descriptor_header_extension_for_pipe_consumer = caps
+        .header_extensions
+        .iter()
+        .find(|ext| ext.uri == RtpHeaderExtensionUri::DependencyDescriptor);
+
+    if let Some(dependency_descriptor) = dependency_descriptor_header_extension_for_pipe_consumer {
```
ibc, line 411: Here this condition is missing:

```ts
&& dependencyDescriptorHeaderExtensionForPipeConsumer.direction !== 'sendrecv'
```

See `ortc.ts`: https://github.com/versatica/mediasoup/pull/1632/files#diff-73792083f369cdd571ad32270bd78e55f49b59041095a0fda217a688870001d7R328-R331

ibc (changes requested):

rust/src/ortc.rs
jmillan: good catch.

rust/src/ortc.rs
```diff
@@ -380,6 +397,28 @@ pub(crate) fn generate_router_rtp_capabilities(
...
+        if dependency_descriptor.direction != RtpHeaderExtensionDirection::SendRecv {
```
ibc, line 412: Do we really need this additional if condition within the parent if condition? Can't we merge both in same if line?

jmillan: The way to do it would be 

```rust
if let Some(dependency_descriptor) = dependency_descriptor_header_extension_for_pipe_consumer
    if dependency_descriptor.direction != RtpHeaderExtensionDirection::SendRecv
{
```

, which doesn't seem very readable to me. I've modified the previous block to not only check for the DD but also make sure the direction is the desired one. Let me know what you prefer.

NOTE: This is not correct, updating...

jmillan: It could be achieved by:

```rust
    match dependency_descriptor_header_extension_for_pipe_consumer {
        Some(dependency_descriptor)
            if dependency_descriptor.direction != RtpHeaderExtensionDirection::SendRecv =>
        {
            let mut cache = CACHE.lock();
            cache.dependency_descriptor_header_extension_parameters_for_pipe_consumer =
                Some(RtpHeaderExtensionParameters {
                    uri: dependency_descriptor.uri,
                    id: dependency_descriptor.preferred_id,
                    encrypt: dependency_descriptor.preferred_encrypt,
                });
        }
        _ => {}
    }
```

But it's not more readable. I would prefer keeping the original code or the one commited last.

ibc: Ok, I assu,e this is not valid Rust, right? XD

```rs
if let Some(dependency_descriptor) =
    dependency_descriptor_header_extension_for_pipe_consumer &&
    dependency_descriptor.direction != RtpHeaderExtensionDirection::SendRecv {
```

Assuming it's not valid Rust, please go ahead with your preferred syntax.

ibc (approved):
I'm approving but please add CHANGELOG.

node/src/ortc.ts
```diff
@@ -300,6 +312,30 @@ export function generateRouterRtpCapabilities(
 		}
 	}
 
+	// TODO: Remove this if we switch to 'sendrecv' in Dependency-Descriptor header
```
jiyeyuran: Is the reason we can't use sendrecv right now a bug in libwebrtc?

PR #1635, opened 2025-10-28
SeqManager: Fix, properly account out of order drops

### LINKED ISSUE #1615: VP9: Unexpected packet loss statistics on layer change
# Bug Report
Hello and thank you for all your work!

I was looking at RTC getStats() reports for meetings using VP9 codec. I noticed that packetloss % went up a lot when dropping to a lower spatial layer. I'm not sure there's a problem here, but I'm reporting this as an unexpected behaviour for now.

I have connected with one client that have 3% packet loss on outbound UDP traffic. From another client, I'm looking at graphs for inbound RTP traffic in about://webrtc-internals. As I drop to the lowest layers (spatial 0, temporal 0) at about 11:31 the rate of lost packets starts increasing a bit. 

As we now receive less packets in general the packet loss % change is dramatic, which is how I noticed it.

Image

Image

## Your environment

- Operating system: MacOS 15.6.1, Chrome Version 140.0.7339.186
- https://demo.mediasoup.org/
- mediasoup version: 3.19.3
- mediasoup-client version: 3.16.4

jmillan:
Fixes #1615

Current implementation does only consider dropped inputs when they are higher than the highest input seen.

Before this PR ❌

|input|output|drop|
|-|-|-|
|0|0||
|2||x|
|1||x|
|3|2||

Due to this issue, we were generating a lot of sequence number jumps (considered packet loss at receiver side, which generated unnecessary and unprocessable NACK requests) when RTP is received out of order to the producer and many packets are being discarded in consumer (ie: SVC, targeting low layers).

The following solution allows for properly accounting out of order drops until a newer input is forwarded.

After this PR ✅ 

|input|output|drop|
|-|-|-|
|0|0||
|2||x|
|1||x|
|3|1||

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- SeqManager: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)), thanks to @pnts-se-whereby for reporting.
+- `SeqManager`: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)), thanks to @pnts-se-whereby for reporting.
 
 ### 3.19.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- SeqManager: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)).
+- SeqManager: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)), thanks to @pnts-se-whereby for reporting.
 
 ### 3.19.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- SeqManager: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)).
+
 ### 3.19.6
 
 - AV1: Set DependencyDescriptor Header Extension to 'recvonly' but forward it between pipe transports ([#1632](https://github.com/versatica/mediasoup/pull/1632)).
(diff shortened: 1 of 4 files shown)


ibc:
> This is a WIP. It does not work properly for the scenario described in #1615. There are more frame drops than in v3.

What

worker/src/RTC/SeqManager.cpp
```diff
@@ -61,12 +61,20 @@ namespace RTC
...
+			this->maxDropped = input;
 			// Insert input in the last position.
 			// Explicitly indicate insert() to add the input at the end, which is
 			// more performant.
 			this->dropped.insert(this->dropped.end(), input);
 
+			ClearDropped();
+		}
+		// Input lower than the max input, which is already dropped.
+		else if (this->maxInput == this->maxDropped)
```
ibc: I don't understand this condition here which is not using the current input in the condition at all.

jmillan: * **if**: `input` is higher than `this->maxInput` -> store `input` in `this->dropped`.
* **else if**:  (`input` is NOT higher than `this->maxInput`) and `this->maxInput `was dropped -> store `input` in `this->dropped`.

In other words, if the highest seen input was dropped (this is, it was not forwarded) and `input` is lower than `this->maxInput` it means that this is a dropped input that came out of order.

We cannot store a dropped input if we already forwarded an input higher that that, which would generate duplicate and buggy sequence numbers.

jmillan: > I don't understand this condition here which is not using the current input in the condition at all.

It is using it implicitly because we are in an `else if`.

jmillan: Inline comment updated.

jmillan:
> It does not work properly for the scenario described in https://github.com/versatica/mediasoup/issues/1615. There are more frame drops than in v3.

I found the problem.., working on it.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- SeqManager: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)), thanks to @pnts-se-whereby for reporting.
```
```suggestion
- `SeqManager`: Fix, properly account out of order drops until an input is forwarded ([#1635](https://github.com/versatica/mediasoup/pull/1635)), thanks to @pnts-se-whereby for reporting.
```

ibc (approved):
Amazing. Just a cosmetic thing in CHANGELOG.

jmillan:
Before merging this I'm  testing a long standing session with both v3 and this branch and see if there is any difference in the `freezedCount` webrtc-internals stats.

jmillan:
Merging

PR #1637, opened 2025-10-30
[Simulcast|Svc]Consumer: use Layers type to avoid member duplication

jmillan:
And enhance readability.

Changed this:
```c++
		int16_t preferredSpatialLayer{ -1 };
		int16_t preferredTemporalLayer{ -1 };
		int16_t provisionalTargetSpatialLayer{ -1 };
		int16_t provisionalTargetTemporalLayer{ -1 };
		int16_t targetSpatialLayer{ -1 };
		int16_t targetTemporalLayer{ -1 };

```

Into this:
```c++
		VideoLayers preferredLayers;
		VideoLayers provisionalTargetLayers;
		VideoLayers targetLayers;
```

Apart from the obvious encapsulation, `Layers` contains overloads for `==` and `!=` so we don't need to compare spatial and and temporal layers of two different instances manually.

In general it enhances readability and reduces the possibility of human errors.

This is a previous cleanup for a further refactor in direction of merging all consumers into a single one.

NOTE: This changes have been made with automatic renaming, reviewed by myself carefully line by line and tested.

### DIFF
--- a/worker/include/RTC/Consumer.hpp
+++ b/worker/include/RTC/Consumer.hpp
@@ -43,6 +43,34 @@ namespace RTC
 		{
 			int16_t spatial{ -1 };
 			int16_t temporal{ -1 };
+
+			Layers() : spatial(-1), temporal(-1)
+			{
+			}
+
+			Layers(int16_t spatial, int16_t temporal) : spatial(spatial), temporal(temporal)
+			{
+			}
+
+			Layers(const Layers& other) : spatial(other.spatial), temporal(other.temporal)
+			{
+			}
+
+			bool operator==(const Layers& other) const
+			{
+				return spatial == other.spatial && temporal == other.temporal;
+			}
+
+			bool operator!=(const Layers& other) const
+			{
+				return !(*this == other);
+			}
+
+			void Reset()
+			{
+				spatial  = -1;
+				temporal = -1;
+			}
 		};
 
 	private:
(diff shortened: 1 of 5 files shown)


worker/include/RTC/Consumer.hpp
```diff
@@ -43,6 +43,34 @@ namespace RTC
 		{
 			int16_t spatial{ -1 };
 			int16_t temporal{ -1 };
+
+			Layers() : spatial(-1), temporal(-1)
```
ibc: Perhaps it should be VideoLayers?

ibc (approved):

worker/include/RTC/Consumer.hpp
jmillan: Yes :+1:

ibc: So it has the same name as the TS type

jmillan: In TS it's called `ConsumerLayers`.., I think `VideoLayers` is more adequate. Whether they are used in consumers is just a usage of the type IMO, but it could be used to represent the spatial/temporal layers for a specific RTP packet payload, regardless it's being received or it's being sent.

Having said this I don't have a strong orientation for one or another.

ibc: Nah, `VideoLayers` is better.

PR #1640, opened 2025-10-31
Consumers|PayloadDescriptorHandler: use VideoLayers everywhere

jmillan:
For encapsulation and readability.


worker/src/RTC/SvcConsumer.cpp
```diff
@@ -538,17 +538,7 @@ namespace RTC
 		// Reset provisional target layers.
 		this->provisionalTargetLayers.Reset();
 
-		if (!IsActive())
-		{
-			return;
-		}
```
jmillan, line 544: Removing this redundant condition since line 534 asserts `IsActive()`.

ibc (approved):

PR #1641, opened 2025-11-03
cleanup: clang-tidy pass

jmillan:


ibc (approved):
Wondering if we should have this in CI somehow.

ibc:
Why is this a draft yet?

ibc:
Also, update the branch with v3, please.

jmillan:
> Wondering if we should have this in CI somehow.

this is quite manual and clang-tidy has many false possitives. This is why I do this from time to time.

jmillan:
>  Why is this a draft yet?

I'm doing more changes.., but I'll do another PR.

nazar-pc:
> I'm doing more changes..

Why opening PR if it is not ready for review? Pushing commits one by one generates tons of email notifications.

jmillan:
> Why opening PR if it is not ready for review? Pushing commits one by one generates tons of email notifications.

I did not intend to continue adding changes and I created the PR. But some minutes later I decided to keep doing some more changes... there are trillion changes that can be done regarding clang-tidy, and I'm not going to do them all at once.

Having said this, I've marked the PR as Draft. Once I consider it done, I'll mark as ready for review.

ibc (approved):

worker/src/RTC/Serializable.cpp
```diff
@@ -127,7 +126,7 @@ namespace RTC
 
 		AssertNotFrozen();
 
-		this->consolidatedListener = std::move(listener);
+		this->consolidatedListener = listener;
```
jmillan: `std::move()` has no effect in `const`.

worker/src/handles/UdpSocketHandle.cpp
```diff
@@ -106,7 +106,14 @@ UdpSocketHandle::~UdpSocketHandle()
 
 	if (!this->closed)
 	{
-		InternalClose();
+		try
```
jmillan, line 109: `InternalClose()` can throw so we should catch it here, since no one is handing it.

worker/include/RTC/WebRtcServer.hpp
```diff
@@ -98,7 +98,7 @@ namespace RTC
 
 	public:
 		// Passed by argument.
-		const std::string id;
+		std::string id;
```
ibc, line 105: Why this?

worker/src/RTC/SctpAssociation.cpp
```diff
@@ -449,7 +449,7 @@ namespace RTC
 		// via onSendSctpData.
 		this->listener->OnSctpAssociationBufferedAmount(this, this->sctpBufferedAmount);
 
-		const int ret = usrsctp_sendv(
+		const ssize_t ret = usrsctp_sendv(
```
ibc, line 452: Why this? The function literally returns int which maybe it could be any negative value while ssize_t can only be -1 and 0 and positive integers.

worker/src/RTC/Serializable.cpp
ibc: What does it mean? Implications? It was a move for a reason that ofc I can't remember now

worker/include/RTC/WebRtcServer.hpp
jmillan: The clang-tidy rule that suggested to remove it is [this](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html). There is further reading [here](https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#data-members-never-const), which argues why having const members is never a good idea.

Having said this, this change should go along making `id` private and adding a `GetId()` API, which is how we usually do it rather than having a public `id`. 

Are we fine making `id` private and adding a `GetId()` method?

worker/src/RTC/SctpAssociation.cpp
jmillan: Indeed the `usersctp_sendv(` returns `ssize_t`, hence the change.

worker/include/RTC/WebRtcServer.hpp
ibc: I'm ok with GetId() but those links basically also recommend to avoid local scope constants and many things we are doing on purpose to, ironically, make clang-tidy happy.

worker/src/RTC/Serializable.cpp
jmillan: `std::move(const ConsolidatedListener&& listener)`  as no effect as it's expected, as you are trying to change (move) a `const` value.

If you want it demonstrated take the `CosolidatedListener` definition:
`using ConsolidatedListener = std::function<void()>;`

Now check the `std::function` move constructor definition:
```cpp
  _LIBCPP_HIDE_FROM_ABI function() _NOEXCEPT {}
  _LIBCPP_HIDE_FROM_ABI _LIBCPP_HIDE_FROM_ABI function(nullptr_t) _NOEXCEPT {}
  _LIBCPP_HIDE_FROM_ABI function(const function&);
  _LIBCPP_HIDE_FROM_ABI function(function&&) _NOEXCEPT;
```

There is a `function(function&&)` definition but not a `function(const function&&)`.

So the call to `std::move(listener);` was basically making a copy anyway. Calling `std::move()` does not mean the value is moved, the corresponding move constructor must exist, otherwise it will create a copy.

worker/include/RTC/WebRtcServer.hpp
jmillan: There is no clang-tidy warning after the change. This rule is very specific about not making members private. Otherwise you can not, for example, have this:

```c++
struct Employee {
    std::string name;
    const std::string id;
};

Employee e1 = {"Alice", "12345"};
Employee e2 = {"Bob", "24601"};
e1 = e2; // oops
std::swap(e1, e2); // oops
```

I'm going to create `GetId()`

worker/src/RTC/Serializable.cpp
ibc: But then maybe we should remove the const from the method argument and keep the move.

worker/src/RTC/ActiveSpeakerObserver.cpp
```diff
@@ -49,16 +47,13 @@ namespace RTC
 		return t;
 	}
 
-	inline double ComputeActivityScore(
+	static inline double computeActivityScore(
```
ibc, line 50: I'm running clang-tidy and I see lot of warnings about:

```
method 'AddDeltas' can be made static [readability-convert-member-functions-to-static]
```

Honestly I wouldn't like to go that way. I don't care if a class method is not accessing internal members. That's not a reason to make it static method. For example, `SrtpSession::EncryptRtp()` (etc methods) do not currently use class instance members inside so clang-tidy is warning about them ("make them static"). I don't want them to be static, they should not be static.

jmillan: clan-tidy is based on rules. We can enable/disable rules to our wish. We do not need to make static every method that can be made so.

PR #1642, opened 2025-11-05
Refactor: Use nullToUndefined for optional FlatBuffers fields

Fnoz:
This PR improves FlatBuffers null handling by introducing a type-safe utility function and applying it consistently across the codebase.


jmillan:
Thanks @Fnoz, but honestly we'd get the same with this:

```ts
rtpMapping: data.rtpMapping()?.unpack() || undefined,
```

I don't think we need an extra function just for this, considering it's used in 2/3 places.

ibc:
I agree we don't need that extra function but the thing is: are we converting those fields to undefined? Or are they exposed to the user as null? Because if the latter then we have to fix it.

PR #1643, opened 2025-11-05
Worker: modernize clang-format

jmillan:
We were using a very old clang-format out of npm [clang-format package](https://www.npmjs.com/package/clang-format) which is out of date.

Use a modern clang-format and define a specific version (21) to be run .

Also just run lint GH action for macos. We are not trying to test clang-format on different OS.

### DIFF
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -156,7 +156,7 @@ jobs:
         run: pip3 install --break-system-packages invoke
 
       # Run clang-format on macOS.
-      - if: ${{ startsWith(matrix.build.os, 'macos'}}
+      - if: ${{ startsWith(matrix.build.os, 'macos') }}
         name: invoke -r worker lint
         run: |
           brew install clang-format@21
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -159,6 +159,7 @@ jobs:
       - if: ${{ startsWith(matrix.build.os, 'macos') }}
         name: invoke -r worker lint
         run: |
+          npm ci --prefix worker/scripts
           brew install clang-format@21
           invoke -r worker lint
 
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -22,7 +22,6 @@ jobs:
           - os: ubuntu-22.04
             cc: gcc
             cxx: g++
-            run-lint: true
             run-test: true
             run-test-asan-address: false
             run-test-asan-undefined: false
@@ -30,7 +29,6 @@ jobs:
           - os: ubuntu-22.04
             cc: clang
             cxx: clang++
-            run-lint: true
             run-test: true
             run-test-asan-address: false
             run-test-asan-undefined: false
@@ -38,8 +36,6 @@ jobs:
           - os: ubuntu-22.04-arm
             cc: gcc
             cxx: g++
-            # No clang-format for ARM.
-            run-lint: false
             run-test: true
             run-test-asan-address: false
             run-test-asan-undefined: false
@@ -48,7 +44,6 @@ jobs:
             cc: gcc
             cxx: g++
             pip-break-system-packages: true
-            run-lint: true
             run-test: true
             run-test-asan-address: false
             run-test-asan-undefined: false
@@ -59,7 +54,6 @@ jobs:
             cc: clang
             cxx: clang++
             pip-break-system-packages: true
-            run-lint: true
             run-test: true
             run-test-asan-address: true
             run-test-asan-undefined: true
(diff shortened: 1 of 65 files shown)


.github/workflows/mediasoup-worker.yaml
```diff
@@ -22,24 +22,20 @@ jobs:
           - os: ubuntu-22.04
             cc: gcc
             cxx: g++
-            run-lint: true
```
ibc, line 25: Can you instead write `run-lint: false` in those cases to be explicit, please?

.github/workflows/mediasoup-worker.yaml
```diff
@@ -118,8 +104,6 @@ jobs:
           - os: windows-2025
             cc: cl
             cxx: cl
-            # No clang-format for Windows.
-            run-lint: false
```
ibc, line 122: Same, please don't remove those comments, let's be explicit so we know while reading those lines.

.github/workflows/mediasoup-worker.yaml
```diff
@@ -171,14 +155,13 @@ jobs:
...
         run: pip3 install --break-system-packages invoke
 
-      # We need to install npm deps of worker/scripts/package.json.
-      - if: ${{ matrix.build.run-lint }}
-        name: npm ci --prefix worker/scripts
-        run: npm ci --prefix worker/scripts --foreground-scripts
-
-      - if: ${{ matrix.build.run-lint }}
+      # Run clang-format on macOS.
+      - if: ${{ startsWith(matrix.build.os, 'macos') }}
```
ibc: Don't do this please. Main purpose of all those `run-lint`, `run-test`, etc variables in each matrix host was to avoid direct OS based conditions here. Let's please use `run-lint` as it was. And ok, we can also check the OS to run `brew` command, but let's not remove the `run-lint` condition.

ibc: There could be an extra option like in CI host like "install-clang-format-with-brew: true" so we don't need to do OS conditions later in steps.

jmillan: but `brew` is only for `macos`, so a common variable  `install-clang-format-with-brew` makes no sense IMHO.

IMO Better would be to do:

```yaml
# Install clang-format for macos.
- if: matrix.build.run-lint && matrix.build.os && startsWith(matrix.build.os, 'macos')
- run: brew install clang-format-21

# Install clang-format for linux.
- if: matrix.build.run-lint && matrix.build.os && startsWith(matrix.build.os, 'ubuntu')
- run: apt-get install clang-format-21

# Run clang-format
- if: matrix.build.run-lint
- run: invoke -r worker lint
```

We could do it in a near future, but nowadays ubuntu-24 does not include clang-format-21. We could try to install it anyways. I can try this.

ibc: `brew` also exists for Linux. But ok, I like proposal above.

jmillan: The problem I see is that [clang-format](https://formulae.brew.sh/formula/clang-format) brew formula, which is actively maintained uses the latest clang-format stable version, which right now is 21 (it does not store previous versions). Ubuntu 24 comes with the version 20. IMHO, for CI we could just chose a OS which we use to run clang-format (we are not testing clang-format here), and run it once for the hole CI, as we do in this PR.

Otherwise we'll have to complicate this yaml file to install a newer clang-format on ubuntu. I don't think it's worth it. What do you think?

ibc: I was not suggesting that we should enable clang-format via brew for Linux, not even that we should enable clang-format in CI for Linux.

jmillan: My comment was for my own suggestion [here](https://github.com/versatica/mediasoup/pull/1643#discussion_r2495257254), where I indicated that we could install and run clang-format for macos and linux, but I see that we don't need to do it in both OS, and that installing the very same version (which we must) would complicate things with no benefit.

ibc: But I insist that I never meant that we should run clang-format in Linux in CI.

jmillan: .., and since `clang-format` brew formula has the most updated version I think it makes sense to use it, so we always use the last stable version.

jmillan: > But I insist that I never meant that we should run clang-format in Linux in CI.

I know, I know.

jmillan:
@ibc I've split the steps for `run-lint` so it's reusable if we ever want to run it on a different or multiple OS [here](https://github.com/versatica/mediasoup/pull/1643/commits/e28e1c815d32770e636265ce5439330b06314ca1).

Right now the action will fail if we set `run-lint` on a non macos OS, for explicitness.

ibc (approved):

PR #1645, opened 2025-11-06
Worker: clang-tidy pass

jmillan:
Getting closer to have clang-tidy on CI...


worker/include/RTC/SCTP/packet/errorCauses/NoUserDataErrorCause.hpp
```diff
@@ -68,11 +68,11 @@ namespace RTC
 			NoUserDataErrorCause(uint8_t* buffer, size_t bufferLength);
 
 		public:
-			virtual ~NoUserDataErrorCause() override;
+			~NoUserDataErrorCause() override;
```
jmillan, line 71: `virtual` and `override` are redundant.

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -85,7 +85,14 @@ TcpConnectionHandle::~TcpConnectionHandle()
 
 	if (!this->closed)
 	{
-		InternalClose();
+		try
+		{
+			InternalClose();
```
ibc: There are tons of calls to `InternalClose()`. Wouldn't it be better if we add try/catch into it instead?

worker/src/handles/UnixStreamSocketHandle.cpp
```diff
@@ -123,7 +123,14 @@ UnixStreamSocketHandle::~UnixStreamSocketHandle()
 
 	if (!this->closed)
 	{
-		Close();
+		try
+		{
+			Close();
```
ibc: Same, don't we prefer to add try/catch within `Close()` method body?

And BTW not sure why clang-tidy says this can throw. It cannot. If any, it aborts. But anyway...

worker/src/RTC/RtpStreamRecv.cpp
```diff
@@ -5,6 +5,7 @@
 #include "Logger.hpp"
 #include "Utils.hpp"
 #include "RTC/Codecs/Tools.hpp"
+#include <algorithm> // std::max, std::min
```
ibc: This is not needed. It's already included in `common.hpp`.

ibc:
I'm updating the branch. Also, this PR is not supposed to enabled clang-tidy in CI, right? It's only one step in that direction I assume.

jmillan:
> this PR is not supposed to enabled clang-tidy in CI, right? It's only one step in that direction I assume.

Correct.

worker/src/RTC/RtpStreamRecv.cpp
ibc: NOTE:

```
ack "include \<algorithm\>"
RTC/RtpStreamSend.cpp
12:#include <algorithm> // std::max, std::min

RTC/SimulcastConsumer.cpp
13:#include <algorithm> // std::max, std::min

RTC/SenderBandwidthEstimator.cpp
7:#include <algorithm> // std::max, std::min

RTC/ActiveSpeakerObserver.cpp
7:#include <algorithm> // std::max, std::min

RTC/Producer.cpp
17:#include <algorithm> // std::max, std::min

RTC/SvcConsumer.cpp
13:#include <algorithm> // std::max, std::min
```

Those can be removed because they all include `common.hpp`.

worker/src/handles/TcpConnectionHandle.cpp
jmillan: There are no more cases which can throw.

worker/src/handles/UnixStreamSocketHandle.cpp
jmillan: Ok, I can add try/catch inside `Close()`.

worker/src/handles/TcpConnectionHandle.cpp
jmillan: I'll do that.

ibc (approved):

worker/src/handles/UnixStreamSocketHandle.cpp
jmillan: > And BTW not sure why clang-tidy says this can throw. It cannot. If any, it aborts. But anyway...

Not correct, it will throw if #MS_EXECUTABLE is not defined https://github.com/versatica/mediasoup/blob/v3/worker/include/Logger.hpp#L485. This is why it complains.

jmillan: I'm correctly handling this in another PR.

PR #1646, opened 2025-11-07
Worker: clang-tidy pass

jmillan:
Another clang-tidy pass.

### DIFF
--- a/worker/src/RTC/RtpStreamRecv.cpp
+++ b/worker/src/RTC/RtpStreamRecv.cpp
@@ -36,14 +36,15 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		size_t spatialLayer  = packet->GetSpatialLayer();
-		size_t temporalLayer = packet->GetTemporalLayer();
+		auto spatialLayer  = packet->GetSpatialLayer();
+		auto temporalLayer = packet->GetTemporalLayer();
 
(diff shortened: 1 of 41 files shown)


ibc:
Upps: https://github.com/versatica/mediasoup/actions/runs/19167202659/job/54790605337?pr=1646

worker/src/RTC/RtpStreamRecv.cpp
```diff
@@ -36,20 +36,14 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		auto spatialLayer  = packet->GetSpatialLayer();
-		auto temporalLayer = packet->GetTemporalLayer();
+		size_t spatialLayer  = packet->GetSpatialLayer();
+		size_t temporalLayer = packet->GetTemporalLayer();
```
ibc: I'm totally confused. Why this change?

ibc (approved):

worker/src/RTC/RtpStreamRecv.cpp
jmillan: in order to use `std::min()`, since  `this->spatialLayerCounters.size()` is `size_t`. I may also cast the value within the call to `std::min()`.

jmillan: Done, cast is done now when needed.

PR #1648, opened 2025-11-07
Clang tidy pass

jmillan:


ibc (approved):
My conclusion is that now I don't know anything about C++

worker/include/RTC/PipeConsumer.hpp
```diff
@@ -10,6 +10,13 @@ namespace RTC
 {
 	class PipeConsumer : public RTC::Consumer, public RTC::RtpStreamSend::Listener
 	{
+	private:
+		static void StorePacketInTargetLayerRetransmissionBuffer(
```
ibc, line 14: I don't like that we are moving method to static methods just because clang-tidy complains.

jmillan: It's fine. We can move that back. There are tens of clang-tidy rules that we are already not applying.

PR #1649, opened 2025-11-10
Worker: clang tidy pass

jmillan:


worker/include/RTC/RTCP/ReceiverReport.hpp
```diff
@@ -86,10 +86,18 @@ namespace RTC
...
+				// Get the limit value for possitive totalLost.
+				if (totalLost >= 0)
+				{
+					clamp = (totalLost > 0x07FFFFF) ? 0x07FFFFF : totalLost;
+				}
+				// Get the limit value for negative totalLost.
+				else
+				{
+					clamp = (-totalLost > 0x0800000) ? 0x0800000 : -totalLost;
+				}
```
ibc: This must go away once PR https://github.com/versatica/mediasoup/pull/1650 is merged.

worker/include/Utils.hpp
```diff
@@ -85,6 +85,9 @@ namespace Utils
 				case AF_INET6:
 					std::memcpy(std::addressof(copiedAddr), addr, sizeof(struct sockaddr_in6));
 					break;
+
+				default:
+					break;
```
ibc: Let's not add `break` in `default`. We can do `default:;` as in many others.

worker/src/RTC/RTCP/Feedback.cpp
```diff
@@ -251,8 +251,6 @@ namespace RTC
 					break;
 
 				case FeedbackRtp::MessageType::PS:
-					break;
```
ibc, line 254: Why is this `break` removed? It shouldn't

jmillan: Here clang-tidy annotates: "Switch has 2 consecutive identical branches (clang-tidy bugprone-branch-clone)".  Which interprets both branches are having the same behaviour and could better just have one execution block rather than two. Having two identical blocks may be error prone as we could be copy&pasting them.

In this case the execution block is nothing. So we could:
* leave it as it's here.
* add the `break` back and add a // NOLINT tag

ibc: There are many linters in many languages than complain if a `case` has no `break` and here we are doing the opposite. I know it's correct, but I also understand the default settings of those linters than complain about it.

ibc: I mean, I would remove that rule.

jmillan: Done.

worker/include/RTC/RTCP/Sdes.hpp
```diff
@@ -214,7 +214,7 @@ namespace RTC
 				// A serialized packet can contain a maximum of 31 chunks.
 				// If number of chunks exceeds 31 then the required number of packets
 				// will be serialized which will take the size calculated below.
-				size_t size = Packet::CommonHeaderSize * ((this->GetCount() / (MaxChunksPerPacket + 1)) + 1);
+				size_t size = Packet::CommonHeaderSize * ((this->GetCount() / (maxChunksPerPacket + 1)) + 1);
```
ibc: NOTE: Just a matter of taste but when I use a class static method or variable, I like to always add the `ClassName::` prefix.

worker/include/RTC/RTCP/XrDelaySinceLastRr.hpp
```diff
@@ -60,6 +60,7 @@ namespace RTC
 
 				void Dump(int indentation = 0) const;
 				size_t Serialize(uint8_t* buffer);
+				// NOLINTNEXTLINE (readability-convert-member-functions-to-static)
```
ibc: Can't we just remove/disable that rule from clang-tidy settings? Honestly we don't have any general problem of "class instance methods that could be static methods". The ones we have is because we want them to be like that and I wouldn't like if I have to add `// NOLINTNEXTLINE (readability-convert-member-functions-to-static)` in tons of places.

jmillan: Yes, we can filter that rule. I'll do it.

ibc: And remove all those NOLINT lines plz?

jmillan: OFC

worker/include/RTC/RtpDictionaries.hpp
```diff
@@ -20,6 +20,7 @@ namespace RTC
 		};
 	};
 
+	// NOLINTNEXTLINE (cppcoreguidelines-pro-type-member-init)
```
ibc, line 23: What is clang-tidy complaining about here about?

worker/src/RTC/RateCalculator.cpp
```diff
@@ -2,6 +2,7 @@
 // #define MS_LOG_DEV_LEVEL 3
 
 #include "RTC/RateCalculator.hpp"
+#include "DepLibUV.hpp"
```
ibc, line 5: Why this? `DepLibUV.hpp` is already included by `RTC/RateCalculator.hpp`.

worker/include/RTC/RtpDictionaries.hpp
jmillan: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html

The check flags user-provided constructor definitions that do not initialize all fields that would be left in an undefined state by default construction.

In this case `type` and `subType` are unset on construction. We later on call `SetMimeType()` which gives value to those.

worker/src/RTC/RateCalculator.cpp
jmillan: I removed it in this PR since it's not used in `RTC/RateCalculator.hpp`, but here it is.

ibc: Ah ok ok

ibc (approved):

PR #1653, opened 2025-11-12
Fix RTX packets containing non yet seen RTP packets being discarded

### LINKED ISSUE #906: consumer NACK with the specified sequence number packet but the retransmission fails

ibc:
# Details

- Fixes #906.
- Credits to @penguinol and @mengbieting.
- Allow RTX packets containing non yet seen RTP seq numbers pass though.
- Accept RTX packets containing seq numbers in our NACK list despite the NACK was not sent yet.

## More info

The STR this PR fixes is the following:

1. Client sends RTP seq 1, mediasoup receives it.
2. Client sends RTP seq 2, mediasoup does NOT receive it.
3. Client sends a probation RTX packet that contain the original packet seq 2, mediasoup receives it.

In this case we want to accept that RTP packet with seq 2 received via RTX in bullet 2 rather than discarding it as we do in v3 branch before this PR.

## Notes

- I'm marking PR as draft until we finish the ongoing discussion in the issue #906.
- The code diff between what @penguinol uses and current mediasoup v3 is here: https://www.diffchecker.com/T6y4Osuz/


worker/src/RTC/NackGenerator.cpp
```diff
@@ -68,22 +68,20 @@ namespace RTC
...
 				  packet->GetSsrc(),
 				  packet->GetSequenceNumber(),
 				  isRecovered ? "true" : "false");
 
-				auto retries = it->second.retries;
-
-				this->nackList.erase(it);
+				// NOTE: Even if we didn't send yet the NACK requesting this packet (this
+				// is, if nackInfo.retries == 0), we accept this packet and let it go
+				// through because we would request it later anyway.
```
jmillan: Little rewording which makes it more clear IMO.

```suggestion
				// NOTE: Accept the packet since it was in the `nackList`, regardless
				// the NACK requesting this packet was not sent yet (this is, if nackInfo.retries == 0)
				// because we would request it later anyway.
```

jmillan:
I see it OK so far.

ibc:
> I see it OK so far.

No no, see last comment in the related issue.

PR #1655, opened 2025-11-13
Clang tidy pass

jmillan:


worker/include/RTC/RTCP/Feedback.hpp
```diff
@@ -36,19 +36,19 @@ namespace RTC
 			}
 			uint32_t GetSenderSsrc() const
 			{
-				return uint32_t{ ntohl(this->header->senderSsrc) };
+				return ntohl(this->header->senderSsrc);
```
jmillan, line 39: This is a redundant cast. `ntohl` already returns `uint32_t`.

worker/src/RTC/Codecs/VP8.cpp
```diff
@@ -318,22 +319,14 @@ namespace RTC
...
+				// - Temporal layer is higher than target.
+				// - Temporal layer is higher than current and sync flag is not set.
 				// clang-format off
-				else if (
-					this->payloadDescriptor->tlIndex > context->GetCurrentTemporalLayer() &&
-					!this->payloadDescriptor->y
+				if (
+				  this->payloadDescriptor->tlIndex > context->GetTargetTemporalLayer() ||
+				  (this->payloadDescriptor->tlIndex > context->GetCurrentTemporalLayer() &&
+				   !this->payloadDescriptor->y)
```
jmillan, line 329: We were basically having two duplicate if/else bodies.

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1241,6 +1241,7 @@ namespace RTC
 		}
 	}
 
+	// NOLINTNEXTLINE (misc-no-recursion)
```
jmillan, line 1244: `SetTimeout` may call `OnTimer` which may call `SetTimeout`. I think we had this super tested and reviewed. Do you remember it @ibc?

worker/src/RTC/DtlsTransport.cpp
```diff
@@ -1695,6 +1696,7 @@ namespace RTC
 		// callback).
 	}
 
+	// NOLINTNEXTLINE (misc-no-recursion)
```
jmillan, line 1699: Same here, `OnTimer` may call `SetTimeout` which may call `OnTimer`..

worker/src/handles/TcpConnectionHandle.cpp
```diff
@@ -384,7 +381,14 @@ void TcpConnectionHandle::InternalClose()
 
 		if (err != 0)
 		{
-			MS_ABORT("uv_shutdown() failed: %s", uv_strerror(err));
+			try
```
jmillan, line 384: `~TcpConnectionHandle` kept warning because we were missing this try/catch here.

worker/src/handles/UnixStreamSocketHandle.cpp
```diff
@@ -179,7 +176,14 @@ void UnixStreamSocketHandle::Close()
 
 		if (err != 0)
 		{
-			MS_ABORT("uv_shutdown() failed: %s", uv_strerror(err));
+			try
```
jmillan, line 179: Same as in `~TcpConnectionHandle`

worker/src/RTC/DtlsTransport.cpp
ibc: Technically `SetTimeout()` can immediately call `onTimer()` and vice-versa. It's never been a problem obviously but indeed there is a potential bug here.

jmillan: I'll add a note inline so we remove this tag once re-checked.

jmillan: I'll open an issue.

jmillan: Here https://github.com/versatica/mediasoup/issues/1656

ibc (approved):

PR #1658, opened 2025-11-14
Clang tidy

jmillan:
This is probably the last clang-tidy related commit in the history of mediasoup.

EDIT: Since we'll add it to CI and currently there are 0 warnings in local env (At leas in MacOS)

### DIFF
--- a/worker/include/RTC/Codecs/DependencyDescriptor.hpp
+++ b/worker/include/RTC/Codecs/DependencyDescriptor.hpp
@@ -31,9 +31,9 @@ namespace RTC
 			static std::unordered_map<DecodeTargetIndication, std::string> dtiToString;
 
 		private:
-			struct FameDependencyTemplate
+			struct FrameDependencyTemplate
 			{
-				FameDependencyTemplate(uint32_t spatialLayer, uint32_t temporalLayer)
+				FrameDependencyTemplate(uint32_t spatialLayer, uint32_t temporalLayer)
 				  : spatialLayer(spatialLayer), temporalLayer(temporalLayer)
 				{
 				}
@@ -52,7 +52,7 @@ namespace RTC
 				uint32_t temporalLayers{ 0 };
 				uint8_t templateIdOffset{ 0 };
 				uint8_t decodeTargetCount{ 0 };
-				std::vector<FameDependencyTemplate> templateLayers;
+				std::vector<FrameDependencyTemplate> templateLayers;
 			};
 
 		public:
--- a/worker/include/RTC/Codecs/DependencyDescriptor.hpp
+++ b/worker/include/RTC/Codecs/DependencyDescriptor.hpp
@@ -33,6 +33,11 @@ namespace RTC
 		private:
 			struct FameDependencyTemplate
 			{
+				FameDependencyTemplate(uint32_t spatialLayer, uint32_t temporalLayer)
+				  : spatialLayer(spatialLayer), temporalLayer(temporalLayer)
+				{
+				}
+
 				uint32_t spatialLayer;
 				uint32_t temporalLayer;
 				std::vector<DecodeTargetIndication> decodeTargetIndications;
--- a/worker/src/RTC/Codecs/VP8.cpp
+++ b/worker/src/RTC/Codecs/VP8.cpp
@@ -200,6 +200,12 @@ namespace RTC
 		{
 			MS_TRACE();
 
+			if (data == nullptr)
+			{
+				MS_ERROR("data is nullptr");
+				return;
+			}
+
 			// Nothing to do.
 			if (!this->extended)
 			{
--- a/worker/src/RTC/Codecs/VP8.cpp
+++ b/worker/src/RTC/Codecs/VP8.cpp
@@ -202,8 +202,7 @@ namespace RTC
 
 			if (data == nullptr)
 			{
-				MS_ERROR("data is nullptr");
-				return;
+				MS_ABORT("data is nullptr");
 			}
 
 			// Nothing to do.
--- a/worker/src/RTC/PipeTransport.cpp
+++ b/worker/src/RTC/PipeTransport.cpp
@@ -388,6 +388,7 @@ namespace RTC
 						MS_THROW_TYPE_ERROR("missing port");
 					}
 
+					// NOLINTNEXTLINE (bugprone-unchecked-optional-access)
 					port = body->port().value();
 
 					int err;
--- a/worker/src/RTC/PlainTransport.cpp
+++ b/worker/src/RTC/PlainTransport.cpp
@@ -105,6 +105,7 @@ namespace RTC
 		if (options->enableSrtp())
 		{
 			auto srtpCryptoSuite = options->srtpCryptoSuite();
+
 			if (!srtpCryptoSuite.has_value())
 			{
 				MS_THROW_TYPE_ERROR("missing srtpCryptoSuite");
--- a/worker/src/RTC/PlainTransport.cpp
+++ b/worker/src/RTC/PlainTransport.cpp
@@ -104,13 +104,14 @@ namespace RTC
 
 		if (options->enableSrtp())
 		{
-			if (!options->srtpCryptoSuite().has_value())
+			auto srtpCryptoSuite = options->srtpCryptoSuite();
+			if (!srtpCryptoSuite.has_value())
 			{
 				MS_THROW_TYPE_ERROR("missing srtpCryptoSuite");
 			}
 
 			// NOTE: The SRTP crypto suite may change later on connect().
-			this->srtpCryptoSuite = SrtpSession::CryptoSuiteFromFbs(options->srtpCryptoSuite().value());
+			this->srtpCryptoSuite = SrtpSession::CryptoSuiteFromFbs(srtpCryptoSuite.value());
 
 			switch (this->srtpCryptoSuite)
 			{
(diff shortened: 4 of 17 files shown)


worker/src/RTC/Codecs/VP8.cpp
```diff
@@ -200,6 +200,12 @@ namespace RTC
 		{
 			MS_TRACE();
 
+			if (data == nullptr)
```
jmillan: If `data` was `nullptr`, which is extremely rare, near to impossible, then we would dereference a null pointer leading to undefined behaviour.

worker/src/RTC/PlainTransport.cpp
```diff
@@ -104,13 +104,14 @@ namespace RTC
 
 		if (options->enableSrtp())
 		{
-			if (!options->srtpCryptoSuite().has_value())
+			auto srtpCryptoSuite = options->srtpCryptoSuite();
+			if (!srtpCryptoSuite.has_value())
```
jmillan: `options->srtpCryptoSuite()` returns an optional, meaning previous code:

```cpp
if (!options->srtpCryptoSuite().has_value())
{ 
  this->srtpCryptoSuite = SrtpSession::CryptoSuiteFromFbs(options->srtpCryptoSuite().value());
```

Here the call to `options->srtpCryptoSuite()` in the conditional statement and the call to it in the conditional body return **different** objects, hence there is no way to static analize the second as being previously checked before accessing.

The solution is to store the result of `options->srtpCryptoSuite()` into a variable and then the condition can be statically analysed and assure that we are accessing a checked option value, since the object in the condition statement and body are the same.

worker/src/RTC/PipeTransport.cpp
```diff
@@ -388,6 +388,7 @@ namespace RTC
 						MS_THROW_TYPE_ERROR("missing port");
 					}
 
+					// NOLINTNEXTLINE (bugprone-unchecked-optional-access)
```
jmillan: The reason why I add this label here rather than doing as in the next cases is because the variable `port` is already taken and cannot be reused (different types). And it feels a bit dirty naming it differently. In this case the label means: "we know what we are doing (TM)"

worker/src/Worker.cpp
```diff
@@ -176,13 +176,13 @@ flatbuffers::Offset<FBS::Worker::ResourceUsageResponse> Worker::FillBufferResour
...
 	int err;
-	uv_rusage_t uvRusage; // NOLINT(cppcoreguidelines-pro-type-member-init)
+	uv_rusage_t uvRusage{}; // NOLINT(cppcoreguidelines-pro-type-member-init)
 
 	err = uv_getrusage(std::addressof(uvRusage));
 
 	if (err != 0)
 	{
-		MS_THROW_ERROR("uv_getrusagerequest() failed: %s", uv_strerror(err));
+		MS_THROW_ERROR("uv_getrusage() failed: %s", uv_strerror(err));
```
jmillan: old time typo.

worker/include/RTC/Codecs/DependencyDescriptor.hpp
```diff
@@ -33,6 +33,11 @@ namespace RTC
 		private:
 			struct FameDependencyTemplate
 			{
+				FameDependencyTemplate(uint32_t spatialLayer, uint32_t temporalLayer)
```
ibc: "Fame"? You mean "Fake"?

worker/src/RTC/Codecs/VP8.cpp
```diff
@@ -200,6 +200,12 @@ namespace RTC
...
+			{
+				MS_ERROR("data is nullptr");
+				return;
+			}
```
ibc, line 206:
```suggestion
			if (data == nullptr)
			{
				MS_ERROR("data is nullptr");

				return;
			}
```

But honestly we never use `MS_ERROR` for these things. This should be `MS_ABORT` instead.

worker/src/RTC/RTCP/CompoundPacket.cpp
```diff
@@ -68,6 +68,7 @@ namespace RTC
 
 			if (this->xrPacket.Begin() != this->xrPacket.End())
 			{
+				// NOLINTNEXTLINE [clang-analyzer-deadcode.DeadStores]
```
ibc: What? Am I supposed to write that NOLINTNEXTLINE every time I write I don't know exactly what?

worker/src/RTC/RtpDictionaries/RtpCodecParameters.cpp
```diff
@@ -24,9 +24,10 @@ namespace RTC
 		this->clockRate = data->clockRate();
 
 		// channels is optional.
-		if (data->channels().has_value())
+		auto channels = data->channels();
+		if (channels.has_value())
```
ibc: What is going on in this PR? We always add blank lines after variable declaration and before `if` conditions and before `return`.


```suggestion
		auto channels = data->channels();

		if (channels.has_value())
```

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
```diff
@@ -16,9 +16,10 @@ namespace RTC
 		MS_TRACE();
 
 		// ssrc is optional.
-		if (data->ssrc().has_value())
+		auto ssrc = data->ssrc();
+		if (ssrc.has_value())
```
ibc:
```suggestion
		auto ssrc = data->ssrc();

		if (ssrc.has_value())
```

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
```diff
@@ -28,9 +29,10 @@ namespace RTC
 		}
 
 		// codecPayloadType is optional.
-		if (data->codecPayloadType().has_value())
+		auto codecPayloadType = data->codecPayloadType();
+		if (codecPayloadType.has_value())
```
ibc: Buff there are too many like this...

ibc (changes requested):

worker/include/RTC/Codecs/DependencyDescriptor.hpp
jmillan: No, "Frame". I'll handle all concerns on Monday.

jmillan: No, it's 'Frame'.

worker/src/RTC/RTCP/CompoundPacket.cpp
jmillan: No, in this case the static analysis flags that `offset` is being modified and it's value is not used. I've removed the // NOLINTNEXTLINE comment and fixed the problem.

worker/src/RTC/RtpDictionaries/RtpCodecParameters.cpp
jmillan: You're right, I was unconsciously doing that within this PR because such variable `channels` only concerns the conditional block and tightening it to the condition statement made sense to me.

For such cases like this one C++17 exposes the [if statements](https://en.cppreference.com/w/cpp/language/if.html) with initializers which would lead to this outcome:

```c++
// `channels` variable scope is limited to the condition block.
if (auto channels = data->channels(); channels.has_value())
{
  this->channels = channels.value();
}
```

This also allow us to reuse the `channels` name for variables outside the condition.

I'm fine with this for this cases. WDYT @ibc?

worker/src/RTC/RtpDictionaries/RtpEncodingParameters.cpp
jmillan: Yes, yes.., I'll handle that all together.

worker/src/RTC/RtpDictionaries/RtpCodecParameters.cpp
ibc: Fine

ibc:
Please the blank line before return; and before if conditions everywhere

worker/src/RTC/PlainTransport.cpp
```diff
@@ -592,6 +594,7 @@ namespace RTC
 							MS_THROW_TYPE_ERROR("missing port");
 						}
 
+						// NOLINTNEXTLINE (bugprone-unchecked-optional-access)
```
jmillan, line 597: Here I'm ignoring (bugprone-unchecked-optional-access) rule because  `port` is already defined as `uint16_t` and we cannot override it with an option value. Renaming it to another name does not feel good either. So here we're basically telling the static analyser that we know what we're doing.

ibc (approved):

ibc:
I'm seeing some "const auto" changed to "auto". Which is the policy here? Should we use "const string/uint32_t/etc" (when possible) and never "const auto" or what? Or should we use only "auto" if the right side of the assignment returns a "const xxxx"?

jmillan:
> I'm seeing some "const auto" changed to "auto". Which is the policy here? Should we use "const string/uint32_t/etc" (when possible) and never "const auto" or what? Or should we use only "auto" if the right side of the assignment returns a "const xxxx"?

In `SimulcastConsumer` and `SvcConsumer` we have this change:

```diff
- const flatbuffers::Optional<int16_t> preferredTemporalLayer{ this->preferredLayers.temporal };
+ preferredTemporalLayer     = this->preferredLayers.temporal;
```

The reason is that such `preferredTemporalLayer` is now created few lines before, so we are simply reusing it.

We should always use `const` when the value does not change, otherwise the static analyser will warn.

PR #1660, opened 2025-11-17
Remove build-type from matrix

jmillan:
Do not compile every single matrix build element in Release and Debug.

Just target one element to compile in Debug, and compile in Release mode every other.

NOTE: This PR removes 10 mediasoup-worker runs, which should significantly reduce the overal CI run time.

### DIFF
--- a/.github/workflows/mediasoup-worker.yaml
+++ b/.github/workflows/mediasoup-worker.yaml
@@ -22,6 +22,7 @@ jobs:
           - os: ubuntu-22.04
             cc: gcc
             cxx: g++
+            build-type: Release
             run-lint: false
             run-test: true
             run-test-asan-address: false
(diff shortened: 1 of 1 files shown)


.github/workflows/mediasoup-worker.yaml
```diff
@@ -55,9 +59,20 @@ jobs:
             run-test-asan-thread: false
             # Let's just compile with all Meson option flags enabled once with gcc.
             meson_args: '-Dms_log_trace=true -Dms_log_file_line=true -Dms_rtc_logger_rtp=true -Dms_disable_liburing=true -Dms_sctp_stack=true'
```
ibc: Don't we want to also replicate this in Debug mode?

ibc (approved):

.github/workflows/mediasoup-worker.yaml
jmillan: We can, yes. I'll add them.

PR #1666, opened 2025-11-20
Only look up the RTP packet’s RID extension if the packet doesn’t have MID extension

ibc:
# Details

- When a RTP packet is received by `RtpListener.getProducer(packet)`, and assuming that the packet's SSRC is not yet in the `ssrcTable`, only read the RID extension of the packet if it doesn't have MID extension.
- It fixes a bug exposed in the forum here: https://mediasoup.discourse.group/t/producer-simulcast-ssrc-mixed-up-race-condition/6917/16
- Thanks [Hugo Mallet](https://mediasoup.discourse.group/u/hugo) for reporting the issue.

## The problem

- Imagine client sends simulcast in 2 producers A and B.
- Producer A has MID "A" and 2 RIDs "r0" and "r1".
- Producer B has MID "B" and 2 RIDs "r0" and "r1".
- After a while, Producer A is closed in mediasoup server but not in the client which keeps sending packets from its local Producer A.
- When a packet from Producer A arrives, `RtpListener` won't find it in its `ssrcTable`, neither in its `midTable`. However it will find it in the `ridTable` because both Producers have same RID values, so the `ridTable` contains entries in the `ridTable` with keys "r0" and "r1" and value pointing to Producer 2 (because Producer 2 was created after Producer 1 so its `rid` values override the previous ones of Producer 1, this is ok and expected).
- And hence `RtpListener` will do this: `this->ssrcTable[packet->GetSsrc()] = producer`, and from now on all packets from Producer 1 will be identified as packets of Producer 2 in `RtpListener` and hence a new `RtpStreamRecv` will be created in Producer 2 and it will forward those wrong packets to its Consumers, producing frozen video and different and unexpected issues.

## The solution

- Once Producer 1 has been closed in mediasoup server, there are no entries in the tables of `RtpListener` pointing to it.
- Then a late packet of Producer 1 arrives to mediasoup (or maybe client didn't really close its Producer 1).
- `RtpListener` cannot find an entry in `ssrcTable` neither in `midTable`.
- When it comes to look for an entry in `ridTable`, now we won't do it because the packet also contains MID extension so it should have been identified in the block before when looking in the `midTable`. But it's not in the `midTable` because Producer 1 was closed in mediasoup server.

## Rationale

- All RTP packets that include RID extension also include MID extension.
- Of course, client needs to negotiate MID and RID.
- If a client sends RID without MID and creates different Producers in the same Transport with same RID values, then it has a problem, but it's not a realistic scenario at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Only look up the RTP packet’s RID extension if the packet doesn’t have MID extension ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
+
 ### 3.19.11
 
 - Node: Add `workerBin` optional field in `createWorker()` ([PR #1660](https://github.com/versatica/mediasoup/pull/1660)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Only look up the RTP packet’s RID extension if the packet doesn’t have MID extension ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
+- Only look up the RTP packet’s RID extension if the packet doesn’t have MID extension ([PR #1666](https://github.com/versatica/mediasoup/pull/1666)).
 
 ### 3.19.11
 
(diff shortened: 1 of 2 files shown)


jmillan (approved):
:+1:

PR #1669, opened 2025-11-26
Worker: Perf, avoid dynamic memory allocation in TransportTuple hash

jmillan:
For every packet arriving from network we are generating a tuple, and upon creation we create its hash.

Before this PR we were using a `std::vector` which we later fill it with the tuple info. This generates dynamic memory allocation which must later be released.

This PR makes use of `std::array` with a fixed length (we know it in advance), so we store the info only in the stack.

This only change improves CPU performance in a little more than 1%.

Before `TransportTuple::GenerateHash()` took around 1.4%:
Screenshot 2025-11-26 at 15 28 25

After  `TransportTuple::GenerateHash()` takes around 0.2%:
Screenshot 2025-11-26 at 15 17 35

### DIFF
--- a/worker/src/RTC/TransportTuple.cpp
+++ b/worker/src/RTC/TransportTuple.cpp
@@ -3,6 +3,7 @@
 
 #include "RTC/TransportTuple.hpp"
 #include "Logger.hpp"
+#include <cstring> // std::memcpy()
 
 namespace RTC
 {
--- a/worker/src/RTC/TransportTuple.cpp
+++ b/worker/src/RTC/TransportTuple.cpp
@@ -3,7 +3,6 @@
 
 #include "RTC/TransportTuple.hpp"
 #include "Logger.hpp"
-#include <array>
 
 namespace RTC
 {
@@ -155,10 +154,8 @@ namespace RTC
 
 		// Maximum buffer length for two IPv6 addresses and ports plus protocol.
 		static constexpr size_t BufferSize = ((16 + 2) * 2) + 1;
-		std::array<uint8_t, BufferSize> buffer{};
-		// NOTE: Windows cannot deduce the type if using 'auto'.
-		// NOLINTNEXTLINE (hicpp-use-auto)
-		std::array<uint8_t, BufferSize>::iterator it = buffer.begin();
+		uint8_t buffer[BufferSize]         = {};
+		size_t idx                         = 0;
 
 		auto appendSockAddr = [&](const sockaddr* addr)
 		{
@@ -168,11 +165,10 @@ namespace RTC
 				const auto* ip      = reinterpret_cast<const uint8_t*>(&in->sin_addr.s_addr);
 				const uint16_t port = ntohs(in->sin_port);
 
-				it  = std::copy(ip, ip + 4, it);
-				*it = (port >> 8) & 0xFF;
-				it++;
-				*it = port & 0xFF;
-				it++;
+				std::memcpy(buffer + idx, ip, 4);
+				idx += 4;
+				buffer[idx++] = (port >> 8) & 0xFF;
+				buffer[idx++] = port & 0xFF;
 			}
 			else if (addr->sa_family == AF_INET6)
 			{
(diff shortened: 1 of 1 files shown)


worker/src/RTC/TransportTuple.cpp
```diff
@@ -153,7 +153,10 @@ namespace RTC
 		const auto* localSockAddr  = GetLocalAddress();
 		const auto* remoteSockAddr = GetRemoteAddress();
 
-		std::vector<uint8_t> buffer;
+		// Maximum buffer length for two IPv6 addresses and ports plus protocol.
+		static constexpr size_t BufferSize = ((16 + 2) * 2) + 1;
+		std::array<uint8_t, BufferSize> buffer{};
```
ibc: We should include std::array at the top of this file

worker/src/RTC/TransportTuple.cpp
```diff
@@ -163,26 +166,26 @@ namespace RTC
 				const auto* ip      = reinterpret_cast<const uint8_t*>(&in->sin_addr.s_addr);
 				const uint16_t port = ntohs(in->sin_port);
 
-				buffer.insert(buffer.end(), ip, ip + 4);
-				buffer.push_back((port >> 8) & 0xFF);
-				buffer.push_back(port & 0xFF);
+				it    = std::copy(ip, ip + 4, it);
+				*it++ = (port >> 8) & 0xFF;
```
ibc: Wow I don't really know what this `*it++ =` does

jmillan: Assign the value of the right operand to `it` and increment it.

worker/src/RTC/TransportTuple.cpp
ibc: 👆🏻👆🏻👆🏻

ibc (approved):

worker/src/RTC/TransportTuple.cpp
nazar-pc: Why not write separate operations on separate lines, so it is easier to read?

jmillan: If `it` if for example `buffer.begin()`, this is index 0, then it will assign `(port >> 8) & 0xFF` to the first position of the array and `it` will point now to the position immediately after the written data (hence the ++), this is, index 1.

nazar-pc: The fact that two people ask what this code does is a pretty strong indication that less fancy code that takes 3 lines instead of one would be much more maintainable.

jmillan: Yes, it definitely makes sense to make it a bit more verbose.

jmillan:
Windows failing.

worker/src/RTC/TransportTuple.cpp
jmillan: > The fact that two people ask what this code does is a pretty strong indication that less fancy code that takes 3 lines instead of one would be much more maintainable.

BTW I responded to Iñaki before reading your comment.

ibc:
> * mediasoup-worker / ci (windows-2025, cl, cl, Release, false, false, false, false, false, 24) (pull_request)

Here:

```
../../../src/RTC/TransportTuple.cpp(159): error C3535: cannot deduce type for 'auto *' from 'std::_Array_iterator<_Ty,37>'
        with
        [
            _Ty=uint8_t
        ]
../../../src/RTC/TransportTuple.cpp(159): error C2440: 'initializing': cannot convert from 'std::_Array_iterator<_Ty,37>' to '<error type> *'
        with
        [
            _Ty=uint8_t
        ]
../../../src/RTC/TransportTuple.cpp(159): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
[1296/1386] Compiling C++ object mediasoup-worker.exe.p/src_RTC_SimulcastConsumer.cpp.obj
```

jmillan:
@ibc, @nazar-pc I split the write and increment of the iterator in two lines. I hope this makes it more readable.

ibc (approved):

ibc:
Please add the #include corresponding to std::memcpy() in the file

PR #1670, opened 2025-11-28
Worker: Simplify RtpPacket::GetNextMediasoupPacketId

jmillan:
Just that. I was curious how it could take 0.1% of CPU. It now takes 0% and most importantly it's cleaner.

### DIFF
--- a/worker/src/RTC/RtpPacket.cpp
+++ b/worker/src/RTC/RtpPacket.cpp
@@ -16,7 +16,8 @@ namespace RTC
 {
 	/* Class variables. */
 
-	thread_local uint32_t RtpPacket::nextMediasoupPacketId{ 0u };
+	thread_local uint32_t RtpPacket::nextMediasoupPacketId{ Utils::Crypto::GetRandomUInt(
+		1u, std::numeric_limits<uint32_t>::max() / 2) };
 
 	/* Class methods. */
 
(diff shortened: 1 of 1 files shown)


worker/src/RTC/RtpPacket.cpp
```diff
@@ -16,32 +16,16 @@ namespace RTC
...
-			RtpPacket::nextMediasoupPacketId++;
-
-			if (RtpPacket::nextMediasoupPacketId == 0)
-			{
-				RtpPacket::nextMediasoupPacketId = 1;
-			}
-		}
-
-		return RtpPacket::nextMediasoupPacketId;
+		return RtpPacket::nextMediasoupPacketId++;
```
ibc, line 28: But now it can be 0 which was not possible before. And if zero is a valid value, then why do we start with a random value between 1 and N?

nazar-pc: I kind of get it now why Rust doesn't have `++x` and `x++` operators. Combining them with other operations is confusing. I'd write it on separate lines for clarity.

ibc: Please let's not limit C++ benefits and optimisations with the excuse of the limitations in Rust. We want complex C++ code to prove how smart we are.

BTW ++ operator exists in every language.

jmillan: > But now it can be 0 which was not possible before

What's the problem with being 0 upon wrap up? Should we avoid 0 and not 1 or 2 for any rason?

ibc: I didn't say that 0 is a problem. But you are literally creating an initial value between 1 and N instead of between 0 and N, which doesn't make sense given that it will become 0 eventually.

jmillan: > I kind of get it now why Rust doesn't have ++x and x++ operators. Combining them with other operations is confusing. I'd write it on separate lines for clarity.

This is just returning the value and incrementing itself by one. I could argue that `*it++` may seem confusing, but `foo++` is just post incremental. It's C++ basics. I honestly don't see a reason to split it in this case.

But if you guys prefer it I'll do it.

ibc: ++ operator is clear and obvious in this case.

jmillan: > But you are literally creating an initial value between 1 and N instead of between 0 and N, which doesn't make sense given that it will become 0 eventually.

It makes sense consistency wise. Shall we create an initial value between 0 and N then?

ibc: After 6 comments suggesting exactly that, let me think... yes?

ibc (approved):

PR #1671, opened 2025-11-28
Utils::Number: remove duplicated static member Mask

jmillan:
`MaxValue` and `Mask` are exactly the same.

Remove Mask for being duplicated, and use MaxValue in its place which is also more expressive.

### DIFF
--- a/worker/include/Utils.hpp
+++ b/worker/include/Utils.hpp
@@ -296,8 +296,6 @@ namespace Utils
 	{
 	private:
 		static constexpr T MaxValue = (N == 0) ? std::numeric_limits<T>::max() : ((1 << N) - 1);
-		static constexpr T Mask =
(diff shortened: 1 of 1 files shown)


worker/include/Utils.hpp
```diff
@@ -307,8 +305,8 @@ namespace Utils
 			    std::is_same_v<T, uint64_t>,
 			  "T must be uint8_t, uint16_t, uint32_t or uint64_t");
 
-			lhs &= Mask;
-			rhs &= Mask;
+			lhs &= MaxValue;
+			rhs &= MaxValue;
```
jmillan, line 309: We are indeed applying a maximum value mask to avoid the number begin higher than it should.

ibc (approved):
Ok ok

PR #1673, opened 2025-12-01
Test receive rtx before corresponding rtp

jmillan:
This test which is defined in https://github.com/versatica/mediasoup/issues/906 description demonstrates that https://github.com/versatica/mediasoup/pull/1653 fixes it. The test fails in v3 and succeeds in that branch.

NOTE: This PR targets https://github.com/versatica/mediasoup/pull/1653. It will be part of that branch once merged.

### DIFF
--- a/worker/test/src/RTC/TestRtpStreamRecv.cpp
+++ b/worker/test/src/RTC/TestRtpStreamRecv.cpp
@@ -174,7 +174,7 @@ SCENARIO("receive RTP packets and trigger NACK", "[rtp][rtpstream]")
 		REQUIRE(listener.nackedSeqNumbers.empty());
 	}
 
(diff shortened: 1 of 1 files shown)


worker/test/src/RTC/TestRtpStreamRecv.cpp
```diff
@@ -172,6 +174,39 @@ SCENARIO("receive RTP packets and trigger NACK", "[rtp][rtpstream]")
 		REQUIRE(listener.nackedSeqNumbers.empty());
 	}
 
+	SECTION("Receive RTX before corresponding RTP")
```
ibc: "receive" in low case please

ibc (approved):

PR #1680, opened 2025-12-17
Worker: enhance SeqManager tests

jmillan:
Calling `REQUIRE()` inside a global `validate()` function makes any error in the test point to the same line. The aim of this PR is that the error points to the exact line in the failing test.

We cannot make `validate()` a macro because the function is a template.

New `validate()` returns a pair of numbers which are later compared in each call. If a test fails the exact failing test and line are now reported.

Indeed there is no much gain as catch2 reports the line of the REQUIRE failing (always inside `validate()` before), and the line of the test failing (it was useful already).

So within this PR both the line of the REQUIRE failing and the line of the test failing do point to the test failing.

Maybe this PR is useless, or not much useful.

### DIFF
--- a/worker/test/src/RTC/TestSeqManager.cpp
+++ b/worker/test/src/RTC/TestSeqManager.cpp
@@ -25,7 +25,7 @@ struct TestSeqManagerInput
 };
 
 template<typename T, uint8_t N>
-void validate(SeqManager<T, N>& seqManager, std::vector<TestSeqManagerInput<T>>& inputs)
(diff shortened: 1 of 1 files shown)


worker/test/src/RTC/TestSeqManager.cpp
```diff
@@ -44,15 +44,23 @@ void validate(SeqManager<T, N>& seqManager, std::vector<TestSeqManagerInput<T>>&
 
 			seqManager.Input(element.input, output);
 
-			// Covert to string because otherwise Catch will print uint8_t as char.
```
jmillan, line 47: The string conversion does not have effect in current Catch. I've tested it.

ibc (approved):

PR #1683, opened 2026-01-05
Worker: RtpStreamSend, do not send duplicated RTP packets

### LINKED ISSUE #1675: Side-effect of sending duplicate packets to clients
In #906, we are considering not filtering duplicate rtp packets(received after rtx packets) and instead sending them to client.
If we enable transport-cc, a `transport-wide-cc-id` will be set to each video packet.
On the client side, duplicate packets will be dropped by libsrtp when it attempts to decrypt them, and these packets will be marked as NR in tcc feedback.
Below is a simple example, i just motified the simulcast consumer, and sent each rtp packet twice.

Image

Below is the pcap file:
[tcc.zip](https://github.com/user-attachments/files/23899167/tcc.zip)

jmillan:
Fixes #1675

If the packet is already present in the retransmission buffer then discard it.

### DIFF
--- a/worker/src/RTC/RtpStreamSend.cpp
+++ b/worker/src/RTC/RtpStreamSend.cpp
@@ -121,8 +121,6 @@ namespace RTC
 			  packet->GetSsrc(),
 			  packet->GetSequenceNumber(),
 			  packet->GetSize());
-
-			return ReceivePacketResult::DISCARDED;
 		}
 
 		bool stored{ false };
--- a/worker/src/RTC/RtpStreamSend.cpp
+++ b/worker/src/RTC/RtpStreamSend.cpp
@@ -112,15 +112,40 @@ namespace RTC
 			return ReceivePacketResult::DISCARDED;
 		}
 
+		// Check the packet size.
+		if (packet->GetSize() > RTC::Consts::MtuSize)
+		{
+			MS_WARN_TAG(
+			  rtp,
+			  "packet too big [ssrc:%" PRIu32 ", seq:%" PRIu16 ", size:%zu]",
+			  packet->GetSsrc(),
+			  packet->GetSequenceNumber(),
+			  packet->GetSize());
+
+			return ReceivePacketResult::DISCARDED;
+		}
+
 		bool stored{ false };
 
 		// If NACK is enabled, store the packet into the buffer.
 		if (this->retransmissionBuffer)
 		{
-			if (StorePacket(packet, sharedPacket))
+			// Check if the packet is already stored.
+			if (this->retransmissionBuffer->Get(packet->GetSequenceNumber()))
 			{
-				stored = true;
+				// Packet already stored, do not resend it since the receiver will
+				// fail decrypting it and this packet will be considered not received
+				// in the RTCP transport feedback affecting the bandwidth estimation.
+
+				MS_DEBUG_DEV(
+				  "packet already stored in retransmission buffer [ssrc:%" PRIu32 ", seq:%" PRIu16 "]",
+				  packet->GetSsrc(),
+				  packet->GetSequenceNumber());
+
+				return ReceivePacketResult::DISCARDED;
 			}
+
+			stored = this->retransmissionBuffer->Insert(packet, sharedPacket);
 		}
 
 		// Increase transmission counter.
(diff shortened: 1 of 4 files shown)


worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -112,15 +112,40 @@ namespace RTC
...
+		{
+			MS_WARN_TAG(
+			  rtp,
+			  "packet too big [ssrc:%" PRIu32 ", seq:%" PRIu16 ", size:%zu]",
+			  packet->GetSsrc(),
+			  packet->GetSequenceNumber(),
+			  packet->GetSize());
+
+			return ReceivePacketResult::DISCARDED;
+		}
```
ibc: Are we sure we want this? We don't tell clients about the maximum packet size we allow, MtuSize is just a constant we made up. We have very big buffers to store packets and it's never been a problem. And we don't have real data to know if this change can affect real world RTP clients. Honestly I don't think we should do this change.

ibc: We use our hardcoded MTU size limit to filter and split what we sent, not to allow what we can receive.

jmillan: That's right. I'm reverting this specific part.

worker/src/RTC/RtpStreamSend.cpp
```diff
@@ -112,15 +112,38 @@ namespace RTC
 			return ReceivePacketResult::DISCARDED;
 		}
 
+		// Check the packet size.
+		if (packet->GetSize() > RTC::Consts::MtuSize)
```
ibc: I'm also wondering why we want to drop these packets. If a big packet is received we accept it but then we drop it in each Consumer. I don't know why this is god or needed.

jmillan: We are not dropping the packet, just logging that the packet is bigger than our defined MTU size. This log was already here, I've just moved the code in this exact case.

Doing this here seems redundant as the code will be repeated (and it was already) for every consumer. But since we are going to make each consumer set its own RTP header extension set then it makes sense to do this comparison here (as a consumer may include 1 header extension and another one 5 instead, hence the packet size will be different).

jmillan: Having said this, I'm fine just getting rid of the log totally.

ibc: I vote for it. We shouldn't warn the user just based on a hardcoded MTU size estimation.

ibc: Said that it would be great to verify the max packet size in the RtpStreamSend class and verify that we are NOT attempting to store the packet if bigger than our per packet buffer size. Probably we do it already but I am on vacation drinking Bordeaux wines and cannot check.

jmillan: Log removed.

>  verify that we are NOT attempting to store the packet if bigger than our per packet buffer size

We are not storing the packet in any buffer in RtpStreamSend. We are using a shared pointer, so that does not apply.

ibc: Oh yes, I meant the buffer where we clone the received packet, but that's big enough and we check the size anyway AFAIR.

ibc (commented):
Without looking at the code, should we also do this check in RtxStream::ReceivePacket()? In other words: are we checking if packets received via RTX are duplicated?

jmillan:
> Without looking at the code, should we also do this check in RtxStream::ReceivePacket()? In other words: are we checking if packets received via RTX are duplicated?

It does not apply. When receiving a RTX packet and calling `RtxStream::ReceivePacket()` this method does decode the packet into the original one. Original packet duplication is fixed in this PR.

ibc:
> It does not apply. When receiving a RTX packet and calling `RtxStream::ReceivePacket()` this method does decode the packet into the original one. Original packet duplication is fixed in this PR.

What do you mean? `Producer::ReceiveRtpPacket()` may call to `rtpStream->ReceiveRtxPacket()` if this is a RTX packet and there we are not checking if the decoded RTP packet is duplicated.

jmillan:
> What do you mean? Producer::ReceiveRtpPacket() may call to rtpStream->ReceiveRtxPacket() if this is a RTX packet and there we are not checking if the decoded RTP packet is duplicated.

The duplicated packet will reach `RtpStreamSend::ReceivePacket()` and we'll discard it for being already buffered for retransmissions. This is what this PR does.

ibc:
> The duplicated packet will reach RtpStreamSend::ReceivePacket()

True, now I got it

ibc (approved):

PR #1697, opened 2026-01-20
New `RTC::ICE::StunPacket`

ibc:
# Details

- Replace current `RTC/StunPacket` class with a new `RTC/ICE/StunPacket` class that uses `Serializable` class.
- Move ICE related classes to `RTC/ICE/` folder.
- Create the `RTC::ICE` C++ namespace.
- Add tests for the new `RTC::ICE::StunPacket` class and adapt STUN fuzzer.

## More details about the new `ICE::StunClass`

- It doesn't have any private member other than `klass` and `method` (because they are hard to compute). But they are just simple numeric members.
- It doesn't use `std::string` or `char*` at all in its public API. Instead it uses `std::string_view` which is 200 times more flexible and useful.
- When calling any `stunPacket->AddXxxx(xxx)` method such as `AddUsername(std::string_view username)` etc, the STUN attribute is immediately added to the buffer of the `StunPacket`.
- To read attributes (parsed or added via API), the `StunPacket` uses a `map` member that holds `Attribute` structs with an `offset` field that references the offset (in bytes) of the beginning of the attribute in the buffer counting from the beginning of the attributes.


ibc:
@jmillan this PR is still a draft because the new `ICE::StunPacket` class is not yet used in "production" (`IceServer`, `WebRtcTransport`, etc don't use it yet, they use the old one). However PR review is appreciated (without approving it yet).

ibc:
This is done. In order to easily see the parsed STUN requests and sent STUN responses, add this in `WebRtcTransport.cpp` (assuming `WebRtcServer` usage as the demo does by default):

```c++
void WebRtcTransport::ProcessStunPacketFromWebRtcServer(
  RTC::TransportTuple* tuple, const RTC::ICE::StunPacket* packet)
{
	MS_TRACE();

	// TODO: REMOVE
	MS_DUMP_CLEAN(0, "");
	MS_DUMP_CLEAN(0, "<<<< WebRtcTransport::ProcessStunPacketFromWebRtcServer():");
	packet->Dump();
	MS_DUMP_CLEAN(0, "");

	// Pass it to the IceServer.
	this->iceServer->ProcessStunPacket(packet, tuple);
}
```

```c++
inline void WebRtcTransport::OnIceServerSendStunPacket(
  const RTC::ICE::IceServer* /*iceServer*/,
  const RTC::ICE::StunPacket* packet,
  RTC::TransportTuple* tuple)
{
	MS_TRACE();

	// TODO: REMOVE
	MS_DUMP_CLEAN(2, "");
	MS_DUMP_CLEAN(2, ">>>> WebRtcTransport::OnIceServerSendStunPacket():");
	packet->Dump(2);
	MS_DUMP_CLEAN(2, "");

	// Send the STUN response over the same transport tuple.
	tuple->Send(packet->GetBuffer(), packet->GetLength());

	// Increase send transmission.
	RTC::Transport::DataSent(packet->GetLength());
}
```

Real example using the mediasoup-demo:

```
<<<< WebRtcTransport::ProcessStunPacketFromWebRtcServer():
<ICE::StunPacket>
  length: 124 (buffer length: 124)
  method and class: Binding request
  transaction id: 0x356A6279426D566C67704C6C
  attributes length: 104
  <Attributes>
    username: "zgpmlltwylgavhv98d27uqmpbmxfcuk6:Xt/d"
    priority: 1350508287
    ice controlling: 6851224161021668132
    message integrity: 7e8987e9056f9b6356122ac6e35846890695db34
    fingerprint: 3462804419
  </Attributes>
<ICE::StunPacket>

    
    >>>> WebRtcTransport::OnIceServerSendStunPacket():
    <ICE::StunPacket>
      length: 64 (buffer length: 65536)
      method and class: Binding success response
      transaction id: 0x356A6279426D566C67704C6C
      attributes length: 44
      <Attributes>
        xor mapped address: IPv4 192.168.1.41:57712
        message integrity: 110e15b7862d25e21b1df1240da9f4f6d615833e
        fingerprint: 3775910733
      </Attributes>
    <ICE::StunPacket>
    

<<<< WebRtcTransport::ProcessStunPacketFromWebRtcServer():
<ICE::StunPacket>
  length: 128 (buffer length: 128)
  method and class: Binding request
  transaction id: 0x334C6D334B355755537A6A4A
  attributes length: 108
  <Attributes>
    username: "zgpmlltwylgavhv98d27uqmpbmxfcuk6:Xt/d"
    priority: 1350508287
    ice controlling: 6851224161021668132
    use candidate: yes
    message integrity: a6e0483c248633533d6e179c46d660f262a83713
    fingerprint: 3513627022
  </Attributes>
<ICE::StunPacket>

    
    >>>> WebRtcTransport::OnIceServerSendStunPacket():
    <ICE::StunPacket>
      length: 64 (buffer length: 65536)
      method and class: Binding success response
      transaction id: 0x334C6D334B355755537A6A4A
      attributes length: 44
      <Attributes>
        xor mapped address: IPv4 192.168.1.41:57712
        message integrity: 2ee4cd7b7a7bc4a5cb55efb5905c9f18883899a0
        fingerprint: 1097096247
      </Attributes>
    <ICE::StunPacket>
```

ibc:
This is done:

- Tested with all browsers using the demo app.
- Deployed in test.mediasoup.org.

worker/include/RTC/ICE/IceCandidate.hpp
ibc, line 1: This file has been moved to `ICE` folder, just that.

worker/include/RTC/ICE/IceServer.hpp
ibc, line 1: This file has been moved to `ICE` folder and modified to use the new `RTC::ICE::StunPacket` class instead of the former `RTC::StunPacket` class.

worker/include/RTC/ICE/StunPacket.hpp
ibc, line 1: This is the new `RTC::ICE::StunPacket` class which uses `Serializable`.

worker/include/RTC/WebRtcServer.hpp
ibc, line 1: This file has been modified to use the new `RTC::ICE::StunPacket` class instead of the former `RTC::StunPacket` class.

worker/include/RTC/WebRtcTransport.hpp
ibc, line 1: This file has been modified to use the new `RTC::ICE::StunPacket` class instead of the former `RTC::StunPacket` class.

worker/include/Utils.hpp
```diff
@@ -242,6 +243,8 @@ namespace Utils
 
 		static const uint8_t* GetHmacSha1(const std::string& key, const uint8_t* data, size_t len);
 
+		static void WriteRandomBytes(uint8_t* buffer, size_t len);
```
ibc, line 246: This is a new util to generate the 12 bytes `TransactionID` of STUN packets.

worker/src/RTC/RTP/Packet.cpp
ibc, line 1: Here I just added a flag in the `Validate()` method because it not only validates but also parses and **stores** RTP header extensions, but I want it to be able to validate the packet and the extensions **without** storing them again. Same in the new `ICE::StunPacket::Validate()` method.

worker/src/RTC/Serializable.cpp
```diff
@@ -129,6 +129,8 @@ namespace RTC
 
 		if (serializable->GetBufferLength() < this->length)
 		{
+			const auto bufferLength = serializable->GetBufferLength();
```
ibc, line 132: This was a bug. Before this addition, the `bufferLength` logged in the line below was the **private member** of this `Serializable` class, but we want to log the buffer length of the `serializable` instance given to `CLoneInto()` method.

worker/src/Utils/IP.cpp
```diff
@@ -112,7 +112,7 @@ namespace Utils
 		}
 	}
 
-	void IP::NormalizeIp(std::string& ip)
+	std::string IP::NormalizeIp(std::string& ip)
```
ibc, line 115: Just for convenience, I've made this function ALSO return the modified string rather than just modifying the given one.

worker/test/src/RTC/ICE/TestStunPacket.cpp
ibc, line 1: We didn't have STUN tests.

worker/src/Utils/IP.cpp
jmillan: What was the reason for this, out of curiosity?

worker/test/include/RTC/ICE/iceCommon.hpp
```diff
@@ -0,0 +1,126 @@
...
+	namespace ICE
+	{
+		// NOTE: We need to declare them here with `extern` and then define them in
+		// common.cpp.
+		extern thread_local uint8_t FactoryBuffer[66661];
+		extern thread_local uint8_t ResponseFactoryBuffer[66661];
+		extern thread_local uint8_t SerializeBuffer[66662];
+		extern thread_local uint8_t CloneBuffer[66663];
+		extern thread_local uint8_t DataBuffer[66664];
+		extern thread_local uint8_t ThrowBuffer[66665];
```
jmillan, line 28: Why these increasing sizes? A comment indicating the reason would help future us.

worker/src/Utils/IP.cpp
ibc: See usage in ICE/TestPacket.cpp

jmillan (approved):
I like it. CI is failing, maybe with an unrelated error.

worker/test/include/RTC/ICE/iceCommon.hpp
ibc: This is a copy paste of the rtpCommon.cpp that was already there. The random size is because... why not.

ibc:
> I like it. CI is failing, maybe with an unrelated error.

It's Node pipeToRouter() the test that is failing no idea why and didn't fail on previous days. Zero changes.

worker/test/include/RTC/ICE/iceCommon.hpp
jmillan: Unless there is a reason not to use the same values I think it's more readable using the same. Otherwise a comment here would save us time in the future IMHO. "// Random buffer sizes."

ibc: https://github.com/versatica/mediasoup/pull/1697/changes/80d7b69ff9b6a1f8c132e0bd6b14207436efc942

Issue #1711, opened 2026-02-04
Enable clang-tidy in `worker/test` and `worker/fuzzer`


ibc:
Currently we only run `clang-tidy` on `worker/src`.

- We should run it also on `worker/test` and `worker/fuzzer`.
- Same changes should be done in `mediasoup-worker-clang-tidy.yaml` (see `on.pull_requests.paths` and `include` in `ZedThree/clang-tidy-review@v0.22.1` step.

PR #1713, opened 2026-02-05
clang-tidy: only output diagnose from our code

jmillan:
Ignore external libaries.

### DIFF
--- a/worker/.clang-tidy
+++ b/worker/.clang-tidy
@@ -1,5 +1,4 @@
 ---
-HeaderFilterRegex: '.*/worker/(include|src)/.*'
 Checks: "*,\
   -altera*,\
   -android*,\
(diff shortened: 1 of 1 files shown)


ibc (commented):
I don't love it. This change goes directly against https://github.com/versatica/mediasoup/issues/1711 task.

jmillan:
We don't want to be warned about external libraries. This RegExp just needs to be adjusted. On it.

worker/.clang-tidy
```diff
@@ -83,6 +83,7 @@ Checks: "*,\
   -readability-simplify-boolean-expr,\
   -readability-uppercase-literal-suffix,\
   "
+HeaderFilterRegex: '.*/worker/(include|src|test|fuzzer)/.*'
```
ibc: Why is `src` included? Also, not sure if possible but this should be way easier and way more specific:

- worker/include
- worker/test/include
- worker/fuzzer/include

jmillan: I can remove `src`. The filter regex is what we have to indicate so.

ibc: But can't it be like this?

```
HeaderFilterRegex: '.*/worker/(include|test\/include|fuzzer\/include)/.*'
```

jmillan: Done. Only worker header folders are indicated.

ibc (approved):

PR #1719, opened 2026-02-06
clang-tidy for RTP tests

ibc:
# Details

- Enable RTP tests in clang-tidy task.
- This PR follows same guidelines as previous one (RTCP tests):
  - If there is a single `SCENARIO` in a test file, declare variables, structs and classes in the `SCENARIO` block and also use lambda functions there (such as `validate()`).
  - If there are N `SCENARIOs` in a test file and they need access to same variables/classes/structs/functions, then declare those within an anonymous C++ namespace.

### DIFF
--- a/worker/test/include/RTC/RTP/rtpCommon.hpp
+++ b/worker/test/include/RTC/RTP/rtpCommon.hpp
@@ -11,22 +11,19 @@
 
 using namespace RTC::RTP;
 
-namespace RTC
+namespace RTP_COMMON
 {
-	namespace RTP
-	{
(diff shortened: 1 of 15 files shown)


worker/test/include/RTC/RTP/rtpCommon.hpp
```diff
@@ -11,22 +11,19 @@
 
 using namespace RTC::RTP;
 
-namespace RTC
+namespace RTP_COMMON
```
ibc, line 14: Stop defining **same** `RTC` namespace in tests, it was just a mess.

jmillan (approved):

PR #1723, opened 2026-02-06
Worker: Remove all `using namespace`

ibc:
# Details

- It doesn't make sense that we only use `using namespace` in `worker/test/`.

## Bonus tracks

- Remove `worker/scripts/normalize-compile-commands.mjs` file and integrate it in `worker/scripts/clang-scripts.mjs` in a new `npm run normalize-compile-commands` NPM script. Why? Because it doesn't depend on `cwd`. It generates the same correct normalized `compile_commands.json` file no matter where `npm run normalize-compile-commands` is called from.


ibc:
I know I know, in same cases it looks convenient to use `using namespaces RTC::RTCP`, etc. But then you look at the code and you see a terrible and unclear mix of symbols with a namespace prefix and others that don't have it.

jmillan (approved):

PR #1727, opened 2026-02-11
Worker: Improve `Channel` classes

ibc:
# Details

- Cosmetic minimal changes.


jmillan (approved):

ibc:
NOTE: I've reverted this:

> - Use always a `thread_local static flatbuffers::FlatBufferBuilder bufferBuilder` class member instead of using a instance member.

There is no reason to have a static member in those classes since they are singleton class instances (there is only one per `Worker`).

PR #1728, opened 2026-02-12
Merge `Channel::ProducerSocket` and `Channel::ConsumerSocket` into a single bidirectional `Channel::Socket`

ibc:
# Details

- In theory we don't need separate UNIX sockets to send and receive messages between Node and Worker.
- Many years ago we had a single bidirectional UNIX socket but we split it into "producer socket" and "consumer socket" due to some problems in Windows.
- Hopefully those problems have gone away in recent Windows/Node versions.
- Make `WORKER_CLOSE` be a notification instead of a request because once we send it we don't care about the response (and we cannot care about the response because the UNIX socket is closed anyway).
- Worker executable: No need to wait 200ms after Worker instance is destroyed (because we are not awaiting for any response).


ibc:
Closing because indeed same problems happen in Node+Windows. This is presumably because of course there is no Unix Socket in Windows and what libuv provides on Windows to interconnect processes is not a bidirectional channel. So I'm closing this PR and will write another one with just minor improvements such as moving `WORKER_CLOSE` to notifications.

Node test failing in Windows:

- https://github.com/versatica/mediasoup/actions/runs/21951216192/job/63402333881?pr=1728
- https://github.com/versatica/mediasoup/actions/runs/21951216192/job/63402333839?pr=1728

All tests fail with `PROCESSWRAP` or `PIPEWRAP`.

PR #1731, opened 2026-02-16
Worker: New Consumer class handling all possible producer stream modes.

jmillan:
New `ProducerStreamManager` interface with the following implementations:
- `SimpleProducerStreamManager`: handles a single stream with no spatial or temporal layers.
- `SimulcastProducerStreamManager`: handles N streams with 1 spatial and N temporal layers.
- `SvcProducerStreamManager`: handles 1 stream with N spatial and M temporal layers.

This new class handles the different flavours of producers streams making `Consumer` agnostic. This allows:
- Cleaner implementation: Producer streams details are hidden in the corresponding implementation.
- Reduced  code size: No more duplicated full consumer implementation.
- Will allow to use the same Consumer and switch the ProducerStreamManager that sources it.

`Consumer + SimpleProducerStreamManager` replaces `SimpleConsumer`.
`Consumer + SimulcastProducerStreamManager `replaces `SimulcastConsumer`.
`Consumer + SvcProducerStreamManager` replaces `SvcConsumer`.
`Consumer + PipeProducerStreamManager` replaces `PipeConsumer`.

 
FUTURE (Other PR):
- Make consumer pipe-able.


worker/include/RTC/Consumer.hpp
```diff
@@ -86,12 +92,9 @@ namespace RTC
 		{
 			return this->type;
 		}
-		virtual RTC::ConsumerTypes::VideoLayers GetPreferredLayers() const
+		VideoLayers GetPreferredLayers() const
```
ibc: Can we please keep RTC::ConsumerTypes::VideoLayers instead of just VideoLayers? We have a general problem about how we include namespaces or not randomly in the whole code.

jmillan: sure, no problem.

jmillan: done

worker/include/RTC/Consumer.hpp
```diff
@@ -16,14 +17,19 @@
...
+#include <map>
+#include <memory>
 #include <string>
 #include <vector>
 
 namespace RTC
 {
-	class Consumer : public Channel::ChannelSocket::RequestHandler
+	class Consumer : public Channel::ChannelSocket::RequestHandler,
+	                 public RTC::RTP::RtpStreamSend::Listener,
```
ibc, line 31: Does it make sense that `Consumer` still owns `RtpStreamSend` streams and hence listens directly for their events/callbacks? Shouldn't this be done now by the `ProducerStreamManager`? I mean, "producer stream manager" really sounds like if it's something that manages producer streams.

worker/include/RTC/Consumer.hpp
```diff
@@ -196,13 +219,21 @@ namespace RTC
 		struct TraceEventTypes traceEventTypes;
 
 	private:
+		// Allocated by this.
+		RTC::RTP::RtpStreamSend* rtpStream{ nullptr };
 		// Others.
+		std::vector<RTC::RTP::RtpStreamSend*> rtpStreams;
```
ibc, line 224: Exactly here. This should exclusively exist in `ProducerStreamManager`, shouldn't it?

worker/include/RTC/Consumer.hpp
jmillan: > "producer stream manager" really sounds like if it's something that manages producer streams.

But `RtpStreamSend` streams are unrelated to producer or producer streams. They are consumer streams indeed.

These streams in the vector are only used by `PipeConsumer`.  I believe we'll need to use them in `Consumer` now in order to make it pipe-able.

worker/include/RTC/Consumer.hpp
jmillan: As said, logically they are consumer streams rather than producer streams, so they should be in Consumer. Anyway, since these are only used for pipe-ing and such a feature is not yet done, I'll suggest that we come back to this item afterwards.

I'd like to have this PR reviewed  before so it serves as a base for the new_consumer featue. Pipe-ing can later be done in a separate PR (or even here) pointing to this branch.

ibc: Why should they be in the Consumer. Please consider the case where we will need to switch the entire ProducerStreamManager when changing the Producer source of a Consumer. In that scenario I'd like to have to replace as less Consumer members as possible. Perhaps the send RTP stream must never be replaced, not sure.

Anyway I got it wrong since I thought these were the Producer' streams.

worker/test/src/RTC/TestSimpleProducerStreamManager.cpp
```diff
@@ -0,0 +1,540 @@
...
+		0x80, 0x01, 0x00, 0x08,
+		0x00, 0x00, 0x00, 0x04,
+		0x49, 0x96, 0x02, 0xD2, // SSRC: 1234567890.
+		// Payload (4 bytes).
+		0xFF, 0xFF, 0xFF, 0xFF,
+		// Extra buffer for cloning.
+		0xFF, 0xFF, 0xFF, 0xFF,
+		0xFF, 0xFF, 0xFF, 0xFF,
+		0xFF, 0xFF, 0xFF, 0xFF,
+	};
```
ibc: Please don't do this. I commented about it in many recent commits. Let's not parse buffers to create RTP packets. Let's use Factory() instead as done in RTC/RTP/TestPacket.cpp.

ibc (dismissed):

worker/test/src/RTC/TestSimpleProducerStreamManager.cpp
ibc: Also what does "extra space for cloning" mean? That's not a thing in the new RTP Packet class where a separate buffer must be given for cloning purposes.

jmillan: Done 👍

jmillan: > Also what does "extra space for cloning" mean? T

I was immersed in the past.

worker/test/src/RTC/TestSimpleProducerStreamManager.cpp
```diff
@@ -1,8 +1,7 @@
 #include "RTC/RTP/Codecs/PayloadDescriptorHandler.hpp"
-#include "RTC/RTP/Packet.hpp"
 #include "RTC/RTP/RtpStreamRecv.hpp"
+#include "RTC/RTP/rtpCommon.hpp"
 #include "RTC/SimpleProducerStreamManager.hpp"
-#include <catch2/catch_test_macros.hpp>
```
ibc: Why have you removed catch2 include???

worker/test/src/RTC/TestSimpleProducerStreamManager.cpp
```diff
@@ -185,23 +182,12 @@ namespace
...
-	};
-	// clang-format on
+	std::unique_ptr<RTC::RTP::Packet> packet(
+	  RTC::RTP::Packet::Factory(rtpCommon::FactoryBuffer, sizeof(rtpCommon::FactoryBuffer)));
 
-	const size_t packetLength{ 16 };
-	const size_t bufferLength{ packetLength + 12 };
+	packet->SetPayloadType(1);
+	packet->SetSsrc(mappedSsrc);
+	packet->SetPayloadLength(sizeof(rtpCommon::FactoryBuffer) - RTC::RTP::Packet::FixedHeaderMinLength);
```
ibc: This is too risky. If you add CCSR or header extensions then the length of the packet will exceed the buffer length (it will throw). No need to add so much big payload here.

You can safely add just N bytes:

```c++
packet->SetPayloadLength(40);
```

worker/test/src/RTC/TestSimpleProducerStreamManager.cpp
jmillan: editor did it. putting back.

ibc:
Please CHANGELOG entry.

worker/include/RTC/Consumer.hpp
```diff
@@ -16,15 +17,25 @@
 #include "RTC/RTP/RtpStreamSend.hpp"
 #include "RTC/RTP/SharedPacket.hpp"
 #include "RTC/RtpDictionaries.hpp"
+#include "RTC/SeqManager.hpp"
+#include "Shared.hpp"
 #include "SharedInterface.hpp"
+#include <ankerl/unordered_dense.h>
 #include <bitset>
+#include <map>
+#include <memory>
```
ibc: Please don't include headers already present in `common.hpp`.

worker/include/RTC/ProducerStreamManager.hpp
```diff
@@ -0,0 +1,183 @@
+#ifndef MS_RTC_PRODUCER_STREAM_MANAGER_HPP
+#define MS_RTC_PRODUCER_STREAM_MANAGER_HPP
+
+#include "common.hpp"
+#include "Logger.hpp"
```
ibc: Logger.hpp should never be included in .hpp files.

worker/include/RTC/ProducerStreamManager.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include "RTC/ConsumerTypes.hpp"
+#include "RTC/RTP/Codecs/PayloadDescriptorHandler.hpp"
+#include "RTC/RTP/Packet.hpp"
+#include "RTC/RTP/RtpStreamRecv.hpp"
+#include "RTC/RtpDictionaries.hpp"
+#include "SharedInterface.hpp"
+#include <memory>
```
ibc: Already in common.hpp. Remove please.

worker/src/RTC/SimulcastProducerStreamManager.cpp
```diff
@@ -0,0 +1,1069 @@
+#define MS_CLASS "RTC::SimulcastProducerStreamManager"
+// #define MS_LOG_DEV_LEVEL 3
+
+#include "RTC/SimulcastProducerStreamManager.hpp"
+#include "Logger.hpp"
+#include <utility>
```
ibc: Already in common.hpp. Remove please.

worker/src/RTC/Transport.cpp
```diff
@@ -12,7 +12,8 @@
 #include "FBS/transport.h"
 #include "RTC/BweType.hpp"
 #include "RTC/Consts.hpp"
-#include "RTC/PipeConsumer.hpp"
+#include "RTC/Consumer.hpp"
+// #include "RTC/PipeConsumer.hpp" // TODO: PipeConsumer not yet ported to new Consumer.
```
ibc: Is this still true? I mean, PipeConsumer class has been deleted in this PR.

worker/test/src/RTC/TestPipeProducerStreamManager.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+namespace
+{
+	using RtpPacketProcessResult = RTC::ProducerStreamManager::RtpPacketProcessResult;
+
+	// NOLINTBEGIN(readability-identifier-naming)
+	constexpr uint32_t mappedSsrc0 = 1000;
+	constexpr uint32_t mappedSsrc1 = 2000;
+	constexpr uint32_t mappedSsrc2 = 3000;
+	// NOLINTEND(readability-identifier-naming)
```
ibc: Why don't we use uppercase for initial char here?

worker/test/src/RTC/TestPipeProducerStreamManager.cpp
```diff
@@ -0,0 +1,473 @@
...
+		}
+
+		const RTC::ConsumerTypes::VideoLayers preferredLayers;
+
+		return std::make_unique<RTC::PipeProducerStreamManager>(
+		  consumableRtpEncodings, preferredLayers, nullptr, kind, keyFrameSupported, listener, &shared);
+	}
+} // namespace
+
+SCENARIO("PipeProducerStreamManager", "[rtp][producer-stream-manager][pipe]")
```
ibc: I know there is no real guidelines here but in many other tests we don't split words with "-" => `"producerstreammanager"`.

ibc (changes requested):
Initial review.

worker/src/RTC/Transport.cpp
jmillan: Implemented in the corresponding PipeConsumeProducerStreamManager. Removing the outdated comment.

worker/test/src/RTC/TestPipeProducerStreamManager.cpp
jmillan: removed NOTLINTBEGIN

ibc: and NOLINTEND right? But then tidy will complain because those consts should start with upper letter.

jmillan: Of course I renamed the variables.

ibc (approved):

PR #1740, opened 2026-02-23
Enhance RTCP.md with TMMBR explanation and mediasoup strategy

Kartik-619:
Expanded section on TMMBR and its implications in SFU topologies, detailing why mediasoup ignores TMMBR and the principles behind its congestion control strategy.


ibc (commented):
AI generated PRs not allowed, sorry.

PR #1749, opened 2026-03-12
Fix "SCTP failed" if no DataChannel is created on a Transport with `enableSctp: true`

### LINKED ISSUE #1748: If a `Transport` is created with `enableSctp: true` and no DataChannel is created, after ~4 minutes it emits `sctpstatechange` with state "failed"
## STR

1. Create a `WebRtcTransport` in mediasoup with `enableSctp: true`.
2. Connect a client and do not create any `DataProducer/DataConsumer`.

To reproduce it in the mediasoup-demo (locally), modify `Room.ts` in the server by setting `enableSctp: true` instead of `enableSctp: Boolean(sctpCapabilities)`.

Then in the browser add `&datachannel=false` to the URL and reload. After 4 minutes you will see this warning error in the server terminal:

```
mediasoup:WARN:Channel [pid:54917] RTC::SctpAssociation::OnUsrSctpReceiveSctpNotification() | SCTP setup failed:  
```

## More info

When `enableSctp` is set, `Transport.cpp` constructor creates `this->sctpAssociation = new RTC::SctpAssociation(...)`. Somehow, if no `DataProducer/DataConsumer` is created within ~4 minutes, `usrsctp` emits `SCTP_CANT_STR_ASSOC` event, which in `SctpAssociation` triggers a call to `OnSctpAssociationFailed()`:

```c++
case SCTP_CANT_STR_ASSOC:
{
	if (notification->sn_header.sn_length > 0)
	{
		static const size_t BufferSize{ 1024 };
		thread_local char buffer[BufferSize];

		const uint32_t len =
		  notification->sn_assoc_change.sac_length - sizeof(struct sctp_assoc_change);

		for (uint32_t i{ 0 }; i < len; ++i)
		{
			std::snprintf(
			  buffer, BufferSize, " 0x%02x", notification->sn_assoc_change.sac_info[i]);
		}

		MS_WARN_TAG(sctp, "SCTP setup failed: %s", buffer);
	}

	if (this->state != SctpState::FAILED)
	{
		this->state = SctpState::FAILED;
		this->listener->OnSctpAssociationFailed(this);
	}

	break;
}
```

This is totally undesirable. We shouldn't mandate any quick creation of `DataProducer/DataConsumer` to avoid this.

AFAIS the problem is that `SctpAssociation` constructor immediately sends or tries to send the SCTP INIT message. Maybe we should do that only when the first `DataProducer/DataConsumer` is created.

**UPDATE:** No, the problem is that we call `usrsctp_connect()` in `SctpAssociation::TransportConnected()` even if there are no `DataProducer/Consumers` yet.

We should call `usrsctp_connect()` in two cases:
1. `SctpAssociation::TransportConnected()` is called and **there are already** `DataProducer/Consumers`.
2. The first `DataProducer` or `DataConsumer` is created **after** `TransportConnected()` was called time ago.

ibc:
# Details

- Fixes https://github.com/versatica/mediasoup/issues/1748
- Bonus: Don't even try to start DTLS handshake if ICE is not yet connected/completed.

### DIFF
--- a/worker/src/RTC/SctpAssociation.cpp
+++ b/worker/src/RTC/SctpAssociation.cpp
@@ -290,60 +290,16 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		// Just run the SCTP stack if our state is 'new'.
-		if (this->state != SctpState::NEW)
-		{
-			return;
-		}
+		this->transportConnected = true;
 
-		try
-		{
-			int ret;
-			struct sockaddr_conn rconn{}; // NOLINT(cppcoreguidelines-pro-type-member-init)
+		MayConnect();
+	}
 
-			std::memset(&rconn, 0, sizeof(rconn));
-			rconn.sconn_family = AF_CONN;
-			rconn.sconn_port   = htons(5000);
-			rconn.sconn_addr   = reinterpret_cast<void*>(this->id);
-#ifdef HAVE_SCONN_LEN
-			rconn.sconn_len = sizeof(rconn);
-#endif
+	void SctpAssociation::TransportDisconnected()
+	{
+		MS_TRACE();
 
-			ret = usrsctp_connect(this->socket, reinterpret_cast<struct sockaddr*>(&rconn), sizeof(rconn));
-
-			if (ret < 0 && errno != EINPROGRESS)
-			{
-				MS_THROW_ERROR("usrsctp_connect() failed: %s", std::strerror(errno));
-			}
-
-			// Disable MTU discovery.
-			sctp_paddrparams peerAddrParams{}; // NOLINT(cppcoreguidelines-pro-type-member-init)
-
-			std::memset(&peerAddrParams, 0, sizeof(peerAddrParams));
-			std::memcpy(&peerAddrParams.spp_address, &rconn, sizeof(rconn));
-			peerAddrParams.spp_flags = SPP_PMTUD_DISABLE;
-
-			// The MTU value provided specifies the space available for chunks in the
-			// packet, so let's subtract the SCTP header size.
-			peerAddrParams.spp_pathmtu = SctpMtu - sizeof(struct sctp_common_header);
-
-			ret = usrsctp_setsockopt(
-			  this->socket, IPPROTO_SCTP, SCTP_PEER_ADDR_PARAMS, &peerAddrParams, sizeof(peerAddrParams));
-
-			if (ret < 0)
-			{
-				MS_THROW_ERROR("usrsctp_setsockopt(SCTP_PEER_ADDR_PARAMS) failed: %s", std::strerror(errno));
-			}
-
-			// Announce connecting state.
-			this->state = SctpState::CONNECTING;
-			this->listener->OnSctpAssociationConnecting(this);
-		}
-		catch (const MediaSoupError& /*error*/)
-		{
-			this->state = SctpState::FAILED;
-			this->listener->OnSctpAssociationFailed(this);
-		}
+		this->transportConnected = false;
 	}
 
 	flatbuffers::Offset<FBS::SctpParameters::SctpParameters> SctpAssociation::FillBuffer(
@@ -369,12 +325,17 @@ namespace RTC
 		  this->isDataChannel);
 	}
 
-	void SctpAssociation::ProcessSctpData(const uint8_t* data, size_t len) const
+	void SctpAssociation::ProcessSctpData(const uint8_t* data, size_t len)
 	{
 		MS_TRACE();
 
+		this->sctpDataReceived = true;
+
+		MayConnect();
+
 #if MS_LOG_DEV_LEVEL == 3
-		MS_DUMP_DATA(data, len);
+		// NOTE: Only uncomment this during local debugging if needed.
+		// MS_DUMP_DATA(data, len);
 #endif
 
 		usrsctp_conninput(reinterpret_cast<void*>(this->id), data, len, 0);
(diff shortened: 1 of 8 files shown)


worker/src/RTC/SctpAssociation.cpp
```diff
@@ -522,6 +496,96 @@ namespace RTC
...
+	void SctpAssociation::MayConnect()
+	{
+		MS_TRACE();
+
+		// Just run the SCTP stack if our state is 'new'.
+		// Notice that once MayConnect() is called (and the code below is executed),
+		// SCTP state will no longer be "NEW".
+		if (this->state != SctpState::NEW)
+		{
+			MS_DEBUG_DEV("ignoring because SCTP state is not NEW");
```
jmillan: Cosmetic for this and the following logs:


```suggestion
			MS_DEBUG_DEV("SCTP state is not NEW, ignoring");
```

worker/src/RTC/SctpAssociation.cpp
```diff
@@ -522,6 +496,96 @@ namespace RTC
...
+			MS_DEBUG_DEV("ignoring because transport is not connected");
+
+			return;
+		}
+		// If there are no SCTP streams yet and no SCTP data has been yet received
+		// from the remote peer, don't do anything.
+		// This is because the peer may never create a DataChannel so we shouldn't
+		// try to connect SCTP (SCTP INIT chunk, etc) since it will timeout and
+		// trigger "SCTP failed".
+		else if (!this->firstStreamCreated && !this->sctpDataReceived)
```
jmillan, line 525: By design we cannot receive SCTP data unless:
- `transportConnected=true` && 
- `firstStreamCreated=true`

So we don't need to check for `sctpDataReceived` in order to connect SCTP. We can connect as long as both `transportConnected` and `firstStreamCreated` is true.

## `DataConsumer` creation
### Transport is NOT yet connected
1. `DataConsumer` is created in server (`firstStreamCreated = true`).
   -  `DataConsumerOptions` are sent to client. 
2. `transport.consumeData()` is called in client, which awaits the method.
   -  Client signals server transport to connect
3.  `transport.connect()` is triggered in server side (`transportConnected = true`).
    -  Server responds the client about transport connection.
    -  Once server notification arrives to client `transport.consumeData()` is resolved in client.

NO SCTP data can arrive if `transportConnected = false`.

### Transport is ALREADY connected
1. `DataConsumer` is created in server (`firstStreamCreated = true`).
   -  `DataConsumerOptions` are sent to client. 
2. `transport.consumeData()` is called in client.

NO SCTP data can arrive if `transportConnected = false`.

## `DataProducer` creation
### Transport is NOT yet connected
1. `transport.produceData()` is called in client, which awaits the method.
   -  Client signals server transport to connect
2.  `transport.connect()` is triggered in server side (`transportConnected = true`).
     - Server notifies the client about transport connection.
3. Client request server to create the `DataProducer`.
   - Server responds client about `DataProducer` creation.
   - `transport.produceData()` is resolved in client.

### Transport is ALREADY connected
1. `transport.produceData()` is called in client, which awaits the method.
2. Client request server to create the `DataProducer`.
   - Server responds client about `DataProducer` creation.
   - `transport.produceData()` is resolved in client.

NO SCTP data can arrive if `transportConnected = false`.

ibc:
Some comments before I enter into details:

- "transport connected" doesn't mean that transport.connect() was called. In a WebRtcTransport it means that DTLS is connected. In a PlanTransport it is always true from the beginning. Please see in Transport.cpp and WebRtcTransport.cpp when TransportConnected() is called.
- We need "sctpDataReceived" because the transport may receive SCTP data (INIT chunk) **before** the DataProducer/Consumer request arrives to the worker (typical race condition between signaling and media) so we optimize it as follows: if SCTP data is received it means that the peer has enabled DataChannel so we allow such a received INIT chunk and start connecting our side. Otherwise the received SCTP INIT chunk could be ignored, so peer would resend it after some time etc.

worker/src/RTC/SctpAssociation.cpp
ibc: > transport.connect() is triggered in server side (transportConnected = true).
> Server responds the client about transport connection.
> Once server notification arrives to client transport.consumeData() is resolved in client.

BTW this flow has never been like this. The client app never waits for transport.connect() to resolve in server before it starts sending, for example, RTP. Same here. This has never been a requirement.

jmillan: Probably I'm missing  something here?

- `transport.consumeData()` -> 
- `handle->receiveDataChannel()` -> 
- `handle->setupTransport()` ->
- `emit('@connect')` ->
- `transport.on('connect', resolve,  reject)` -> 
- Tell server to connect ->
- Wait for server connect response -> 
- `resolve()` -> 
- `transport.consumeData()` returns.

EDIT: Same applies to `DataProducer`

jmillan: > BTW this flow has never been like this. The client app never waits for transport.connect() to resolve in server before it starts sending, for example, RTP. Same here. This has never been a requirement.

That applies to RTP which is sent by client's WebRTC engine, but in case of DataChannels which are sent by the client app, it cannot send anything through a `DataProducer` that is not yet created. Am I wrong?

ibc: Peer CANNOT send SCTP DataChunks (user/app messages) until SCTP is connected. But peer CAN send SCTP INIT chunk to connect the SCTP and we want to optimize it and allow SCTP connection start even if the remote INIT chunk is received by the mediasoup Transport before the `PRODUCE/CONSUME_DATA` signaling request.

ibc: The SCTP association corresponds to the SINGLE `m=application` section in the SDP. There is a SINGLE `m=application` section despite there are 0, 1 or 1000 DataProducers/DataConsumers. The SCTP connection is NOT related to the creation of DataProducers/Consumers. The SCTP connection establishes the SCTP association and then SCTP streams are created on it, and those streams correspond to our DataProducer/Consumers.

jmillan: That's correct, yes. Totally agreed. :+1:

jmillan (approved):

PR #1771, opened 2026-04-08
worker: relax log severity for legitimate situations

jmillan:
Do not use `warn` for legitimate situations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Worker: Relax log severity for legitimate situations ([PR #1771](https://github.com/versatica/mediasoup/pull/1771).
+
 ### 3.19.19
 
 - Worker: Ensure 4-byte alignment for network packet receive buffers and test buffers to avoid undefined behavior ([PR #1756](https://github.com/versatica/mediasoup/pull/1756).
--- a/worker/src/RTC/Producer.cpp
+++ b/worker/src/RTC/Producer.cpp
@@ -570,7 +570,7 @@ namespace RTC
 
 		if (!rtpStream)
 		{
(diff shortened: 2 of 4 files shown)


ibc (commented):
Those are indeed legitimate situations that shouldn't happen if everything is good. Yes, sometimes the RTP reaches mediasoup before the Producer is created and those warnings show up, but that's ok. If we followed the rationale of this PR in the whole code then everything should be "debug" severity but a few catastrophic errors/aborts.

jmillan:
> Those are indeed legitimate situations that shouldn't happen if everything is good. Yes, sometimes the RTP reaches mediasoup before the Producer is created and those warnings show up, but that's ok. If we followed the rationale of this PR in the whole code then everything should be "debug" severity but a few catastrophic errors/aborts.

The client may start sending RTP before producer is ready in server side, for example. Or producer may be removed from server side before doing so in client side. Those normal situations (normal because there is always a signaling delay between client and server) generate lots of warnings which is OK in dev environments but simply unusable in big production environments where one must filter those messages in a post process pipeline otherwise.

I'm fine closing this PR. One can always temporarily enable 'warn' log level and filter at another place.

ibc:
If the client sends RTP packets with wrong PT or SSRC etc that don't match the given RTP parameters (due to bad setup), we want to see warnings, right?

jmillan:
> If the client sends RTP packets with wrong PT or SSRC etc that don't match the given RTP parameters (due to bad setup), we want to see warnings, right?

Depending on what kind of error you are looking for. If you have a specific problem to troubleshoot you may not want to see those, which may be very well generated for every RTP producer (depending on how many producers we are talking about these may be a lot of logs).

So in summary we cannot make assumptions on what to filter on the library side. The app/service needs to filter as it needs.

PR #1777, opened 2026-04-19
Worker: Add `useBuiltInSctpStack` setting

ibc:
# Details

- Worker: Add `useBuiltInSctpStack` setting (defaults to `false`) to enable mediasoup built-in SCTP stack (which is still WIP of course).
- So remove `-Dms_sctp_stack` meson arg.

# TODO

- [x] Document it in the website.


jmillan (approved):

worker/src/RTC/Transport.cpp
```diff
@@ -1474,7 +1505,9 @@ namespace RTC
 
 			case Channel::ChannelRequest::Method::TRANSPORT_CLOSE_DATAPRODUCER:
 			{
-				if (!this->sctpAssociation)
+				if (
+				  (Settings::configuration.useBuiltInSctpStack && !this->sctpAssociation) ||
+				  (!Settings::configuration.useBuiltInSctpStack && !this->oldSctpAssociation))
```
jiyeyuran, line 1510: DirectTransport doesn't have oldSctpAssociation. We shouldn't exit here prematurely, otherwise the dataConsumers will not receive notifications.

ibc: Sorry @jiyeyuran, I don't understand. What do you mean? When `useBuiltInSctpStack` is `false` (default value) both conditions are the same:

BEFORE:
```c++
if (!this->sctpAssociation)
```

AFTER:
```c++
if (
				  (Settings::configuration.useBuiltInSctpStack && !this->sctpAssociation) ||
				  (!Settings::configuration.useBuiltInSctpStack && !this->oldSctpAssociation))
```

Are you reporting a problem which happens **after** this PR? Or something that was already there before?

ibc: Ok, it was a regression introduced in previous versions. Fixing.

ibc: Fixed here https://github.com/versatica/mediasoup/pull/1780

ibc: BTW I've added you @jiyeyuran  as contributor in the mediasoup website:

- https://mediasoup.org/about/#contributors-and-acknowledgements
- https://mediasoup.org/github/#3rd-party-contributions

PR #1778, opened 2026-04-20
Attempt to fix package-lock.json issue due to `knip` dep version update

ibc:
# Details

- Issue reported in [knip](https://github.com/webpro-nl/knip/issues/1691).
- Trying to fix this issue in [GitHub CI](https://github.com/versatica/mediasoup/actions/runs/24653385862/job/72082694105):
  ```
  npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
  npm error
  npm error Missing: @emnapi/core@1.10.0 from lock file
  npm error Missing: @emnapi/runtime@1.10.0 from lock file
  ```
- It seems to happen because, due to the upgrade from `knip` 6.4.1 to 6.5.1, `package-lock.json` becomes different depending on the OS.
- Here I've invoked `npm install` in Linux (Docker) so the generated `package-lock.json` is different.

## IMPORTANT

This PR fixes the issue. `npm ci` works fine in all OS. **Problem** is that, from now on, if we want to update some dep in our package.json, then we have to run `npm install xxxx` or `npm install` in Linux (or Linux Docker) rather than in macOS.

Another option is that we downgrade `knip` from 6.5.0 to 6.4.1 until they fix the problem (if it happens, because the issue could be in another dep of `knip`).

## UPDATE

Finally we have decided to downgrade `knip` to 6.4.1.


jmillan:
I'd personally use the old version of knip until fixed.

jmillan (dismissed):

ibc:
Ok, the issue happens if I upgrade knip from 6.4.1 to 6.5.0 or if I upgrade eslint from 10.2.0 to 10.2.1. Great.

ibc:
I cannot fix it at all even downgrading deps... I'm closing this PR.

ibc:
Ok, no idea how but I've tried the latest knip 6.11.0 and the issue went away. I generate the `package-lock.json` of mediasoup in macOS and now it contains those deps so `npm ci` doesn't complain on Linux anymore:

```json
"node_modules/@emnapi/core": {
	"version": "1.10.0",
	"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
	"integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
	"dev": true,
	"license": "MIT",
	"optional": true,
	"dependencies": {
		"@emnapi/wasi-threads": "1.2.1",
		"tslib": "^2.4.0"
	}
},
"node_modules/@emnapi/runtime": {
	"version": "1.10.0",
	"resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
	"integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
	"dev": true,
	"license": "MIT",
	"optional": true,
	"dependencies": {
		"tslib": "^2.4.0"
	}
},
```


```
npm ls @emnapi/core
mediasoup@3.19.22 /Users/ibc/src/mediasoup
├── @emnapi/core@1.10.0 extraneous
└─┬ @napi-rs/wasm-runtime@1.1.4 extraneous
  └── @emnapi/core@1.10.0 deduped
```

PR #1784, opened 2026-04-24
Worker: Add interfaces for Timer classes

ibc:
# Details

- Add `TimerHandleInterface`.
- Add `BackoffTimerHandleInterface`.

## Notes

This PR replaces https://github.com/versatica/mediasoup/pull/1058 which is too old and we never paid attention, but we should check if we can use things from it. What I don't like is that such a PR uses C macros for the code to behave in one way or another when in production builds or when running tests.

## TODO

- [ ] Pass somehow factory functions to classes that need to create a timer. Otherwise these classes are useless. For example libwebrtc/dcsctp passes a `callbacks` object to all classes and it includes a `createTimeout()` function.
  - _NOTE:_ We have the proper place for this: `RTC::Shared` instance that it's created in each `Worker`. Currently we pass it to many classes but not to all them. We should propagate it to all classes and include on `Shared` functions such as `CreateTimer()`, etc. But this should be part of a separate story.
- [ ] Implement mock timers. Specially useful for adding missing tests in new SCTP classes that require a timer.

### DIFF
--- a/worker/include/DepUsrSCTP.hpp
+++ b/worker/include/DepUsrSCTP.hpp
@@ -3,13 +3,13 @@
 
 #include "common.hpp"
 #include "RTC/SctpAssociation.hpp"
-#include "handles/TimerHandle.hpp"
+#include "handles/TimerHandleInterface.hpp"
 #include <absl/container/flat_hash_map.h>
 
 class DepUsrSCTP
 {
 private:
-	class Checker : public TimerHandle::Listener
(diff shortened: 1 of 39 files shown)


ibc:
Or perhaps we could complete the TODO items in a separate PR becase we will also probably need interfaces for `DepLibUV::GetTimeMs()`, etc.

jmillan (approved):
I think the factory stuff can be do in subsequent PRs.

Issue #1370, opened 2024-04-09
Use uint64_t for containers with RTP seq number as key


ibc:
As discussed in issue https://github.com/versatica/mediasoup/issues/1366 (this comment by @penguinol:  https://github.com/versatica/mediasoup/issues/1366#issuecomment-2045075046) we may use `uint64_t` in C++ maps and sets in which the key is a RTP seq number or similar. This would allow us to use our `SeqManager` compare functions into those containers without violating transitivity, i.e. `comp(a,b) && comp(b,c) -> comp(a,c)`, which currently we are violating (see the referenced issue for details).

libwebrtc uses `uint64_t` for seq nums, it takes more memory and cpu usage, but it's easier to understand and more reliable. `uint64_t` is big enougth to avoid wrap around. When receiving a packet, unwrap `uint16_t` seq num to `uint64_t` by `roc * 65535 + seq` (where `roc` is the cycle number), and then store and use the `uint64_t` value as in https://webrtc.googlesource.com/src/+/refs/heads/main/rtc_base/numerics/sequence_number_unwrapper.h. libwebrtc creates a `SeqNumUnwrapper` object for every stream which stores the `uint64_t` value of the last packet. When a new packet arrives, it calculates the distance between the seq of the packet and the last seq in `SeqNumUnwrapper` by doing something like our `SeqManager::SeqLowerThan()` to determine which cycle does the packet belongs to.

ibc:
**NOTE:** We have this now:

- https://github.com/versatica/mediasoup/blob/v3/worker/include/Utils/UnwrappedSequenceNumber.hpp
- https://github.com/versatica/mediasoup/blob/v3/worker/test/src/Utils/TestUnwrappedSequenceNumber.cpp

PR #1807, opened 2026-05-22
perf(SeqManager): Use std::vector rather than std::set

jmillan:
It increases performance greatly[*] thanks to the cache locality of the vector.

Scenarios where this is noticeable is when there are SeqManager drops due to, ie: Codec implementation or SimulcastConsumer dropping a packet.

[*]: Using mediasoup-demo with 1 audio|video producer and 100 audio|video consumers the CPU usage of SeqManager went down from >10% to <2%


worker/include/RTC/SeqManager.hpp
```diff
@@ -2,8 +2,9 @@
 #define RTC_SEQ_MANAGER_HPP
 
 #include "common.hpp"
-#include <limits> // std::numeric_limits
-#include <set>
+#include <algorithm> // for lower_bound, binary_search, distance, find_if
```
ibc: Please don't include algorithm. It's already in common.hpp

worker/src/RTC/SeqManager.cpp
```diff
@@ -63,9 +63,13 @@ namespace RTC
 		{
 			this->maxInput   = input;
 			this->maxDropped = input;
-			// Insert input in the last position.
-			// Explicitly insert at the end, which is more performant.
-			this->dropped.insert(this->dropped.end(), input);
+			// Insert input in sorted order, if not present.
+			SeqLowerThan seqLowerThan;
```
ibc: Blank lines after variable declaration please.

worker/src/RTC/SeqManager.cpp
```diff
@@ -63,9 +63,13 @@ namespace RTC
...
+			auto it = std::lower_bound(this->dropped.begin(), this->dropped.end(), input, seqLowerThan);
```
ibc: Also here

worker/src/RTC/SeqManager.cpp
```diff
@@ -122,12 +132,10 @@ namespace RTC
...
 		else
 		{
-			auto droppedCount = this->dropped.size();
-
-			// Get the first dropped input which is higher than or equal 'input'.
-			auto it = this->dropped.lower_bound(input);
-
-			droppedCount -= std::distance(it, this->dropped.end());
+			// Count dropped numbers less than input.
+			auto droppedCount = std::distance(
```
ibc, line 143: Also after this please. And probably add const here and in other ones above.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### NEXT
 
 - Node: Update TypeScript to v6 ([PR #1790](https://github.com/versatica/mediasoup/pull/1790)).
+- Worker: SeqManager, use std::vector rather than std::set ([1807](https://github.com/versatica/mediasoup/pull/1807)).
```
```suggestion
- Worker: `SeqManager`, use `std::vector` rather than `std::set` ([1807](https://github.com/versatica/mediasoup/pull/1807)).
```

ibc (approved):

penguinol:
Not sure whether this works properly in all situations.
It seems `std::binary_search` and `std::lower_bound` have different requirement to compare function.
https://www.cppreference.com/cpp/algorithm/lower_bound
> Unlike std::binary_search, std::lower_bound does not require operator< or comp to be asymmetric (i.e., a < b and b < a always have different results). In fact, it does not even require value < *iter or comp(value, *iter) to be well-formed for any iterator iter in [first, last).

https://en.cppreference.com/cpp/named_req/Compare

Maybe realted to #1370

ibc:
What about if we used the new Utils::UnwrappedSequenceNumber class that I borrowed from libwebrtc/dcsctp?

- https://github.com/versatica/mediasoup/blob/v3/worker/include/Utils/UnwrappedSequenceNumber.hpp
- https://github.com/versatica/mediasoup/blob/v3/worker/test/src/Utils/TestUnwrappedSequenceNumber.cpp

jmillan:
> What about if we used the new Utils::UnwrappedSequenceNumber class that I borrowed from libwebrtc/dcsctp?
> 
> - https://github.com/versatica/mediasoup/blob/v3/worker/include/Utils/UnwrappedSequenceNumber.hpp
> - https://github.com/versatica/mediasoup/blob/v3/worker/test/src/Utils/TestUnwrappedSequenceNumber.cpp

That is totally unrelated to this PR. Here I'm just using a more efficient container for the job after measuring the performance.

I believe there is an open issue with your suggestion which is another different topic than this one.

jmillan:
> Not sure whether this works properly in all situations.
> It seems `std::binary_search` and `std::lower_bound` have different requirement to compare function.
> https://www.cppreference.com/cpp/algorithm/lower_bound
> > Unlike std::binary_search, std::lower_bound does not require operator< or comp to be asymmetric (i.e., a < b and b < a always have different results). In fact, it does not even require value < *iter or comp(value, *iter) to be well-formed for any iterator iter in [first, last).
> 
> https://en.cppreference.com/cpp/named_req/Compare
> 
> Maybe realted to #1370
> 
> 
> 

Thanks, I'll check this info.

jmillan:
> Unlike std::binary_search, std::lower_bound does not require operator< or comp to be asymmetric (i.e., a < b and b < a always have different results). In fact, it does not even require value < *iter or comp(value, *iter) to be well-formed for any iterator iter in [first, last).

This this no trouble. In our case if `a<b` `b>a`, so I don't see any problem. Besides, al tests pass. Do you see a specific issue that is not present in current v3?

penguinol:
`std::set` uses a red-black tree while `std::vector` uses binary search, so they use different algorithms.
I am concerned about whether they behave the same with `SeqLowerThan`, which does not satisfy the C++ Compare standard.

jmillan:
> std::set uses a red-black tree while std::vector uses binary search, so they use different algorithms.
I am concerned about whether they behave the same with SeqLowerThan, which does not satisfy the C++ Compare standard.

We've plenty of tests, all passing: https://github.com/versatica/mediasoup/blob/v3/worker/test/src/RTC/TestSeqManager.cpp#L1

Both use the same `comp`. The difference is that binary_search requires it to be asymmetric (i.e., a < b and b < a always have different results), which is the case of `SeqLowerThan` anyways.

Any specific failing case/test on top of your mind?

jmillan:
Ok, so  IsLowerThan **not strictly asymmetric**: for some values, both `IsLowerThan(a, b)` and `IsLowerThan(b, a)` can be false (e.g., when `a == b` or when the difference is exactly half the range).

I'll avoid the binary_search.

jmillan:
@penguinol please check the[ latest commit]([8a204f0](https://github.com/versatica/mediasoup/pull/1807/commits/8a204f02bb49e2772a70b9ed911b5af3948d8a80)).

worker/src/RTC/SeqManager.cpp
ibc: Reminder: this auto should be const auto. clang-tidy should complain.

ibc (approved):

PR #1809, opened 2026-05-26
New releases for mediasoup NPM and mediasoup Rust crates

ibc:
# Details

- mediasoup Node 3.2.0
- mediasoup Rust crate 0.22.0
- mediasoup-sys Rust crate 0.12.0
- mediasoup-types Rust crate 0.4.0


ibc:
Optimistically merging because it should be ok.

PR #1823, opened 2026-06-06
TransportTuple: replace uint64_t hash with new `PortManager::PortRangeKeyHash`

### LINKED ISSUE #1815: TransportTuple: must replace uint64_t hash with new `PortManager::PortRangeKeyHash`
Same as done in PR https://github.com/versatica/mediasoup/pull/1812 (see related issue #1805) must be done in `TransportTuple` to replace its `uint64_t hash` member.

Read more detail sin https://github.com/versatica/mediasoup/pull/1812#issuecomment-4573063335.

CC @penguinol @999purple999 to be in sync about this.

penguinol:
# Details

- Implement `TransportTuple::TupleKey` and `TransportTuple::TupleKeyHash` Rewrite `TransportTuple::Compare()`.
- Related to issue #1805 and PR #1812.


worker/src/RTC/TransportTuple.cpp
```diff
@@ -143,54 +126,105 @@ namespace RTC
...
+		      reinterpret_cast<const sockaddr*>(std::addressof(this->localAddr)),
+		      reinterpret_cast<const sockaddr*>(std::addressof(other.localAddr))))
+		{
+			return false;
+		}
+
+		if (!Utils::IP::CompareAddresses(
+		      reinterpret_cast<const sockaddr*>(std::addressof(this->remoteAddr)),
+		      reinterpret_cast<const sockaddr*>(std::addressof(other.remoteAddr))))
+		{
```
ibc: I think we should compare this first, because when using WebRtcServer the local address is the same in all tuples.

ibc:
Please run "npm run lint" and "npm run format:node" to fix. And "npm run release:check" to validate everything.

penguinol:
`npm run lint:node` does not report any error.
```
H:\svn\OpenSource\mediasoup>npm run lint:node

> mediasoup@3.20.1 lint:node
> node npm-scripts.mjs lint:node

npm-scripts.mjs [INFO] [lint:node]
npm-scripts.mjs [INFO] [lint:node] lintNode()
npm-scripts.mjs [INFO] [lint:node] executeCmd(): eslint-config-prettier eslint.config.mjs
No rules that are unnecessary or conflict with Prettier were found.
npm-scripts.mjs [INFO] [lint:node] executeCmd(): eslint -c eslint.config.mjs --max-warnings 0 --ignore-pattern node/src/fbs eslint.config.mjs jest.config.mjs knip.config.mjs node/src npm-scripts.mjs worker/scripts
npm-scripts.mjs [INFO] [lint:node] executeCmd(): prettier --check CHANGELOG.md CONTRIBUTING.md README.md doc eslint.config.mjs jest.config.mjs knip.config.mjs node/src npm-scripts.mjs package.json tsconfig.json worker/scripts
Checking formatting...
All matched files use Prettier code style!
npm-scripts.mjs [INFO] [lint:node] executeCmd(): knip --config knip.config.mjs --treat-config-hints-as-errors
✂️  Excellent, Knip found no issues.
```

ibc:
It's failing in CI: https://github.com/versatica/mediasoup/actions/runs/27118410859/job/80029913564#step:10:1

Can you please merge v3 branch?

node/src/WebRtcServer.ts
ibc, line 1: I'm confused, why does this PR diff show **so many changes** in `WebRtcServer.ts`???

rust/src/webrtc_server.rs
ibc, line 1: Also here, why so many changes?

worker/include/RTC/TransportTuple.hpp
```diff
@@ -23,28 +23,55 @@ namespace RTC
...
+			bool operator==(const TupleKey& other) const noexcept;
+			bool operator!=(const TupleKey& other) const noexcept
+			{
+				return !(*this == other);
+			};
+
+		private:
+			Protocol protocol{ Protocol::UDP };
+			sockaddr_storage localAddr{};
+			sockaddr_storage remoteAddr{};
```
ibc: Let's please follow code changes in the former PR https://github.com/versatica/mediasoup/pull/1812/changes. For example, no need to add `{}` in these members.

worker/include/RTC/TransportTuple.hpp
```diff
@@ -74,7 +101,31 @@ namespace RTC
...
+				}
+			}
+			else
+			{
+				if (this->tcpConnection != tuple->tcpConnection)
+				{
+					return false;
+				}
+			}
+			return true;
```
ibc, line 134:
```suggestion
			}

			return true;
```

worker/include/RTC/TransportTuple.hpp
```diff
@@ -74,7 +101,31 @@ namespace RTC
...
+					return false;
+				}
+			}
+			else
+			{
+				if (this->tcpConnection != tuple->tcpConnection)
+				{
+					return false;
+				}
+			}
```
ibc: Can we please use `switch (this->protocol)` instead? Also, instead of several `if (xxx) conditions, let's please use `if ()` and `else if ()` following style in https://github.com/versatica/mediasoup/pull/1812/changes#diff-4d144a346d9fd1a104b09481b3bdb84ac06625f500a1485357fcc3e03a3ba021R743.

worker/include/Utils.hpp
```diff
@@ -562,6 +562,15 @@ namespace Utils
...
 	};
 
+	class Hash
+	{
+	public:
+		static void Combine(size_t& seed, size_t value)
+		{
+			seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+		};
+	};
```
ibc: Can we have some jsdoc style documentation above this function please? What does this `Combine()` function do? Why `0x9e3779b9`? Why `seed <<6` etc?

ibc: ok, I see we already had a similar function without any comment XD

worker/src/RTC/TransportTuple.cpp
```diff
@@ -143,54 +126,102 @@ namespace RTC
...
 	}
 
-	void TransportTuple::GenerateHash()
+	bool TransportTuple::TupleKey::operator==(const TupleKey& other) const noexcept
 	{
 		MS_TRACE();
 
-		const auto* localSockAddr  = GetLocalAddress();
-		const auto* remoteSockAddr = GetRemoteAddress();
+		if (this->protocol != other.protocol)
```
ibc, line 137: same here. Let's please use `if` and `else if` etc, without blank lines between them. Same as in https://github.com/versatica/mediasoup/pull/1812/changes#diff-4d144a346d9fd1a104b09481b3bdb84ac06625f500a1485357fcc3e03a3ba021R743

node/src/WebRtcServer.ts
penguinol: My editor automatically‌ converted file format to CRLF, already fixed.

worker/include/Utils.hpp
```diff
@@ -562,6 +562,16 @@ namespace Utils
 		uint32_t offset{ 0 };
 	};
 
+	class Hash
+	{
+	public:
+		static void Combine(size_t& seed, size_t value)
+		{
+			// The implementation is the same as boost::hash_combine
```
ibc: Cosmetic: add () and final dot to the comment.

ibc (commented):
Thanks. I am on vacation so probably won't be able to review until next Monday but everything looks good so far.

worker/src/RTC/WebRtcServer.cpp
```diff
@@ -562,29 +544,33 @@ namespace RTC
 	{
 		MS_TRACE();
 
-		if (this->mapTupleWebRtcTransport.find(tuple->hash) != this->mapTupleWebRtcTransport.end())
+		auto tupleKey = tuple->GetTupleKey();
+
+		if (this->mapTupleWebRtcTransport.contains(tupleKey))
 		{
 			MS_WARN_TAG(ice, "tuple hash already exists in the table");
```
jmillan: ? 

```suggestion
			MS_WARN_TAG(ice, "tuple key already exists in the table");
```

worker/src/RTC/WebRtcServer.cpp
```diff
@@ -562,29 +544,33 @@ namespace RTC
...
 	  RTC::WebRtcTransport* /*webRtcTransport*/, RTC::TransportTuple* tuple)
 	{
 		MS_TRACE();
 
-		if (this->mapTupleWebRtcTransport.find(tuple->hash) == this->mapTupleWebRtcTransport.end())
+		auto it = this->mapTupleWebRtcTransport.find(tuple->GetTupleKey());
+
+		if (it == this->mapTupleWebRtcTransport.end())
 		{
 			MS_DEBUG_TAG(ice, "tuple hash not found in the table");
```
jmillan: ?

```suggestion
			MS_DEBUG_TAG(ice, "tuple key not found in the table");
```

jmillan (approved):

worker/include/RTC/TransportTuple.hpp
```diff
@@ -147,12 +193,20 @@ namespace RTC
 			}
 		}
 
-	private:
-		void SetHash();
-		void GenerateHash();
-
-	public:
-		uint64_t hash{ 0u };
+		TupleKey GetTupleKey() const
```
ibc: Just a question: Why are not we keeping a computed `TupleKey` private member in `TransportTuple` class instead of having to compute it every time by calling `GetTupleKey()`?

worker/src/RTC/TransportTuple.cpp
```diff
@@ -143,54 +126,100 @@ namespace RTC
...
-				buffer[idx++] = port & 0xFF;
+				const auto* localIn6 = reinterpret_cast<const sockaddr_in6*>(std::addressof(key.localAddr));
+				const auto* remoteIn6 = reinterpret_cast<const sockaddr_in6*>(std::addressof(key.remoteAddr));
+
+				const auto* addr = localIn6->sin6_addr.s6_addr;
+
+				uint64_t hi;
+				uint64_t lo;
+				std::memcpy(std::addressof(hi), addr, sizeof(uint64_t));
+				std::memcpy(std::addressof(lo), addr + sizeof(uint64_t), sizeof(uint64_t));
```
ibc:
```suggestion
				uint64_t hi;
				uint64_t lo;

				std::memcpy(std::addressof(hi), addr, sizeof(uint64_t));
				std::memcpy(std::addressof(lo), addr + sizeof(uint64_t), sizeof(uint64_t));
```

worker/src/RTC/TransportTuple.cpp
```diff
@@ -3,7 +3,6 @@
 
 #include "RTC/TransportTuple.hpp"
 #include "Logger.hpp"
-#include <cstring> // std::memcpy()
```
ibc, line 6: Why is this include removed? We are using `std::memcpy()` in lines 193, 104, 204 and 205 below.

worker/src/RTC/WebRtcServer.cpp
```diff
@@ -450,7 +432,7 @@ namespace RTC
 		}
 
 		// First try doing lookup in the tuples table.
-		auto it1 = this->mapTupleWebRtcTransport.find(tuple->hash);
+		auto it1 = this->mapTupleWebRtcTransport.find(tuple->GetTupleKey());
```
ibc, line 440: So as said above, here we are computing the `TupleKey` by calling `tuple->GetTupleKey()`. Why don't we store it in `TransportTuple` as a private member?

ibc (commented):
Some questions made. Also, could you please introduce an item in the CHANGELOG similar to the recent one?:

```
- `PortManager`: Replace `uint64_t` hash token with exact-tuple `PortRangeKey` ([PR #1812](https://github.com/versatica/mediasoup/pull/1812), by @999purple999 and @penguinol).
```

worker/include/RTC/TransportTuple.hpp
```diff
@@ -74,7 +107,26 @@ namespace RTC
...
+
+			switch (this->protocol)
+			{
+				case Protocol::UDP:
+				{
+					return (
+					  this->udpSocket == tuple->udpSocket &&
+					  Utils::IP::CompareAddresses(this->udpRemoteAddr, tuple->udpRemoteAddr));
+				}
+				case Protocol::TCP:
```
ibc, line 128:
```suggestion
				}

				case Protocol::TCP:
```

worker/include/RTC/WebRtcServer.hpp
```diff
@@ -119,7 +120,8 @@ namespace RTC
 		// Map of WebRtcTransports indexed by local ICE usernameFragment.
 		ankerl::unordered_dense::map<std::string, RTC::WebRtcTransport*> mapLocalIceUsernameFragmentWebRtcTransport;
 		// Map of WebRtcTransports indexed by TransportTuple.hash.
-		ankerl::unordered_dense::map<uint64_t, RTC::WebRtcTransport*> mapTupleWebRtcTransport;
+		ankerl::unordered_dense::map<TransportTuple::TupleKey, RTC::WebRtcTransport*, TransportTuple::TupleKeyHash>
```
ibc: Please always add the `RTC::` namespace prefix con using symbols outside the current class.
```suggestion
		ankerl::unordered_dense::map<RTC::TransportTuple::TupleKey, RTC::WebRtcTransport*, RTC::TransportTuple::TupleKeyHash>
```

ibc:
I'm worried about this last commit "Save `sockaddr*` instead of `sockaddr_storage` in `TupleKey`": https://github.com/versatica/mediasoup/pull/1823/changes/fb424cda858d90792910d8ecbe36fd2897e7fa54

We definitely **need** to store the full address and not just a pointer to it because such as pointer may be later invalidated. This happens specially in TCP when the TCP connection is closed. If there is a `TransportTuple` associated to that TCP connection stored somewhere (in `IceServer`, in some `XxxxTransport`, etc) it will now contain an invalid/deallocated `sockaddr*` that points to garbage. I'm not 100% sure about this. Can you check and confirm please? But I'm 100% sure that in v3 branch we store/clone the `sockaddr*` into a `sockaddr_storage` member for that exact reason.

UPDATE: Actually I see that in v3 we only store/clone the address when it's UDP, in `StoreUdpRemoteAddress()` method which already exists in this PR, so I think I maybe wrong in my above assumption.

worker/include/RTC/TransportTuple.hpp
```diff
@@ -53,6 +80,12 @@ namespace RTC
 			if (protocol == TransportTuple::Protocol::UDP)
 			{
 				StoreUdpRemoteAddress();
+				tupleKey = TupleKey(protocol, udpSocket->GetLocalAddress(), udpRemoteAddr);
+			}
+			else
+			{
+				tupleKey =
```
ibc: Please always use `this->` for class instance members (that are not methods).
```suggestion
				this->tupleKey = TupleKey(protocol, udpSocket->GetLocalAddress(), udpRemoteAddr);
			}
			else
			{
				this->tupleKey =
```

worker/include/RTC/TransportTuple.hpp
```diff
@@ -23,28 +23,55 @@ namespace RTC
...
 		};
 
+		class TupleKey
+		{
+		private:
+			friend class TransportTuple;
+			friend class TupleKeyHash;
+
+		public:
+			TupleKey() = default;
```
ibc, line 33: Why is this default constructor needed?

worker/include/RTC/TransportTuple.hpp
```diff
@@ -53,6 +80,12 @@ namespace RTC
 			if (protocol == TransportTuple::Protocol::UDP)
 			{
 				StoreUdpRemoteAddress();
+				tupleKey = TupleKey(protocol, udpSocket->GetLocalAddress(), udpRemoteAddr);
```
ibc: Here we are passing `udpRemoteAddr` to the `TupleKey` so the `TupleKey` will store the original pointer given to the `TransportTuple` constructor. And this is wrong because such a pointer maybe deallocated at any time. Here we should pass the **locally cloned/stored** `this->udpRemoteAddr` which was created in the line above `StoreUdpRemoteAddress()`. Right?

worker/src/RTC/WebRtcServer.cpp
```diff
@@ -303,26 +303,8 @@ namespace RTC
...
-
 		return FBS::WebRtcServer::CreateDumpResponseDirect(
-		  builder,
-		  this->id.c_str(),
-		  &udpSockets,
-		  &tcpServers,
-		  &webRtcTransportIds,
-		  &localIceUsernameFragments,
-		  &tupleHashes);
+		  builder, this->id.c_str(), &udpSockets, &tcpServers, &webRtcTransportIds, &localIceUsernameFragments);
```
ibc: Please don't use `&foo`. Use `std::addressof(foo)` instead (everywhere).

ibc (changes requested):

worker/src/RTC/WebRtcServer.cpp
ibc: NOTE: I know that `&` was already there in v3 branch...

worker/include/RTC/TransportTuple.hpp
jmillan: But we are storing `udpRemoteAddr` in `TransportTuple` anyway, right? Meaning if it was invalidated we would already have problems.

worker/include/Utils.hpp
```diff
@@ -562,6 +562,16 @@ namespace Utils
...
 
+	class Hash
+	{
+	public:
+		static void Combine(size_t& seed, size_t value)
+		{
+			// The implementation is the same as boost::hash_combine().
+			seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+		};
+	};
```
ibc: Please move this to the `Utils::Crypto` class (no need to add a new class `Hash` in `Utils`) and rename the static method from `Combine()` to `HashCombine()`.

ibc (dismissed):

worker/include/RTC/TransportTuple.hpp
ibc: > But we are storing `udpRemoteAddr` in `TransportTuple` anyway, right?

Nope, we are not storing `udpRemoteAddr` as it is in `TransportTuple`.

jmillan: As far as I see, we are:

https://github.com/penguinol/mediasoup/blob/5be14553e1b593ac5a0a40d1124309e8a5a89894/worker/include/RTC/TransportTuple.hpp#L60
https://github.com/penguinol/mediasoup/blob/5be14553e1b593ac5a0a40d1124309e8a5a89894/worker/include/RTC/TransportTuple.hpp#L75

ibc: See the call to `StoreUdpremoteAddress()` and what it does in https://github.com/penguinol/mediasoup/blob/5be14553e1b593ac5a0a40d1124309e8a5a89894/worker/include/RTC/TransportTuple.hpp#L73C3-L87C7.

Yes, we only store it when **copying** a tuple. I don't remember why.

jmillan: We also store it as it is in regular constructor:

https://github.com/penguinol/mediasoup/blob/5be14553e1b593ac5a0a40d1124309e8a5a89894/worker/include/RTC/TransportTuple.hpp#L60

ibc: What does it mean "also store as it is"? Please read my previous comment to be in sync.

BTW the only place where we use the copy constructor (in which we invoke `StoreUdpRemoteAddress()` method to **store/clone** the given sockaddr) is `PlainTransport.cpp`. Why don't we need to store it in `IceServer`? I don't remeber but I'm 100% sure there is a reason for it and I spent long time on it plus I fixed related crashes in the past.

worker/include/RTC/TransportTuple.hpp
penguinol: It's because in `explicit TransportTuple(const TransportTuple* tuple)` tuplekey will be constructed with default constructor fisrt, and then be replaced after `StoreUdpRemoteAddress` is invoked.

worker/include/RTC/TransportTuple.hpp
ibc: Please don't forget this.

ibc:
@penguinol I am gonna investigate the cases where we need to store/clone the UDP sockaddr to remember why it was the way it was and verify if current PR does it good. May I have write permission in this PR please?

worker/include/RTC/TransportTuple.hpp
penguinol: > Here we are passing `udpRemoteAddr` to the `TupleKey` so the `TupleKey` will store the original pointer given to the `TransportTuple` constructor. And this is wrong because such a pointer maybe deallocated at any time. Here we should pass the **locally cloned/stored** `this->udpRemoteAddr` which was created in the line above `StoreUdpRemoteAddress()`. Right?

Yes, here `udpRemoteAddr` means `this->udpRemoteAddr`, i forgot to add `this->`, input params is `tuple->udpRemoteAddr`.

ibc: Notice that if you don't add this-> then it points to the original argument given to the method which is not the same as the stored in the class (because both have the same name).

ibc:
> @penguinol I am gonna investigate the cases where we need to store/clone the UDP sockaddr to remember why it was the way it was and verify if current PR does it good. May I have write permission in this PR please?

NOTE: I forgot that `IceServer.cpp` clones the UDP remote sockaddr when it stores the tuple in `AddTuple()` method:

```c++
// If it is UDP then we must store the remote address (until now it is
// just a pointer that will be freed soon).
if (storedTuple->GetProtocol() == TransportTuple::Protocol::UDP)
{
	storedTuple->StoreUdpRemoteAddress();
}
```

penguinol:
> @penguinol I am gonna investigate the cases where we need to store/clone the UDP sockaddr to remember why it was the way it was and verify if current PR does it good. May I have write permission in this PR please?

Yes, do i need to set something, like permission?

penguinol:
Well, i see it, `Allow edits and access to secrets by maintainers` is checked.

penguinol:
`TcpConnectionHandle` will be destroyed after invoking `UserOnTcpConnectionClosed` , and `TransportTuple` will be removed  in `UserOnTcpConnectionClosed`.
https://github.com/versatica/mediasoup/blob/ed3a2cf333182f3a88301a468457fcda637e57fe/worker/src/handles/TcpServerHandle.cpp#L267-L281

ibc:
Ok, so here my investigation (this is about **v3** branch):

## Summary of the `TransportTuple` lifetime

The design is correct, there is no use-after-free. A `TransportTuple` never owns its socket, so safety relies on destruction/removal ordering and that ordering holds everywhere.

**`this->udpRemoteAddr` / `this->udpRemoteAddrStorage` members:**
- Normal UDP constructor does not clone (points to external memory).
- Copy constructor calls `StoreUdpRemoteAddress()` when in UDP, so it repoints to own storage.
- The implicit copy constructor (used by `std::list::push_front()` in `IceServer`) copies the storage bytes but leaves `udpRemoteAddr` dangling at the source. Every persisting site fixes this manually: `IceServer::AddTuple()` calls `StoreUdpRemoteAddress()` after `std::list::push_front()`, `PlainTransport`/`PipeTransport` use the copy constructor which does it internally.
- `IceServer::tuples` is `std::list` (not vector) so node addresses and `selectedTuple` member stay stable across insert/erase.

**`this->udpSocket` / `this->tcpConnection` access after close:**
- `PlainTransport`/`PipeTransport`: `this->tuple` and `this->udpSocket` are members of the same transport. Destructor deletes the socket then the tuple, and `~TransportTuple()` never dereferences the socket.
- `WebRtcTransport` destructor: deletes `this->iceServer` member (which closes TCP tuples) before `udpSockets`/`tcpServers` mebers, so connections are alive when touched.
- TCP close path: `OnTcpConnectionClosed()` -> `RemoveTuple()` (synchronously removes the stored tuple) -> then delete `tcpConnection`. `TriggerClose()` is idempotent (closed guard), so the re-entrant `CloseTcpConnection()` is a no-op.
- `WebRtcServer`: Destructor closes all transports first, then its UDP sockets and TCP servers (sockets must stay open for DTLS Close Alert), so the server always outlives the tuples referencing it. On transport closure the hash is erased from `this->mapTupleWebRtcTransport` before `CloseTcpConnection()`, so the later `OnRtcTcpConnectionClosed()` finds nothing.
- `this->mapTupleWebRtcTransport` stores only hash (uint64), no lifetime issue. Consent timer only removes tuples, never sends.

**Only fragility:** Correctness depends on keeping these destructor/removal orders. Reordering the deletes in `~WebRtcTransport`, `~WebRtcServer` or `TcpServerHandle::OnTcpConnectionClosed()` would introduce a use-after-free, but it's ok in v3 branch.

## Regarding this PR

We must ensure that this PR doesn't break the weak safety above.

worker/include/RTC/TransportTuple.hpp
jmillan: > What does it mean "also store as it is"?

I mean that it stores the raw pointer, which  may be dangerous since we are not controlling 100% the ownership or lifetime of it, and C++ does not expose a way to do it.

So IMO the safest way would be to store a copy of if in every `TransportTuple` constructor. Unless we are 100% that the tuple is temporal and will not outlive the current libuv iteration.

ibc:
I've found a regression in this PR. Details in a few minutes here.

worker/include/RTC/TransportTuple.hpp
jmillan: > Unless we are 100% that the tuple is temporal and will not outlive the current libuv iteration.

In which case rather than passing a pointer we should std::move an instance to clearly express it's lifetime.

jmillan: Reading @ibc's comments [here](https://github.com/versatica/mediasoup/pull/1823#issuecomment-4727940723) now.

ibc:
I found a real bug introduced by this PR:

### use-after-free in the UDP + WebRtcServer

The PR replaces `hash` (a `uint64_t` stored by value) with a `TupleKey` that holds pointers to `localAddr`/`remoteAddr`, and both `TupleKey::operator==()` and `TupleKeyHash::operator()` (`TransportTuple.cpp`lines  133-225) dereference those pointers. That turns any dangling pointer in the key into a UAF on every map operation.

**The bug flow:**

`TransportTuple::StoreUdpRemoteAddress()` repoints `this->udpRemoteAddr` to its own `this->udpRemoteAddrStorage` but it does not update `this->tupleKey`. The explicit copy constructor compensates by rebuilding `this->tupleKey` right afterwards (`TransportTuple.hp` lines 83-84), but `IceServer::AddTuple()` does not use that constructor:

```c++
// IMPLICIT copy: copies tupleKey as it is:
this->tuples.push_front(*tuple);              
storedTuple = std::addressof(*this->tuples.begin());

if (UDP) {
  // fixes udpRemoteAddr, does NOT fix tupleKey.
  storedTuple->StoreUdpRemoteAddress();     
}
```

Result for a stored UDP tuple: `this->udpRemoteAddr` ends up correct (points to its own storage), but `this->tupleKey.remoteAddr` still points to the original `remoteAddr` which is the `sockaddr*` that libuv passes into `uv_udp_recv_cb`, which is a `sockaddr_storage` on libuv's stack, valid only during the callback (`UdpSocketHandle.cpp` lines 351-376).

That `storedTuple` is the one that reaches `WebRtcServer::OnWebRtcTransportTransportTupleAdded()` where it is stored as the map key:

```c++
// WebRtcServer.cpp line 561
this->mapTupleWebRtcTransport[tupleKey] = webRtcTransport;
```

In other words, the `WebRtcServer` map key holds a pointer into libuv's transient stack. As soon as the receive callback returns, that pointer dangles. From then on:

- `std::list::find(tuple->GetTupleKey())` in `OnStunDataReceived()`/`OnNonStunDataReceived()`/`OnRtcTcpConnectionClosed()` (in `WebRtcServer.cpp`) compares against that key so it reads foreign/reused stack memory.
- Any `std::list::insert()`/`erase()` from any transport triggers a map rehash, which re-hashes that key so it dereferences the dangling pointer -> **use-after-free / map corruption**.

STUN/RTP packets from an already-established peer will start failing as "unknown tuple", routing to the wrong transport, or crashing.

**Scope of the bug:**

- UDP only and `WebRtcServer` mode only. For TCP, `this->tupleKey` points to `tcpConnection->GetLocalAddress()/GetPeerAddress()` (stable while the connection lives) and the TCP copy constructor rebuilds it, so it's fine. In `WebRtcTransport` without a `WebRtcServer` there is no map using `GetTupleKey()` so the stale key stays latent but harmless (`Compare()` uses `udpRemoteAddr`, not `tupleKey`).
- It's a new bug from the PR: in `v3`branch the key was a `uint64_t` by value and nothing was ever dereferenced.

**Proposed fix:**

Make `StoreUdpRemoteAddress()` rebuild `this->tupleKey` too, since it has access to `this->udpSocket` and to the already-corrected `this->udpRemoteAddr`:

```c++
void StoreUdpRemoteAddress()
{
    this->udpRemoteAddrStorage = Utils::IP::CopyAddress(this->udpRemoteAddr);
    this->udpRemoteAddr =
      reinterpret_cast<struct sockaddr*>(std::addressof(this->udpRemoteAddrStorage));
    this->tupleKey =
      TupleKey(this->protocol, this->udpSocket->GetLocalAddress(), this->udpRemoteAddr);
}
```

This fixes the `IceServer::AddTuple()` path, and lines 83-84 of the copy constructor become redundant so must be removed (no need to reassign `this->tupleKey =` again since `StoreUdpRemoteAddress()` already did it).

worker/include/RTC/TransportTuple.hpp
ibc: > I mean that it stores the raw pointer, which may be dangerous since we are not controlling 100% the ownership or lifetime of it, and C++ does not expose a way to do it.
> 
> So IMO the safest way would be to store a copy of if in every `TransportTuple` constructor. Unless we are 100% that the tuple is temporal and will not outlive the current libuv iteration.

While I agree the design is "weak", it's optimized. And verified, see https://github.com/versatica/mediasoup/pull/1823#issuecomment-4727940723.



> > Unless we are 100% that the tuple is temporal and will not outlive the current libuv iteration.
> 
> In which case rather than passing a pointer we should std::move an instance to clearly express it's lifetime.

We cannot use `std::move` here. First of all because this is a raw C pointer and there is no `std::move` semantics for raw C pointers. And secondly because we don't want to transfer ownership of the `sockaddr`. The `UdpSocket` and `TcpConnection` classes **need** to access those private members to receive and send packets.

ibc:
@jmillan here an explanation about why we do NOT want to clone the UDP `sockaddr` in every `TransportTuple`:

This applies to both v3 and the current branch.

**Why the non-copy constructors don't need to clone `this->udpRemoteAddr` and `this->udpRemoteAddrStorage`?**

There are two ways the normal UDP constructor `TransportTuple(RTC::UdpSocket*, const struct sockaddr*)` is used, and neither needs cloning:

1. Transient tuple in the stack: The common case. On every received UDP packet, `OnUdpSocketPacketReceived()` builds a `RTC::TransportTuple tuple(socket, remoteAddr)` on the stack, where `remoteAddr` is libuv's transient data. That tuple lives only during the callback (used for find/compare, then discarded). The pointer is valid for the tuple's whole short life, so cloning would be wasted work.

2. Persisted tuple. In `PlainTransport`/`PipeTransport` in non-comedia mode:
   ```c++
   this->tuple = new RTC::TransportTuple(
     this->udpSocket, reinterpret_cast<struct sockaddr*>(&this->remoteAddrStorage));
   ```

Here `udpRemoteAddr` points to `this->remoteAddrStorage`, a member of the transport itself that outlives the tuple. No clone needed because the pointed-to memory is already stable.

**Why the copy constructor does clone (it calls `StoreUdpRemoteAddress()`)?**

`TransportTuple(const TransportTuple*)` is used to persist a tuple copied from a transient one, e.g. `PlainTransport` in comedia mode where `tuple` is the stack tuple from the libuv callback:

   ```c++
   this->tuple = new RTC::TransportTuple(tuple);
   ```

The source's `udpRemoteAddr` points to libuv's transient buffer, so the copy constructor must clone it via `StoreUdpRemoteAddress()`.

Regarding `IceServer::AddTuple()`, `IceServer` does not use the explicit copy constructor. It uses `this->tuples.push_front(*tuple)`, which invokes the implicit copy constructor `TransportTuple(const TransportTuple&)`, which does not clone. That's why `AddTuple()` calls `tuple->StoreUdpRemoteAddress()` manually right after.

So there are two "persist a tuple" routes:

1. `new TransportTuple(const TransportTuple*)`, it clones inside the constructor itself (`PlainTransport/PipeTransport`).
2. `push_front(*tuple)` + explicit `StoreUdpRemoteAddress()`in `IceServer`.

And route 2 is exactly where the current PR has the bug: `StoreUdpRemoteAddress()` doesn't rebuild `this->tupleKey` in `TransportTuple`, whereas route 1's copy constructor does.

**What if all constructors called `StoreUdpRemoteAddress()`?**

It would be correct but less efficient. If the normal UDP constructor always cloned, every transient case tuple would do a `Utils::IP::CopyAddress()` (a memcpy, etc) on every incoming UDP datagram for tuples that are almost always discarded immediately. Pure overhead.

jmillan:
@ibc, yes, I see. Basically whoever needs to store the tuple it needs to make sure to make a copy of the remote address , in order to own it, by the means indicated.

ibc:
@penguinol I've fixed the regression in this commit: https://github.com/versatica/mediasoup/pull/1823/changes/4c927e62dcb8c2a2957642ce6d940d08450643e5

ibc:
@penguinol I've pushed an improvement:

https://github.com/versatica/mediasoup/pull/1823/commits/18b476924498c2f139dbc71fe7b58be4a3e3750f

> Identify TupleKey local endpoint by socket/connection pointer
>
> `TransportTuple::TupleKey` now identifies the local endpoint by the `UdpSocket`/`TcpConnection` pointer instead of by the local address value. This matches `TransportTuple::Compare()` internal logic and removes the asymmetry between the two
compare criteria.
>
> It is also more efficient: comparing and hashing an 8-byte pointer avoids dereferencing `GetLocalAddress()` and processing a full `sockaddr` on every map operation. For TCP the connection pointer alone identifies the whole tuple, so the remote address is not needed.

ibc:
@penguinol anything you want to check or add in this PR? Is it totally ready to merge?

penguinol:
> @penguinol anything you want to check or add in this PR? Is it totally ready to merge?

I think it's ready to ship

ibc:
Let's merge.

ibc (approved):

ibc:
Thanks a lot @penguinol!

PR #1838, opened 2026-06-17
Add `rust-scripts.mjs`

ibc:
# Details

- Add `rust-scripts.mjs` script which "release:check" and "release" actions.
- They are controlled via `npm` scripts:: `npm run release:rust:check` and `npm run release:rust`.
- They are similar to the ones in `npm-scripts.mjs` and run lint, test, doc, changelog file, etc.
- The "release" one also publishes the modified crates.
- The "release" script creates not only a `rust-x.x.x` tag in GitHub but also a `rust-x.x.x` release.

## Bonus tracks

- Rename `KEEP_BUILD_ARTIFACTS` env (using when building mediasoup Rust) to `MEDIASOUP_LOCAL_DEV` so both Node and Rust use the same env to avoid rebuilding everything from scratch when in local development.
- `npm-scripts.mjs`: Make "release:check" **also** verify that `CHANGELOG.md` is in sync with version in `package.json`.

### DIFF
--- a/.github/workflows/mediasoup-rust.yaml
+++ b/.github/workflows/mediasoup-rust.yaml
@@ -51,7 +51,7 @@ jobs:
     timeout-minutes: 60
 
     env:
-      KEEP_BUILD_ARTIFACTS: '1'
+      MEDIASOUP_LOCAL_DEV: 'true'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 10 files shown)


jmillan (approved):

ibc:
TODO: Make "release" script verify that Cargo.lock is in sync (so "cargo build" was run already). Also check that things are pushed and not only committed!

ibc:
> TODO: Make "release" script verify that Cargo.lock is in sync (so "cargo build" was run already). Also check that things are pushed and not only committed!

Done.

PR #1846, opened 2026-06-25
Handle new STUN "NOMINATION" attribute `0x0030`

### LINKED ISSUE #1766: STUN NOMINATION attribute code should not be `0xC001`
# Context

STUN `NOMINATION` attribute is not standard. It's used by libwebrtc (and then others) and uses code `0xC001` (in mediasoup too).

# Problem

However the value `0xC001` has already been assigned by IANA to `ENF-FLOW-DESCRIPTION` (see https://www.iana.org/assignments/stun-parameters/stun-parameters.xml), so this means conflict.

# More info

- Issue reported in libwebrtc: https://issues.webrtc.org/issues/496629058 (we must monitor it).
- For example, Pion uses `0x0030`, which is currently unused.

ibc:
# Details

- The STUN "NOMINATION" attribute is not yet a standard specification. In libwebrtc native code, `STUN_ATTR_NOMINATION` is defined as `0xC001`. However that value is not defined in any RFC/draft, not even in the draft that defines the NOMINATION attribute. See the old draft https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination.
- So there is an ongoing effort to standardize it (see https://issues.webrtc.org/issues/496629058 and the ongoing draft https://juberti.github.io/draft-renomination/draft-thatcher-tsvwg-renomination.html).
- The ongoing draft defines the "NOMINATION" attribute with value `0x0030`.
- In this PR we make `IceServer` handle both, the old value `0xC001` and the new one `0x0030`.
- Fixes #1766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: Handle new STUN "NOMINATION`" attribute `0x0030` ([XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
+- Worker: Handle new STUN "NOMINATION" attribute `0x0030` ([1846](https://github.com/versatica/mediasoup/pull/1846)).
 
 ### 3.20.9
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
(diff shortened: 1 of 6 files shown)


ibc:
@penguinol, you maybe interested in this PR.

Issue #1359, opened 2024-03-26
Must join all XxxConsumer classes into a single one


ibc:
### Details

- In C++ we currently have a parent abstract `Consumer` class and then `SimpleConsumer`, `SimulcastConsumer`, `SvcConsumer` and `PipeConsumer` that inherit from `Consumer`.
- Those classes are intended to handle different producing scenarios and associated codecs.
- For example a video transmission could use VP8 in single stream mode so it would use `SingleConsumer`.
- Or with N streams in simulcast mode so it would use `SimulcastConsumer`.

### Problems

- Those classes do not represent real scenarios. 
- For example, a producer may perfectly use VP8 with a single stream (so this is not simulcast) and temporal layers. We currently manage it in `SimulcastConsumer` since it's the one ready for VP8 codec, but this is SVC.
- VP9 is supposed to be SVC, however in libwebrtc now it's possible to produce real simulcast using VP9 (probably also with temporal layers): https://groups.google.com/g/discuss-webrtc/c/-QQ3pxrl-fw

### Proposal

- Have a single `Consumer` class ready to deal with all use cases and all supported codecs.
- This would also help in the future when we implement the ability to switch from a producer to another in the same consumer.
- We should also reconsider those `producer.type` and `consumer.type` ("single", "simulcast", "svc" and "pipe") since they don't make any sense. However this would introduce breaking changes. Let's see how to do it.

jmillan:
I think we can already close this issue.

PR #1851, opened 2026-07-02
Worker: Enable SVC for VP8 and H264

jmillan:
Related to #1359

Rather than relying on Producer's type, do calculate the type based on `consumableRtpEncodings`.

Adapted VP8 and H264 codecs accordingly.

### How to test it with mediasoup-demo:

1. **&forceVP8=true&numWebcamSimulcastStreams=1&webcamScalabilityMode=L1T3**
   This will create a single encoding with L1T3. It will use SVC.
2. **&forceVP8=true&numWebcamSimulcastStreams=3&webcamScalabilityMode=L1T3**
   This will create three encodings with L1T3. It will use Simulcast.

NOTE: this PR does not pretend to handle #1359 entirely but just what indicates. It uses no SVC for VP8 and H264 when there is a single stream with temporal layers.

### DIFF
--- a/worker/include/RTC/RTP/Codecs/Tools.hpp
+++ b/worker/include/RTC/RTP/Codecs/Tools.hpp
@@ -159,7 +159,9 @@ namespace RTC
 								{
 									switch (mimeType.subtype)
 									{
+										case RTC::RtpCodecMimeType::Subtype::VP8:
 										case RTC::RtpCodecMimeType::Subtype::VP9:
+										case RTC::RtpCodecMimeType::Subtype::H264:
 										case RTC::RtpCodecMimeType::Subtype::AV1:
 										{
 											return true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+Worker: Enable SVC for VP8 and H264 ([PR #1851](https://github.com/versatica/mediasoup/pull/1851)).
+
 ### 3.20.10
 
 - Worker: Handle new STUN "NOMINATION" attribute `0x0030` ([PR #1846](https://github.com/versatica/mediasoup/pull/1846)).
--- a/node/src/test/test-Consumer.ts
+++ b/node/src/test/test-Consumer.ts
@@ -1163,6 +1163,132 @@ test('transport.consume() for a simulcast producer honors preferredLayers', asyn
 	});
 }, 2000);
 
+test('transport.consume() from a single-encoding no temporal layer producer is of type simple', async () => {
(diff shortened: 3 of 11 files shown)


worker/include/RTC/RTP/Codecs/Tools.hpp
ibc, line 135: But VP8 and H264 still exist here. Maybe I don't understand what this PR does. Does it change something for real? Or is it just a "preparation" work to make changes later?

Ok, if the Producer uses VP8 with 3 encodings and `scalabilityMode` in each encoding (AKA simulcast stream) is "L1T1", then which type is selected, SIMULCAST or SVC?

And if the Producer uses VP8 with 1 encoding with `scalabilityMode` "L1T3"?

And if the Producer uses VP8 with 3 encodings and each encoding has `scalabilityMode` "L1T3"?

jmillan: I'll update the PR description with more info.

jmillan: > But VP8 and H264 still exist here. Maybe I don't understand what this PR does. Does it change something for real? 

They are still there because VP8 and H264 can be used in SIMULCAST as before. The change is that now they can also be used in SVC.

> Does it change something for real? Or is it just a "preparation" work to make changes later?

Yes, you will now use SVC when consuming a single encoding and multiple layer stream.

ibc: The description is not ok:

> `&forceVP8=true&numWebcamSimulcastStreams=3&webcamScalabilityMode=L1T2``
>
> This will create a single encoding with L1T3. It will use Simulcast.

I assume you mean "3 encodings each one with L1T3".

jmillan: Yes, typo. Fixed.

ibc:
There are Node tests that check `producer.type` and `consumer.type`. It would be great to update them to handle these new cases (also in Rust).

BTW, now imagine this scenario:

**Producer has VP8 with 1 encoding with L1T3**

- Will `producer.type` be "svc"? or "simulcast"?
- Will `consumer.type` be "svc"? or "simulcast"?

Obviously both should report same type. Tests would help here.

jmillan:
I've added tests for Node. Once confirmed upon review, I'll add them to Rust.

ibc:
> I've added tests for Node. Once confirmed upon review, I'll add them to Rust.

New Node tests LGTM.

jmillan:
Rust tests done.

ibc (approved):

PR #1852, opened 2026-07-03
Add `NotFoundError`

ibc:
# Details

- Add `NotFoundError`, thrown when the referenced entity in the Worker doesn't exist.
- Expose mediasoup Node `errors` via `mediasoup/errors` (in EMS).
- **Breaking change:** Rename `InvalidStateError` to `WorkerClosedError`.

## TODO

- [x] Add "Breaking change" in CHANGELOG and PR description above.
- [x] Rust.
- [x] Document errors in the website.

### DIFF
--- a/worker/src/RTC/Router.cpp
+++ b/worker/src/RTC/Router.cpp
@@ -726,7 +726,7 @@ namespace RTC
 
 		if (mapProducersIt == this->mapProducers.end())
 		{
-			MS_THROW_ERROR("Producer not found [producerId:%s]", producerId.c_str());
+			MS_THROW_NOT_FOUND_ERROR("Producer not found [producerId:%s]", producerId.c_str());
 		}
 
 		auto* producer              = mapProducersIt->second;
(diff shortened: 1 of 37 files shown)


worker/src/RTC/Router.cpp
```diff
@@ -439,55 +436,61 @@ namespace RTC
...
-	{
-		if (this->mapRtpObservers.find(rtpObserverId) != this->mapRtpObservers.end())
-		{
-			MS_THROW_ERROR("an RtpObserver with same id already exists");
-		}
-	}
-
-	RTC::Transport* Router::AssertAndGetTransportById(const std::string& transportId) const
+	RTC::Transport* Router::AssertAndGetTransportById(
+	  const std::string& transportId, const std::string& message) const
```
jmillan, line 445: Why not calling the argument `method`?

```suggestion
	  const std::string& transportId, const std::string& method) const
```

ibc: In notifications we don't have "method" but "event" and those methods could also be called when handling notifications.

ibc: But then the log says `[method:xxxx]` so it's wrong.

ibc: I've renamed it to `method` in v3.

PR #1859, opened 2026-07-20
Encode subchannels in SCTP messages

ibc:
# Details

- Encode `subchannels` within SCTP messages so they survive pipe transports.
- `DataConsumer`: Do not require that the transport is connected at SCTP to be able to send a message. It's not a real requirement! the SCTP stack is ready to buffer sending messages until SCTP is established.

## TODO

* [x] Fix documentation.
* [x] Node and Rust tests.


ibc:
Ready. Rust tests should pass now.

jmillan (approved):

PR #1870, opened 2026-07-24
Demonstrate sequrity advisory GHSA-wq22-prq7-m8pc

jmillan:
https://github.com/versatica/mediasoup/security/advisories/GHSA-wq22-prq7-m8pc

Imagine a chat application that provides with a chat room and also allows for direct messages. Each participant has an associated subchannel but is the application the one that controls which participants can directly message with others. Ie: Only participants in the same organization teams can message each others directly.

With the current code a participant can send direct messages to anyone regardless off the limitations defined by the application.

### DIFF
--- a/node/src/test/test-PipeTransport.ts
+++ b/node/src/test/test-PipeTransport.ts
@@ -1290,7 +1290,7 @@ test('crafted SubchannelsCodec MagicToken payload injected via piped DataChannel
 
 		ctx.dataProducer!.send(craftedPayload);
 
-		// If no message arrives within the timeout, consumer5 correctly rejected
+		// If no message arrives within the timeout, dataConsumer correctly rejected
 		// the crafted payload.
 		setTimeout(resolve, 1000);
 	});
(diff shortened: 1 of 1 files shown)


ibc:
This is not a good scenario because it assumes that the application controls subchannels but it doesn't. The application doesn't send messages with subchannels or requiredSubchannel options. If it did it, the DataConsumer in the pipe transport 1 would add a magic token header with those application decided subchannels that would invalidate the malicious ones introduced by the malicious data producer. And that's exactly my point.

jmillan:
> This is not a good scenario because it assumes that the application controls subchannels but it doesn't.

I get it. I don't have a full app logic to demonstrate it, if it is even possible. So there is no point of wasting more time :+1:.

ibc:
There is only one scenario: an app that sends messages via a direct DataProducer with subchannels (for specific destination consumers via piped transports) and sends messages without subchannels (to be received by all consumers). But in this scenario it doesn't make sense that a browser sends messages via WebRTC that should only be intended for certain consumers because we don't have any api to assign subchannels to a message received via WebRTC.

PR #1875, opened 2026-07-29
Handle subchannels in pipe `DataConsumers`

### LINKED ISSUE #1874: Must handle subchannels in pipe `DataConsumers`
# Feature Request

In order to avoid broadcasting all messages to all pipe `DataConsumers` when not needed, allow these to subuscribe to the desired subchannels. If a pipe `DataConsumer` is subscribed to subchannels then honor them as with regular `DataConsumer`s, filtering messages accordingly. Otherwise send all messages as we do now.

ibc:
# Details

- Fixes #1874.
- Verify/check given subchannels and/or required subchannels in messages received by piped data consumers (`DataConsumers` created on `PipeTransports`).
- However the behavior is different than in data consumers created in **non** `PipeTransports`: If the piped data consumer is not subscribed to **any** subchannel then it lets all messages pass through despite the message comes with subchannels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### 3.23.1
 
-Worker: Fix, use `thread_local` buffer on `MS_ABORT()` ([PR#1873](https://github.com/versatica/mediasoup/pull/1873)).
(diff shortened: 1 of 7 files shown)


jmillan (approved):

jmillan:
Note: using `string_view` as label (argument and member) would avoid memory allocations.

PR #1876, opened 2026-07-30
DataProducerOptions: make new_pipe_transport() public

jmillan:
It was missing. It must be public the same way the analogous `ProducerOptions` function is.

### DIFF
--- a/rust/src/router/data_producer.rs
+++ b/rust/src/router/data_producer.rs
@@ -52,7 +52,8 @@ pub struct DataProducerOptions {
 
 impl DataProducerOptions {
     #[must_use]
-    pub(super) fn new_pipe_transport(
+    /// Create data producer options that will be used with Pipe transport.
+    pub fn new_pipe_transport(
         data_producer_id: DataProducerId,
         sctp_stream_parameters: SctpStreamParameters,
     ) -> Self {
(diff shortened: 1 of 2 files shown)


rust/src/router/data_producer.rs
```diff
@@ -52,7 +52,8 @@ pub struct DataProducerOptions {
 
 impl DataProducerOptions {
     #[must_use]
-    pub(super) fn new_pipe_transport(
+    /// Create data producer options that will be used with Pipe transport.
+    pub fn new_pipe_transport(
```
ibc, line 56: Definitely I'm not familiar with this code in `data_producer.ts` but... why is there a `new_pipe_transport()` within `DataProducerOptions`? I mean, 100% sure this PR is correct but I don't understand the related code.

ibc (approved):

rust/src/router/data_producer.rs
jmillan: This kind of methods are helpers that pre-fill the struct. Ie: Rather than creating an empty `DataProducerOptions` and fill it one by one, `new_pipe_transport()` will explicitly fill the required fields related to the pipe. There are plenty of usages in the code and it's a widely used pattern in rust.

ibc: 👍🏻

PR #1892, opened 2026-08-17
Simulcast and SVC: Limit temporal layer to the preferred one

ibc:
# Details

- Modify `IncreaseLayer()` in `SimulcastProducerStreamManager` and `SvcProducerStreamManager` so temporal layer is always capped at `min(stream temporal layers - 1, preferred.temporal)`.
- `IncreaseLayer()` no longer stops below the preferred spatial layer when no lower one is usable, and can still raise the temporal layer of the one it already holds.
- Of course, if `consumer.setPreferredLayers()` is called without specifying any `temporalLayer`, then it's assumed that the highest possible temporal layer is desired (no behavior change here).

## Notes

- Acknowledgements to @ggarber who wrote a similar PR very long ago but for whatever reason I don't remember it was closed. See PR #665.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
+- **Breaking change:** Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
 
 ### 3.25.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
+
 ### 3.25.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
```
ibc: I need to add BREAKING here.

node/src/Consumer.ts
```diff
@@ -339,7 +339,7 @@ export class ConsumerImpl<ConsumerAppData extends AppData = AppData>
 		if (typeof spatialLayer !== 'number') {
 			throw new TypeError('spatialLayer must be a number');
 		}
-		if (temporalLayer && typeof temporalLayer !== 'number') {
+		if (temporalLayer != null && typeof temporalLayer !== 'number') {
```
ibc, line 342: A guard because temporal layer can be 0 and that's a valid value.

rust/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
```
ibc: Must add BREAKING here.

jmillan (approved):

PR #1899, opened 2026-09-02
Worker: New `RateCalculator`

jmillan:
- Items hold just their count and the oldest one is derived from the ring geometry, replacing four correlated position members with one.
- The ring is sized to always cover the window, so the #1316 overrun, its count leak and the rate it over-reported become unrepresentable rather than worked around.
- Degenerate constructor arguments are clamped, closing a division by zero, an out of bounds access and an unclamped rate cast.
- The memoized rate is keyed by total count as well as time, so `Update()` needs no invalidation store.
- Roughly 2x faster per call and about half the memory per instance, with identical output in the shipped configurations.
- Added regression tests covering the invariants above.

Note: I was willing to rewrite this class since a long time as it became more and more complex with the time. It comes out we've made it ~2x faster. This class is a hot path as its methods (`Update()`, `GetRate()`) are called several times per packet. The performance gain may not be that much in the big picture, comparing with packet send/recv and `srtp`, but this class always showed up in performance tests, so it's a noticiable performance enhancement too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: New RateCalculator ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
+- Worker: New RateCalculator ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
 
 ### 3.26.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Worker: New RateCalculator ([PR #XXXX](https://github.com/versatica/mediasoup/pull/XXXX)).
+
 ### 3.26.0
 
 - **Breaking change:** Simulcast and SVC: Limit temporal layer to the preferred one ([PR #1892](https://github.com/versatica/mediasoup/pull/1892)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: New RateCalculator ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
+- Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
 
 ### 3.26.0
 
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT
 
-- Worker: New RateCalculator ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
+- Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
 
 ### 0.27.0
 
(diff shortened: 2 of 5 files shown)


rust/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Worker: New RateCalculator ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
```
ibc:
```suggestion
- Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEXT
 
+- Worker: New RateCalculator ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
```
```suggestion
- Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
```

worker/include/RTC/RateCalculator.hpp
```diff
@@ -35,42 +45,38 @@ namespace RTC
...
-		std::optional<uint64_t> newestItemStartTimeMs{ std::nullopt };
-		// Index for the last item in the time window.
-		int32_t newestItemIndex{ -1 };
-		// Time (in milliseconds) for oldest item in the time window.
-		std::optional<uint64_t> oldestItemStartTimeMs{ std::nullopt };
-		// Index for the oldest item in the time window.
-		int32_t oldestItemIndex{ -1 };
-		// Total count in the time window.
+		// Item size (in milliseconds). Always >= 1.
+		size_t itemSizeMs{ 1u };
```
ibc: Can we avoid using `u` for unsigned integers? It doesn't look nice and it's only needed in rare cases when entering very big unsigned integer literals.

worker/src/RTC/RateCalculator.cpp
```diff
@@ -3,129 +3,76 @@
 
 #include "RTC/RateCalculator.hpp"
 #include "Logger.hpp"
-#include "Utils.hpp"
 #include <cmath>   // std::trunc()
-#include <cstring> // std::memset()
+#include <limits>  // std::numeric_limits()
+#include <utility> // std::cmp_less()
```
ibc: Remove please. `utility` is already included in `common.hpp`.

worker/src/RTC/RateCalculator.cpp
```diff
@@ -3,129 +3,76 @@
...
-	  : windowSizeMs(windowSizeMs),
-	    scale(scale),
-	    windowItems(windowItems),
-	    itemSizeMs(std::max(windowSizeMs / windowItems, size_t{ 1 }))
 	{
 		MS_TRACE();
 
-		this->buffer.resize(windowItems);
+		// Clamp the given values so every derived value is safe to use.
+		this->windowSizeMs = std::max(windowSizeMs, size_t{ 1u });
```
ibc: Is this `size_t{ 1u }` needed here? If so, can we remove the `u` suffix please? (and everywhere).

ibc (approved):

ibc (approved):

PR #1901, opened 2026-09-02
Worker: Mark `common.hpp` as an umbrella header

jmillan:
- Its includes exist for its consumers, they are not used in the file, so include-cleaner (clang) reported all of them as unused, and also reported `common.hpp` itself as unused in every file including it.
- `IWYU pragma: begin_exports` states that intent, silencing both cases without having to disable the diagnostic.


jmillan:
brew clang-format@22 just vanished in favour of @23.., but now 23 is not found in linux... I'll check it.

ibc (approved):

ibc (approved):

PR #1907, opened 2026-09-03
Worker: fix DataConsumer::Send() response generation

jmillan:
Worker and Node were using `FbsDataConsumer.GetBufferedAmountResponse` rather than `FbsDataConsumer.SendResponse`as response to `FbsDataConsumerSend` request.

It worked because they have exactly the same schema, but would fail in the future if changed.


ibc (approved):
good catch, but do we really need to document this in the CHANGELOG? it0's something totally internal and it was not even a real bug yet.

jmillan:
we can skeep it, yes. I'll remove it

ibc (approved):

PR #1908, opened 2026-09-03
Enhance comments on mediasoup-worker-prebuild.yaml

jmillan:
Just avoid future me re-investigating about this.

### DIFF
--- a/.github/workflows/mediasoup-worker-prebuild.yaml
+++ b/.github/workflows/mediasoup-worker-prebuild.yaml
@@ -80,11 +80,6 @@ jobs:
           # `getWorkerPrebuildTarName()` in npm-scripts.mjs), so hosts running
           # a kernel 7 look for a different tarball than those running a
           # kernel 6.
-          #
-          # TODO: That per kernel naming was introduced to ship a separate
-          # binary with `io_uring` support (which required kernel >= 6), but
-          # `io_uring` support was removed in PR #1832, so this split is no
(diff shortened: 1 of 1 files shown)


.github/workflows/mediasoup-worker-prebuild.yaml
```diff
@@ -47,18 +47,44 @@ jobs:
...
+          # A separate build is needed because prebuilt tarballs are named
+          # after the kernel major version of the build host (see
+          # `getWorkerPrebuildTarName()` in npm-scripts.mjs), so hosts running
+          # a kernel 7 look for a different tarball than those running a
+          # kernel 6.
+          #
+          # TODO: That per kernel naming was introduced to ship a separate
+          # binary with `io_uring` support (which required kernel >= 6), but
+          # `io_uring` support was removed in PR #1832, so this split is no
+          # longer strictly needed.
```
ibc: I would remove this TODO. Yes, we added kernel version to the tarball because of io_uring support, but still it's useful to keep it. Otherwise we would have a single worker prebuild for **all** Linux versions/kernels, meaning that we would use the worker prebuild generated above with Ubuntu 22.04 also for newer Ubuntu 26 hosts.

jmillan: My fault, it was not a TODO but rather a NOTE. let's remove it.

ibc (approved):

PR #1910, opened 2026-09-03
Remove unused FBS.RtpParameters.EncodingMapping.scalability_mode field

jmillan:
`FBS.RtpParameters.EncodingMapping.scalability_mode` was serialized by both Node and Rust on every `transport.produce()`, but the worker never read it. `Producer.cpp` only reads `ssrc`, `mapped_ssrc` and the *presence* of `rid` from that table, and it already passed `nullptr` for this field when echoing `rtpMapping` back in `producer.dump()`.

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### NEXT
 
 - Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
-- **Breaking change:** Remove unused `scalability_mode` field from `FBS.RtpParameters.EncodingMapping` ([PR #1910](https://github.com/versatica/mediasoup/pull/1910)).
+- Remove unused `scalability_mode` field from `FBS.RtpParameters.EncodingMapping` ([PR #1910](https://github.com/versatica/mediasoup/pull/1910)).
 
 ### 0.27.0
 
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### NEXT
 
 - Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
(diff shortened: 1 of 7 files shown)


rust/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### NEXT
 
 - Worker: New `RateCalculator` ([PR #1899](https://github.com/versatica/mediasoup/pull/1899)).
+- Remove unused `scalability_mode` field from `FBS.RtpParameters.EncodingMapping` ([PR #1910](https://github.com/versatica/mediasoup/pull/1910)).
```
ibc: I insist that IMHO we don't need to include those super internal changes in public CHANGELOG.

ibc:
> `FBS.RtpParameters.EncodingMapping.scalability_mode` was serialized by both Node and Rust on every `transport.produce()`, but the worker never read it. `Producer.cpp` only reads `ssrc`, `mapped_ssrc` and the _presence_ of `rid` from that table, and it already passed `nullptr` for this field when echoing `rtpMapping` back in `producer.dump()`.

I haven't looked in depth but I fail to understand this. Worker needs scalability mode to know how many temporal layers a stream declares.

jmillan:
> I haven't looked in depth but I fail to understand this. Worker needs scalability mode to know how many temporal layers a stream declares.

1. **`RtpEncodingParameters.scalability_mode`** — inside `RtpParameters.encodings`, i.e. the producer's actual RTP parameters. **This is the one that declares the layers, and it is NOT touched.** It's still parsed in `RtpEncodingParameters.cpp:54-79`, where the `L{spatial}T{temporal}` regex fills `spatialLayers`, `temporalLayers` and `ksvc`.

2. **`EncodingMapping.scalability_mode`** — inside `RtpMapping.encodings`, the producer→router SSRC mapping table. **This is the one removed.**

ibc (approved):
Approved but see new concern above.

Issue #1913, opened 2026-09-06
Worker: convert time to `int64_t`


ibc:
# Context

Same as we are doing for the new mediasoup BWE in https://github.com/versatica/mediasoup/issues/1895, we should migrate from `uint64_t` to `int64_t` when it comes to time (which usually is `uint64_t` in ms).

Rationale is that many mathematical operations involve subtractions of time or bitrate, and subtracting A - B, when both are unsigned and B > A, produces a giant positive integer. However, if A and B were signed, the result could easily be determined to be negative and discarded.

## More info

- This must be done in the whole SCTP stack, in RTP, RTCP, etc.

ibc:
Some partial work has been done in https://github.com/versatica/mediasoup/pull/1917.

Issue #1895, opened 2026-09-01
New Bandwidth Estimation stack


ibc:
# Overview

We must drop `libwebrtc` dependency, which is very old and it's only used for BWE purposes. It's also not ideal for a SFU since it's designed for a browser/device that produces up to 2-3 audio/video streams and can change their bitrate at any time to accomplish with network limits dynamically.

So we need to write our own BWE stack.

# Ongoing work

- [x] https://github.com/versatica/mediasoup/pull/1896
- [x] https://github.com/versatica/mediasoup/pull/1897
- [x] https://github.com/versatica/mediasoup/pull/1900
- [x] https://github.com/versatica/mediasoup/pull/1903
- [x] https://github.com/versatica/mediasoup/pull/1911
- [x] https://github.com/versatica/mediasoup/pull/1914

ggarber:
This looks fun :P Just make sure you run a suite of automated tests in different network conditions at the end of the refactor and compare the results before and after.

ibc:
> This looks fun :P Just make sure you run a suite of automated tests in different network conditions at the end of the refactor and compare the results before and after.

And if something goes wrong I can randomly change the value of some of those 84-97 trial fields across those 12-19 BWE related classes and try again.

PR #1917, opened 2026-09-07
Worker: Use real receive time of packets

ibc:
# Details

- Until now the worker read the clock at the point of use whenever it needed a packet's arrival time, which measures when the packet was processed rather than when it arrived. The timestamp is now taken as close to the socket read as possible. `RTC::Transport::ReceiveRtpPacket()`, `RTC::Transport::ReceiveRtcpPacket()` and `RTC::Transport::ReceiveSctpData()` are now provided with `int64_t receiveAtUs` argument, no matter whether the packet is RTP, RTCP, STUN, DTLS or SCTP.
- `Channel::ChannelNotification` takes the time at which it was read from the channel, so a RTP packet injected via `producer.send()` also carries a real arrival time.
- The whole SCTP stack is converted from `uint64_t` milliseconds to `int64_t` microseconds so that sub-millisecond times are not lost.
- **Fix:** `RTC::SCTP::RetransmissionTimeout::ObserveRtt()` discarded any measurement equal to zero, which in milliseconds meant every sub-millisecond RTT. Over a LAN or over loopback that discarded all of them, so the RTO never left its initial value.
- **Fix:** `RTC::RTP::RtpStreamRecv::CalculateJitter()` computed the RFC 3550 interarrival jitter from the clock read while processing the packet rather than from its arrival time, so our own processing jitter was reported to the sender as network jitter in every Receiver Report. It now takes the arrival time, and in microseconds instead of milliseconds.
- **Fix:** `SenderReportTiming::receivedAtUs` (in `RTC::RTP::RtpStreamRecv` and `RTC::RTP::RtxStream`) is now the arrival time of the Sender Report, and the DLSR reported back to the sender is computed in microseconds, matching its 1/65536 second unit.
- **Fix:** The RTT computed from Receiver Reports and from XR DLRR blocks was measured against the clock read while processing them rather than against their arrival time, adding our own processing delay to a value that feeds the bandwidth estimator, the retransmission decisions and `RTC::NackGenerator`. Same for the arrival time the DLRR reported back to the sender is measured from.
- The whole RTCP generation and capture instant paths are converted to `int64_t` microseconds too, so the NTP timestamps we emit and the capture instants used for A/V sync no longer lose precision. Includes new `Utils::Time` microsecond conversions and `DepLibUV::GetNtpOffsetUs()`, which no longer throws away the microseconds `uv_gettimeofday()` already provides.
- `mocks::MockShared` takes the clock in microseconds and derives the millisecond ones from it, so it no longer truncates the clock of the tests that use it.
- The whole new ongoing/WIP BWE (see #1895) now uses microseconds as well.
- This PR _partially_ complemtes ticket https://github.com/versatica/mediasoup/issues/1913.
- **Fix:** The wall clock instant `RTC::RTP::RtpStreamSend` stored for the last Sender Report it emitted was our own monotonic clock rather than the NTP one it announced, so the same field meant a different thing in a send and in a receive stream. _NOTE:_ This code was not really used but it makes sense that it exists.
- `RTC::RTP::RtpStreamRecv::InterpolateRemoteCaptureAtUs()` takes the max distance in microseconds, so the interpolation limit is no longer truncated to whole milliseconds.
- Remove unused `RTC::RTP::RtpStream::GetMaxPacketAtUs()`.


jmillan (approved):

worker/include/SharedInterface.hpp
```diff
@@ -52,33 +52,20 @@ class SharedInterface
...
-
-	/**
-	 * @remarks
-	 * - Used within libwebrtc dependency which uses int64_t values for time
-	 *   representation.
-	 *
-	 * @todo Remove once not needed.
+	 * Get current time in milliseconds in int64_t.
 	 */
 	virtual int64_t GetTimeMsInt64() = 0;
```
jmillan, line 57: Not for this PR, but do we need to add the return type on the method name?

jmillan (approved):

worker/include/SharedInterface.hpp
ibc: Yes because we have all those methods since ever. If in the future we can get rid of getTimeMs() then we can remove the suffix in others.

PR #1918, opened 2026-09-08
Worker: Use `int64_t` for time everywhere

### LINKED ISSUE #1913: Worker: convert time to `int64_t`

ibc:
# Details

- Fixes #1913
- Also add `Ms` or `Us` suffix to all variables, constants, arguments, fields and functions/methods that hold a time or duration.

### DIFF
--- a/worker/fbs/transport.fbs
+++ b/worker/fbs/transport.fbs
@@ -161,7 +161,7 @@ table Dump {
 
 table Stats {
     transport_id: string (required);
-    timestamp: int64;
+    timestamp: uint64;
     sctp_state: FBS.SctpAssociation.SctpState = null;
     bytes_received: uint64;
     recv_bitrate: uint32;
@@ -256,7 +256,7 @@ table BweTraceInfo {
 
 table TraceNotification {
     type: TraceEventType;
-    timestamp: int64;
+    timestamp: uint64;
     direction: FBS.Common.TraceDirection;
     info: TraceInfo;
 }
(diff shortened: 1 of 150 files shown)


ibc:
_NOTE:_ I'm renaming `GetTimeXxInt64()` to `GetTimeXx()` everywhere (where `Xx` can be `Ms` or `Us`). Will push soon.

worker/fbs/transport.fbs
```diff
@@ -256,7 +256,7 @@ table BweTraceInfo {
 
 table TraceNotification {
     type: TraceEventType;
-    timestamp: uint64;
+    timestamp: int64;
```
jmillan: Why are we changing this? A signed timestamp may be useful inside the worker for internal operations with BWE etc, but why would we expose that to Node or Rust? A timestamp determining when an event has occurred cannot be negative by definition.

ibc: Reverted and explained in this commit: https://github.com/versatica/mediasoup/pull/1918/commits/41f961cd4fb7eb302062c349c8a48b2c365c1c8d

jmillan (approved):

